PR #24, opened 2021-06-09
feat(server): specify API methods

crowlKats:
This allows to create exported function with appropriate method names. Default export will catch any routes that do not have a handler for them. As per default denosaurs router behaviour, hitting an endpoint that doesnt have handling for the requested method will just return a 405 with the possible methods for that endpoint as per spec.


src/server/routes.ts
```diff
@@ -14,15 +14,27 @@ export interface Page {
...
+export type ApiRouteModule = {
+  [
+    K in
+      | "default"
+      | "GET"
+      | "HEAD"
+      | "POST"
+      | "PUT"
+      | "DELETE"
+      | "CONNECT"
```
lucacasonato: Remove `CONNECT`. Deno native http / deploy will likeky block that method soon :-)

lucacasonato: Actually lets remove `TRACE` too. Not useful and is blocked by `fetch`.

src/server/routes.ts
```diff
@@ -49,7 +61,7 @@ export function processRoutes(routes: Routes): [Page[], ApiRoute[]] {
         route,
         url,
         name,
-        handler: module.default as router.MatchHandler,
+        handlers: module as ApiRouteModule,
```
lucacasonato: Can you filter out only the supported methods here?

src/server/routes.ts
```diff
@@ -14,15 +14,27 @@ export interface Page {
...
+      | "PATCH"
+  ]?: router.MatchHandler;
+};
 
 export interface ApiRoute {
   route: string;
   url: string;
   name: string;
-  handler: router.MatchHandler;
+  handlers: ApiRouteModule;
```
lucacasonato:
```suggestion
  handlers: Record<string, router.MatchHandler>;
```

lucacasonato (commented):
Very nice. Just some minor comments.

lucacasonato (commented):
Sorry to bug, but could you add an additional test that "NOTAREALMETHOD" response a 405, even if there is an exported handler for it?

lucacasonato (approved):
LGTM, thanks :-)

PR #25, opened 2021-06-09
feat: add _document route

### LINKED ISSUE #19: Add custom render function support
The `start` function that starts the server in `main.ts` should take an optional `render` function. This function can be used to create the JIT server side rendered static HTML page for a specific route. It gets a page component, some items to put into the `<head>`, and some items to put at the end of the `<body>` in its parameters, and returns a string.

There is a default implementation of this render function. It can be used for advanced use-cases like #7

lucacasonato:
This resolves #19, but I am not quite happy with the implementation. I think I will do a second pass on it after implementing some other things, like suspense capable (async) rendering.﻿ I am unsure about this whole `_document` situation. Maybe it would make more sense to give users "setup", "pre render" and "post render" hooks (which can trigger multiple times when rendering asynchronously), and "postprocess" which can be used by the user to adjust the content of items in `<head>`. This would also be more compatible with a "helmet" module that you could use to update items in `<head>`.

Not merging for now, but opening for visibility.


lucacasonato:
Alternate implementation in #30 supersedes this.

PR #55, opened 2021-09-25
BREAKING: disable runtime JS by default

lucacasonato:
This commit disables runtime JS by default. This stems from the observation that most pages are static, and thus do not need runtime JS.


satyarohith:
@lucacasonato, does this mean we have to opt-in for SSR?

lucacasonato:
@satyarohith No you opt in to client side rendering.

PR #64, opened 2021-09-26
Fix build on windows

### LINKED ISSUE #63: On windows - Broken count buttons in the example project, and init
When trying to run the example, the count buttons are not enabled.

It seems the js script is not found.

image

_the init script should probably have `runtime: true` in the generated index.tsx as well_

sylc:
Fix https://github.com/lucacasonato/fresh/issues/63

The issue is that on windows the bundle.outputFiles path are absolute path, while on linux they are relative path.
So on windows the url path do not match the bundle cache path.

I have made esbuild output absolute path from for both linux and windows, based on cwd, then the cache keys are recalculated by removing the cwd path. (similar to the route gen)

tested on windows and linux.
external ref:
https://github.com/evanw/esbuild/blob/9e569c40f9f642139c59372e7321d90baeb507c1/internal/bundler/bundler.go#L1810
https://pkg.go.dev/path/filepath#Join

### DIFF
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -64,7 +64,7 @@ async function init(directory: string) {
   )}";\n`;
   await Deno.writeTextFile(join(directory, "deps.ts"), DEPS_TS);
   const PAGES_INDEX_TSX = `/** @jsx h */
-import { h, IS_BROWSER, useState } from "../deps.ts";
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM, thanks. I really need to improve the testing situation around `runtimeJS: true`.

PR #75, opened 2021-11-02
[quenio-lt] Patch: Log Error Message If Head Component Fails

lt-hacker:
In order to aid in debugging, appending the error message if the Head component fails.

### DIFF
--- a/src/runtime/head.ts
+++ b/src/runtime/head.ts
@@ -10,7 +10,8 @@ export function Head(props: HeadProps) {
   let context: ComponentChildren[];
   try {
     context = useContext(HEAD_CONTEXT);
-  } catch {
+  } catch (err) {
+    console.log("Rendering of Head component failed:\n", err);
     throw new Error(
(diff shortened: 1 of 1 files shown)


src/runtime/head.ts
```diff
@@ -10,9 +10,9 @@ export function Head(props: HeadProps) {
   let context: ComponentChildren[];
   try {
     context = useContext(HEAD_CONTEXT);
-  } catch {
+  } catch (err) {
     throw new Error(
-      "<Head> component is not supported in the browser, or during suspense renders.",
+      "<Head> component is not supported in the browser, or during suspense renders: " + err.message,
```
lucacasonato: Seems like a prime candidate for the new `Error.cause` feature in JS: https://github.com/tc39/proposal-error-cause

```suggestion
      "<Head> component is not supported in the browser, or during suspense renders.",
      { cause: err },
```

lucacasonato (approved):
LGTM, thanks @quenio-lt. Sorry for the slow review!

PR #81, opened 2021-12-24
fix for issue:80

boazblake:
this should help

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -148,7 +148,7 @@ export class ServerContext {
       const url = new URL(req.url);
       if (url.pathname.length > 1 && url.pathname.endsWith("/")) {
         url.pathname = url.pathname.slice(0, -1);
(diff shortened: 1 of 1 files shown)


src/server/context.ts
```diff
@@ -148,7 +148,7 @@ export class ServerContext {
       const url = new URL(req.url);
       if (url.pathname.length > 1 && url.pathname.endsWith("/")) {
         url.pathname = url.pathname.slice(0, -1);
-        return Response.redirect(url, 307);
+        return Response.redirect(JSON.stringify(url), 307);
```
lucacasonato:
```suggestion
        return Response.redirect(url.href, 307);
```

I think this should work too.

boazblake: ok updated 👍

lucacasonato:
Thank you

PR #82, opened 2021-12-30
feat: allow passing additional args to HandlerContext.render

crowlKats:
Which then get passed to the page component


src/runtime/types.ts
```diff
@@ -15,6 +15,11 @@ export interface PageProps {
    * be `{ path: [ 'bar', 'baz' ] }`.
    */
   params: Record<string, string | string[]>;
+
+  /**
+   * Additional arguments were passed to HandlerContext.render
+   */
+  renderArgs?: T;
```
lucacasonato:
```suggestion
  /**
   * Additional data passed into `HandlerContext.render`. Not present by
   * default.
   */
  renderData?: T;
```

lucacasonato (commented):
Looks good. Just some minor name bikeshedding :-)

lucacasonato (approved):
LGTM, thx!

PR #89, opened 2022-02-11
feat: allow render to return string

crowlKats:


src/server/render.tsx
```diff
@@ -310,6 +310,7 @@ export async function suspenseRender(
     function render() {
       try {
         body = renderToString(vnode);
+        return body;
```
lucacasonato, line 322: In the "catch a promise path" this should return an empty string.

lucacasonato (commented):
Missing tests

src/server/render.tsx
crowlKats: should i change the typings to RendererModule.render to return only string and not null then?

lucacasonato: Yes

lucacasonato (approved):
LGTM

Issue #50, opened 2021-09-03
Development address not obvious


sylc:
When starting, the `0.0.0.0:8000` address appear in the terminal.
![image](https://user-images.githubusercontent.com/10268458/131979627-0befc6f8-a561-4ba8-b17b-0696be78c085.png)

 Although when navigating to it, there is nothing.
The [documentation ](https://github.com/lucacasonato/fresh/blob/main/docs/getting-started.md#running-the-project-locally )mention `localhost:8080`

The correct address is `localhost:8000`

lucacasonato:
Resolved now!

PR #93, opened 2022-02-21
Fix default port number in docs

kidonng:
Despite #50, the Markdown docs still says 8080 instead of 8000 :(


lucacasonato (approved):
Thanks!

PR #94, opened 2022-02-28
feat: add etag support to static file handler

piscisaureus:
I guess I should add tests?


lucacasonato (commented):
This is going to be really slow, as startup time will significantly suffer due to all of the file fetching. The etag does not need to be a file hash. It could also just be a hash of "path + deployment ID", as deployments are immutable. This is how client side JS works for example (`/_frsh/$deployment_id`) path for example.

Also: tests

lucacasonato:
Landed in #95

PR #95, opened 2022-03-01
feat: static file etag

crowlKats:


src/server/context.ts
```diff
@@ -192,17 +194,28 @@ export class ServerContext {
...
           "application/octet-stream";
+        const etag = await crypto.subtle.digest(
+          "SHA-1",
+          encoder.encode(BUILD_ID + path),
+        ).then((hash) =>
+          Array.from(new Uint8Array(hash))
+            .map((byte) => byte.toString(16).padStart(2, "0"))
+            .join("")
+        );
+        console.log(etag);
```
lucacasonato: stray `conosle.log`

src/server/context.ts
```diff
@@ -192,17 +194,27 @@ export class ServerContext {
         includeDirs: false,
         followSymlinks: false,
       });
+      const encoder = new TextEncoder();
       for await (const entry of entires) {
```
lucacasonato, line 198: Unrelated: we should parallelise this with `Promise.all` or `pooledMap` in a follow up

lucacasonato (approved):
LGTM!

PR #97, opened 2022-03-03
Reimagining fresh to be "island" based

lucacasonato:
This PR changes makes some fundamental changes to `fresh`. I will try to explain the changes in detail here, and give rationale for why I am making these changes.

### The status quo

Currently a typical fresh app will have a `pages/` directory containing one file per rendered route. You can also create non page routes in the pages directory (POST, api routes etc). The rendered are always SSR'd, but can optionally also be completely client side rendered by enabling the `runtimeJS` option in the page config. This will bundle the entire source code required for rendering a page into client side JS on demand.

When using just SSR, so without runtime JS enabled, this is an awesome way to build websites. It is fast, and very efficient. Once you enable runtime JS though, the story changes. You are now bundling a whole bunch of stuff into your client bundles to do rendering of things that really don't need to be re-rendered on the client. For example on a blog page, you don't need to re-render the entire blog entry from markdown on the client.

Instead you only want to render certain pieces of the page interactively. The page is a sea of content, with little "islands" of interactivity.

### All in on islands

The new version of fresh embraces this island architecture. Pages are now always completely rendered server side and never in their entirety client side. You can however create "island" components that live in a new special `islands/` folder. These islands of interactivity are shipped to the client to hydrate SSR'd DOM nodes to provide interactivity. This minimizes the JS you ship to the client to only the code you actually need, while still making it trivial to provide interactivity.

This commit also removes `useData` and all notions of streaming HTML responses (and `<Suspense>`). I might re-add this in a follow up commit. For now it made it easier to integrate this new island architecture. To get asynchronous data now, you should use the `handler` and specify route data.

Some additional changes:

- `pages/` has been renamed to `routes/`
- `routes.gen.ts` has been renamed to `fresh.gen.ts`
- `fresh routes` has been renamed to `fresh manifest`


examples/counter/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+```
+deno run -A --watch main.ts
+```
+
+After adding, removing, or moving a page in the `routes` or directory, or
+adding, removing, or moving an island in the `islands` directory, run:
+
+```
+fresh maniofest
```
sylc:
```suggestion
fresh manifest
```

src/cli/init.ts
```diff
@@ -168,16 +184,17 @@ Start the project:
 deno run -A --watch main.ts
 \`\`\`
 
-After adding, removing, or moving a page in the \`pages\` directory, run:
+After adding, removing, or moving a page in the \`routes\` or directory, or adding,
+removing, or moving an island in the \`islands\` directory, run:
 
 \`\`\`
-fresh routes
+fresh maniofest
```
sylc:
```suggestion
fresh manifest
```

crowlKats (approved):
LGTM! This is a great change & improvement!

PR #99, opened 2022-03-04
chore: update esbuild

lucacasonato:
Update esbuild to 0.14.25.


lucacasonato:
I'm vendoring in `esbuild-wasm` until https://github.com/evanw/esbuild/commit/536ebd3b906de8e7a74f309ffa0a8a79ba1fcc22 is released in an esbuild release.

PR #114, opened 2022-03-06
refactor: detach request from context

### LINKED ISSUE #113: Change handler signature from `(ctx) => resp` to `(req, ctx) => resp`
The current signature is somewhat clumsy. It does not align nicely with the signature for `serve` from deno.land/std.

The `req` property from `ctx` should be removed. The `match` property from `ctx` should be renamed to `params`.

This is a prerequisite for #96.

crowlKats:
Closes #113

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -33,30 +33,34 @@ export interface IslandModule {
 }
 
 export interface HandlerContext<T = unknown> {
-  req: Request;
   match: Record<string, string>;
   render: (data?: T) => Response;
 }
 
 export interface UnknownHandlerContext {
-  req: Request;
   render: () => Response;
 }
 
 export interface ErrorHandlerContext {
-  req: Request;
   error: unknown;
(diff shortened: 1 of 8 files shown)


src/server/types.ts
```diff
@@ -33,30 +33,34 @@ export interface IslandModule {
 }
 
 export interface HandlerContext<T = unknown> {
-  req: Request;
   match: Record<string, string>;
```
lucacasonato, line 36: Want to do this rename in a follow up?

lucacasonato (commented):
www/ and examples/ and `fresh init` need to be updated.

src/server/types.ts
crowlKats: yea, was also considering doing #96 in this PR, but maybe its just cleaner to have these things in separate PRs

lucacasonato:
Thx :-)

PR #121, opened 2022-03-07
feat: pass ConnInfo to various handler contexts

### LINKED ISSUE #96: Expose source IP to middleware and handlers

crowlKats:
Closes #96


src/server/types.ts
```diff
@@ -33,20 +33,24 @@ export interface IslandModule {
 }
 
 export interface HandlerContext<T = unknown> {
+  connInfo: ConnInfo;
```
lucacasonato: I think `HandlerContext` should implement `ConnInfo` rather than it containing a `ConnInfo` object. That way you can do `ctx.sourceIp` rather than having to do `ctx.connInfo.sourceIp`.

crowlKats: done. Sadly `ConnInfo` isnt a clean interface like that, and is overly "complex" imo. maybe we should do something a bit different? (maybe only `remoteAddr` instead?)

lucacasonato: What isn't clean about it? It's an interface with two fields :-)

https://doc.deno.land/https://deno.land/std@0.128.0/http/mod.ts/~/ConnInfo

crowlKats: yea, but i dont really like it. not sure why 😅. eh, whatever

lucacasonato (approved):
LGTM

lucacasonato:
actually, can you add a test?

lucacasonato (approved):
LGTM, thanks!

PR #122, opened 2022-03-07
Add error page

### LINKED ISSUE #109: Show page build errors (500 errors) as a nice HTML error page during development
Currently when a page fails to render (for example it `throw`s), an error is logged to the console and the HTTP request is responded to with a '500 Internal Server Error' response. After this plain text response is rendered in the browser, the hot-reload breaks because the `reload.js` script is not running in the browser anymore (because the 500 error is a plain text page, not an HTML page).

Ideally the default 500 error page should be a regular HTML page, that contains the `reload.js` script as usual.

One can probably do this by removing this try-catch block here: https://github.com/lucacasonato/fresh/blob/fc25d2dfbf58d3ad2f9aba4b49925b12a8abafc9/src/server/context.ts#L312-L333
and then improving the default error page here to include a `component` and changing the handler to be `(ctx) => ctx.render()`:
https://github.com/lucacasonato/fresh/blob/fc25d2dfbf58d3ad2f9aba4b49925b12a8abafc9/src/server/context.ts#L506-L514

bartlomieju:
Closes #109


bartlomieju:
Still requires CSS update before landing.

lucacasonato:
image

lucacasonato (approved):
LGTM, thanks Bartek!

PR #130, opened 2022-03-10
fix: 128 routes/[id]/[path] catches request for _frsh/refresh.js

### LINKED ISSUE #128: `./routes/[id]/[path].tsx` catches request for `_frsh/refresh.js`
Should we move this block up?
https://github.com/lucacasonato/fresh/blob/026ed86f4009fc0a35463d43218bf8c34aead98e/src/server/context.ts#L395

kuboon:
Fixes #128
Requests with INTERNAL_PREFIX should be handled before all other routes.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -282,6 +282,42 @@ export class ServerContext {
   #routes(): [router.Routes, RequestHandler, router.ErrorHandler] {
     const routes: router.Routes = {};
 
+    routes[`${INTERNAL_PREFIX}${JS_PREFIX}/${BUILD_ID}/:path*`] = this
+      .#bundleAssetRoute();
+
+    if (this.#dev) {
+      routes[REFRESH_JS_URL] = () => {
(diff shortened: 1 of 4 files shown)


lucacasonato (commented):
Could you add a test for this to the `tests/` directory?

kuboon:
I added this new route on `tests/fixture/fresh.gen.ts`
```
    "./routes/[...all].ts": $16,
```
and found that these 2 tests fails.
```
/books/:id page - /books/abc
/api/xyz not found
```
404 expected / 200 actual (because new route catches it)

Which is better?
1. change above 2 expectations to 200
2. prepare another fixture set
3. anything else

lucacasonato:
Could you either do it in the other fixture (fixture_error), or add another fixture? If you'd rather I do this, just let me know 🙂

lucacasonato (approved):
LGTM, thanks!

PR #133, opened 2022-03-13
feat: Add support for cached static asset

### LINKED ISSUE #111: Add support for cached static asset handling
Currently, assets stored in the `static/` directory in the project are served through fresh automatically. These assets are served without any cache headers though. This is not ideal (especially not for assets which rarely change). It would not be safe to just plainly add long duration cache headers, but it would be reasonable to also host static assets at a seperate URL endpoint that includes the deployment ID in the path. For that "hashed" path, it would be safe to include a very long duration cache header.

I imagine the API would look like this:

```js
...
import { asset } from "../client_deps.ts";

function Page() {
  return <link type="stylesheet" href={asset("/style.css")} />
}
```

The output of the `asset("/style.css")` call would be something like `/_frsh/static/${DEPLOYMENT_ID}/style.css`. The cache header could be immutable for at least 1 year.

sylc:
Closes https://github.com/lucacasonato/fresh/issues/111

### Note
I have made the use of the `asset` function mandatory, i.e a static file is not served anymore on its normal path like `https://fresh.deno.dev/style.css`. 

The other alternative would be to serve the static file under 2 paths, the current one and the 'hashed' one.
However serving under to 2 path has the disadvantage that it will be really easy for a user to forget to use the `asset` function. Maybe in that option, the `assets function should be call `withCache`.
I opted to the fail fast option, i.e only serving the static file under its cache version. 

I think the best would be to have no need  for the user to use the `asset` function, and fresh will automatically serve the file with a `hash` version and provide a `noCache` function if the user do not want a hash path. But this is harder to do.

I'm happy to update if you believe serving 2 path is actually better.


lucacasonato (commented):
This is very nice @sylc. Thanks for working on it.

I think it would make sense if _both_ the existing functionality, and the new functionality with the `asset` functions were available out of the box. Sometimes you want to just serve a static asset at a predetermined path, for example a PDF file or a slide deck that you want to link to directly.

One thing I think that could be done for a follow up to mitigate your concern that people will forget (or be unaware) to use `asset()`, is to have `fresh` automatically install a service worker client sides that intercepts static file requests on "unhashed" static file paths, and rewrites them inside of the service worker to hashed paths. The service worker would then be responsible for caching the pages in sync with the build id of the requesting page. Details would still need to be figured out (disclaimer: I just came up with this on the fly, so this may or or may not work out - experimentation will be required).

Something more short term (for a follow up PR) would be to use preact's "options hooks" (specifically the vnode creation hook) to automatically rewrite things like `src` props on `img` elements to use the "hashed" static files, or inject preload headers.

lucacasonato (commented):
I am not quite sure of the exact approach you are taking for the build ID here. Two main concerns:

- There are two `asset()` functions. The one for the server, and the one for the client. Ideally there is just one used by both, that works exactly the same in both environments.
- The client side asset function does not return the "real" hashed URL on the server. It returns a placeholder URL that will later be replaced during rendering. This is not ideal, as it will break if you try to return the URL in a "raw" response from a handler function, that doesn't go through the HTML rendering (for example an API endpoint).

Instead, I suggest you try the following:

- In `src/runtime/utils.ts` you add a single `asset()` function that takes a path. This function internally returns `${INTERNAL_PREFIX}${STATIC_PREFIX}/${globalThis.__FRSH_BUILD_ID}${path}`.
- In `src/server/constants.ts`, you define `globalThis.__FRSH_BUILD_ID` to be the build ID. This will cover all server side use cases, because `src/server/constants.ts` will always be loaded.
- During the esbuild bundling for client code (`src/server/bundle.ts`) you esbuild's [`define`](https://esbuild.github.io/api/#define) option to replace `globalThis.__FRSH_BUILD_ID` with the real build from the server constants ID. This means that you don't need the `<script id="__FRSH_BUILD_ID">` anymore, and instead the build ID will be directly inlined into the JS files.

sylc:
@lucacasonato Thanks for pointing me in the right direction. I have now updated as per your suggestion. Please another look. cheers.

lucacasonato (approved):
LGTM, thanks!

PR #139, opened 2022-03-19
feat: Layered middleware and passing of data through them

sylc:
Closes: 
- https://github.com/lucacasonato/fresh/issues/131
-  https://github.com/lucacasonato/fresh/issues/136

Notes: 
- I had to modify the `crux-land/router` to be able to pass the `state` to the handler. However I cannot find a GitHub repo for it. So I have added it to the repo for now. How would you prefer to handle this? 
- There is now a type error in the tailwind setup which is also happening on main, after a `--reload`. The error does not seem correct, so I disabled it for now.


sylc:
@lucacasonato this is ready for you to review. Please see the notes in the PR description. thanks.

src/server/context.ts
```diff
@@ -536,6 +613,33 @@ const DEFAULT_ERROR: ErrorPage = {
...
+export function selectMiddlewares(url: string, middlewares: MiddlewareRoute[]) {
+  const selectedMw: [number, Middleware][] = [];
+  const reqURL = new URL(url);
+
+  for (const { route, handler } of middlewares) {
+    const regRoute = pathToRoute(route).slice(0, -"_middleware".length);
+    // check for each subpath of url if it is matching the middleware path
+    const pattern = new URLPattern({
+      pathname: `${regRoute}*`,
+    });
```
lucacasonato: We really don't want to be recompiling the `URLPattern` in each iteration. This should be done only once.

src/server/context.ts
```diff
@@ -536,6 +613,33 @@ const DEFAULT_ERROR: ErrorPage = {
...
+    const res = pattern.exec(reqURL);
+    if (res) {
+      selectedMw.push([route.split("/").length, { handler }]);
+    }
+  }
+
+  // sort middlewares to apply by their layer level from the deepest layer to the root
+  const mws = selectedMw.sort((a, b) => b[0] - a[0]).map((mw) => mw[1]);
```
lucacasonato: The ordering of middlewares relative to each other can also be done ahead of time to prevent some overhead here, as the ordering is fixed.

docs/concepts/middleware.md
```diff
@@ -1 +1,36 @@
...
+```
+
+The middlewares are scoped and layered. So for example if there is both a
+`routes/_middleware.ts` and a `routes/admin/_middleware.ts`, then a request to
+`/admin/foobar` would first go to `routes/_middleware.ts`, then through
+`routes/admin/_middleware.ts`, before finally being handled by a
+`routes/admin/[name].tsx` handler.
+
+Even if an argument is not passed to the `ctx.handle()` function, the state from
+the upper layer middleware will be passed down to the lower layer middleware.
```
lucacasonato: It is currently not explicit if state is merged. With the mutable `ctx.state` idea, this would be irrelevant because merging behaviour would be user defined.

src/server/context.ts
```diff
@@ -265,16 +269,75 @@ export class ServerContext {
...
+            }),
+          );
+        handlers.push(handleItem);
+      }
+
+      return shallowMw!.handler(req, {
+        handle: handlers.pop()!,
+        ...connInfo,
+        state: {},
+      });
```
lucacasonato: This can be simplified a lot when using a mutable `ctx.state`:

```suggestion
      const handlers = [];

      const ctx = {
        next() {
          const handler = handlers.shift();
          return handler();
        },
        ...connInfo,
        state: {},
      }; 
     
      for (const mw of mws) {
        handlers.push(() => mw.handler(req, ctx));
      }
      
      // It'd be nice to not pass `connInfo` && `ctx.state` here, and instead just pass `ctx` directly.
      handlers.push(() => inner(req, connInfo, ctx.state));
      
      const handler = handlers.shift();
      return handler();
```

lucacasonato (commented):
Great work @sylc, thanks! I have some comments, but it's looking very promising already.

My primary change revolves around that I think we could simplify this a lot if instead of passing state down by giving it to `ctx.handle()`, we just had folks set it right onto the `ctx.state` object (so have `ctx.state` be mutable). So for example this:

```js
const session = extractSession(req);
return ctx.handle({ session });
```

would turn into

```js
ctx.state.session = extractSession(req);
return ctx.next();
```

This would not solve the issues with `router` though (cc @crowlKats), but it would significantly simplify the internal logic in `fresh`.

sylc:
I have updated to have a mutable context and reduce to a minimum the changes to the crux-router. (i.e the logic is not changed, i have just modified some signature to allow for more than connInfo as the second parameter. eg: `type HandlerContext<T = unknown> = T & ConnInfo;`

src/server/types.ts
```diff
@@ -105,20 +118,33 @@ export interface Renderer {
...
+export interface MiddlewareRoute extends Middleware {
+  /**
+   * path-to-regexp style url path
+   */
+  route: string;
+  /**
+   * URLPattern of the route
+   */
+  // deno-lint-ignore no-explicit-any
+  regPattern: any;
```
lucacasonato:
```suggestion
  pattern: URLPattern;
```

sylc: On the server, `URLPattern` is declared in globals as `any` and so cannot be used as a type. I could declare it as a class instead but, I'm not sure there is lots of advantage. Having the proper definition would better, but i'm not too sure how.

sylc:
@crowlKats Could you have a look at the the `crux-router.ts` file and advise if the changes could be integrated into the real router ? thanks.

crowlKats:
Hey, @sylc, yes, I will do that first thing in the morning

crowlKats:
@sylc https://crux.land/router@0.0.7, sorry for the delay

lucacasonato (approved):
Thanks @sylc. Sorry for taking this long to review this. This is a great addition!

The tests are very nice btw :-)

PR #140, opened 2022-03-20
test: add integration test

NotFounds:
Closes:
- https://github.com/lucacasonato/fresh/issues/104

Notes:
- It runs `fresh init` and checks if generated files exist simply(does not check the content of the files); access to the server using `curl`. It also checks type.

### DIFF
--- a/.github/scripts/check-generated-files.sh
+++ b/.github/scripts/check-generated-files.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+declare DIRNAME=$1
+if ! [ -d $DIRNAME ]; then
+  exit 1
+fi
+cd $DIRNAME
+
+# Check files
+declare -a TARGET_FILES=(
+  "README.md"
+  "client_deps.ts"
+  "fresh.gen.ts"
+  "main.ts"
+  "server_deps.ts"
+)
+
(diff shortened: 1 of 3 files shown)


lucacasonato (commented):
Hey @NotFounds, thanks for working on this! I realize I didn't specify this in the issue, but it'd be great if this test was written in TypeScript so it would run when we do `deno test` in the repo.

I imagine maybe something like this:

- Have a `tests/cli_test.ts` file
- Have a `Deno.test("fresh init", async () => { ... });`
- Create a temp dir with `Deno.createTempDir`
- Run `deno -A --no-check cli.ts init <tmp>` using `Deno.run` and await the process finishing
- Check all the necessary files exist
- Run `deno run -A main.ts` on the generated output with tmp path as CWD
- Wait until `Listening on ...` is printed by subprocess
- `fetch` the address to check that the project is functional
- Kill the subprocess & wait for it to exit

(essentially porting your shell script to Deno :-) )

If you don't want to / have time to do this, just let me know and I can do the porting over next week!

NotFounds:
Thank you for your review!
I should have confirmed a way to implement it, and I'll try to re-write it in Deno as another PR!

NotFounds:
https://github.com/lucacasonato/fresh/pull/141

PR #141, opened 2022-03-21
test: add integration test

NotFounds:
Closes:
- https://github.com/lucacasonato/fresh/issues/104

Notes:
- https://github.com/lucacasonato/fresh/pull/140#pullrequestreview-915155162
> * Have a `tests/cli_test.ts` file
> * Have a `Deno.test("fresh init", async () => { ... });`
> * Create a temp dir with `Deno.createTempDir`
> * Run `deno -A --no-check cli.ts init <tmp>` using `Deno.run` and await the process finishing
> * Check all the necessary files exist
> * Run `deno run -A main.ts` on the generated output with tmp path as CWD
> * Wait until `Listening on ...` is printed by subprocess
> * `fetch` the address to check that the project is functional
> * Kill the subprocess & wait for it to exit


tests/cli_test.ts
```diff
@@ -0,0 +1,115 @@
...
+              ],
+            },
+            { "type": "file", "name": "index.tsx" },
+          ],
+        },
+        { "type": "file", "name": "server_deps.ts" },
+      ],
+    },
+    { "type": "report", "directories": 3, "files": 9 },
+  ];
```
NotFounds: It uses the results of the tree command.
Because I thought it would be better to compare the results based on the automatically generated ones rather than changing them manually.

tests/cli_test.ts
```diff
@@ -0,0 +1,115 @@
...
+    const lsof = Deno.run({
+      cmd: ["lsof", "-ti", `:${port}`],
+      stdout: "null",
+    });
+    const { code: lsofCode } = await lsof.status();
+    lsof.close();
+    if (lsofCode === 0) break;
+    await new Promise((r) => setTimeout(r, 500));
+  }
+};
```
NotFounds: I think it could be better, but I have no idea...

tests/cli_test.ts
```diff
@@ -0,0 +1,115 @@
...
+  });
+
+  await t.step("start up the server and access the root page", async () => {
+    Deno.chdir(tmpDirName);
+    const serverProcess = Deno.run({
+      cmd: ["deno", "run", "-A", "--no-check", "main.ts"],
+      stdout: "null",
+    });
+
+    await waitForServer(8000);
```
lucacasonato:
```suggestion
      stdout: "piped",
    });

    const lines = serverProcess.stdout.readable
     .pipeThrough(new TextDecoderStream())
     .pipeThrough(new TextLineStream());
     
   for await (const line of lines) {
     if (line.includes("Listening on http://") break;
   }
```

Instead of using `lsof`, which doesn't work on Linux, you can inspect the stdout and wait for the "Listening on" message to show up.

lucacasonato (commented):
Just a minor comment, overall looking good!

lucacasonato:
The test seems to be hanging.

lucacasonato (approved):
LGTM, thanks!

PR #142, opened 2022-03-25
feat: async renderer

so1ve:
If the render function can't be async, I can't use unocss :(


so1ve:
BTW, using unocss instead of twind seems like a good idea.

lucacasonato (commented):
Can you add a test for this? To do this you can just make the render function in `./tests/fixture/routes/_render.ts` async and add an `await new Promise((r) => r());` in the function body.

so1ve:
ok, like this?

lucacasonato (approved):
LGTM, thanks!

lucacasonato:
If you get unocss working, it'd be fantastic if you could add a page to the `docs/` folder about how users can use UnoCSS with fresh. I think people would really find that very useful :-)

PR #146, opened 2022-04-01
Ignoring .git directorys while initalizing a fresh project

RoeHH:
With this change, the fresh init command would ignore .git directories

### DIFF
--- a/src/cli/init.ts
+++ b/src/cli/init.ts
@@ -48,15 +48,8 @@ async function init(directory: string) {
 
   try {
     const dir = [...Deno.readDirSync(directory)];
-    if (dir.length === 1 && dir[0].name === ".git" && dir[0].isDirectory) {
-      console.log(
-        `%cThe directory '${directory}' is a git repository.`,
-        "color: yellow; font-weight: bold",
-      );
-    } else {
-      if (dir.length > 0) {
-        error("Directory is not empty.");
-      }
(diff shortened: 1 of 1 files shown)


RoeHH:
@lucacasonato open for review 😸

src/cli/init.ts
```diff
@@ -48,7 +48,7 @@ async function init(directory: string) {
 
   try {
     const dir = [...Deno.readDirSync(directory)];
-    if (dir.length > 0) {
+    if (dir.length > 0 && !confirm("This is no Empty directory, some Files could get deleted, do you want to continue?")) {
```
lucacasonato:
```suggestion
    if (dir.length > 0 && !confirm("The target directory is not empty (files could get overwritten). Do you want to continue anyway?")) {
```

src/cli/init.ts
```diff
@@ -48,7 +48,7 @@ async function init(directory: string) {
 
   try {
     const dir = [...Deno.readDirSync(directory)];
-    if (dir.length > 0) {
+    if (dir.length > 0 && !confirm("The target directory is not empty (files could get overwritten). Do you want to continue anyway?")) {
```
lucacasonato:
```suggestion
    if (
      dir.length > 0 &&
      !confirm(
        "This is no Empty directory, some Files could get deleted, do you want to continue?",
      )
    ) {
```

lucacasonato (commented):
Thanks @RoeHH!

lucacasonato (approved):

PR #151, opened 2022-04-07
chore: update esbuild

tylerbrostrom:
- Update esbuild to 14.34.
- Remove `esbuild-wasm` from vendored dependencies (see [comment from #99](https://github.com/lucacasonato/fresh/pull/99#issuecomment-1059620606))


tylerbrostrom:
Seems Deno Deploy could unknowingly break—`esbuild-wasm` is not used in testing environments. (I ran tests locally using `esbuild-wasm` before pushing this PR, btw).

Given that Deno Deploy is the likeliest target environment for `fresh` users, maybe tests should use `esbuild-wasm` instead of native `esbuild`?

lucacasonato:
> `esbuild-wasm` is not used in testing environments.

It's not used in the CI tests, but each PR is deployed to Deno Deploy and then GitHub Actions will run Lighthouse against the preview URL. So that is sort of a smoke test.

I'd be happy to merge a PR that updates CI to also run the tests with WASM :-)

lucacasonato (approved):
Thanks @tylerbrostrom.

PR #163, opened 2022-04-23
www: use green as accent color on docs page

lucacasonato:
This adds a green accent color for the docs page header. Not sure if I like it
yet. WDYT @hashrock?

image


hashrock (approved):

hashrock:
LGTM!

PR #172, opened 2022-04-24
feat: differential JS bundle based on user agent

### LINKED ISSUE #38: Optimized bundles
Because bundles are generated with esbuild JIT, we can optimize bundles for each browser we are serving content to. For example the bundle shipped to Firefox 88 would need transpilation for private class methods, whereas a bundle shipped to the latest Chrome would not.

Browser version can be determined by User-Agent string. The remaining question is just if this is even a good idea.

lucacasonato:
This commit adds support for serving a different bundle to different
user agents. This means that every client will get a JS bundle that is
perfectly optimized for the features it supports.

Closes #38

### DIFF
--- a/src/server/bundle.ts
+++ b/src/server/bundle.ts
@@ -22,13 +22,16 @@ async function ensureEsbuildInialized() {
 
 export class Bundler {
   #islands: Island[];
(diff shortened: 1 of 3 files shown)


iuioiua:
I'm going to close this PR, as discussed offline. I won't delete the branch in case we want to revisit this sometime in the future.

Issue #161, opened 2022-04-23
www: lemondrop animation renders slightly differently SSR than on first client render


lucacasonato:
See the below screenshot. The left side is the SSR with JS disabled, the right side is the client side render. The white part of the lemon drop animation is not visible at all on the SSR render, and but averages around 50% of the SVG height on the client render. I think it would make sense if the "start state" of the client animation matched the SSR render.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/7829205/164934361-06978053-a408-4c4a-9eb4-ad0b698a7a02.png">

cc @hashrock

hashrock:
I'll try it!

PR #174, opened 2022-04-24
Fix lemondrop rendering

### LINKED ISSUE #173: Disable animation on home page when `prefers-reduced-motion` is set
[prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) is an accessibility feature that lets developers know that a user prefers to not see motion on the page (like animations).

We should disable the animation on the home page when a user has this set.

cc @hashrock

hashrock:
Closes #173 #161

- Do not call rAF when prefers-reduced-motion is set.
- Since the value of window.innerWidth cannot be obtained during SSR, now it doesn't depend on innerWidth.

### DIFF
--- a/www/islands/LemonDrop.tsx
+++ b/www/islands/LemonDrop.tsx
@@ -16,13 +16,14 @@ function easeInCirc(x: number) {
 const waveTank = new WaveTank();
 
 function LemonDrop() {
+  const SVG_WIDTH = 100;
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
Looks fantastic 💯. The load is soooo seamless now - you can't tell when hydration happens.

It's a masterpiece 😍

PR #179, opened 2022-04-27
fix: update start task with no-check

cknight:
The default task `start` fails to run as a remote dependency has a type issue.  This provides a poor experience to first time users who may struggle to understand why the default example doesn't work.  This issue is resolved by adding the `--no-check` flag to the deno run command.


lucacasonato (approved):
Thanks

PR #180, opened 2022-04-28
feat: group deps files together by filename

dsherret:
Not sure if you would prefer `deps_server.ts`. I used `x.x.ts` because it's similar to `fresh.gen.ts`.


dsherret:
Actually, this isn't so nice because `deps.test.ts` matches as a test file for deno test and in vscode. Looking at what I've done previously elsewhere, I've also done a `test_deps.ts` kind of format.

PR #184, opened 2022-05-10
feat: Automatically apply asset caching to img src URLs

### LINKED ISSUE #167: Automatically apply asset caching to `img` src URLs
This would use the work from #133 and preact's "options hooks" (specifically the vnode creation hook) to automatically rewrite things like src props on img elements to use the "hashed" version of static files.

This would be done only for asset URLs that we definitely know to be static assets (the server knows the paths of all assets in the `static/` directory).

sylc:
fix https://github.com/lucacasonato/fresh/issues/167

> This would be done only for asset URLs that we definitely know to be static assets (the server knows the paths of all assets in the static/ directory).

I stumble on this because the preact hook needs to run on the client when an island is revived. This would require to pass the list of asset to the client, maybe using the island props, which would be cumbersome. Instead the hook assume that if the src path is starting with `/` then it must be a static asset as this is what fresh also assume.  WDYT ?

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -31,14 +31,23 @@ export function asset(path: string) {
   }
 }
 
-// apply the asset function a srcset
-export function assetSrcSet(srcset: string) {
-  const srcsetParts = srcset.split(",");
-  return srcsetParts.map((part) => {
-    const sections = part.trim().split(" ");
-    sections[0] = asset(sections[0]);
-    return sections.join(" ");
-  }).join(", ");
+/** Apply the `asset` function to urls in a `srcset` attribute. */
+export function assetSrcSet(srcset: string): string {
+  if (srcset.includes("(")) return srcset; // Bail if the srcset contains complicated syntax.
+  const parts = srcset.split(",");
+  const constructed = [];
+  for (const part of parts) {
+    const trimmed = part.trimStart();
+    const leadingWhitespace = part.length - trimmed.length;
+    if (trimmed === "") return srcset; // Bail if the srcset is malformed.
+    let urlEnd = trimmed.indexOf(" ");
+    if (urlEnd === -1) urlEnd = trimmed.length;
+    const leading = part.substring(0, leadingWhitespace);
+    const url = trimmed.substring(0, urlEnd);
+    const trailing = trimmed.substring(urlEnd);
+    constructed.push(leading + asset(url) + trailing);
+  }
+  return constructed.join(",");
 }
 
 export function assetHashingHook(
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -17,8 +17,8 @@ export function asset(path: string) {
 export function assetHashingHook(vnode: VNode) {
   if (vnode.type === "img") {
     const props = (vnode.props as HTMLImageElement);
-    // deno-lint-ignore no-explicit-any
     if (
+      // deno-lint-ignore no-explicit-any
       props.src && !(props as any)["data-no-auto-hashing"] &&
       // do not apply the for assets that are already targetting the a frsh special handling
       !props.src.startsWith(INTERNAL_PREFIX) &&
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -14,16 +14,18 @@ export function asset(path: string) {
   return `${INTERNAL_PREFIX}${STATIC_PREFIX}/${__FRSH_BUILD_ID}${path}`;
 }
 
-export function assetHashingHook(vnode: VNode){
+export function assetHashingHook(vnode: VNode) {
   if (vnode.type === "img") {
-    const props = (vnode.props as HTMLImageElement)
+    const props = (vnode.props as HTMLImageElement);
     // deno-lint-ignore no-explicit-any
-    if (props.src && !(props as any)['data-no-auto-hashing'] 
+    if (
+      props.src && !(props as any)["data-no-auto-hashing"] &&
       // do not apply the for assets that are already targetting the a frsh special handling
-      && !props.src.startsWith(INTERNAL_PREFIX)
+      !props.src.startsWith(INTERNAL_PREFIX) &&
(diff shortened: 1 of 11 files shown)


lucacasonato (commented):
Very nice! Tests are failing though.

sylc:
fixed

thegaryroberts:
Thanks for documenting this. Had to do a lot of digging across examples to work out where to put the favicon. Should probably put a placeholder one (the fresh lemon) in the initial app to make this a bit easier. But the docs will help in any case.

src/runtime/utils.ts
```diff
@@ -11,3 +13,45 @@ export const IS_BROWSER = typeof document !== "undefined";
...
+    return sections.join(" ");
+  }).join(", ");
+}
+
+export function assetHashingHook(vnode: VNode) {
+  if (vnode.type === "img" || vnode.type === "source") {
+    const props = (vnode.props as HTMLImageElement);
+    // deno-lint-ignore no-explicit-any
+    if ((props as any)["data-no-auto-hashing"]) {
+      // skip
```
thegaryroberts: This should just be an early return, you can skip the next else and reduce the nesting

src/runtime/utils.ts
```diff
@@ -11,3 +13,45 @@ export const IS_BROWSER = typeof document !== "undefined";
...
+    } else {
+      // apply for src
+      if (
+        props.src &&
+        // do not apply the for assets that are already targetting a fresh special handling
+        !props.src.startsWith(INTERNAL_PREFIX) &&
+        // Only apply for assets that is referenced from the static folder, i.e path starting by '/'
+        props.src.startsWith("/")
+      ) {
+        (vnode.props as HTMLImageElement).src = asset(props.src);
```
thegaryroberts: You already have the typed props constant available.

src/runtime/utils.ts
```diff
@@ -11,3 +13,43 @@ export const IS_BROWSER = typeof document !== "undefined";
...
+    sections[0] = asset(sections[0]);
+    return sections.join(" ");
+  }).join(", ");
+}
+
+export function assetHashingHook(vnode: VNode) {
+  if (vnode.type === "img" || vnode.type === "source") {
+    const props = (vnode.props as HTMLImageElement);
+    // deno-lint-ignore no-explicit-any
+    if ((props as any)["data-no-auto-hashing"]) return;
```
thegaryroberts: You should be able to get rid of the explicit any with this:
 
```
    if (props.dataset.noAutoHashing) return;
```
As per:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset

thegaryroberts: I've not managed to test this at runtime though, but it does stop the typescript error.

lucacasonato: That probably doesn't work, considering this is not a "real" `HTMLImageElement` on the server.

lucacasonato (commented):
This looks good, thanks @sylc. I want to do some minor refactors to the `asset` function before landing this, to make this more robust.

tests/main_test.ts
```diff
@@ -309,3 +358,22 @@ Deno.test({
     assertEquals(resp.headers.get("server"), "fresh test server");
   },
 });
+
+Deno.test("applyAssetSrcSet", () => {
```
lucacasonato: Can you move this to `runtime/utils_test.ts`?

src/runtime/utils.ts
```diff
@@ -28,3 +30,43 @@ export function asset(path: string) {
...
+
+    // apply for src
+    if (
+      props.src &&
+      // do not apply the for assets that are already targetting a fresh special handling
+      !props.src.startsWith(INTERNAL_PREFIX) &&
+      // Only apply for assets that is referenced from the static folder, i.e path starting by '/'
+      props.src.startsWith("/")
+    ) {
```
lucacasonato: With the new `asset` changes, this can be simplified to `if (typeof props.src === "string")` I think.

lucacasonato:
The srcset parsing worries me slightly. I'll have to find some more test cases to check that it is robust enough.

sylc:
an image with a src defined as `src="user.png"`  (no leading `/`) will be correctly served by fresh from the static folder but the hook will not apply. I'm not sure if this is just fine for now or a bug or a feature (the `data-no-auto-hashing` could be removed to leverage this. but the difference between the 2 path is very subtle.)
it is ready for another look. thanks.

lucacasonato:
> an image with a src defined as src="user.png" (no leading /) will be correctly served by fresh from the static folder but the hook will not apply.

I think that is fine. We can always improve this in the future. We should be as conservative as possible in the beginning so we don't introduce bugs.

lucacasonato:
@sylc I have some edits in the docs page, that I unfortunately can not push to your fork. Could you enable the "Allow maintainers to push to branch" option on this PR?

sylc:
@lucacasonato  Sorry, Is that what you are referring to? it is already ticked 
image

lucacasonato:
Mhh, weird. I can't push. Must be something on my side. Sorry for the trouble, I'll try figure it out.

lucacasonato:
Oh, seems like it was an issue with my configuration (something to do with `git-lfs`).

lucacasonato:
I have updated the `assetSrcSet` function to be a little more robust.

~~My final concern is that the vnode hooks are not installed client side yet, and `` components rendered from islands will have different `src` paths server side and client side. We thus need to install this vnode hook on the client too.~~ Retracted, I just missed where you do this during review.

lucacasonato (approved):
LGTM! Thanks @sylc. As a follow up we should add a test that the asset mapping in islands also works correctly client side.

sylc:
> As a follow up we should add a test that the asset mapping in islands also works correctly client side.

@lucacasonato I agree, I was thinking to do that after https://github.com/lucacasonato/fresh/pull/197 is merged so that I can leverage the islands test setup in that PR

PR #188, opened 2022-05-14
www: add frontmatter and meta description handling

### LINKED ISSUE #165: www: add meta description to docs page
There is no `<meta name="description" />` on the docs pages. One should be added. The descriptions should be manually added for each docs page in the front matter (I will do this). I don't want automatic generation of the descriptions based on content - it never contains the right info that way.

bartlomieju:
Closes https://github.com/lucacasonato/fresh/issues/165

### DIFF
--- a/www/routes/docs/[...slug].tsx
+++ b/www/routes/docs/[...slug].tsx
@@ -27,7 +27,7 @@ interface Data {
 
 interface Page extends TableOfContentsEntry {
   markdown: string;
-  frontMatterData: Record<string, string>;
+  data: Record<string, unknown>;
 }
 
 export const handler: Handlers<Data> = {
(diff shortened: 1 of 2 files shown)


lucacasonato:
It looks like the docs pages are crashing with this change: https://fresh-wawt210e3rw0.deno.dev/docs/introduction.

bartlomieju:
Fixed now

www/routes/docs/[...slug].tsx
```diff
@@ -27,6 +27,7 @@ interface Data {
 
 interface Page extends TableOfContentsEntry {
   markdown: string;
+  frontMatterData: Record<string, string>;
```
Comment (lucacasonato) [inline www/routes/docs/[...slug].tsx]: Can you rename to just `data`?

Also should really be `unknown` or `Record<string, unknown>` as yaml can have non string values. This is not too important tho.

lucacasonato (approved):
Thanks Bartek! :tada:

PR #192, opened 2022-05-16
refactor: asset caching via query param

lucacasonato:
This updates the asset caching to use a cache busting query string,
rather than a totally seperate path. This makes it safer to apply
caching automatically to things like `` tags, as the path does not
change anymore.

cc @sylc

Ref #184

### DIFF
--- a/src/runtime/utils.ts
+++ b/src/runtime/utils.ts
@@ -20,7 +20,11 @@ export function asset(path: string) {
     }
     url.searchParams.set(ASSET_CACHE_BUST_KEY, __FRSH_BUILD_ID);
     return url.pathname + url.search + url.hash;
-  } catch {
+  } catch (err) {
+    console.warn(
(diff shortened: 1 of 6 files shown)


src/runtime/utils.ts
```diff
@@ -1,13 +1,26 @@
...
+    if (
+      url.protocol !== "https:" || url.host !== "freshassetcache.local" ||
+      url.searchParams.has(ASSET_CACHE_BUST_KEY)
+    ) {
+      return path;
+    }
+    url.searchParams.set(ASSET_CACHE_BUST_KEY, __FRSH_BUILD_ID);
+    return url.pathname + url.search + url.hash;
+  } catch {
+    return path;
```
sylc, line 28: maybe add a `console.warn` otherwise it will be hard for anyone to spot that it is not working as expected.

sylc (commented):
Interesting. LGTM.

Issue #168, opened 2022-04-24
Support nesting island components


lucacasonato:
If an island is nested inside of another island, no special work should be done to re-hydrate the inner component.

mohsen1:
How this should work if children is not serializable?

codemaster138:
Is there any workaround for this until #517 is merged?

PR #197, opened 2022-05-18
Fix: 2 islands on 1 page are revived

### LINKED ISSUE #196: Bug: Hydration fails on > 1 Island
Thought I was losing my mind. Tried writing my own island to try it out, but no matter what I did, it just seemed to never run in the browser.

After testing, it turns out it wasn't my code.  Steps to reproduce:

1. `fresh init testDir`
2. in routes/index.tsx, copy `<Counter start={3} />` so there are 2 instances.
3. `deno task start`

Expected:

Both components should hydrate and be interactive.

Actual:

Only the first one works, the second one is only rendered server-side.  No errors in the console to indicate there's an issue.

--- 

This seems like a super cool project and I'm surprised I hadn't heard about it before. Looking forward to using it once it's production ready! Thanks! :D

sylc:
Fix: https://github.com/lucacasonato/fresh/issues/196

The island end comment node was removed before the call to node.nextSibling, stopping the walk. I have re-ordered the deletion and added a test.


tests/islands_test.ts
```diff
@@ -0,0 +1,69 @@
...
+    }
+    if (!started) {
+      throw new Error("Server didn't start up");
+    }
+
+    // verify the island is revived.
+    const browser = await puppeteer.launch({ args: ["--no-sandbox"] });
+    const page = await browser.newPage();
+
+    await delay(500);
```
lucacasonato: Why a delay?

tests/islands_test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    await t.step("Ensure 2 islands on 1 page are revived", async () => {
+      await page.goto("http://localhost:8000/islands");
+      await delay(500);
```
lucacasonato: Use `page.waitForSelector` instead?

lucacasonato:
Thanks for the fix. I think this should generally work, except in the case of nested islands. Nested islands will not work like you are expecting. In the test with `<Inner>` and `<Outer>`, the inner component is never directly hydrated. Instead it is hydrated as part of `<Outer>`. In this case this results in the correct behavior, but it may not always, as the props that are serialized into the HTML will not be used to revive the `<Inner>` island.

lucacasonato:
I guess this does fix nested islands like I described in #168, so we can probably land this. The only concern is that there is a little bit of inefficiency because we are shipping the props for the `<Inner>` component, even though we are not hydrating it.

sylc:
> I guess this does fix nested islands like I described in #168, so we can probably land this. The only concern is that there is a little bit of inefficiency because we are shipping the props for the `<Inner>` component, even though we are not hydrating it.

This intend of this PR is not really to fix #168 but to fix #196 (the reviving issue when 2 island are in one page). I added to test for the nested island because i though they would be needed in the future anyway and to make sure there was no regression.

nested island still don't work for example with a thin outer as per reported in https://github.com/lucacasonato/fresh/issues/178

if you think, that the nested island test should be added in a PR with a more holistic way around the nested island issue, then they can be removed.

lucacasonato (approved):
LGTM.

I have removed the nested islands test for now.

Thanks for working on this.

PR #209, opened 2022-05-23
Use import maps instead of deps files

lucacasonato:
This commit updates fresh to use import maps instead of deps files in generated projects.


vibegui:
Is this being supported by deno deploy now?

lucacasonato:
Yes. See the deploy preview :)

vibegui:
Great!!

PR #211, opened 2022-05-23
Docs fix typo

antoninbeaufort:
Fix some typos in docs, "middlware" becoming "middleware", an "peform" becoming "perform".

### DIFF
--- a/docs/concepts/middleware.md
+++ b/docs/concepts/middleware.md
@@ -48,7 +48,7 @@ specific first).
 
 For example, take a project with the following routes:
 
-- `routes/_middlware.ts`
+- `routes/_middleware.ts`
 - `routes/index.ts`
 - `routes/admin/_middleware.ts`
 - `routes/admin/index.ts`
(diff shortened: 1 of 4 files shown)


hashrock:
I'm trying to fix this CI error(should be not related in this PR)

https://github.com/lucacasonato/fresh/runs/6586752640?check_suite_focus=true#step:6:22

When I run ci on macOS and ubuntu at the same time, esm.sh appears to return a 403 forbidden error.

lucacasonato:
cc @ije: have you seen this esm.sh issue before?

ije:
Screen Shot 2022-05-27 at 09 30 59

seems it was blocked by Cloudflare Security

ije:
but i never add any Firewall filters in fact, i don't know why

hashrock (commented):
I think your changes are right, but I got an error here:

> resp.headers.set("X-Custom-Header", "Hello");

`resp` type is `Response | Promise<Response>` , maybe we should use `await` at `ctx.render()`?

lucacasonato (approved):
LGTM, thank you @AntoninBeaufort.

@hashrock I addressed the missing `await` now.

@ije Thanks for looking into the esm.sh issue :)

PR #215, opened 2022-05-27
fixing function name in dynamic-routes doc

oldmanfleming:
In this section we are adding a new Greeting Page. I think the function for this page should be named appropriately to the content it is rendering.


lucacasonato (approved):
Thanks!

PR #216, opened 2022-05-28
www: update eta from May 2022 to "very soon"

lucacasonato:


thegaryroberts:
I'm still finding this project stable enough to build good things, even if a little regular effort is required by consumers to keep up to date with the updates (which are very useful!).

Now the future date for stability is a bit less concrete, would it be possible you could add beta/rc tags to help with fixing imports?

If this is an admin headache, a bit of guidance in vendoring would be gratefully received: https://github.com/lucacasonato/fresh/discussions/218

PR #222, opened 2022-06-01
fix: webp image source

Trugamr:
Fix broken image source when serving webp format
image


Trugamr:
Fixed in preview deployment
image

lucacasonato (approved):
Thank you!

PR #229, opened 2022-06-13
Remove curly braces in import for Countdown

CosminHorjea:
Since Countdown is a default export, putting curly braces here will raise an error


lucacasonato (approved):
Thanks!

PR #234, opened 2022-06-14
soften warning

ry:
Fresh is already used in production on deno.land - the warning doesn't need to be so stark
https://news.ycombinator.com/item?id=31737833


lucacasonato (commented):
Do you want to change this warning too: https://fresh.deno.dev/#getting-started ?

lucacasonato (approved):

PR #242, opened 2022-06-16
www: use `tabular-nums` on Counter count

boyum:
`font-variant-numeric: tabular-nums` makes each number glyph take up the same amount of space, which in turn makes the add and subtract buttons move less about when altering the count. This has a slight accessibility benefit, but adds some extra whitespace between some combinations of numbers when the count is 10 or above.

Before: 

After:

This change is purely for improving(?) UX on the Fresh landing page and I won't be offended if it's not the kind of submissions you consider adding. Thanks anyways for making such a cool framework! Can't wait to see its future.

### DIFF
--- a/www/islands/Counter.tsx
+++ b/www/islands/Counter.tsx
@@ -22,7 +22,7 @@ export default function Counter(props: CounterProps) {
       >
         <IconMinus />
       </RoundedButton>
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
Thank you!

PR #245, opened 2022-06-17
test: add test to ensure asset mapping in islands also works correctly client side

sylc:
Follow-up of https://github.com/lucacasonato/fresh/pull/184 as per https://github.com/lucacasonato/fresh/pull/184#pullrequestreview-978959802

Add a test to ensure an image 'hashed' src path in an island is revived correctly


lucacasonato (approved):
Thank you.

PR #267, opened 2022-06-21
refactor: dev mode also depends on DENO_ENV variable

dogukanakkaya:
Check for `Deno.env.get("DENO_ENV")` for prod mode instead of depending only to `Deno.env.get("DENO_DEPLOYMENT_ID")`
So we can just add:
```
DENO_ENV=prod
```
Also this should be solution for: https://github.com/lucacasonato/fresh/issues/84


lucacasonato (commented):
Not super in favor of this, because really if you want to use Fresh in prod mode, you should set a unique ID based on the "version" of the application to `DENO_DEPLOYMENT_ID`. If you do not do this, caching does not work as efficiently as possible.

PR #270, opened 2022-06-22
Replace the lemon squash illustration with SVG

### LINKED ISSUE #246: Just a very small note of "original.jpeg" binary file
https://github.com/lucacasonato/fresh/blob/main/www/static/illustration/original.jpeg

For now, the image is used on the front page and it's a .jpeg...!
This means when the user agent uses some weird `background-color`, it's kind look weird, too...!

I just one to add a `// TODO: ` somewhere, so...!

Please todo this...?
Great project, btw...!

Screenshot from my mac using the `plash` app

hashrock:
Close #246 

old: 

image

new: 

image

We don't need multiple images for different screens anymore :)

image

I don't think we need dark mode support yet, but I did just in case.


lucacasonato (approved):
LGTM, thanks!

Suya1671:
> Close #246
> 
> old:
> 
> image
> 
> new:
> 
> image
> 
> We don't need multiple images for different screens anymore :)
> 
> image
> 
> I don't think we need dark mode support yet, but I did just in case.

I liked the hand-drawn aesthetic, I'll miss it

PR #271, opened 2022-06-22
www: fix footer to the bottom when content is short

magurotuna:
This commit fixes the footers to the bottom of the page when the content is short relative to the screen height.

Currently the footers are floating up if the screen height is big, as seen in the following screenshot:

image

With this commit, they will be fixed to the bottom of the page, which looks even nicer IMO:

image


lucacasonato (approved):
Thanks

PR #276, opened 2022-06-22
test: add test case of fresh-init

shinshin86:
I added a test case for an invalid `fresh-init` command.
What do you think?

### DIFF
--- a/tests/cli_test.ts
+++ b/tests/cli_test.ts
@@ -292,65 +292,60 @@ Deno.test({
   sanitizeResources: false,
 });
 
-Deno.test({
-  name: "fresh-init error(help)",
-  async fn(t) {
-    const includeText = "fresh-init";
+Deno.test("fresh-init error(help)", async function (t) {
+  const includeText = "fresh-init";
 
-    await t.step(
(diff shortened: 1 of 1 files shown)


tests/cli_test.ts
```diff
@@ -290,3 +291,66 @@ Deno.test({
...
+
+        const rawError = await cliProcess.stderrOutput();
+        const errorString = new TextDecoder().decode(rawError);
+
+        assertStringIncludes(errorString, includeText);
+      },
+    );
+  },
+  sanitizeOps: false,
+  sanitizeResources: false,
```
lucacasonato: is this needed for this test?

lucacasonato (commented):
Thanks!

tests/cli_test.ts
shinshin86: @lucacasonato Thanks for the code review!
Oh sorry,  this option was not necessary for this test.

I see that you have fixed this code on your end.
https://github.com/lucacasonato/fresh/pull/276/commits/386a8e3589175112f517330bc61e119c82e657cd

Thank you very much.

Issue #252, opened 2022-06-18
Button presses should not trigger zoom


oaustegard:
On the homepage at https://fresh.deno.dev/, viewed in Chrome on iOS iPhone 12, when tapping either of the Example’s plus or minus buttons repeatedly, if your tap speed hits the double-tap threshold the screen zooms unexpectedly to the width of the current number and the button you pressed. On zoom in the two increment/decrement presses are counted, then the zoom happens, on zoom out the zoom out happens first and “eats” the increment/decrement. 

Both behaviors are surprising to a user and deserve a more consistent behavior.

lucacasonato:
I'd be happy to take a PR to fix this. It seems like some oddity in Apple's Safari.

TommyDew42:
Lemme take this up! 🚀

TommyDew42:
@lucacasonato 

I am testing the change in my mobile device. Does fresh server by default listen to network request too? I don't seem to be able to connect to my local server from my mobile device.

lucacasonato:
@TommyDew42 It binds to `localhost` by default I think. You may need to explicitly bind to `hostname: "0.0.0.0"` (you can specify this as an option to `start` in `//www/main.ts`.

lino-levan:
This should probably be closed since #277 got merged.

iuioiua:
Closing, as this has been completed.

PR #277, opened 2022-06-22
make RoundedButton not zoom when double click in mobile devices

TommyDew42:
### What this PR is for?
Fix issue #252.

Stop the homepage count +/- buttons from zooming in/out when mobile users double clicks them.

### How it's tested?
Tested in my mobile device chrome & safari browser.
Before:
<video src="https://user-images.githubusercontent.com/46992350/175068913-916e3ba4-d435-41a2-afb8-c2fdcf9095ff.mov">

After:
<video src="https://user-images.githubusercontent.com/46992350/175068941-dc82b4af-2b62-40dc-bc15-8f1b0649380b.mov">


www/components/Button.tsx
```diff
@@ -5,6 +5,9 @@ import { tw } from "../utils/twind.ts";
 export function RoundedButton(props: h.JSX.HTMLAttributes<HTMLButtonElement>) {
   return (
     <button
+      style={{
+        touchAction: 'manipulation',
+      }}
```
TommyDew42, line 10: Seems to me `twind` has not supported the [Touch Action](https://tailwindcss.com/docs/touch-action) classes from tailwind 3 yet?

lucacasonato: Seems like it indeed.

lucacasonato (approved):
Thanks!

PR #278, opened 2022-06-22
init: add favicon.ico to static

satotake:
Browswers try to get `/favicon.ico` by default.
This request enters `[name].tsx` and returns greet unexpectedly.
To guard it, favicon is fetched from `https://fresh.deno.dev` and save into `static`.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -233,6 +233,13 @@ await Deno.writeTextFile(
   STATIC_LOGO,
 );
 
+const faviconArrayBuffer = await fetch("https://fresh.deno.dev/favicon.ico")
+  .then((d) => d.arrayBuffer());
+await Deno.writeFile(
+  join(directory, "static", "favicon.ico"),
+  new Uint8Array(faviconArrayBuffer),
+);
+
 let MAIN_TS = `/// <reference no-default-lib="true" />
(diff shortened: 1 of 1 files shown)


init.ts
```diff
@@ -233,6 +233,13 @@ await Deno.writeTextFile(
   STATIC_LOGO,
 );
 
+const faviconArrayBuffer = await fetch("https://fresh.deno.dev/favicon.ico")
+  .then((d) => d.arrayBuffer());
+await Deno.writeFile(
+  join(directory, "static", "favicon.ico"),
+  new Uint8Array(faviconArrayBuffer),
+);
```
lucacasonato: Could you wrap this in a try catch, so if there is a network error it will just silently ignore the favicon? Makes it a little more resilient to slow networks.

satotake: done! thanks

lucacasonato (approved):

lucacasonato:
Thank you!

thegaryroberts:
@satotake thanks for this, as a bonus it makes it easy to understand where to put images in a new project.

aashutoshrathi:
Hey! Is there a way I can use png as a favicon?

PR #280, opened 2022-06-23
test: remove tmp dir

shinshin86:
I was looking at the test code and noticed one thing, so I am sending you a PR.

The `Deno.makeTempDir()` creates a temporary directory for testing, but maybe we need to handle deleting this directory.
My understanding is that the directory is increasing each time the test is run on the PC.
(Also, the Deno documentation says `It is the caller's responsibility to remove the directory when no longer needed.`)

So I added a process to remove temporary directories.

If I have missed something and there is no need to remove it in the first place, you can close this PR👍

Thank you.


tests/cli_test.ts
```diff
@@ -286,6 +288,8 @@ Deno.test({
       await lines.cancel();
       serverProcess.kill("SIGTERM");
       serverProcess.close();
+
+      await Deno.remove(tmpDirName, { recursive: true });
```
lucacasonato: Please do this outside of the test step (in the top level test itself). Just move the code 1 line down.

shinshin86:
@lucacasonato Thanks for your comment!
I fixed the code.

lucacasonato (approved):
Thanks.

hashrock:
This PR looks good, but seems to cause problems with CI on Windows.

```
fresh-init => ./tests/cli_test.ts:37:6
error: Error: The process cannot access the file because it is being used by another process. (os error 32), remove 'C:\Users\RUNNER~1\AppData\Local\Temp\e4f7d61c'
error: Test failed
    await Deno.remove(tmpDirName, { recursive: true });
    ^
    at async Object.remove (deno:runtime/js/30_fs.js:161:5)
    at async fn (file:///D:/a/fresh/fresh/tests/cli_test.ts:171:5)

fresh-init --twind --vscode => ./tests/cli_test.ts:177:6
error: Error: The process cannot access the file because it is being used by another process. (os error 32), remove 'C:\Users\RUNNER~1\AppData\Local\Temp\25fc6cf1'
    await Deno.remove(tmpDirName, { recursive: true });
    ^
    at async Object.remove (deno:runtime/js/30_fs.js:161:5)
    at async fn (file:///D:/a/fresh/fresh/tests/cli_test.ts:326:5)
```

shinshin86:
@hashrock Thanks for your comment!
I have tried on a Windows machine and was able to reproduce this issue.
I will try to investigate the cause.

shinshin86:
@lucacasonato @hashrock 
I have fix the issue that was occurring with Windows on CI.

The cause of the error was that when `Deno.remove` was performed, the `serverProcess` that had been started immediately before was still holding onto the process related to the directory to be removed.
And on Windows, it seemed to take a little time for this process to be released.
Therefore, i modified it so that `Deno.remove` is executed after `delay` is executed and the process is completely released.

hashrock:
Thank you! Looks good, but I'll look into a way to do it without delay just in case. If not, I'll merge it in.

kt3k (approved):

PR #285, opened 2022-06-25
Fix a few small typos

nathanjsharpe:
I worked through the documentation recently and found a few small typos.

Thank you for an interesting project that makes things fun again!


lucacasonato (approved):
Thank you.

PR #288, opened 2022-06-25
feat: allow lang configuration

benjamingeets:
In render function, can set HTML lang using `ctx.lang = "fr"`, by default it remains `en`


bartlomieju:
Looks good! Can you please add a test case @benjamingeets?

benjamingeets:
(hope i've done it the right way, honestly I've never done testing before😅)

lucacasonato (approved):
Thank you

PR #295, opened 2022-06-28
docs: fix typos

jimt:


jimt:
It was also tempting to move the page to `architecture` without the extra `h`.

lucacasonato:
> It was also tempting to move the page to architecture without the extra h.

Aw crap. I'll need to fix that. But we'll need a redirect from `architechture` to `architecture`

lucacasonato (approved):
Thanks

PR #303, opened 2022-06-29
docs: fix typos

rwaskiewicz:
this commit fixes a few small typos in the docs. I didn't try to editorialize anything here, just a few single word replacements that I picked up on while going through the docs

### DIFF
--- a/docs/concepts/data-fetching.md
+++ b/docs/concepts/data-fetching.md
@@ -3,10 +3,10 @@ description: |
   Data fetching in Fresh happens inside of route handler functions. These can pass route data to the page via page props.
 ---
 
-Server side data fetching in Fresh is acomplished through asynchronous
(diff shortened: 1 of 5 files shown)


lucacasonato (approved):
Thanks!

santuan:
Hi!
Could it be that in the "Running locally" page at the docs.
There is a _comma_ instead of a _dot_ in the `--watch=static` flag
They both work, so i'm not sure.

With Comma (_currently in the doc_)
`deno run --allow-net --allow-read --allow-env --allow-run --watch=static/,routes/ main.ts`

With a dot
`deno run --allow-net --allow-read --allow-env --allow-run --watch=static/.routes/ main.ts`


Thanks for this project! I already love it :)

lucacasonato:
@santuan No, the comma is correct. It's watching whatever --watch watches by default, and then `static/` and `routes/` directories.

PR #320, opened 2022-06-29
Update README.md

PyBaker:
Fixed some grammar :).


lucacasonato:
The current wording is already grammatically correct. I prefer the current wording.

Thank you though

PR #331, opened 2022-06-29
Update data-fetching.md

TomLorenzi:
Fixing this link to a non existing routes, looks like the best suiting route for it ?


lucacasonato:
Thanks, but closed in favor of #400.

PR #333, opened 2022-06-29
Update adding-interactivity.md

umfy:
The previous solution did not clear interval after the time has passed.
The "now" variable was not being updated in useEffect().


lucacasonato (approved):
Thank you!

Issue #350, opened 2022-06-30
Add a shared component living in `components/` to the starter project


lucacasonato:
See https://github.com/denoland/fresh/discussions/326

asrvd:
working on it

briancbarrow:
@lucacasonato with #353 being merged, i'm assuming we can close this now, right?

iuioiua:
Closing as completed.

PR #351, opened 2022-07-01
Update example: Add code element around file location; Add a a shared component #350

omar2205:
Add a code element around file location
image

Add a shared component inside `components` #350


lucacasonato:
Sorry, this has gone stale. Closing.

Issue #342, opened 2022-06-30
rename example command dir from "my-app" to "deno-fresh-demo"


ry:

lucacasonato:
It has been renamed to `my-project`.

PR #353, opened 2022-07-01
feat: add component/Button in starter project

asrvd:
added a new Button component as requested in issue #350 , has been tested.

@lucacasonato shall I update the start command to watch the `components/` folder too ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You can scaffold a new project by running the Fresh init script. To scaffold a
 project in the `my-project` folder, run the following:
 
 ```sh
-deno run -A -r https://fresh.deno.dev my-project
+deno run -A -r https://fresh.deno.dev deno-fresh-demo
 ```
 
 Then navigate to the newly created project folder:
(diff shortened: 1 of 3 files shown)


asrvd:
updated example project dir name to `deno-fresh-demo` following #342

santosant (approved):

lucacasonato (commented):
Thanks, but this is a kind of weird if a user chooses not to use `twind` (because then the wrapper does nothing). Maybe add the `disabled` stuff to the `Button` component too?

asrvd:
> Thanks, but this is a kind of weird if a user chooses not to use `twind` (because then the wrapper does nothing). Maybe add the `disabled` stuff to the `Button` component too?

yes, will do 👍🏼

asrvd:
@lucacasonato the `disabled` thing is already being passed as a prop, the import was unnecessary, removed it.

lucacasonato (approved):
I made some tweaks, but LGTM now.

Issue #364, opened 2022-07-02
Cannot find module 'preact' or its corresponding type declarations


andrewasd:
Hi, im trying to run the code in this page after following all the docs but i don't know how to install the preact module
https://fresh.deno.dev/docs/getting-started/create-a-route
![image](https://user-images.githubusercontent.com/44606697/176985854-1f0715ef-e507-42d7-bc8f-fade4486371c.png)

excessivecoding:
The Preact module should import automatically when starting the project with `deno task start`.

Deno automatically cache all dependencies so there is no "installation" to do.

I think you are just missing some VSCode settings (I'm assuming from the screenshot that this is VSCode):
1. Make sure to install the Deno VSCode extension
2. Enable the Deno extension in your VSCode settings or by creating a `.vscode/settings.json` file with:
```json
{
  "deno.enable": true
}
```
3. Restart VSCode to make sure the settings are loaded correctly

fdemir:
> 

These instructions should be added to the documentation. So many people have issues like that.

lucacasonato:
The required VS Code files are now added in the init script.

nhasdarjian:
I was still having this problem, but it resolved when I loaded the project folder itself in VSCode and not a parent folder. Just a tip for anyone else who might still get this error.

teneburu:
I have the same problem along with other errors like 
`'React' refers to a UMD global, but the current file is a module. Consider adding an import instead.`
I just completed the new project script from fresh and i have not modified anything.

I'm using VSCodium last version. The extension is installed and I can see Deno Language Server in Output picking up the configuration file. 

In .vscode/settings.json

`{
  "deno.enable": true,
  "deno.lint": true,
  "editor.defaultFormatter": "denoland.vscode-deno",
  "[typescriptreact]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "[typescript]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "[javascriptreact]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "[javascript]": {
    "editor.defaultFormatter": "denoland.vscode-deno"
  },
  "css.customData": [
    ".vscode/tailwind.json"
  ]
}
`
The selected language is typescriptreact.
I expected it to work after restart but it does not. 

![qw](https://github.com/user-attachments/assets/63572f2d-ec51-4c26-83ee-3da769e09d7f)

PR #367, opened 2022-07-02
feat: add vscode files on init

excessivecoding:
I'm guessing a lot of people are coming to Fresh without any knowledge on Deno or never used Deno before.

This PR will add some VSCode files to ease their way into the framework by recommanding the Deno extension and enable Deno's extension by default when they open the project for the first time.

Related issue: https://github.com/denoland/fresh/issues/364

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -36,6 +36,7 @@ USAGE:
 OPTIONS:
     --force   Overwrite existing files
     --twind   Setup project to use 'twind' for styling
+    --vscode  Setup project for VSCode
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
Maybe should be locked behind a "Do you use VS Code?" prompt.

excessivecoding:
Good call 👍
Updated to behave like the Twind prompt

lucacasonato (approved):
Thanks!

Issue #338, opened 2022-06-30
Document passing a custom app wrapper


jukefr:
I think fresh should allow passing a custom document wrapper in the same manner next does.
See https://github.com/denoland/fresh/blob/8d118a2d3055a7b5e734a9f2af838fdc14fe5c32/src/server/render.tsx#L270.
This would for example allow to set twind classes on the html element to support light/dark color schemes at a system level via media preferences, and also manually (see https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually=)

It would probably need abstracting of this template function some more since it contains parts that you might want to hide from the user (i see some dangerouslySetInnerHTML etc)

To give a bit more of an idea about the context the first thing I tried after generating a project with twind is changing the darkMode to `media` and setting a dark mode background class property on the page, however to handle changing the background of the html element i would need to set the class on there or on the body at least. Currently the only way this would be possible is to make a container that has the same size as body which I find a bit "hacky" and not really intended practice from what I gather from tailwind docs.

It would probably allow for more flexibility outside of this specific use case also.

See https://nextjs.org/docs/advanced-features/custom-document for inspiration maybe.

**EDIT: this appears to actually be possible already although not documented (as far as I could tell) https://github.com/denoland/fresh/commit/04a32f34987fbd59ce646967f216b2cca072a58f**

excessivecoding:
Facing the same issue with dark mode and also repeating element like `title`, etc.

I would love to have a more Remix approach where the DOM is completely exposed rather than the Next.js one with a `_app.jsx`, but this is just an opinion

m4rocks:
+1 would be cool to see this. but it's already kind of possible by using a custom wrapper component for every route

excessivecoding:
There is limitation on what you can do with custom wrapper though. For example changing property on the root `html` tag like `lang` and `dir` property

Edit: seems like it's possible to change `lang` just not `dir` and other html attributes

EccentricVamp:
> Facing the same issue with dark mode and also repeating element like `title`, etc.

Same here. I was excited to see first class support for Twind, but Fresh doesn't seem to allow developers to easily set the CSS class of the `<html>` or `<body>` so no global styles for me.

> I would love to have a more Remix approach where the DOM is completely exposed

I agree. An exposed template also seems like a much more maintainable approach than adding fields to `TemplateOptions` for everything that could possibly be changed in the template.

[...]

PR #380, opened 2022-07-02
[proposal] feat: add overridable root document

excessivecoding:
Hello,

I was tinkering a little with the Fresh codebase. As mentioned in https://github.com/denoland/fresh/issues/338 , I think having the ability to customize the root document of your Fresh app could be a nice feature to add and could be mandatory in some situation.

Example in the Fresh website repo to illustrate (obviously, this is just to illustrate how the PR works):
`./routes/_root.tsx`
```ts
/** @jsx h */
import { h } from "preact";
import { Body, Links, Meta, Scripts, Styles } from "$fresh/server.ts";

export default function Root({ lang }: { lang: string }) { // I kept the `lang` options for retrocompatibility
  return (
    <html lang={lang}>
      <head>
        <meta charSet="UTF-8" />
        <meta http-equiv="X-UA-Compatible" content="IE=edge" />
        <Meta />
		{/* We are now able to customize element of the HTML directly in the component giving full freedom to the end developer */}
		{/* This will be the default title if none is specify in the route <Head /> */}
        <title>Hello from ./routes/_root.tsx</title>
		{/* We can even import global stylesheet from the `static` folder */}
        <link rel="stylesheet" href="/styles/test.css" />
        <Links />
        <Scripts />
        <Styles />

      </head>

      <Body />
    </html>
  );
}

```

The codebase being pretty clean, I was able to make a prototype of the feature copying Remix-style root (https://remix.run/docs/en/v1/guides/migrating-react-router-app#creating-the-root-route) without too much modification.

The PR is pretty rough around the edge and need some clean up but I first wanted to validate the idea if this is something that could be worth pursuing first.

If this is something that is completely out of scope or not wanted, feel free to close the PR.

Happy to hear any feedback !

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -26,12 +26,15 @@ import {
   RenderFunction,
   Route,
   RouteModule,
+  Root,
+  RootModule,
   UnknownPage,
   UnknownPageModule,
 } from "./types.ts";
 import { render as internalRender } from "./render.tsx";
 import { ContentSecurityPolicyDirectives, SELF } from "../runtime/csp.ts";
 import { ASSET_CACHE_BUST_KEY, INTERNAL_PREFIX } from "../runtime/utils.ts";
+import { DefaultRoot } from "./template.tsx"
 
 interface RouterState {
   state: Record<string, unknown>;
@@ -61,6 +64,7 @@ export class ServerContext {
   #app: AppModule;
   #notFound: UnknownPage;
   #error: ErrorPage;
+  #root: Root
 
   constructor(
     routes: Route[],
@@ -72,6 +76,7 @@ export class ServerContext {
     notFound: UnknownPage,
     error: ErrorPage,
     importMapURL: URL,
+    root: Root,
   ) {
     this.#routes = routes;
     this.#islands = islands;
@@ -83,6 +88,7 @@ export class ServerContext {
     this.#error = error;
     this.#bundler = new Bundler(this.#islands, importMapURL);
     this.#dev = typeof Deno.env.get("DENO_DEPLOYMENT_ID") !== "string"; // Env var is only set in prod (on Deploy).
+    this.#root = root;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


EccentricVamp:
Awesome PR! I look forward to it getting merged.

I do have a few questions/suggestions:

Do you think "Layout" or "DefaultTemplate" would be better names for the root component? I know it's called the "document root" in the DOM model, but Fresh seems to abstract most of the DOM away so developers don't really have to deal with or know about it.

Since root is a component shouldn't it go in the components folder and use [pascal case](https://en.wiktionary.org/wiki/Pascal_case) (`Root` instead of `_root`)?

excessivecoding:
I guess Layout could work, but this PR is mainly to allow to expose the DOM abstraction. So root seems to fit better.

The Root is a component that same way 404 page are components. It's still is declared as a "special route" using the `_` to mark it as such. So to keep the same convention, it's better to keep it as `_root`

EccentricVamp:
> this PR is mainly to allow to expose the DOM abstraction

If this is focused on the DOM abstraction and conventions from existing projects then we may also want to consider the name `_document` from Next.js. I know that Remix refers to this as `root`, but Next.js uses the same `_` convention as Fresh and it would be more familiar to a significantly higher number of developers.

> The Root is a component that same way 404 page are components

Thanks, I misunderstood what the components were for. They are definitely both routes, not components.

> It's still is declared as a "special route" using the _ to mark it as such.

Good point. Error pages confused me a bit. After reading through the documentation, I realized that this would be similar to the existing `_middleware` route so the convention makes sense.

excessivecoding:
Open on changing the name to anything that would fit better for the root, `_document` seems to fit quite well as well.

excessivecoding:
It seems like Fresh already have a `routes/_app.tsx` feature which is undocumented and achieve the same thing as this PR.

Closing as it is redundant

related: https://github.com/denoland/fresh/pull/494

PR #386, opened 2022-07-03
docs(form-submissions): Improve style

miso-devel:
thanks for awesome framework !!!

## URL
https://fresh.deno.dev/docs/getting-started/form-submissions

## motivation
The input button was difficult to understand, so I improved it.

## before
スクリーンショット 2022-07-03 17 22 48

## after
スクリーンショット 2022-07-03 17 20 29
スクリーンショット 2022-07-03 17 21 12

### DIFF
--- a/docs/getting-started/form-submissions.md
+++ b/docs/getting-started/form-submissions.md
@@ -34,6 +34,7 @@ server side:
 
 /** @jsx h */
 import { h } from "preact";
+import { tw } from "@twind";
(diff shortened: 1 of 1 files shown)


lucacasonato:
@nisi0929 Thanks, the styling looks good, but I want to keep the code in the docs as simple as possible as to avoid confusing folks about what parts are required for functionality, and what is just for styling.

Maybe we can add a "Include styling" switch in the docs at some point.

I'll close this without a merge for now.

PR #391, opened 2022-07-03
Added hyphen to directions to match the example

Chris-Gillis:


lucacasonato:
Thanks, this is already fixed on main :)

PR #399, opened 2022-07-04
Remove border of counter

roman-koshchei:
Border at counter on site main page doesn't match design!


lucacasonato:
I like it better with the border. The border matches the border on the footer.

PR #412, opened 2022-07-04
Printing real directory name

alpineQ:
Printing real directory name after default project creation


lucacasonato:
Thanks, but closing in favor of #415.

PR #415, opened 2022-07-04
fix: display real value for the directory to cd to after init

dikarel:
| Before | After |
| - | - |
| Screen Shot 2022-07-04 at 3 23 05 PM | Screen Shot 2022-07-04 at 3 22 31 PM |

Fixes the output of the `init` script, so that it instructs the user to `cd` into the directory that the user specifies (rather than a hard-coded `my-project`)

**Why this change?**

It can be confusing to the user to see a prompt to `cd` to `my-project`, when the user specifies a different directory for `init`. It's extra confusing for folks who are newer to programming or are trying to on-board by following tutorials, so I think this might hurt adoption if not fixed.

**Test plan:**

- [x] Run `deno fmt init.ts` to catch lint violations
- [x] Run `deno task test` to catch test failures
- [x] Run `deno run -A init.ts todo-list` to verify that the `cd` message is correct

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -56,11 +56,11 @@ if (flags._.length !== 1) {
   error(help);
 }
 
-let directory = Deno.args[0];
-directory = resolve(directory);
+const unresolvedDirectory = Deno.args[0];
+const resolvedDirectory = resolve(unresolvedDirectory);
 
 try {
-  const dir = [...Deno.readDirSync(directory)];
+  const dir = [...Deno.readDirSync(resolvedDirectory)];
   const isEmpty = dir.length === 0 ||
     dir.length === 1 && dir[0].name === ".git";
   if (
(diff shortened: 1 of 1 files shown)


init.ts
```diff
@@ -83,13 +83,15 @@ const useVSCode = flags.vscode === null
...
-await Deno.mkdir(join(directory, "islands"), { recursive: true });
-await Deno.mkdir(join(directory, "static"), { recursive: true });
+await Deno.mkdir(join(resolvedDirectory, "routes", "api"), { recursive: true });
+await Deno.mkdir(join(resolvedDirectory, "islands"), { recursive: true });
+await Deno.mkdir(join(resolvedDirectory, "static"), { recursive: true });
 if (useVSCode) {
-  await Deno.mkdir(join(directory, ".vscode"), { recursive: true });
+  await Deno.mkdir(join(resolvedDirectory, ".vscode"), { recursive: true });
+}
+if (useTwind) {
```
dikarel, line 92: This re-format came from `deno fmt`. I think the rename of `directory` has caused the original line to exceed a column limit?

init.ts
```diff
@@ -106,7 +108,10 @@ if (useTwind) {
   importMap.imports["twind/"] = "https://esm.sh/twind@0.16.17/";
 }
 const IMPORT_MAP_JSON = JSON.stringify(importMap, null, 2) + "\n";
-await Deno.writeTextFile(join(directory, "import_map.json"), IMPORT_MAP_JSON);
+await Deno.writeTextFile(
```
dikarel, line 111: This re-format also came from `deno fmt`

init.ts
```diff
@@ -370,14 +375,16 @@ const VSCODE_EXTENSIONS = JSON.stringify(vscodeExtensions, null, 2) + "\n";
...
-const manifest = await collect(directory);
-await generate(directory, manifest);
+const manifest = await collect(resolvedDirectory);
+await generate(resolvedDirectory, manifest);
 
+// Specifically print unresolvedDirectory, rather than resolvedDirectory in order to
+// not leak personal info (e.g. `/Users/MyName`)
 console.log("\n%cProject created!", "color: green; font-weight: bold");
-console.log("\`cd my-project\` to enter to the project directory.");
+console.log(`\`cd ${unresolvedDirectory}\` to enter to the project directory.`);
```
dikarel, line 389: **Alternate approach: Why not just print `Deno.args[0]` here?**

I think that would make the concern of reading from command line args less DRY. "reading from command line args" feels like one of those areas where making a mistake would be impactful, so it felt important to optimize for DRY-ness there.

lucacasonato (approved):
Thanks.

PR #416, opened 2022-07-04
♻️  implement constants for HTTP Status Codes

ajf-sa:
make source code more readable 
I get confuse when see http status code as number 😶‍🌫️ 
I hope this RP Accept !


ajf-sa:
sorry for format !  format on save in vscode  make some  [that](https://github.com/denoland/fresh/blob/ae9b8aa6b3345bc94ee60d67d32f4ad20be0e626/src/server/context.ts#L669) 😞

PR #417, opened 2022-07-04
♻️ Using deno Std  HTTP Status Codes

ajf-sa:
I notes the fresh source code is fresh and easy for read but, I get confuse when see http status code as number 😶‍🌫️
I hope this RP Accept !

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -472,7 +472,10 @@ export class ServerContext {
         },
       );
 
-    const errorHandlerRender = genRender(this.#error, Status.InternalServerError);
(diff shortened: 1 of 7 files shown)


lucacasonato:
Can you use https://doc.deno.land/https://deno.land/std@0.146.0/http/http_status.ts/~/Status instead?

ajf-sa:
I see.. that will be nice if we can get Status code in Application  by `import {Status} from "$fresh/server.ts"`

un-versed (approved):

lucacasonato (approved):
Thanks.

PR #423, opened 2022-07-05
fix(init): add import map to vscode config

synecdokey:
Without it, you're met with a bunch of errors on first opening the project

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -351,6 +351,7 @@ await Deno.writeTextFile(
 
 const vscodeSettings = {
   "deno.enable": true,
+  "deno.importMap": "./import_map.json"
 };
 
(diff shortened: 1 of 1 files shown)


lucacasonato (commented):
That's surprising... the editor should discover the import map via the `deno.json` file in the root of the project.

nikableh:
> Without it, you're met with a bunch of errors on first opening the project

Can this be because [the official Deno extension for VSCode](https://marketplace.visualstudio.com/items?itemName=denoland.vscode-deno) is not installed?

synecdokey:
Did another brand new install, and it seemed to work just fine. Might have just been the ts server getting spun somehow, I should've looked more closely because I can't reproduce, but I suspect a TS server running concurrently somehow. (I already had the extension installed, so that wasn't that)

PR #434, opened 2022-07-05
feat: add markdown file from URL example

un-versed:
An example of how to render a Markdown file from a URL


ajf-sa (approved):

examples/markdown-file/import_map.json
```diff
@@ -0,0 +1,11 @@
+{
+  "imports": {
+    "$fresh/": "https://deno.land/x/fresh@1.0.1/",
```
lucacasonato:
```suggestion
    "$fresh/": "../../",
```

To use the in-tree version of fresh.

examples/markdown-file/routes/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const main = tw`py-8 overflow-hidden`;
+  const title = tw`text(4xl gray-900) tracking-tight font-extrabold mt-6`;
+  const body = tw`mt-6`;
+  const html = gfm.render(data?.markdown ?? "");
+
+  return (
+    <>
+      <Head>
+        <title>Markdown</title>
+        <link rel="stylesheet" href={`/gfm.css?build=${__FRSH_BUILD_ID}`} />
```
lucacasonato:
```suggestion
        <link rel="stylesheet" href={asset("/gfm.css")} />
```

Could this work instead?

examples/markdown-file/routes/index.tsx
```diff
@@ -0,0 +1,62 @@
+/** @jsx h */
+/** @jsxFrag Fragment */
+import { Fragment, h } from "preact";
+import { Handlers, PageProps } from "$fresh/server.ts";
+import { frontMatter, gfm } from "../utils/markdown.ts";
+import { tw } from "@twind";
+import { Head } from "https://deno.land/x/fresh@1.0.1/src/runtime/head.ts";
```
lucacasonato: Import from `$fresh/runtime.ts`

examples/markdown-file/routes/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    if (resp.status !== 200) {
+      return ctx.render(null);
+    }
+
+    const markdown = await resp.text();
+    const { content, data } = frontMatter(markdown) as {
+      data: Record<string, string>;
+      content: string;
+    };
```
lucacasonato, line 28: If we are going to support frontmatter, we should probably use a markdown file that actually contains some frontmatter. What about putting a file into this example folder, and then reading it with `Deno.readFileText`?

examples/markdown-file/utils/markdown.ts
```diff
@@ -0,0 +1,6 @@
+export * as gfm from "https://deno.land/x/gfm@0.1.19/mod.ts";
+import "https://esm.sh/prismjs@1.27.0/components/prism-jsx.js?no-check";
```
lucacasonato: Please add `https://esm.sh/prismjs@1.27.0` to the import map, and import `prism/components/prism-jsx.js?no-check` instead.

examples/markdown-file/utils/markdown.ts
```diff
@@ -0,0 +1,6 @@
+export * as gfm from "https://deno.land/x/gfm@0.1.19/mod.ts";
```
lucacasonato: Please add to import map

examples/markdown-file/utils/markdown.ts
```diff
@@ -0,0 +1,6 @@
...
+import "https://esm.sh/prismjs@1.27.0/components/prism-jsx.js?no-check";
+import "https://esm.sh/prismjs@1.27.0/components/prism-typescript.js?no-check";
+import "https://esm.sh/prismjs@1.27.0/components/prism-tsx.js?no-check";
+
+export { parse as frontMatter } from "https://deno.land/x/frontmatter@v0.1.4/mod.ts";
```
lucacasonato, line 5: Use https://deno.land/std/encoding/front_matter.ts instead.

examples/markdown-file/routes/index.tsx
un-versed: My idea for the example was specifically fetching the markdown file from a URL...

What if I remove the frontMatter and use the raw markdown file, and make another route in the same example folder using  `Deno.readFileText` and frontMatter?

We cover 2 usecases in the same example :)

examples/markdown-file/utils/markdown.ts
un-versed: See https://github.com/denoland/fresh/pull/434#discussion_r914914084

examples/markdown-file/routes/index.tsx
lucacasonato: Sounds good!

iuioiua:
Hi, @un-versed. Luca and I have discussed this online. We've decided to introduce a new "Examples" section instead of bringing on example apps and the maintenance that goes along with them. I'll close this PR. However, such an example would still be useful in snippet form.

I've opened #1213 to see if we can get this added to the new examples section. A new PR in snippets form would be very welcomed. Either way, thank you for your efforts 🙂

PR #435, opened 2022-07-05
fix: prevent folks messing with the fresh manifest

lucacasonato:
I am looking at you James.


imjamesb:
Wow that was really fast...

crowlKats (approved):
agreed

aapoalas:
> Wow that was really fast...

We're watching history in the making.

imjamesb:
Oh god the name :joy: can't believe this was done in 8 minutes

PR #439, opened 2022-07-06
Rename my-app to my-project in landing page

xyntechx:


nikableh:
While I agree that what fresh is — is more of a _"project"_ than an _"app"_, I think that it's not too important what this suggested name is. And at the same time, changing such things may potentially cause confusion or break something.

lucacasonato (approved):
Thanks, the docs are written using `my-project`, so lets use that uniformly.

xyntechx:
Awesome! Thanks for the merge :) @synalice apologies for not specifying why I made the change!

nikableh:
> Awesome! Thanks for the merge :) @synalice apologies for not specifying why I made the change!

Oh, I didn't realize the documentation was using `my-project` :sweat_smile:

Issue #257, opened 2022-06-20
Document how to use fresh in a sub-directory (e.g `src/`)


lucacasonato:
### Discussed in https://github.com/lucacasonato/fresh/discussions/253

<div type='discussions-op-text'>

<sup>Originally posted by **o-az** June 18, 2022</sup>
Moved `islands`, `routes`, `fresh.gen.ts`, `dev.ts` and `main.ts` to `src` directory.

Updated `start` task in `deno.json` to: `deno run -A --watch=src/static/,src/routes/,src/islands/ --no-check src/dev.ts`.

Project runs and routes work but this error is consistently logged:
![CleanShot 2022-06-18 at 08 40 01](https://user-images.githubusercontent.com/23618431/174440261-6b6b0b03-9c12-4ecb-a3bb-2a025701e871.png)

I'm trying to avoid this error. Is there a proper way to have a `.src/` directory or is it not possible as of now?</div>

o-az:
I think my problem was not being able to run `fresh manifest` in the cli. So when I made directory changes it didn't pick it up right away. Just curious, is `fresh manifest` supposed to run automatically or do I need to install anything? I do have Deno CLI installed and I scaffolded fresh app using the documented command.

lucacasonato:
The `fresh manifest` command does not exist anymore (the entire fresh CLI has been removed).

o-az:
when making updates like directory structure changes how are we supposed to make fresh recognize those updates?

lucacasonato:
Fresh does it itself now (see //www/dev.ts in this repo). https://fresh.deno.dev/docs/getting-started/create-a-project elaborates on what the dev.ts file does.

greym0uth:
Looking at [this line](https://github.com/denoland/fresh/blob/10094f894728e2074e5c10b678d8328258159310/src/dev/mod.ts#L145) it looks like if your `dev.ts` lives in `src/dev.ts` and your `start` task points there (`deno run -A --watch=src/static/,src/routes/ src/dev.ts`) it'll look for the routes directory relative to that entrypoint's base directory (i.e. `<project_root>/src/routes` instead of `<project_root>/routes`). I tested this with locally w/ dev and it works just fine but there may be more to consider for deployments.

greym0uth:
The error seen look specifically related to invalid MIME type configuration in server 
[...]

PR #456, opened 2022-07-06
docs: add src directory guide to create-a-project

greym0uth:
Creates documentation for #257 

This adds a guide on using the `src`, or any other, directory for code, routes, islands, etc. instead of at the root of the project.


lucacasonato (commented):
I don't think this is something we should add to the "Getting Started". A page in concepts would be better I think.

The "Getting Started" guide should have as few decision points for devs as possible.

alirani:
Hello I tried following the steps you documented, but fresh.gen.ts generate `import config from "./deno.json" assert { type: "json" };` which is an error since deno.json is at the root and not in /src directory

solofeed:
Same problem as for @alirani

iuioiua:
Sorry for the delay, @greym0uth. Could you please update and test the doc? Note: imports now live within `deno.json` for new projects instead of `import_map.json`.

iuioiua:
I am closing this in favour of #1286.

Issue #178, opened 2022-04-26
Fresh breaks when Island returns nested Island component directly


vibegui:
Apparently, when an Island returns another Island directly, instead of wrapped
in an element such as div, fresh breaks with this in the console:

```
main.js:1 Uncaught TypeError: Cannot read properties of undefined (reading 'nextSibling')
    at h (main.js:1:99)
    at n (main.js:1:659)
    at n (main.js:1:690)
    at n (main.js:1:717)
    at n (main.js:1:717)
    at S (main.js:1:722)
    at (index):4:1
```

You can find a reproduction here: https://github.com/FutureDrivenDev/fresh-nested-islands/blob/main/islands/Outer.tsx#L10

lucacasonato:
Duplicate of #168 maybe?

vibegui:
Yes! Will close in favour of #168

igorbrasileiro:
> Duplicate of #168 maybe?

@lucacasonato I'm facing this problem without nested islands.
If you get this page document (view-source:https://deco-sites-fashion-fb34r6n15tg0.deno.dev/feminino?filter.category-1=feminino&filter.category-3=calca) exists an island with this comment `<!--frsh-sendeventbutton:9-->` related to this island (https://github.com/deco-sites/fashion/blob/c4672dfa5b93cd65ae74e27d6181b9eae8ccfcf1/components/product/ProductCard.tsx#L97-L115)

Something interesting is that the HTML tree is
```xml
<div>
  <ul />
  <a />
</div>
```

But the dom after fresh revive, be like this:
```html
<div>
   <a> // I don't know which anchor is this
      <ul />
   </a>
   <a /> // the same a from the HTML
</div>
```

You can also test here: https://deco-sites-fashion-fb34r6n15tg0.deno.dev/live/previews/deco-sites/fashion/sections/ProductShelf.tsx?props=eyJwcm9kdWN0cyI6eyJfX3Jlc29sdmVUeXBlIjoiZGVjby1zaXRlcy9zdGQvZnVuY3Rpb25zL3Z0ZXhQcm9kdWN0TGlzdC50cyIsInF1ZXJ5IjoiY2FsY2EifSwidGl0bGUiOiJ0aXRsZSJ9

luanfonceca:
> > Duplicate of #168 maybe?
> 
[...]

PR #459, opened 2022-07-06
feat: support nested islands (#168)

michalvavra:
Collects children of the island (#168) and top level island when nested (#178).

I've added back tests removed in c8afee9.

### DIFF
--- a/src/runtime/main.ts
+++ b/src/runtime/main.ts
@@ -30,17 +30,24 @@ const ISLAND_PROPS: any[] = JSON.parse(
   ISLAND_PROPS_COMPONENT?.textContent ?? "[]",
 );
 
+function matchFreshTag(node: Node) {
+  return node.nodeType === 8 &&
+    (node as Comment).data.match(/^\s*frsh-(.*)\s*$/);
+}
+
 export function revive(islands: Record<string, ComponentType>) {
   function walk(node: Node | null) {
-    const tag = node!.nodeType === 8 &&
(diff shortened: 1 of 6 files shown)


blakechambers:
Nice change.  I was looking into this over the weekend and got partway through a fix for this – but I took a bit of a different path that I thought worth mentioning.  

In my change, I used a context object to track whether the code is in an island or not.  You can see it here: 

https://github.com/denoland/fresh/compare/main...blakechambers:fresh:support-nested-islands#diff-b8e94e8d0640e34d87e2781f4a56590cec48f7704c8d8b72f532b7b750fc2da5R317-R335

This probably comes with a different set of pros and cons, but one thing a context allows is exposing a hook to detect whether you are in an island or not.  If that's something that feels useful, that could be another option.

lucacasonato (commented):
I don't think this does what you think it does. The inner island is going to get hydrated by preact first, at which point it will start controlling the DOM elements.

Then the parent island gets hydrated by preact, which means it will take control over the entire DOM tree underneath the parent island (including the inner island). This means that the inner and outer island are now fighting over who gets to render the inner DOM nodes.

That obviously isn't ideal :)

lucacasonato:
I think @blakechambers's approach is probably more reliable, because we can decide on the server to not hydrate an inner island at all. (I can't confirm this, because the git diff link you shared is not working).

blakechambers:
@lucacasonato I'll clean up my branch and share here soon 👍

blakechambers:
@lucacasonato I've updated my branch and submitted it as PR https://github.com/denoland/fresh/pull/517.

PR #468, opened 2022-07-07
feat: use `PORT` environment variable to change port

m4rocks:
This can be useful when deploying to other providers.

Also more convenient for projects with multiple developers with port preferences


fdemir (approved):

lucacasonato (commented):
This does not do what you think it does. It will still listen on port 8000 only, but it will now print that it's listening on a different port.

m4rocks:
damn i'm stupid. i didn't notice it was inside console.log. thanks for pointing out

src/server/mod.ts
```diff
@@ -50,10 +50,21 @@ export interface Manifest {
 
 export { ServerContext };
 
+const defaultPort = parseInt(Deno.env.get("PORT") || "") || 8000;
```
lino-levan: Couldn't this be written more simply as:

```javascript
const defaultPort = parseInt(Deno.env.get("PORT") || "8000");
```

iuioiua:
This seems like a useful addition. Are you able to add a simple test?

src/server/mod.ts
```diff
@@ -60,9 +60,11 @@ export interface DenoConfig {
 
 export { ServerContext };
 
+const port = parseInt(Deno.env.get("PORT") || "8000");
```
iuioiua: Nit: I think it's fine to make this inline.

iuioiua:
Hi @notangelmario, we're happy to merge this once a test is added.

m4rocks:
sure, will get on it asap

iuioiua (approved):
I've added a test. LGTM now!

marvinhagemeister (approved):

PR #470, opened 2022-07-07
feat: enable deno lint & point editor default format to deno

excessivecoding:
Enable Deno lint & format on save for VSCode.

I wonder if it's better to keep those enable by default or if we should ask via a prompt to enable it.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -356,6 +356,8 @@ await Deno.writeTextFile(
 
 const vscodeSettings = {
   "deno.enable": true,
+  "deno.lint": true,
+  "editor.defaultFormatter": "denoland.vscode-deno",
 };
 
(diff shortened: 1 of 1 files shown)


init.ts
```diff
@@ -356,6 +356,9 @@ await Deno.writeTextFile(
 
 const vscodeSettings = {
   "deno.enable": true,
+  "deno.lint": true,
+  "editor.formatOnSave": true,
```
nikableh: I would leave this up to the developer to decide if he wants to format on save. That's too opinionated. The rest seems pretty good.

excessivecoding: kind of agree with you on this, I'm going to remove it for now

nikableh (approved):

lucacasonato (approved):

lucacasonato:
Thanks

Issue #371, opened 2022-07-02
How to add a CSS file to all pages as the default header ?


razvn:
Hi.

New to Fresh and new to Preact so maybe the question is trivial but, I can't figure out how to customized the main html file in order to include a css valid for all site. 

Thanks

ajf-sa:
make components as layout to wrap the route with children in fresh some thing like this 

```typescript

// componets/layout/container.tsx

/** @jsx h */
/** @jsxFrag Fragment */
import { ComponentChildren, Fragment, h } from "preact";
import { tw } from "@twind";
import { Head } from "https://deno.land/x/fresh@1.0.0/runtime.ts";
import Footer from "../../components/Footer.tsx";

export type Props = {
  children: ComponentChildren;
  title?: string;
  name?: string;
  description?: string;
};

export const Container = ({ children, ...customMeta }: Props) => {
  return (
    <>
      <div style={{ minHeight: "100vh" }}>
        <Seo {...customMeta} />
        <div className="container">{children}</div>
        <Footer />
      </div>
    </>
  );
};

const Seo = ({ ...customMeta }) => {
  const meta = {
    title: " بحوث عروض برمجة",
    description: "بحوث عروض برمجة تصميم تصوير مونتاج",
    type: "website",
    ...customMeta,
  };

  return (
    <Head>
      <title>{meta.title}</title>
      <meta content={meta.description} name="description" />
      <link rel="icon" href="/favicon.ico" />
 {/* 
    add javascript css what ever you want
  */}
    </Head>
  );
};

```
and in routes you can use container components to warp as layout 
```typescript

// routes/index.tsx

import { Container } from "../components/layouts/container.tsx";
export default function Home(props: PageProps<H>) {
  const meta = {
    title: "Some Meta Title",
    description: "I am a description, and I can create multiple tags",
    canonical: "http://example.com/path/to/page",
    meta: {
      charset: "utf-8",
      name: {
        keywords: "react,meta,document,html,tags",
      },
    },
  };
  return (
    <>
      <Container {...meta}>
       
         <h1>fresh is really fresh</h1>
      
      </Container>
    </>
  );
}


```
I hope this help you

razvn:
Thank you very helpful.

HaQadosch:
When I try the example, `<Head>` is being ignored. 🤷🏻 
It does work if I import it this way
```js
import { Head } from "$fresh/runtime.ts";
```

lucacasonato:
[...]

PR #476, opened 2022-07-07
[proposal] feat: Custom Stylesheet injection

Dhanielk:
I think there were multiple requests like #371 to have ability to include custom stylesheets into `head` element of root html. I don't think we need full ability to override whole root template (like #380), but have ability to inject stylesheets would suffices.

With this approach we could only modify render function in `main.ts` to have stylesheets included on every page.

Example:
```typescript
...
function render(ctx: RenderContext, render: InnerRenderFunction) {
  ...
  ctx.styleSheets.push(asset('/my.local.css'));
  ctx.styleSheets.push('https://my.remote/file.css'));
}

await start(manifest, { render });
```

I will leave the decision for this feature to you, but it would be enough for my needs to have this change rather than overriding whole root template.

### DIFF
--- a/src/server/render.tsx
+++ b/src/server/render.tsx
@@ -35,6 +35,7 @@ export class RenderContext {
   #id: string;
   #state: Map<string, unknown> = new Map();
   #styles: string[] = [];
+  #styleSheets: string[] = [];
   #url: URL;
   #route: string;
   #lang: string;
@@ -68,6 +69,16 @@ export class RenderContext {
    */
   get styles(): string[] {
     return this.#styles;
+  }
+
+  /**
+   * All of the CSS stylesheet links that should be included into the document.
+   * Adding to this list across multiple renders is supported (even across
+   * suspense!). The CSS stylesheet links will always be inserted on the client
+   * in the order specified here.
+   */
+  get styleSheets(): string[] {
+    return this.#styleSheets;
   }
 
   /** The URL of the page being rendered. */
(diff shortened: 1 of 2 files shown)


harrel56:
Isn't this issue solved by using __app.tsx_ in conjunction with _Head_ component? Injecting style sheets here seems to work nicely, or am I missing something? Usage example taken from [tests](https://github.com/denoland/fresh/blob/main/tests/fixture/routes/_app.tsx):
```typescript
/** @jsx h */
/** @jsxFrag Fragment */
import { Fragment, h } from "preact";
import { Head } from "$fresh/runtime.ts";

export default function App(props: AppProps) {
  return (
    <>
      <Head>
        <meta name="description" content="Hello world!" />
      </Head>
      <props.Component />
    </>
  );
}
```

Dhanielk:
> Isn't this issue solved by using __app.tsx_ in conjunction with _Head_ component? Injecting style sheets here seems to work nicely, or am I missing something? Usage example taken from [tests](https://github.com/denoland/fresh/blob/main/tests/fixture/routes/_app.tsx):
> 
> ```ts
> /** @jsx h */
> /** @jsxFrag Fragment */
> import { Fragment, h } from "preact";
> import { Head } from "$fresh/runtime.ts";
> import { AppProps } from "$fresh/server.ts";
> 
> export default function App(props: AppProps) {
>   return (
>     <>
>       <Head>
>         <meta name="description" content="Hello world!" />
>       </Head>
>       <props.Component />
>     </>
>   );
> }
> ```

That is actually true! I haven't found it anywhere (I might have missed it in documentation) and it seems the others missed it as well. Thanks a lot for input, I will close this PR then.

Issue #251, opened 2022-06-18
Nested Layouts && Error Boundaries?


CanRau:
Hey Luca, fantastic work, thanks a lot 🙏

I was wondering how you feel about nested layouts and error boundaries or if it's already somehow possible?

Nested layouts are great to group certain pages and with error boundaries I can keep errors for example only within the content and have the group layout unaffected, which could be more granular than a general `_404.tsx` route.  Also I can't get a custom `_404.tsx` working, I basically copied from `fixtures` and customized the text but it's still just rendering the text from

```
return new Response("404 Page not found", {
  status: 404,
});
```

Or how am I supposed to trigger a 404?

Uh I guess the [example here](https://github.com/lucacasonato/fresh/blob/main/docs/getting-started/fetching-data.md) can be an alternative to separate error boundaries, though I just realized that this would be too late to send the proper HTTP status code 🤔

CanRau:
Ok so seams like the custom `_404.tsx` only renders if there's no route file at all for the requested URL.

Also, out of curiosity, as you're active in Remix, especially Deno support, as well, what are you're plans with Fresh, vs Remix with much better Deno integration?

lucacasonato:
> Ok so seams like the custom _404.tsx only renders if there's no route file at all for the requested URL.

Yup. I should have probably used error boundaries + nested routes... The infra is actually all there (middlewares), I just need to plumb it through.

I kind of want to get the 1.0 release out soon, and then this is something I can tackle for a 1.1 (if I can make it work backwards compatibly), or a 2.0 release soon.

CanRau:
Great than I'll keep my eyes open 👀

marvinhagemeister:
Closing as recent Fresh versions support nested layouts + error boundaries.

PR #481, opened 2022-07-08
feat: add `ctx.renderNotFound()` to explicitly render 404 template

excessivecoding:
Hello,

This PR will allow to render the `_404.tsx` if the handler return a 404 status code.

This is kind of mandatory when using dynamic routes where the data doesn't exists and a 404 has to be return from the `handler` function.

I filtered out to only invoke this logic when the `Content-Type` is either `text/html` or `null` (defaulting then to `text/html`) to avoid rendering HTML on `application/json` content type (for example on API endpoints).

I see that some discussion has already been made about this issue in https://github.com/denoland/fresh/issues/251 

Feel free to close if you rather use a cleaner implementation.


iamsahilsonawane (approved):
LGTM

byhemechi:
This will fail if the `Content-Type` header includes the charset (e.g. `Content-Type: text/html; charset=UTF-8`). 

It'd also probably be a good idea to write unit tests

excessivecoding:
Good catch, I will add some testing

excessivecoding:
added some tests and fix Content-Type issue.

thanks @byhemechi !

lucacasonato:
I would rather do this differently. For example by allowing users to return `null` from handlers. Re-interpreting already created `Response` objects with 404 status is not something that will scale long term I think.

excessivecoding:
@lucacasonato I guess I can see how this could be a problem long term. Rather than allowing to return `null`, do you think it would be better to add a function `renderNotFound` on the `ctx` of the `handler` to provide a more explicit way of invoking the `_404` template ? If not, I'm fine with your suggestion, just trying to bounce out ideas.

lucacasonato:
That works too

byhemechi:
Another option would be to `throw` a response object for an error page?

excessivecoding:
I like this idea but it might require more work to make it work. I'll go with the `renderNotFound` method and maybe something can be done later in another PR.

excessivecoding:
It should be good to go, we now have a `renderNotFound` method with some tests and I also added some doc on the new method (would love to have feedback if the doc is clear enough).

There seems to be an issue with the CI on format check which is out of the scope of this PR.

lucacasonato:
> Another option would be to throw a response object for an error page?

Not super fond of this path, as `throw` is always very un-optimized in engines and should not really be used in a "hot path". This is also why I don't like how Reacts' suspense works :)

docs/concepts/error-pages.md
```diff
@@ -8,6 +8,8 @@ Fresh supports customizing the `404 Not Found`, and the
 matching route exists, and when a middleware, route handler, or page component
 throws an error respectively.
 
+### 404
```
lucacasonato:
```suggestion
### 404: Not Found
```

docs/concepts/error-pages.md
```diff
@@ -22,6 +24,39 @@ export default function NotFoundPage({ url }: UnknownPageProps) {
 }
 ```
 
+#### Manually render 404
```
lucacasonato:
```suggestion
#### Manually render 404 pages
```

docs/concepts/error-pages.md
```diff
@@ -22,6 +24,39 @@ export default function NotFoundPage({ url }: UnknownPageProps) {
...
+
+The `_404.tsx` file will be invoke when not routes matches the URL. In some
+cases, you would want to manually trigger the rendering of the 404 page with
+`renderNotFound`.
```
lucacasonato:
```suggestion
The `_404.tsx` file will be invoked automatically when no route matches the URL. In some
cases, one needs to manually trigger the rendering of the 404 page, for example when the route did match, but the requested resource does not exist. This can be achieved with
`ctx.renderNotFound`.
```

docs/concepts/error-pages.md
```diff
@@ -22,6 +24,39 @@ export default function NotFoundPage({ url }: UnknownPageProps) {
...
+  return (
+    <article>
+      <h1>{data.blogpost.title}</h1>
+      {/* rest of your page */}
+    </article>
+  );
+}
+```
+
+### 500
```
lucacasonato:
```suggestion
### 500: Internal Server Error
```

lucacasonato (commented):
Some minor comments, but looks great overall! Well done!

excessivecoding:
Thanks for the documentation review, everything is merged 👍

lucacasonato (approved):
Thanks.

lucacasonato:
Can you rebase and resolve the conflicts?

excessivecoding:
synced w/ main branch ✓

PR #482, opened 2022-07-09
Add .gitignore file for default project

xyntechx:


lucacasonato (commented):
I think users can add this themselves if they want to. I think .vscode should be checked in, and the default configuration does not deal with `.env` out of the box, and this suggests it does.

lucacasonato:
That is to say, thanks, but I don't want to merge this now.

xyntechx:
Noted! Thanks!

Issue #491, opened 2022-07-09
Update documentation to include info about _app.tsx


harrel56:
I'm guessing that lack of such information in docs is not intentional. There are many issues being created asking how to modify html template. Docs should contain complete example how to achieve this using special `<Head>` component.

harrel56:
Created PR for that #494

lucacasonato:
The reason this is not documented yet, is that I am contemplating ripping out `_app.tsx` in favor of a more customisable layouts system that let's you layer layouts, and supports changing the HTML document in it's entirety.

hashrock:
Merged documentation about _app.tsx. Hopefully it will be replaced by a Layout system in the near future.

PR #494, opened 2022-07-10
docs: create docs for _app.tsx

harrel56:
I've decided to write some docs on my own (regarding issue #491) . Keep in mind that I have no technical writing experience and treat it more like a sketch. All suggestions are welcome - most importantly, check if everything I wrote is true.

Additionally, I slipped in 1 typo fix


docs/concepts/app-wrapper.md
```diff
@@ -0,0 +1,51 @@
...
+            <Head>
+                <title>Fresh</title>
+                <link rel="stylesheet" href={asset('style.css')}/>
+            </Head>
+            <body class="bodyClass">
+                <Component/>
+            </body>
+        </html>
+    )
+};
```
EccentricVamp: Adjust code format to stay consistent with the rest of fresh, and remove unused import
```suggestion
// routes/_app.tsx

/** @jsx h */
import { h } from "preact";
import { asset, Head } from "$fresh/runtime.ts";
import { AppProps } from "$fresh/src/server/types.ts";
export default function App({ Component }: AppProps) {
  return (
    <html data-custom="data">
      <Head>
        <title>Fresh</title>
        <link rel="stylesheet" href={asset("style.css")} />
      </Head>
      <body class="bodyClass">
        <Component />
      </body>
    </html>
  );
}
```

docs/concepts/app-wrapper.md
```diff
@@ -0,0 +1,51 @@
...
+import { h } from "preact";
+import { asset, Head } from "$fresh/runtime.ts";
+import { AppProps } from "$fresh/src/server/types.ts";
+
+export default function App({ Component }: AppProps) {
+  return (
+    <html data-custom="data">
+      <Head>
+        <title>Fresh</title>
+        <link rel="stylesheet" href={asset("style.css")} />
```
cdoremus, line 43: There is nothing here on where the CSS file is stored. In the `static` folder? I've seen posts on the discord fresh channel  that a CSS file stored in the `static` folder can ce referenced in the link href attribute without a call to the asset function. Also you should say that the `Component` argument is a placeholder for the route's content.

cdoremus:
You might want the `app-wrapper.md` file to focus on layout (and maybe renamed to `layout.md`) and have `_app.tsx ` as one option. Another option would be to use a Layout component to wrap the app. Both of them could use the Head component to add an html head element  with its child elements to the app.

docs/concepts/app-wrapper.md
harrel56: There is fine documentation about static files and `asset` function in previous chapters, so I feel there is no need for more explanations.

harrel56:
> You might want the app-wrapper.md file to focus on layout (and maybe renamed to layout.md) and have _app.tsx as one option. Another option would be to use a Layout component to wrap the app. Both of them could use the Head component to add an html head element with its child elements to the app.

I'm not sure if splitting this feature into 2 pages is such a good idea. If it's possible, please propose your changes by code suggestions. But all of this doesn't really matter, because it seems this PR is not getting merged. I hope the replacement for `_app.tsx` will get implemented soon, as this feature seems to really needed by users.

ahmetcanaksu:
Thank you I almost given up

lino-levan:
Any update on this getting merged? I think it's pretty clear.

lino-levan:
Not sure if this makes sense to merge given that `_app.tsx` will be removed in favor of layouts (at some distant point in the future). I'd still support it being merged since this is kind of the only way to do it at the moment.

hashrock:
@lino-levan 
I think it would be good to add this until the layout feature comes along. I guess the layout feature is still a ways off.

lino-levan:
Agreed.

hashrock (approved):
LGTM

hashrock:
@lambtron Code example updated to the latest. I have also verified that the description is correct, could you double check?

The following are memo during confirming what has been written: 

> There is also a possibility to modify the document template by using special tags `html`, `Head` or `body`. This can be done in any other Preact component

I tried this. Indeed, I was able to do this from components as written.

> Currently, there is no way of overriding default tags/attributes from provided template.

This is also true. For example, setting `<body class="foo">` in `_app.tsx` and `<body class="bar" data-baz="1">` in `index.tsx` will render `<body class="foo" data-baz="1">`. You can add attribute but not overwrite.

(There is a `<html lang="en">` set before `_app.tsx'`, so `_app.tsx` also cannot override by `lang="ja"`)

lambtron:
LGTM!

PR #499, opened 2022-07-12
chore: update `preact-render-to-string` import with esm.sh `?external` query

ije:
https://esm.sh/preact-render-to-string@5.2.0?external=preact


lucacasonato (approved):
Nice!

PR #512, opened 2022-07-15
feat: add initial plugin support

lucacasonato:
This commit adds initial support for fresh "plugins". These are bundles
of functionality that can be added to a fresh plugin through a simple
addition in the `main.ts` file.

Right now plugins are only able to hook into the render steps. This
means that they can inject styling and scripts. They can not yet hook
into routing.

Things that I think should be possible now:
- styling integrations for twind, unocss, and probably a few other tools

Things that are not possible yet:
- google analytics middleware
- database connectors
- CMS connectors

The routing additions to plugins can be added in a follow-up though.


excessivecoding (approved):

excessivecoding:
Looks pretty awesome already, can't wait to see how much would be possible in the future with this plugin system 🔥

vibegui:
This looks great! And it goes in the direction of this request: https://github.com/denoland/fresh/issues/516

Looking forward to being able to inject middlewares with a line on main.ts too.

PR #513, opened 2022-07-15
feat: twind plugin

lucacasonato:
This PR implements a first party plugin for `twind` integration.

It also switches over `www/` to us this new plugin.


vibegui:
That's great. Question: how would one use twind's shim? I noticed in the other PR you now use the result of the plugin rendering, which today is not possible (so shorthand expansions don't work for me now). Would the call to shim go in `plugins/twind.ts` render()? 

reference for how I'm using it in main right now: https://github.com/deco-cx/live/blob/main/src/main.ts#L28

roj1512:
@lucacasonato sorry for disturbing. can you please take a look into #557? I believe it won't take too much time to fix it.

lucacasonato:
Completely off topic

PR #517, opened 2022-07-16
feat: support nested islands (alternative)

blakechambers:
This branch is an alternate to #459 that adds support for nested islands.  As [discussed](https://github.com/denoland/fresh/pull/459#issuecomment-1179009786), the way this change handles nested islands is to use an island context to track whether the component is within an island or not.  This should address the problems mentioned in https://github.com/denoland/fresh/issues/168 and https://github.com/denoland/fresh/issues/178.

Additionally, this change has a second commit exposing a `useWithinIsland` component which would let components detect whether they are being loaded within an island.  I'm not sure if the addition of this is a good fit for the framework long term.  I could see it being used in cases where a component library (one specifically designed for fresh) would need to do different things on render e.g. modifications to `<head>`.  Since there are likely pros and cons to this, I'm happy to split it into another PR so it can be discussed separately.

Future idea 💡 – I could also see a way that the island context could open a path to get the `<Head>` component [working  in the browser](https://github.com/denoland/fresh/blob/main/src/runtime/head.ts#L15-L18).

### DIFF
--- a/runtime.ts
+++ b/runtime.ts
@@ -1,3 +1,4 @@
 export * from "./src/runtime/utils.ts";
 export * from "./src/runtime/head.ts";
 export * from "./src/runtime/csp.ts";
+export * from "./src/runtime/island.ts";
--- a/src/runtime/island.ts
+++ b/src/runtime/island.ts
@@ -1,0 +1,16 @@
+import { createContext } from "preact";
+import { useContext } from "preact/hooks";
+
+export const ISLAND_CONTEXT = createContext<boolean>(false);
+
+export function useWithinIsland() {
+  const ctx = useContext(ISLAND_CONTEXT);
+
+  if (ctx === undefined) {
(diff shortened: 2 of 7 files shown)


lucacasonato (commented):
Great idea with the context! I'll give this a proper review on Monday.

blakechambers:
@lucacasonato I noticed [the roadmap](https://github.com/denoland/fresh/issues/563) mentioned supporting <head> anywhere in the tree.  

I feel like the context approach (from this PR) to track island during render could still help to accomplish that goal, especially as a way to support indifferent usage of head between server side and client side.

However, in light of the roadmap goals, I was wondering if this PR needs any changes?

marvinhagemeister:
Thanks for the PR! Apologies that we couldn't get it over the finish line. In the end we went with the approach in #1285 to support nested islands.

PR #528, opened 2022-07-20
docs: grammar fix

angellk11:
Changed "it's props" to "its props" to indicate possession instead of a contraction.


lucacasonato (approved):
Thanks

PR #529, opened 2022-07-21
fix: change semver link

justjavac:


lucacasonato:
Done in #561. Thanks though.

PR #536, opened 2022-07-23
feat: parallelize all writes in init script

ZackarySantana:
Small optimization that saves time for machines with slow disk writes.

If parallelizing isn't better- perhaps Deno.writeTextFile should be replaced with Deno.writeTextFileSync to remove await calls.


lucacasonato:
Everyone uses SSDs nowadays, so I don't think this will have noticable impact. I'd like to avoid complications like this, because it makes the code more difficult to read.

PR #538, opened 2022-07-23
[wip] feat: add entrypoint for server-side bodyClass param

brycedorn:
Hello! 👋  First of all, love the framework. I'm using it for a project and it's a joy to work with.

I'm running into an issue with SSR & dark mode though, as there's currently no way to modify the server-rendered HTML to send the correctly themed content down to the client (by adding a class, re: #473). Adding this (optional) parameter to `TemplateOptions` in conjunction with [client hints headers](https://web.dev/user-preference-media-features-headers/) would enable correct theme preference on the server and avoid an annoying screen flicker on the client on pageload.

Let me know if this is something that could be added & if you have other suggestions! Thanks.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -29,7 +29,7 @@ import {
   UnknownPage,
   UnknownPageModule,
 } from "./types.ts";
-import { render as internalRender } from "./render.tsx";
+import { render as internalRender, RenderContext } from "./render.tsx";
 import { ContentSecurityPolicyDirectives, SELF } from "../runtime/csp.ts";
 import { ASSET_CACHE_BUST_KEY, INTERNAL_PREFIX } from "../runtime/utils.ts";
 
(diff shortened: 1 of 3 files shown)


ZackarySantana:
@brycedorn Looking at how lang is developed, I don't think the | undefined should be in a bunch of places. maybe just an empty string and then when rendering, only setting it if it isn't an empty string. Also maybe it should have a setter to set per routes just like lang can. 

I wasn't able to replicate using bodyClass in the opts of start, what's the intended usage so I could give it a test?

brycedorn:
Thanks for taking a look @ZackarySantana! I agree with enforcing the string type, updated.

And re: your other question, I had assumed that any context modifications a middleware applies would propagate to the render fn, but these appear to be different contexts entirely. As it stands the renderer appears isolated with no way to dynamically set something like this.

I've implemented what I understand to be the requirement here: adding header logic within `server/context.ts` so that device theme preference can be detected server-side to be applied to set `<body class="dark">`. I wouldn't consider this mergeable anymore but it helps to show what I'm trying to do.

If anyone has an alternative approach that could accomplish the same thing let me know! Would love to see support for this. Additionally this could be applied to support server-side `lang` logic in response to the `Accept-Language` header, etc.

brycedorn:
Updated approach; closing this for cleaner PR history.

Issue #475, opened 2022-07-07
Behavior of `/[dynamic].ext` routes together with `/static.ext`


nikableh:
Proposing a feature of an error (or a warning) being returned to the developer when trying to create a single `/[dynamic].ext` route in the same folder with `/static.ext` routes, or when creating multiple `/[dynamic].ext` routes in the same folder. Doing this may lead to an unexpected behavior and thus should be discouraged.

![Screenshot from 2022-07-08 00-40-35](https://user-images.githubusercontent.com/61655860/177876068-cbee7846-f5b8-4d03-a9bc-df6624cf9ccf.png)
It should be obvious which file will be returned instead of Fresh choosing one over the other.

Related issue: https://github.com/denoland/fresh/issues/461

iamsahilsonawane:
@synalice I think it uses the static routes first over the dynamic ones (priority wise).

It works fine creating a new project and adding static routes on the relative path to dynamic ones


Recording example: https://user-images.githubusercontent.com/47249618/177854292-b182fb60-6600-4e48-a962-7afbdf2a79cc.mp4

nikableh:
@iamsahilsonawane, yes, I know it does choose static over the dynamic and works fine. What I'm saying is that most of the time it wouldn't be a wise decision to put dynamic routes together with static, and the developer should be warned about what's he doing.

iamsahilsonawane:
> @iamsahilsonawane, yes, I know it does choose static over the dynamic and works fine. What I'm saying is that most of the time it wouldn't be a wise decision to put dynamic routes together with static, and the developer should be warned about what's he doing.

Yes, I see, I think there should be lints implemented for this

nhc:
I would be interested in helping with this (and getting involved generally)

@lucacasonato - you tagged good-first-issue 

Would you (or anyone) be able to give any pointers as to where and how you would like the warning to be shown? I'm currently thinking it would show in the terminal output where the developer has started the process?

iamsahilsonawane:
> Would you (or anyone) be able to give any pointers as to where and how you would like the warning to be shown? I'm currently thinking it would show in the terminal output where the developer has started the process?

IMO yes, there should be a warning in the terminal. Also, these warnings should be reflected in the docs.

nhc:
> IMO yes, there should be a warning in the terminal. Also, these warnings should be reflected in the docs.

Ok sure I can have a go a figuring this out.

Issue #498, opened 2022-07-11
Could you add additional css options to the documentation


timscodebase:
I am not a fan of tailwind css.  Could you add instructions for "vanilla CSS"?

lucacasonato:
Coming soon

timscodebase:
Please let me know if I can help.  This is an awesome project! 🏆🔥

davidhousedev:
It'd be great to see Vanilla Extract represented here as well!

timscodebase:
@davidhousedev What do you me buy "Vanilla Extract"?

timscodebase:
@ry @jed Where would I put a styles directory?

ZackarySantana:
@timscodebase I think currently, a styles directory goes in the static folder

davidhousedev:
@timscodebase see here: https://vanilla-extract.style

nhc:
> @timscodebase see here: https://vanilla-extract.style

+1 for this type of thing. We have a home rolled system using Emotion https://emotion.sh/docs/introduction that does similar and its really nice to have CSS in JS 

Why? (top of my head)

- Variables in the objects 
- set mediaqeries breakpoints once, then use a function in the TS objects for the corresponding breakpoints so 

```typescript
import {Breakpoints as bp} from './comp/mediaqueries'

const styles = createStyles({
    main: {
        width: bp([200, 600, 1000, 1200])
    }
)

// Also programatically identiify devices 

import {isMobile, isTablet, isPortrait } from './comp/devices'

if(isMobile && isPortrait) {
    <RenderThisWickedComponent />
}

```

osdiab:
I also use Emotion (although Vanilla Extract looks suuuper nice!) and it's further useful since it's easy to compose styles freely and decouple styling from the actual HTML markup, while having predictable and easy to understand style scoping. Looking forward to proper support for this stuff here, as it's one of my main blockers for using Fresh seriously.

rista404:
Any progress on this one? Would love to see vanilla-extract 🧁 as it goes well with the Fresh 🍋 philosophy.

ValeriaVG:
I've made [@emotion/css](https://emotion.sh/docs/@emotion/css) work with fresh for one of the projects.
It works for islands and server-side rendered pages, but I **didn't test it enough** to call it production ready just yet.

Here's the gist if anyone is interested: https://gist.github.com/ValeriaVG/fae78394feee2ae44b9c58521a33d3e9

Example usage:
```ts
[...]

PR #465, opened 2022-07-07
Add CONTRIBUTING.md

nikableh:
It's not the best one, but might be a good place to start.

PR #539, opened 2022-07-23
#475 Add warnings to terminal for possible route conflicts. 

nhc:
Hi All. 

I'm new to Deno and Fresh so please go easy on me ! This is my first attempt and contributing to this project. As it is quite new I was not able to find a comprehensive guide to contributing (I could help with that too if there are plans?) so I might have got quite a bit wrong. For example I've not written any tests. 

This is my first attempt at solving issue/475 and I might have completely missed the point completely. 

I'm looking for feedback and help on how to improve what I've done. 

What it does do 
- in a function called `routeWarnings` in `dev/mod.ts` it looks at all the routes from the manifest
-  if there is a dynamic route and a static route in the same directory it adds them to a list
-  then`console.logs` in red the following message from the list we've built up.

Screenshot 2022-07-23 at 21 09 39

- if there are conflicts in a number of places it lists these out 

I realise some things will need changing and improving (like I've used `any` as a type on line 153)

```typescript
(totals: any, p) => ({ ...totals, [p.dir]: (totals[p.dir] || 0) + 1 }),
```

But I was hoping for some feedback in terms of if I'm in the right direction etc? 

Also I have these questions 

1. Why is there no .gitignore ? 
2. Are there any plans or style guide / design system for the docs ? 
3. Is there a contributors guide that I'm missing? 
4. Why is my prettier working differently ? It has formatted a few exisiting things differently? Dont we need a prettier config file in the root of Fresh?
 
Thank you all feedback welcomed.

### DIFF
--- a/src/dev/deps.ts
+++ b/src/dev/deps.ts
@@ -4,6 +4,7 @@ export {
   extname,
   fromFileUrl,
   join,
+  parse as parsePath,
   resolve,
   toFileUrl,
 } from "https://deno.land/std@0.128.0/path/mod.ts";
--- a/src/dev/mod.ts
+++ b/src/dev/mod.ts
@@ -3,6 +3,7 @@ import {
   extname,
   fromFileUrl,
   join,
+  parsePath,
   toFileUrl,
   walk,
 } from "./deps.ts";
@@ -84,25 +85,28 @@ export async function generate(directory: string, manifest: Manifest) {
 // This file is automatically updated during development when running \`dev.ts\`.
 
 ${
-    routes.map((file, i) => `import * as $${i} from "./routes${file}";`).join(
-      "\n",
-    )
+    routes
+      .map((file, i) => `import * as $${i} from "./routes${file}";`)
+      .join("\n")
   }
 ${
-    islands.map((file, i) => `import * as $$${i} from "./islands${file}";`)
+    islands
+      .map((file, i) => `import * as $$${i} from "./islands${file}";`)
       .join("\n")
   }
 
 const manifest = {
   routes: {
     ${
-    routes.map((file, i) => `${JSON.stringify(`./routes${file}`)}: $${i},`)
+    routes
+      .map((file, i) => `${JSON.stringify(`./routes${file}`)}: $${i},`)
       .join("\n    ")
   }
   },
   islands: {
     ${
-    islands.map((file, i) => `${JSON.stringify(`./islands${file}`)}: $$${i},`)
+    islands
+      .map((file, i) => `${JSON.stringify(`./islands${file}`)}: $$${i},`)
       .join("\n    ")
   }
   },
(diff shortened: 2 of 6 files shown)


www/fresh.gen.ts
```diff
@@ -2,19 +2,25 @@
...
-    "./routes/gfm.css.ts": $1,
-    "./routes/index.tsx": $2,
-    "./routes/raw.ts": $3,
+    "./routes/about.tsx": $0,
+    "./routes/docs/[...slug].tsx": $1,
+    "./routes/gfm.css.ts": $2,
+    "./routes/index.tsx": $3,
+    "./routes/level1/level2/[test].tsx": $4,
+    "./routes/level1/level2/test.tsx": $5,
+    "./routes/raw.ts": $6,
```
excessivecoding: Don't commit your test routes here, as this is the documentation website.

src/dev/deps.ts
```diff
@@ -6,6 +6,7 @@ export {
   join,
   resolve,
   toFileUrl,
+  parse as pparse,
```
excessivecoding: Maybe use a more explicit name like `parsePath` ?

src/dev/mod.ts
```diff
@@ -83,28 +84,23 @@ export async function generate(directory: string, manifest: Manifest) {
...
-${
-    islands.map((file, i) => `import * as $$${i} from "./islands${file}";`)
-      .join("\n")
-  }
+${routes
+  .map((file, i) => `import * as $${i} from "./routes${file}";`)
+  .join("\n")}
+${islands
+  .map((file, i) => `import * as $$${i} from "./islands${file}";`)
+  .join("\n")}
```
excessivecoding: This repo use the Deno formatter by default for code format, I did some tests and I don't think those lines needed to be formatted

src/dev/mod.ts
```diff
@@ -135,8 +131,53 @@ export default manifest;
...
+    }
+    return output;
+  };
+
+  const errorsList = errors();
+  if (errorsList.length > 0) {
+    console.log(
+      `%cCheck ${errorsList} for potential routing issues.
+You may have dynamic and static routes overwriting each other.
+Please check the documentation for more information. http://localhost:8000/docs/getting-started/dynamic-routes`,
```
excessivecoding, line 185: You are pointing to your `localhost:8000` here, replace it with the Fresh website.

src/dev/mod.ts
```diff
@@ -135,8 +131,53 @@ export default manifest;
...
+      "color: red; font-weight: bold"
```
excessivecoding: This is more of a personal opinion, but I don't think this message should display in red, as it's more a warning. Might need confirmation from the project maintenainer though

excessivecoding (commented):
Made some feedback. Note that I'm not the maintainer, so take those with a pinch of salt

Edit: I just saw your questions, I'm going to reply at the best of my limited knowledge:


>  Why is there no .gitignore ?

There is none as Fresh doesn't generate bundled code and Deno doesn't install dependencies in a `node_modules`. And I guess no files needed to be ignored.

> Are there any plans or style guide / design system for the docs ?

There is (cf. https://github.com/denoland/fresh/issues/498)

> Is there a contributors guide that I'm missing?

Not yet, one is in the work (cf. https://github.com/denoland/fresh/pull/465)

> Why is my prettier working differently ? It has formatted a few exisiting things differently? Dont we need a prettier config file in the root of Fresh?

Deno comes with a custom `deno fmt` (similar to prettier, but with a different default, [for more info](https://deno.land/manual/tools/formatter)). VSCode should pick it up automatically, but you might need to disable your prettier in this workspace.

nhc:
@xstevenyung - Thank you so much for taking the time to answer my questions. I will fix this up some more. Makes a lot of sense about custom formatter etc, it just didnt cross my mind.

src/dev/mod.ts
nhc: But when they are developing it will always be here and the error is in the terminal?

src/dev/mod.ts
nhc: But when they are developing it will always be  running locally here in the terminal?

excessivecoding: Yes, but this is a link to the documentation, not on your local project

excessivecoding (changes requested):

Mrashes:
[Issue was closed as not planned](https://github.com/denoland/fresh/issues/475) probably worth closing this PR

bartlomieju:
Thanks for the contribution! Unfortunately this PR is from the Fresh 1.x era and the codebase has since been completely rewritten for Fresh 2.x — the files modified here (`src/dev/mod.ts`, `src/dev/deps.ts`) no longer exist, and the routing system is entirely different.

Closing as stale. If route conflict warnings are still a desired feature in Fresh 2.x, it would need to be reimplemented from scratch against the current architecture.

PR #540, opened 2022-07-24
fix: upgrade preact and preact-render-to-string

dogukanakkaya:
Preact@10.8.2 throws an exception (I also upgraded to Deno v1.24 but it does not seems to be related with Deno version)

```
An error occured during route handling or page rendering. TypeError: Cannot read properties of undefined (reading '__H')
```

Upgrading Preact fixed this problem.
This commit and merge might be urgent because I just created a new Fresh app and faced this issue.


lucacasonato:
Done in https://github.com/denoland/fresh/pull/561. Thanks for the PR though.

PR #552, opened 2022-07-26
Add a function to get a server class instance.

Octo8080X:
I cut out the function that gets an instance of the server from the start function.

The purpose is to prepare you to create tests for your application.
The introduction of this feature does not affect fresh.

Example:

```typescript
// main_test.ts

/// <reference lib="dom" />
/// <reference lib="deno.unstable" />

import { getServer } from "$fresh/server.ts";

import { superdeno } from "https://deno.land/x/superdeno/mod.ts";
import { assertEquals } from "https://deno.land/std@0.148.0/testing/asserts.ts";
import {
  Document,
  DOMParser,
} from "https://deno.land/x/deno_dom/deno-dom-wasm.ts";

Deno.test("HTTP assert test.", async (t) => {
  await t.step("#1 GET /", async () => {
    const server = await getServer(manifest);

    await superdeno(server)
      .get("/")
      .expect(200);
  });

  await t.step("#2 GET /Foo", async () => {
    const server = await getServer(manifest);

    const r = await superdeno(server)
      .get("/Foo")
      .expect(200);

    const document: Document = new DOMParser().parseFromString(
      r.text,
      "text/html",
    )!;

    assertEquals(document.querySelector("div")?.innerText, "Hello Foo");
  });

  await t.step("#3 GET /Foo/Bar", async () => {
    const server = await getServer(manifest);

    await superdeno(server)
      .get("/Foo/Bar")
      .expect(404);
  });
});
```

```sh
$ deno test --allow-net --allow-env --allow-read main_test.ts
Checked 11 files
running 1 test from ./main_test.ts
HTTP assert test. ...
  #1 GET / ... ok (58ms)
  #2 GET /Foo ... ok (56ms)
  #3 GET /Foo/Bar ... ok (42ms)
HTTP assert test. ... ok (162ms)

ok | 1 passed (3 steps) | 0 failed (295ms)
```

### DIFF
--- a/src/server/deps.ts
+++ b/src/server/deps.ts
@@ -5,7 +5,7 @@ export {
   toFileUrl,
 } from "https://deno.land/std@0.128.0/path/mod.ts";
 export { walk } from "https://deno.land/std@0.128.0/fs/walk.ts";
-export { serve } from "https://deno.land/std@0.128.0/http/server.ts";
+export { Server } from "https://deno.land/std@0.128.0/http/server.ts";
 export type {
   ConnInfo,
   Handler as RequestHandler,
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -1,5 +1,5 @@
 import { ServerContext } from "./context.ts";
-import { serve } from "./deps.ts";
+import { Server } from "./deps.ts";
 import {
   AppModule,
   ErrorPageModule,
@@ -45,15 +45,37 @@ export interface Manifest {
 
 export { ServerContext };
 
+export async function getServer(
+  routes: Manifest,
+  opts: StartOptions = {},
+): Promise<Server> {
+  const ctx = await ServerContext.fromManifest(routes, opts);
+
+  const server = new Server({
+    handler: ctx.handler(),
+    port: opts.port ?? 8000,
+    hostname: opts.hostname ?? "0.0.0.0",
+    onError: opts.onError,
+    ...opts,
+  });
+
+  if (opts?.signal) {
+    opts.signal.onabort = () => server.close();
+  }
+
+  return server;
+}
+
 export async function start(
   routes: Manifest,
   opts: StartOptions = {},
 ) {
-  const ctx = await ServerContext.fromManifest(routes, opts);
+  const server = await getServer(routes, opts);
   console.log(
     `Server listening on http://${opts?.hostname ?? "localhost"}:${
       opts?.port ?? 8000
     }`,
   );
-  await serve(ctx.handler(), opts);
+
+  return await server.listenAndServe();
 }


lucacasonato:
What if we add a `createHandler(manifest, opts)` function instead? It can just be implemented as follows:

```js
export async function createHandler(
  routes: Manifest,
  opts: StartOptions = {},
) {
  const ctx = await ServerContext.fromManifest(routes, opts);
  return ctx.handler();
}
```

`superdeno` also accepts handler functions instead of a full blown `http.Server`.

Octo8080X:
Thank you for your interest.
I have implemented createHandler(manifest, opts) in src/server/mod.ts.

```typescript src/server/mod.ts
export async function createHandler(
  routes: Manifest,
  opts: StartOptions = {},
) {
  const ctx = await ServerContext.fromManifest(routes, opts);
  return ctx.handler();
}
```

I confirmed that it works by adding --no-check. It seems to work faster.

```sh
$ deno test -A --no-check main_test.ts
Checked 14 files
running 1 test from ./main_test.ts
HTTP assert test. ...
  #1 GET / ... ok (58ms)
HTTP assert test. ... ok (63ms)

ok | 1 passed (1 step) | 0 failed (133ms)
```

The error in the type check was the following.

```sh
$ deno test -A main_test.ts
Checked 14 files
Check file:///usr/src/app/my-project/main_test.ts
error: TS2345 [ERROR]: Argument of type 'Handler' is not assignable to parameter of type 'string | RequestHandlerLike | ListenerLike | ServerLike'.
  Type 'Handler' is not assignable to type 'RequestHandlerLike'.
    await superdeno(handler)
                    ~~~~~~~
```

As a workaround, we can use assertions.

```typescript
import { createHandler } from "$fresh/server.ts";

import { superdeno } from "https://deno.land/x/superdeno/mod.ts";
import {
  type RequestHandlerLike,
} from "https://deno.land/x/superdeno/src/types.ts";

Deno.test("HTTP assert test.", async (t) => {
  await t.step("#1 GET /", async () => {
    const handler = await createHandler(manifest) as RequestHandlerLike;

    await superdeno(handler)
      .get("/")
      .expect(200);
  });
});
```
```sh 
$ deno test -A main_test.ts
Checked 14 files
running 1 test from ./main_test.ts
HTTP assert test. ...
  #1 GET / ... ok (56ms)
HTTP assert test. ... ok (63ms)
```

Errors in type checking should probably be addressed with superdeno.

We think the direction you suggested is better as it also speeds up the process.

Octo8080X:
@lucacasonato 

We have created another pull request based on the ideas we received.
Once the new one is adopted, this pull request will be closed.

[Add a function to create handler.](https://github.com/denoland/fresh/pull/622)

Thank you.

lucacasonato:
Closing in favour of your superseding PR.

Issue #549, opened 2022-07-26
Multiple middleware per route


jcs224:
I wonder if it would be feasible to instantiate multiple middlewares for a single route, instead of just a single `_middleware.ts` file.

For example, in Oak, you can stack an infinite number of middlewares for a single route, or a subset of routes.

My main use-case is to authenticate a route, the authentication middleware depends on a session middleware. Like, if I could export more than one handler per `_middleware.ts`...

```ts
import { MiddlewareHandlerContext } from "$fresh/server.ts";

interface State {
  data: string;
}

export async function handler(
  req: Request,
  ctx: MiddlewareHandlerContext<State>,
) {
  console.log('boot up a session')
  const resp = await ctx.next()
  console.log('clean up session')

  return resp;
}

export async function handler2(
  req: Request,
  ctx: MiddlewareHandlerContext<State>,
) {
  console.log('authenticate route, depends on session')
  const resp = await ctx.next()
  console.log('tear down auth')

  return resp;
}
```

I guess the reason I would do this, is because I don't want to split up the before/after middleware manually. This is how I'd handle it now, splitting all my code into before and after `next()`:

```ts
import { MiddlewareHandlerContext } from "$fresh/server.ts";

interface State {
  data: string;
}

export async function handler(
  req: Request,
  ctx: MiddlewareHandlerContext<State>,
) {
  console.log('boot up a session')
  console.log('authenticate route, depends on session')
  const resp = await ctx.next()
  console.log('tear down auth')
  console.log('clean up session')

  return resp;
}
```

ahuigo:
I think there are many ways to implement this

## Method 1: put multiple middleware file like this
Refer to [middleware-layer](https://github.com/denoland/fresh/tree/main/tests/fixture/routes/layeredMdw/layer2/layer3).

```
/routes/_middleware.ts
/routes/layer2/_middleware.ts
/routes/layer2/layer3/_middleware.ts
/routes/index.ts
```
Sine we may have to write our own **url-pattern**, it is a bit cumbersome to implement.

> Note:  The order of middleware files depends on the `sortRoutes(middlewares)` https://github.com/denoland/fresh/blob/7e2ab7fa50585d0848f2d157c85f789bbf648ab4/src/server/context.ts#L190

## Method 2: wrap your middleware
e.g.
```
export async function handler(
  req: Request,
  ctx: MiddlewareHandlerContext<State>,
) {
  console.log('boot up a session')
[...]

PR #553, opened 2022-07-27
feat: support multiple middlewares with the same matcher

ahuigo:
This `mergeHandler()` could be used to 
1. Merge handlers.
2. Control the order of handlers manually

It is necessary in many situations. https://github.com/denoland/fresh/issues/549


ZackarySantana:
@ahuigo I like the flexibility the merge handler method provides. Do you think it would be worth having each handler handle calling the next handler (similar to how express middleware does it)?

I'm thinking in the context of a handler that should stop the continuation of middleware- whether for performance or since the rest doesn't matter (like your middleware test that does a 204 response, the following middleware wouldn't do much)

lucacasonato (commented):
Seems much easier if we just allow users to export an `Array<Handler>` from middleware routes. WDYT of that?

This would also allow us to avoid adding the `Proxy` complexity.

ahuigo:
Can't agree any more.

> Seems much easier if we just allow users to export an `Array<Handler>` from middleware routes. WDYT of that?
> 
> This would also allow us to avoid adding the `Proxy` complexity.

I think there are many methods, but I don't know which one is better.

## Method 1
Completely exposes `ctx.handlers` to the users. 
```
// src/server/types.ts
export interface MiddlewareHandlerContext<State = Record<string, unknown>>
  extends ConnInfo {
  next: () => Promise<Response>;
  state: State;
  // expose
  handlers: Handler[];
}
```

The downside is that users may misuse it.
For example:
1. Misusing `ctx.handlers.push(handler)`  instead of `ctx.handlers.unshift(handler)` 
2. Misusing `ctx.handlers=[new]` instead of `ctx.handlers=[new, ...old]`


## Method 2
Exposing `ctx.prependHandler(handler)` only is  safer and easier to use(I prefer this).
```ts
export interface MiddlewareHandlerContext<State = Record<string, unknown>>
  extends ConnInfo {
  next: () => Promise<Response>;
  state: State;
 // expose
  prependHandler(handler: Handler):void;
}
```
I pushed a demo here: https://github.com/denoland/fresh/compare/main...ahuigo:fresh:feat/handler?diff=split
**Changes**
I found that `ctx: MiddlewareHandlerContext` is passed to `handler(req, ctx: HandlerContext)`,
so `HandlerContext` and `MiddlewareHandlerContext` should be same object, and I re-alias`MiddlewareHandlerContext` to `HandlerContext` in my code.

lucacasonato:
What I mean is supporting:
```
// routes/_middleware.ts

export const handler = [
  middleware1,
  middleware2,
];
```

jcs224:
> What I mean is supporting:
> 
> ```
> // routes/_middleware.ts
> 
> export const handler = [
>   middleware1,
>   middleware2,
> ];
> ```

This would be my preference. In fact, when I want to bundle some middleware together and use it in various places, this is how I do it in Oak. Each array entry is an Oak middleware:

```ts
export default [
  Session,
  Inertia.initMiddleware(),
  async (ctx: Context, next: () => Promise<unknown>) => {
    ctx.state.inertia.setShared({
      errors: await ctx.state.session.get('errors')
    })
    await next()
  }
] as const
```

ahuigo:
My latest commit has implemented this logic.

> ```
> // routes/_middleware.ts
> 
> export const handler = [
>   middleware1,
>   middleware2,
> ];
> ```

Just two things I'm struggling with.

1. First, wether it is necessary to support  **`Handler[]` only**  instead of  **single `Handler`** , which makes the implementation and usage be more simpler, but  causes **break changes**.

2. Second, is it necessary to support multiple handlers for **component handler**?
For example:
```ts
// routes/index.tsx
  GET:[
      corsHandler,
      logHandler,
      (req, ctx)=> {
          ....
      }
  ],
};
```

lucacasonato:
> First, wether it is necessary to support Handler[] only instead of single Handler , which makes the implementation and usage be more simpler, but causes break changes.

We are past 1.0, so we can not make breaking changes. Both should be supported.

> Second, is it necessary to support multiple handlers for component handler?

I don't think so. If this need comes up in the future, we can look into it then.

lucacasonato (approved):
Thank you!

zenyr:
For the time being (1.0.2) I've adopted original @ahuigo 's solution in my userland code.
```ts
// lib/mergeHandlers.ts
import { MiddlewareHandlerContext } from '$fresh/server.ts';

type Handler<T> = (req: Request, ctx: MiddlewareHandlerContext<T>) => Promise<Response> | Response;

// Merge middleware handlers
export function mergeHandlers<T>(...handlers: Handler<T>[]) {
  return (_req: Request, ctx: MiddlewareHandlerContext<T>) => {
    const innerHandlers = [...handlers];
    const proxy: MiddlewareHandlerContext<T> = new Proxy(ctx, {
      get: (orgCtx, prop: keyof MiddlewareHandlerContext<T>) => {
        if (prop === 'next') {
          const handler = innerHandlers.shift();
          if (handler) {
            return async () => await handler(_req, proxy);
          }
        }
        return orgCtx[prop];
      },
    });
    return proxy.next();
  };
}



// routes/_middleware.ts
import { mergeHandlers } from '$/lib/mergeHandlers.ts';
// ...snip...
export type State = WithRedis & WithSession;
export const handler = mergeHandlers(redisMiddleware, sessionMiddleware);
```

PR #564, opened 2022-07-30
feat: apply `lang` and `bodyClass` set in middleware to template

brycedorn:
Hi again! As followup to #47, setting `ctx.lang` doesn't appear to do anything as `ServerContext` is isolated from `RenderContext`. These changes enable a middleware to apply a `lang` param (e.g. based on `Accept-Language` header) that will be rendered in the server-side template (as `<html lang={lang}>`).

Additionally, I added a `bodyClass` param that has similar behavior, but instead sets a class on the `<body>` tag. This enables a middleware to apply the correct theme during SSR via the `Sec-CH-Prefers-Color-Scheme` client-hint header. This also resolves #473.

This implementation piggybacks on the existing `params` object, though these values could be passed via a separate method. Let me know what you think!

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -29,7 +29,7 @@ import {
   UnknownPage,
   UnknownPageModule,
 } from "./types.ts";
-import { render as internalRender } from "./render.tsx";
+import { render as internalRender, RenderContext } from "./render.tsx";
 import { ContentSecurityPolicyDirectives, SELF } from "../runtime/csp.ts";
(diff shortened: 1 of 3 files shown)


hashrock:
> These changes enable a middleware to apply a lang param

What does this actually do? It would be helpful if you could write a test for example

brycedorn:
Resolved by https://github.com/denoland/fresh/pull/1617.

PR #581, opened 2022-08-03
add title tag, add setPageTitle function

omar2205:
This PR adds `<title>` tag to the HTML template, and `setPageTitle` function to change it (server-side, ie, Handler).

```ts
import { setPageTitle } from "$fresh/server.ts";
export const handler = async (req, ctx) => {
  setPageTitle('Home - Fresh App')
  return await ctx.render()
}
```

### DIFF
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -53,3 +53,5 @@ export async function start(
   const ctx = await ServerContext.fromManifest(routes, opts);
   await serve(ctx.handler(), opts);
 }
+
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
This should be done with `<Head>` from `$fresh/runtime.ts` instead.

omar2205:
This adds a default title tag which can be updated quickly without the need to use a component.

Anyways, you can close this PR if it's not useful.

lucacasonato:
I don't think this is useful.

PR #593, opened 2022-08-06
Update dynamic-routes.md

sebazelonka:
Update docs to match current file structure.

I'm using `fresh@1.0.2` and the boilerplate created have the dynamic route directly into the root of `routes` folder.


iuioiua (approved):
LGTM! Thank you, @sebazelonka.

PR #600, opened 2022-08-08
Remove `deno.unstable` references

kidonng:
Now that Fresh hit 1.0, I feel the `deno.unstable` references seems unnecessary (and does more harm than good to end users).


lucacasonato (approved):
Thanks.

PR #609, opened 2022-08-10
fix: add missing icon bg colors on mobile

michael-pfister:
I noticed that the `FeatureIcons` in the new landing page had missing background colors on mobile, so I added them.

image
image


hashrock:
thank you for catching up!

Is it possible to avoid color repetition? 

image

Ideally, the colors should be arranged as [green, blue, yellow, green, blue, yellow] on mobile. However, I would like to keep the current colors on desktop.

hashrock (approved):

hashrock:
Thank you!

PR #610, opened 2022-08-10
feat: support JSX automatic mode

lucacasonato:
This commit adds support for JSX automatic mode instead of the default
mode that requires the presence of the JSX pragma in each file.

This commit also adds the config file to the `fresh.gen.ts` manifest,
which is used to determine if the client emit should use automatic mode.

The minimum Deno version has also been bumped to 1.25.0.


src/server/context.ts
```diff
@@ -103,7 +105,32 @@ export class ServerContext {
...
     const baseUrl = new URL("./", manifest.baseUrl).href;
-    const importMapURL = new URL("./import_map.json", manifest.baseUrl);
+
+    const config = manifest.config || { importMap: "./import_map.json" };
+    if (typeof config.importMap !== "string") {
+      throw new Error("deno.json must contain an 'importMap' property.");
+    }
+    const importMapURL = new URL(config.importMap, manifest.baseUrl);
+
+    config.compilerOptions ??= {};
```
vwkd, line 115: Incomplete check for existence of `./deno.json`?

This introduces a dependency on the existence of a `./deno.json`. When it doesn’t exist, `config.compilerOptions` is `undefined` and accessing `config.compilerOptions.jsx` throws.

A few lines up its existence is successfully checked when getting the import map path, falling back to the default otherwise.

src/dev/mod.ts
```diff
@@ -83,6 +103,7 @@ export async function generate(directory: string, manifest: Manifest) {
 // This file SHOULD be checked into source version control.
 // This file is automatically updated during development when running \`dev.ts\`.
 
+import config from "./deno.json" assert { type: "json" };
```
vwkd, line 106: A `./deno.json` implies that fresh expects to be in the root directory of the repository. This might not always be the case, for example in a mono repo that contains multiple clients and/or servers.

It would be great to not depend on a `deno.json` at all, or at least make its path configurable.

vwkd (commented):
Currently, this introduces a hard dependency on a `./deno.json` for compiler options.

It would be great if its usage were optional and, if used, its path configurable.

Currently, if it doesn’t exist fresh throws an error.

```txt
error: Uncaught (in promise) TypeError: Module not found "file:[PATH]/deno.json".
    at file:[PATH]/fresh.gen.ts:5:20
  await import(entrypoint);
  ^
    at async dev (https://raw.githubusercontent.com/denoland/fresh/main/src/dev/mod.ts:187:3)
    at async file:[PATH]/dev.ts:5:1
```

PR #612, opened 2022-08-11
added freshhub showcase

harshmangalam:


lucacasonato:
Thanks, but I don't want to put personal sites on the home page for now.

PR #621, opened 2022-08-15
www: use frontmatter parser from std

benny123tw:
Replace the `frontmatter` with the std version, since the package is deprecated: https://deno.land/x/frontmatter@v0.1.5


lucacasonato (approved):
Thanks!

PR #628, opened 2022-08-16
feat: add preflight option in twind plugin

ghost:
This PR adds the ability to use a custom preflight config in the twind plugin. This allows things like injecting custom styles into the head and adding `@font-faces` or `@imports` without having to use an _app.jsx file!

This is an option built into twind itself and the documentation is right here: https://twind.dev/handbook/configuration.html#preflight

Adding a custom CSS `@import` and a `@font-face` in `twind.config.js` would look like this:

```
export default {
  preflight: {
    // Import external stylesheet
    '@import': `url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap')`,
    // Declare font face
    '@font-face': [
      {
        fontFamily: 'Proxima Nova',
        fontWeight: '400',
        src: 'url(/fonts/proxima-nova/400-regular.woff) format("woff")',
      },
      {
        fontFamily: 'Proxima Nova',
        fontWeight: '700',
        src: 'url(/fonts/proxima-nova/500-medium.woff) format("woff")',
      },
    ],
  },
}
```

### DIFF
--- a/plugins/twind/shared.ts
+++ b/plugins/twind/shared.ts
@@ -2,6 +2,7 @@ import { JSX, options as preactOptions, VNode } from "preact";
 import {
   Configuration,
   DarkMode,
+  Preflight,
   setup as twSetup,
   Sheet,
   ThemeConfiguration,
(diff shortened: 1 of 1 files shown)


irvanmalik48 (commented):
You might wanna make the preflight type this way:

```
preflight?: Preflight | boolean | CSSRules;
```

PR #639, opened 2022-08-21
Add support for preact devtools in dev mode

### LINKED ISSUE #611: Debug using preact
Hello, 

I'm new with Deno, but not with Node.js. I started a small project using fresh framework and I'd like to know the best way to debug **Preact**. 

In **Preact** [documentation](https://preactjs.com/guide/v10/debugging/#installation) mention a google chrome extension and `import("preact/debug")` which I couldn't make it work. 

Also try with `--inspect-brk` which works for code but doesn't provide information of states, components, etc, in **Preact**. 

This is my **dev.ts** file: 

```typescript
#!/usr/bin/env -S deno run -A --watch=static/,routes/

import("dotenv");
if (Deno.env.get("ENV") === "DEV") {
  import("preact/debug");
}

import dev from "$fresh/dev.ts";

await dev(import.meta.url, "./main.ts");
```

But doesn't seem to work. Also added at the beginning of **main.ts** file with same result. 

Any advice?

### LINKED ISSUE #3686: DevTools do not work with the new Deno framework, Fresh.
**Describe the bug**
When making a Fresh application, adding "import "preact/debug"" to the entrypoint of the application does nothing. The Fresh framework relies on preact by default, so it should work?

**To Reproduce**
1. Create a new Fresh
```
deno run -A -r https://fresh.deno.dev my-project
```
2. Insert `import preact/debug` to the top of main.ts (entry point for production) and dev.ts (entry point for development).
3. Go to `localhost:8000/` and check the Preact DevTools Extension.
4. It will say "This page doesn't seem to be using Preact. // Import preact/debug somewhere to initialize the connection to the extension. Make sure that this import is the first import in your whole app."

**Expected behavior**
I should be able to view the devtools if I've downloaded the extension and place `import "preact/debug"` into the entrypoint of the application.

marvinhagemeister:
This PR injects `preact/debug` before the main bundle when running in `dev` mode. This enables helpful debug warnings for Preact and sets up the bridge for the DevTools browser extension.

To show the actual component names instead of the minified names, we need to disable esbuild's identifier minification option. The alternative would be to transform every file and append the `displayName` property to all component functions, but since esbuild doesn't expose its AST we'd need to fall back on babel or something similar. That would incur a significant compilation cost and I'm assuming that this wouldn't be in the spirit of this project.

Let me know what you think.

DevTools in action:

Screenshot 2022-08-21 at 08 42 31

Fixes #611 
Fixes https://github.com/preactjs/preact/issues/3686

### DIFF
--- a/src/runtime/main_dev.ts
+++ b/src/runtime/main_dev.ts
@@ -1,0 +1,2 @@
+import "preact/debug";
+export { revive } from "./main.ts";
--- a/src/server/bundle.ts
+++ b/src/server/bundle.ts
@@ -25,16 +25,25 @@ export class Bundler {
   #islands: Island[];
   #plugins: Plugin[];
   #cache: Map<string, Uint8Array> | Promise<void> | undefined = undefined;
+  #dev: boolean;
 
-  constructor(islands: Island[], plugins: Plugin[], importMapURL: URL) {
+  constructor(
+    islands: Island[],
+    plugins: Plugin[],
+    importMapURL: URL,
+    dev: boolean
+  ) {
     this.#islands = islands;
     this.#plugins = plugins;
     this.#importMapURL = importMapURL;
+    this.#dev = dev;
   }
 
   async bundle() {
     const entryPoints: Record<string, string> = {
-      "main": new URL("../../src/runtime/main.ts", import.meta.url).href,
+      main: this.#dev
+        ? new URL("../../src/runtime/main_dev.ts", import.meta.url).href
+        : new URL("../../src/runtime/main.ts", import.meta.url).href,
     };
 
     for (const island of this.#islands) {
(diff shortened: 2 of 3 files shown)


lucacasonato (approved):
LGTM. Thanks Marvin!

moonstripe:
@marvinhagemeister are there any special instructions for implementing this in existing projects?

marvinhagemeister:
@moonstripe Once a new version of fresh is released you just need to upgrade to that and that's it. There are no additional steps required.

labidiaymen:
fresh: 1.3.0
preact: 10.15.1
Deno: 1.34.3

Still doesn't work for me, is there any additional steps? 
Thanks in advance.

image

marvinhagemeister:
@labidiaymen See https://github.com/denoland/fresh/issues/1461 . We're currently affected by a transpilation bug in `esm.sh`

PR #641, opened 2022-08-21
fix: add reference to dom.iterable

rojvv:
This fixes TypeScript errors with things like `URLSearchParams.entries`.

Related: https://github.com/denoland/deno/issues/15517#issuecomment-1221521244


lucacasonato (approved):
Thanks.

PR #647, opened 2022-08-23
chore: use rutt instead of crux.land/router

eliassjogreen:
crux.land/router is becoming denosaurs/rutt. Functionality is the same and everything should keep on working the same!


lucacasonato (approved):
Thanks

PR #682, opened 2022-08-31
fix: Regression issues with Deployment concepts page

lino-levan:
Fixes two of the links not being set up correctly. (Top is before, bottom is after)

Screen Shot 2022-08-31 at 8 23 34 AM
Screen Shot 2022-08-31 at 8 23 19 AM


lucacasonato (approved):
Thanks

PR #683, opened 2022-08-31
feat(plugins/twind): allow all config options except mode and sheet

eliassjogreen:


lucacasonato (dismissed):

lucacasonato:
Actually, this doesn't work. Plugins are functions and can as such not be serialized into the client config.

lucacasonato:
Closing in favor of #705.

PR #689, opened 2022-09-02
docs: image alt for "Made with Fresh" badges

xyntechx:
Included the `alt` attribute in `` tags in README.


hashrock (approved):
Good catch! thank you!

Issue #555, opened 2022-07-27
Use file names for magic instead of directories


shelbyd:
Currently, Fresh uses `routes/_middleware.ts`, `routes/some/path.tsx`, `islands/*.tsx`, and probably more I'm not aware of for certain behaviors. This is often referred to as "magic".

I propose Fresh uses file names instead of directory locations to determine this magic.

Most importantly, islands should not all have to reside in `islands/`, but should instead be in any directory as `some/path/Counter.island.tsx`. This will allow the relevant code for some functionality to be co-located as much as possible and not spread between multiple directories.

For example, `routes/_middleware.ts` should be `routes/auth.middleware.ts` to support multiple middlewares applying to the same route. If the order of middlewares is important, they could be applied in lexicographic order like `routes/01auth.middleware.ts` and `routes/02content_type.middleware.ts`.

I think it does make sense to keep `routes/` as a magic directory. Routing is a global concern, and projects can import components for routes from the route file.

ahuigo:
> Most importantly, islands should not all have to reside in `islands/`, but should instead be in any directory as `some/path/Counter.island.tsx`. This will allow the relevant code for some functionality to be co-located as much as possible and not spread between multiple directories.

You could put most the relevant code to be co-located into `islands/`, `routes` is just `routes`.
I think the reason for this arrangement is that fresh chooses the [islands architecture](https://jasonformat.com/islands-architecture/)

lucacasonato:
I prefer the current setup. If someone else wants to build this, I may consider merging it, but no guarantees. I'll close this because I am not planning to spend any time on this.

TranquilMarmot:
The current setup where all islands need to be in the `islands` directory, which can't have any folders in it, is actually really frustrating to me. I now have code for one piece of my app spread throughout _three_ separate folders:

- `routes` for the actual route
- `components` for various bits and pieces rendered by the route
- `islands` for any JS that I want to send to the client

`routes` and `islands` both import from `components` in different ways (I actually have even _more_ folders that I use to organize, but they're mostly irrelevant here). I feel like I'm zipping around the project's directory layout whenever I'm working on even one, isolated part of the app.

I'd love to keep everything in just `components` and not have to deal with `islands` as a magic folder.

PR #692, opened 2022-09-03
Filename islands

m4rocks:
This is to reduce magic as mentioned in #555. Islands are now grabbed from the components directory and require `.island.{ts,tsx,js,tsx}` extension.

Things to do:
- [x] Add flag to toggle this behaviour
- [ ] Allow islands to be imported from subdirectories
- [ ] Add tests
- [ ] Add example
- [ ] Add documentation


notangelmario:
Update: added options to dev method to toggle this behaviour. The default behaviour is to use the islands directory

iuioiua:
I'm closing this as it hasn't had any progress in the last 9 months. Also, IMO, the current islands folder setup works fine. Thanks anyway 🙂

PR #702, opened 2022-09-06
docs: add plugin docs

lucacasonato:


docs/concepts/plugins.md
```diff
@@ -0,0 +1,93 @@
...
+
+```ts
+// main.ts
+
+// ...
+
+import twindPlugin from "$fresh/plugins/twind.ts";
+import twindConfig from "./twind.config.js";
+
+await start(manifest, {
```
kwnath, line 20: Curious but what is manifest in this context? Might be a bit confusing for new readers.

docs/concepts/plugins.md
```diff
@@ -0,0 +1,93 @@
...
+  plugins: [
+    // This line configures Fresh to use the first-party twind plugin.
+    twindPlugin(twindConfig),
+  ],
+});
+```
+
+Currently, the only available first-party plugin is the Twind plugin.
```
kwnath, line 28: Would be nice to have a list of first-party / third-party plugins in some other readme later down the line

docs/concepts/plugins.md
```diff
@@ -0,0 +1,93 @@
...
+
+// ...
+
+import twindPlugin from "$fresh/plugins/twind.ts";
+import twindConfig from "./twind.config.js";
+
+await start(manifest, {
+  plugins: [
+    // This line configures Fresh to use the first-party twind plugin.
+    twindPlugin(twindConfig),
```
kwnath, line 23: How are plugins run, in order or at random?

kwnath: Additionally I think it might be a good idea to state that third-party plugins are also allowed as the example above can imply that third-party plugins aren't allowed yet.

kwnath (commented):
This looks awesome!

docs/concepts/plugins.md
lucacasonato: It is the `./fresh.gen.ts` file (see imports).

docs/concepts/plugins.md
lucacasonato: Agreed.

docs/concepts/plugins.md
lucacasonato: I see the import was not there. I have added the imports for context now.

PR #704, opened 2022-09-06
Fix HTML meta tag charset property casing

vwkd:
Corrects strange casing of HTML meta tag charset property for cleaner visual appearance.


lucacasonato (approved):
Thx

PR #717, opened 2022-09-08
fixing @preact/signals and signals_core version

flemmingmiguel:
the @preact/signals_core package doesn't have 1.0.2 version so separated the 2 packages to their latest versions

@preact/signals 1.0.3
@preact/signals_core 1.0.1


lucacasonato (approved):
LGTM, thanks!

PR #727, opened 2022-09-08
add todo example

MrSaeedNasiri:


ahuigo:
Maybe, you could deploy your example to https://deno.com and add it to https://fresh.deno.dev/showcase

gitblit:
I think your example could be helpful (I was looking for a good Todo implementation), but it's managing everything client-side in Preact state. Had you considered extending it to get & post to the backend?

PR #739, opened 2022-09-09
fix: sanitize `<script>` in __FRSH_STATE state

wilf312:
Hello, thanks for the great product! 

This PR fixes a problem where text after the script tag is rendered as is if __FRSH_ISLAND_PROPS contains a script tag.
This is happening because the </script> tag is not escaped in JSON.stringify.

1. Routes component's props set  string `</script>` to island component..
2. When the island component is rendered, the value of the island props is embedded in __FRSH_ISLAND_PROPS.
3. `</script>` string is not escaped and is recognized by html as the end of __FRSH_ISLAND_PROPS`</script>`
4. The string after the </script> tag is displayed as is. (ex)`<script/><h1>test</h1>`

> <>\u2028\u2029

Fix to output __FRSH_ISLAND_PROPS through a function that escapes this string.

We can check the reproduced state from the following page 
[demo](https://wilf312-island-props-demo.deno.dev/) 
[demo repository](https://github.com/wilf312/island-props-demo) 

スクリーンショット 2022-09-09 19 28 43

I hope this is not written in a rude way because I am not good at English.


lucacasonato (commented):
Good catch, thanks! Can you add an integration test to verify that this works as intended and does not break client hydration in the browser? Maybe as part of the test in `//tests/island_test.ts`.

wilf312:
@lucacasonato 
Thank you review comment!
Please check back as we have written the test you taught us.

https://github.com/denoland/fresh/pull/739/files#diff-638e97e2a9ebf2444cade01ad51b35c6f71e9ebb986c57be1fe88a485cea4f40R116-R133

lucacasonato (approved):
LGTM, thanks! I ensured the test fails on main, and it does. Great work!

PR #740, opened 2022-09-09
feat: automatic loading of `.env` file

beyazit:
added dotenv option to read environment variables files


src/dev/imports.ts
```diff
@@ -20,3 +20,8 @@ export function twindImports(imports: Record<string, string>) {
   imports["twind"] = `https://esm.sh/twind@${RECOMMENDED_TWIND_VERSION}`;
   imports["twind/"] = `https://esm.sh/twind@${RECOMMENDED_TWIND_VERSION}/`;
 }
+
+export function dotenvImports(imports: Record<string, string>) {
+  // always use latest version
+  imports["$std/"] = `https://deno.land/std/`;
```
lucacasonato: This should be a versioned import.

lucacasonato (commented):
I like this. Let's actually enable dotenv always, with no opt-out in the init script. We should try to keep the options in the init script nice and simple so users don't have to make a bunch of decisions when they are initially setting up their Fresh project :)

thorwebdev:
Just found this while looking for docs for `dotenv`. Wondering what the status here is? Is it worthwhile to contribute some docs or is `dotenv` default landing soon?

src/dev/imports.ts
```diff
@@ -3,6 +3,7 @@ export const RECOMMENDED_PREACT_RTS_VERSION = "5.2.3";
 export const RECOMMENDED_PREACT_SIGNALS_VERSION = "1.0.3";
 export const RECOMMENDED_PREACT_SIGNALS_CORE_VERSION = "1.0.1";
 export const RECOMMENDED_TWIND_VERSION = "0.16.17";
+export const RECOMMENDED_STD_VERSION = "0.155.0";
```
iuioiua:
```suggestion
export const RECOMMENDED_STD_VERSION = "0.187.0";
```
This is the current latest version.

iuioiua (changes requested):
Could you please remove any opt-in/opt-out options, as Luca requested? I'll merge once, the conflict and my suggestion are also addressed 👍🏾

iuioiua (approved):
I've added a small sentence pointing out this new behaviour in the docs. Thanks for this, @0bl!

PR #742, opened 2022-09-09
add im@sparql Data Dashboard to showcase

arrow2nd:
I have created a SPARQL Endpoint data dashboard called [im@sparql](https://sparql.crssnky.xyz/imas/) using Fresh.
So I added it to the showcase.

If there is anything I am doing wrong, I would appreciate it if you could let me know.


hashrock (approved):
LGTM!

PR #745, opened 2022-09-10
Async plugin rendering (addresses #728)

### LINKED ISSUE #728: Asynchronous plugin render functions
Hey @lucacasonato, it's awesome to see the first step towards plugin support released!

I attempted to convert the UnoCSS integration I wrote (back in #290) to a Fresh-loadable plugin, but I ran into one issue: UnoCSS' generator function is asynchronous (and likely always will be: https://github.com/unocss/unocss/issues/26), but Fresh plugins are currently synchronous.

It makes a lot of sense to me to allow plugins to register asynchronous render functions, and I'm sure there'll be other integrations/plugins that'll run into the same issue as I did. Looking at the implementation in https://github.com/denoland/fresh/blob/main/src/server/render.ts#L173-L197, it seems like it should be fairly straightforward to implement async plugin render functions (they're already called from an asynchronous context), unless there's something else I've missed?

I would be happy to open a PR for this and to add an UnoCSS plugin to `$fresh/plugins`.

dragonwocky:
Closes #728

For now, it's added as an extra hook that runs after all other render functions are called. I'm assuming it's alright that the async plugin render functions would be called after `opts.renderFn`, since none of the plugins modify the page's content, only add scripts/styles, and this way the thread local state will be preserved everywhere except in the renderAsync hooks.

If this looks good to you, I'll see if I can start building a couple of plugins.

### DIFF
--- a/docs/concepts/plugins.md
+++ b/docs/concepts/plugins.md
@@ -99,3 +99,29 @@ serializable JavaScript value.
 
 For an example of a plugin that uses the `render` hook, see the first-party
 [Twind plugin](https://github.com/denoland/fresh/blob/main/plugins/twind.ts).
+
+### Hook: `renderAsync`
+
+This hook is largely the same as the `render` hook, with a couple of key
+differences to make asynchronous style and script generation possible. It must
+asynchronously return its
+[`PluginRenderResult`](https://deno.land/x/fresh/server.ts?s=PluginRenderResult),
+either from an `async/await` function or wrapped within a promise.
+
+The render hook is called with the
+[`AsyncPluginRenderContext`](https://deno.land/x/fresh/server.ts?s=AsyncPluginRenderContext)
+object, which contains a `renderAsync()` method. This method must be invoked
+during the render hook to actually render the page. It is a terminal error to
+not call the `renderAsync()` method during the render hook.
+
+This is useful for when plugins are generating styles and scripts with
+asynchronous dependencies based on the `htmlText`. Unlike the synchronous render
+hook, async render hooks for multiple pages can be running at the same time.
+This means that unlike the synchronous render hook, you can not use global
+variables to propagate state between the render hook and the renderer.
+
+The `renderAsync` hooks start before any page rendering occurs, and finish after
+all rendering is complete -- they wrap around the underlying JSX->string
+rendering, plugin `render` hooks, and the
+[`RenderFunction`](https://deno.land/x/fresh/server.ts?s=RenderFunction) that
+may be provided to Fresh's `start` entrypoint in the `main.ts` file.
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -168,18 +168,18 @@ export async function render<Data>(
 
   let bodyHtml: string | null = null;
 
-  function realRender(): string {
+  function renderInner(): string {
     bodyHtml = renderToString(vnode);
     return bodyHtml;
   }
 
-  const plugins = opts.plugins.filter((p) => p.render !== null);
+  const syncPlugins = opts.plugins.filter((p) => p.render);
   const renderResults: [Plugin, PluginRenderResult][] = [];
 
-  function render(): PluginRenderFunctionResult {
-    const plugin = plugins.shift();
+  function renderSync(): PluginRenderFunctionResult {
+    const plugin = syncPlugins.shift();
     if (plugin) {
-      const res = plugin.render!({ render });
+      const res = plugin.render!({ render: renderSync });
       if (res === undefined) {
         throw new Error(
           `${plugin?.name}'s render hook did not return a PluginRenderResult object.`,
@@ -187,7 +187,7 @@ export async function render<Data>(
       }
       renderResults.push([plugin, res]);
     } else {
-      realRender();
+      renderInner();
     }
     if (bodyHtml === null) {
       throw new Error(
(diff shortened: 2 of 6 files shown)


docs/concepts/plugins.md
```diff
@@ -99,3 +99,28 @@ serializable JavaScript value.
...
 For an example of a plugin that uses the `render` hook, see the first-party
 [Twind plugin](https://github.com/denoland/fresh/blob/main/plugins/twind.ts).
+
+### Hook: `renderAsync`
+
+This hook is largely the same as the `render` hook, with a couple of key
+differences to make asynchronous style and script generation possible. It must
+return its
+[`PluginRenderResult`](https://deno.land/x/fresh/server.ts?s=PluginRenderResult)
+**asynchronously** or within a promise.
```
lucacasonato:
```suggestion
within a promise.
```

Asynchronously === within a promise.

docs/concepts/plugins.md
```diff
@@ -99,3 +99,28 @@ serializable JavaScript value.
...
+
+The render hook is called with the
+[`AsyncPluginRenderContext`](https://deno.land/x/fresh/server.ts?s=AsyncPluginRenderContext)
+object, which contains a `renderAsync()` method. This method must be invoked
+during the render hook to actually render the page. It is a terminal error to
+not call the `renderAsync()` method during the render hook.
+
+This is useful for when plugins are generating styles and scripts with
+asynchronous dependencies. However, variables outside of the `renderAsync`
```
lucacasonato:
```suggestion
This is useful for when plugins are generating styles and scripts based on the `htmlText` with
asynchronous dependencies. However, variables outside of the `renderAsync`
```

lucacasonato (commented):
I think the invocation order of the render functions should change slightly. The `renderAsync` hook shouldn't run _after_ page rendering, it should "surrounding" page rending. The flow should be:

1. render async start
2. legacy opts.renderFn start
3. render sync start
4. inner render (preact-render-to-string)
5. render sync end
6. legacy opts.renderFn end
7. render async end

This will be really useful if we do ever get async context tracing (async thread local state) in Deno (through promise hooks for example).

docs/concepts/plugins.md
dragonwocky: True, I was thinking of it as in either as an `async/await` function or as a `new Promise()`. I'll adjust the wording there.

dragonwocky:
> The renderAsync hook shouldn't run after page rendering, it should "surrounding" page rending.

Wouldn't that then interfere with thread local state if the async hooks are started before the synchronous hooks, though? Unless the async hooks are called without `await`-ing them, and then resolved with a `Promise.all` after the legacy opts.renderFn, but that would mean the htmlText from preact-render-to-string couldn't be passed to them.

That does sound like a much better order of execution, but unless I've misunderstood it would require that `ctx.render()`/`ctx.renderAsync()` get called before any/all async tasks run, which isn't guaranteed.

The other option might be, instead of passing a `ctx.render()`/`ctx.renderAsync()` function, to automatically evaluate the inner render functions first and pass the `htmlText` and `requiresHydration` within the `ctx`, since calling `ctx.render()` is enforced anyway? The only downside to that I can think of is if plugins intend to e.g. modify the `preact.options.vnode` hook before each render.

lucacasonato:
> Wouldn't that then interfere with thread local state if the async hooks are started before the synchronous hooks, though?

I don't think so. As long as there is never an async callsite inside of the synchronous hook, the synchronous TLS should be fine. What are you thinking may brake?

> Unless the async hooks are called without await-ing them, and then resolved with a Promise.all after the legacy opts.renderFn, but that would mean the htmlText from preact-render-to-string couldn't be passed to them.

I don't think I've explained well what I mean to happen. Example:

```ts
const plugin1 = {
  name: "plugin 1",
  async renderAsync(ctx) {
    console.log("renderAsync hook start");
    const res = await ctx.renderAsync();
    assert(res.htmlText);
    console.log("renderAsync hook end");
  },
  render(ctx) {
    console.log("render hook start");
    const res = ctx.render();
    assert(res.htmlText);
    console.log("render hook end");
  },
};

async function render(r) {
  console.log("legacy render start");
  const htmlText = r();
  assert(htmlText);
  console.log("legacy render end");
}


main(manifest, {
  plugins: [plugin1],
  render,
})
```

A render with the above setup should log:

```
renderAsync hook start
legacy render start
render hook start
render hook end
legacy render end
renderAsync hook end
```

The rendered HTML should be available in all hooks.

dragonwocky:
I've adjusted the code to work in the order you've suggested, and I've included a (temporary) demo in `tests/fixture_plugin/options.ts` to demonstrate it:

```ts
let renderCount = 0;

const threadLocalStateTest: Plugin = {
  name: "threadLocalStateTest",
  async renderAsync(ctx) {
    renderCount++;
    console.log(renderCount, "renderAsync hook start");
    await new Promise((res) => setTimeout(res, Math.random() * 3000));
    const res = await ctx.renderAsync();
    assert(res.htmlText);
    console.log(renderCount, "renderAsync hook end");
    return {};
  },
  render(ctx) {
    console.log(renderCount, "render hook start");
    const res = ctx.render();
    assert(res.htmlText);
    console.log(renderCount, "render hook end");
    return {};
  },
};

const render: FreshOptions["render"] = (_ctx, render) => {
  console.log(renderCount, "legacy render start");
  const htmlText = render();
  assert(htmlText);
  console.log(renderCount, "legacy render end");
};
```

When a single page is requested, the following is logged, just like you described:

```
1 renderAsync hook start
1 legacy render start
1 render hook start
1 render hook end
1 legacy render end
1 renderAsync hook end
```

When multiple pages are requested at once (or the same page requested twice within close proximity), the 3rd request starts rendering and updates local state before the 2nd request's async renderers are finished:

```
2 renderAsync hook start
3 renderAsync hook start
3 legacy render start
3 render hook start
3 render hook end
3 legacy render end
3 renderAsync hook end
3 legacy render start
3 render hook start
3 render hook end
3 legacy render end
3 renderAsync hook end
```

This is why I was suggesting pre-evaluating inner render functions, so that async code won't interfere with the expected thread state / execution order. Of course, if the `ctx.renderAsync` is called before anything else happens, then it works as it should, but that's not guaranteed + if we depend on it being the first thing called then why not just pre-evaluate the inner render functions anyway.

zhmushan:
Hello @lucacasonato @dragonwocky, I'm trying to build my project for the first time using Fresh and my favorite CSS engine, UnoCSS, so I came here and I'm willing to move forward with this PR if needed. What is it currently blocked for?

lucacasonato (approved):
LGTM

lucacasonato:
Sorry for the long delay on landing.

iuioiua:
I realise the new function is named `renderAsync()`. However, I think it should be `render()`, with its synchronous counterpart being `renderSync()`, to align with the rest of Deno's runtime APIs. This would be a breaking change, but we're dropping a new release soon, so it might be alright to do this now. Sorry, I didn't look at this sooner.

PR #765, opened 2022-09-12
add targetRoute in middleware context

### LINKED ISSUE #521: Root _middleware is called for every request
I'm working on a session middleware which is called in the `routes/_middleware.ts` to be able to access the session across the app.

As I see it, the root `_middleware` is called for every request even the one outside the routes (e.g. static resource request, _frsh/alive, etc.) which could be the intended behavior as it's consider a global middleware.

My issue is that with this behavior, it's impossible to keep a session initialization in the root middleware and implement features like session flash as the multiple request expires the flashed data.

It could be possible to filter out from the request, but seems pretty unreliable.

Is it the expected behavior ? If so, maybe `routes/_middleware` could be reserved only for request routed to a route, and we could find another way to implement global middlewares ?

excessivecoding:
Should fix: #521 

TODO:
- [ ] Need to add some tests for `targetRoute`


lucacasonato (commented):
This change causes us to do two route lookups per request instead of one. This is a performance issue, and we need to avoid it. Instead we need to do one route lookup only and re-use that for both `targetRoute` and the actual routing.

crowlKats:
Closing in favour of #1123

PR #782, opened 2022-09-19
showcase: cool-catboys sample fresh web

CTMarin:
Hi!

I'm a Software Engineering student, and I made this web just for trying Fresh for the first time. My idea is to update the web with some more features.

Thank you!


hashrock:
Thanks for the great site. I have a concern though. The catboys.com you are using the API for does not appear to me to have proper copyright handling. Specifically, they seem to be crawling images from pixiv and storing them on their servers (which is prohibited by Pixiv's terms of service).

I know it is not your fault, but I would prefer not to add this on showcase.
I think you can make a great site with copywrite-free images, so consider it.

CTMarin:
> Thanks for the great site. I have a concern though. The catboys.com you are using the API for does not appear to me to have proper copyright handling. Specifically, they seem to be crawling images from pixiv and storing them on their servers (which is prohibited by Pixiv's terms of service).
> 
> I know it is not your fault, but I would prefer not to add this on showcase. I think you can make a great site with copywrite-free images, so consider it.

Thank you so much for the advise! I wasn't aware of that about the API and I understand your position. I'm investigating about this and I will consider making another copywrite-free web, as you said.

hashrock:
Let me close this PR for clearance. Feel free to make another PR when you make another website. Thanks for your contribution and efforts!

PR #795, opened 2022-09-25
adding screenshots for portfolio template

vorcigernix:


vorcigernix:
I'll update showcase,json as well

PR #796, opened 2022-09-25
Add portfolio site to the showcase

vorcigernix:
Now it should include all required files


vorcigernix:
and I fixed source to display non jpg files as well

vorcigernix:
oh crap, closing, it does not work :/

PR #797, opened 2022-09-25
adding portfolio page to showcase

vorcigernix:
Sorry for the issues, now it should work fine. I will push another change to implement correct <picture> handling.


hashrock:
@vorcigernix 
Hi, I fixed the image so that it looks the same in 1x and 2x. If it does not differ from your intent, I will merge them as is. Please check here:

https://fresh-k9m8t6b4g4t0.deno.dev/showcase

vorcigernix:
Thank you

PR #798, opened 2022-09-25
Return thrown responses without raising an error

macieklad:
This PR itroduces new feature - passing through thrown responses

From time to time there is a need to return from handler with a response to abort the execution or to send arbitrary data. It is a simple task, but when the codebase grows, it calls for refactoring into a separate function and it can become cumbersome. 

Lets see an example:

```ts
const handler: Handler = async (req, ctx.next()) => {
  const text = await req.text()
  if (text.includes("badword")) {
    return new Response("We do not allow badmouthing!", { status: 400 })
  }

  return ctx.next()
}
```

If the conditional logic was bigger, it calls for a separate function, but we have to add additional return types, and that leads to the fun with null and undefined values

```ts
export const handler: Handler = async (req, ctx.next()) => {
  const text = await req.text()
  
  // we will have to check what is returned or this does nothing for us now
  ensureNotBadmouthing(text)

  return ctx.next()
}

function ensureNotBadmouthing(text: string) {
  if (text.includes("badword")) {
    return new Response("We do not allow badmouthing!", { status: 400 })
  }
  return "" // Do what?
}
```
What if we could just throw a response instead and do away with such problems?

```ts
export const handler: Handler = async (req, ctx.next()) => {
  const text = await req.text()
  
  ensureNotBadmouthing(text)

  return ctx.next()
}

function ensureNotBadmouthing(text: string) {
  if (text.includes("badword")) {
    throw new Response("We do not allow badmouthing!", { status: 400 })
  }
}
```

This PR achieves exactly that. There is a question if we want to introduce hidden effects, right now we always know when a response is returned. I think we can do that, other frameworks do it for quite some time and I did not hear about any problems and deno does not use thrown responses for any other purposes as far as I am concerned. On the other hand, the resulting code becomes cleaner and any custom handling can be done from try/catch blocks nested in the user code.


macieklad:
The PR got a bit stale, so I'm bumping it, @lucacasonato, could you take a peek in your free time? Thanks!

itohatweb:
Would love to get this in fresh. Personally I use that very often :)

PR #792, opened 2022-09-24
New Project: Fresh Project Manager

xyntechx:
Hey there! I've made another project for https://fresh.deno.dev/showcase! It's a project manager where you can add/delete/complete projects and tasks all saved in localStorage. Check it out [here](https://xyntechx-project-manager.deno.dev/) 😁

PR #803, opened 2022-09-26
init.ts: improve starter project

xyntechx:
Objective:
- Make the starter project look nicer/more professional
- Introduce styling options for devs not using twind

Changes:
- Modified twind classes for `useTwind`
- Added inline styles for `!useTwind`
- Added `styles/index.ts` for styling the home page (since there are a lot of classes to style) for `!useTwind`


xyntechx:
Also, same problem with the checks as #792

xyntechx:
Hi! This PR is getting some errors which I'm not sure how to fix. I would greatly appreciate any help :))

PR #815, opened 2022-09-30
www: add developer portfolio to showcase

mooxl:


khuongduybui (changes requested):
please do not commit .DS_Store

PR #823, opened 2022-10-07
Added LDkit to showcase

karelklima:
Hi, I would like to add the LDkit project to showcase.

Web: https://ldkit.io
Repo: https://github.com/karelklima/ldkit

LDkit is a Deno / Node / browser library for querying Linked Data / RDF. Its website and docs was built using Fresh.

Thanks!


hashrock:
I like this website / docs!

Issue #861, opened 2022-10-25
Image showing at github not on deno.land/x/fresh


codesculpture:
At README.md, this is a very small mistake that the image is showing on github.com/denoland/fresh but not on https://deno.land/x/fresh@1.1.2 at line :5 
```html
<img align="right" src="./www/static/logo.svg" height="150px" alt="the fresh logo: a sliced lemon dripping with juice">
```
It perfectly works at github.com since the logo.svg is at correct path. But at https://deno.land/x/fresh@1.1.2 its not work as it is.
Because current dir can be relative since both are different domains and deno.land just showing the repository homepage not the actual repository. 

so we need to change URL to 
```html 
<img align="right" src="https://fresh.deno.dev/logo.svg" height="150px" alt="the fresh logo: a sliced lemon dripping with juice">```
## Since it is a very small mistake and this is my first PR

codesculpture:
@hashrock Thanks for pointing this..

hashrock:
Thanks!

PR #862, opened 2022-10-25
Fixing README image

codesculpture:
#861 Fixed this small change :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # fresh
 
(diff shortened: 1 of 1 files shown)


hashrock:
There is still a related Issue, but I'll merge this to separate the problem.
If the gfm side can handle this, it should work even if we revert this.

codesculpture:
@hashrock can u brief what the related issue u are talking about and i will explore the deno-gfm for more deep understanding what exactly it is and Thank You

Issue #403, opened 2022-07-04
Feature: Layouts


einSelbst:
I know this might not be the most important feature at an early stage, but something I would love to see.

It might also be something which could be considered when working on https://github.com/denoland/fresh/pull/380

Nextjs support for layouts wasn't really great and now it's the main thing they are working on, https://nextjs.org/blog/layouts-rfc. Probably something good for inspiration.

Someone who experimented with different ways to define layouts (in nextjs), which might also be worth looking at: https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/

ajf-sa:
We can implement layout like this for now [#371](https://github.com/denoland/fresh/issues/371#issuecomment-1172915981)
✍️

antoninbeaufort:
I was wondering, is there any technical reason not to implement a layout/routing system a la Remix? https://remix.run/docs/en/v1/guides/routing
Or is it a non-technical decision?
For me the ideal framework is based on Deno, without build (like Fresh) but also has this layout system (from Remix) which allows to load only a part of the page when changing URL.
If you like this idea, I would love to work on it!

lucacasonato:
> which allows to load only a part of the page when changing URL.

Fresh would never have this, because it uses full page transitions, not client side routing.

lucacasonato:
But layouts in general might be useful. I'd want us to figure out the exact semantics before anyone starts implementing though.

vorcigernix:
Oh. I misunderstood the impact of what @lucacasonato is saying here. Ouch. It is logical if we think about the nature of the framework, but at the same time it disqualifies the framework from some use cases (or I can't figure out how to do it). What it means is that the page (as it is static) always completely rerenders. So if you have a template with navigation, page1 and page2 and navigate from page1 to page2 it completely reloads the page (including the template part). Also, state of page1 is probably lost. 
I am having this exact problem rn, so I will update here. I am thinking of moving the template to the islands and making the subpages a child components of this component. This will result in all pages having some javascript, but I guess there is no way around it.

iamsahilsonawane:
[...]

PR #870, opened 2022-10-28
add middleware-like layouts in routes using HOCs

ssttevee:
This pr implements layouts as described in #403.

There doesn't seem to have been any productive discussion about this since https://github.com/denoland/fresh/issues/403#issuecomment-1176089873, so I took some inspiration from [QuikCity's layouts](https://qwik.builder.io/qwikcity/layout/overview/).

It works very similarily to QwikCity, except for the usage of [higher-order components](https://reactjs.org/docs/higher-order-components.html) instead of a special `<Slot />` component. Alternatively to HOCs, layouts could be a normal component that only takes children as a prop.

There's a little bit more boilerplate with HOCs, but in return, layouts will have access to `PageProps` and also the ability to transform props before passing them to the child component, which makes them the more powerful option, IMO.

A quick example:

```
# ./routes/any/sub/path/_layout.tsx

export default <P extends PageProps>(Component: ComponentType<P>) => {
  return (props: P) => {
    return <div>hello <Component {...props} someExtraProp="foo" />! We're on {props.url.toString()}!</div>;
  };
};
```

In comparison to a children prop.

```
# ./routes/any/sub/path/_layout.tsx

export default ({ children }: LayoutProps) => {
  return <div>hello {children}! I'm lost :(</div>;
};
```

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -21,6 +21,8 @@ import {
   FreshOptions,
   Handler,
   Island,
+  LayoutRoute,
+  LayoutModule,
   Middleware,
   MiddlewareModule,
   MiddlewareRoute,
@@ -134,6 +136,7 @@ export class ServerContext {
 
     // Extract all routes, and prepare them into the `Page` structure.
     const routes: Route[] = [];
+    const layouts: LayoutRoute[] = [];
     const islands: Island[] = [];
     const middlewares: MiddlewareRoute[] = [];
     let app: AppModule = DEFAULT_APP;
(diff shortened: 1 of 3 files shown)


lino-levan:
@marvinhagemeister I'm a fan of this approach. Any thoughts?

src/server/context.ts
```diff
@@ -150,7 +153,14 @@ export class ServerContext {
       const isMiddleware = path.endsWith("/_middleware.tsx") ||
         path.endsWith("/_middleware.ts") || path.endsWith("/_middleware.jsx") ||
         path.endsWith("/_middleware.js");
-      if (!path.startsWith("/_") && !isMiddleware) {
+      const isLayout = Boolean(path.match(/\/_layout!?.[jt]sx?$/));
+      if (isLayout) {
+        layouts.push({
+          top: baseRoute.endsWith("!"),
```
marvinhagemeister, line 159: What does the `!` do? Is it to overwrite the layout inheritance?

marvinhagemeister (commented):
This is a great feature! Layouts are definitely something we want to have in Fresh. Do you have time to add some tests that verify that the feature works as expected? Good things to tests are nested layouts or nested layouts, but the last one overwrites all preceding ones.

iuioiua:
@marvinhagemeister, should we close this in favour of #1506?

PR #872, opened 2022-10-31
add Fresh Components

hashrock:


lucacasonato (commented):
image

hashrock:
I noticed that Button I made was left disabled (it was using IS_BROWSER on a component that was not called from island). I will fix this as well so converted to draft for now

ry:
Looks good!

Would it be difficult to syntax highlight the code?

What about adding a link to the home page?

www/components/gallery/CodeBox.tsx
```diff
@@ -0,0 +1,43 @@
...
+    props.code,
+    Prism.languages.typescript,
+    "typescript",
+  );
+  const onCopy = () => {
+    navigator.clipboard.writeText(props.code);
+    setCopied(true);
+  };
+
+  const [copied, setCopied] = useState(false);
```
crowlKats, line 21: shouldnt this be in an island?

hashrock: CodeBox.tsx is used in `islands/ComponentGallery.tsx`, so I think it is fine as a component.

crowlKats: ah wasnt aware that works; and yea, islands cant import islands, so this is fine as-is.

kt3k (approved):
LGTM. Nice work!

www/components/gallery/CodeBox.tsx
kt3k: Oh, dependencies of islands can use these hooks. This sounds a tricky situation...

PR #886, opened 2022-11-06
add melon to fresh showcase

kibibites:
Melon is a simple link shortener, without using databases(technically it still does, it uses a JSON file to store data) and analytics. Currently, it is deployed on deno.dev


kt3k (approved):
LGTM

PR #892, opened 2022-11-09
ci: split ci of main and www

hashrock:
Separated yml to speed up CI for www


kt3k (approved):
LGTM

Let's test the actual behavior after landed (I believe this change can be properly tested only after landed)

PR #898, opened 2022-11-10
CI Testing PR(Dummy)

hashrock:


hashrock:
this PR is for CI testing for https://github.com/denoland/fresh/pull/892, but there are problems remain...

It appears that status checking is enabled, so it is not possible to merge into main through the www test only. However, it seems that disabling status checking to speed up CI is not a good idea🤔

hashrock:
There seems to be a work-around to create a task with the same name.

https://docs.github.com/ja/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks

hashrock:
We have stopped adopting this workaround and changed the Required status setting. Hopefully someday GitHub will introduce a mechanism to better handle this issue. We need to check CI status when merging for now.

PR #906, opened 2022-11-16
Update running-locally.md

eraykeskinmac:
Listening port is said to be 8000 Changed from 3000 to 8000


hashrock:
Since the default port is 8000 and this is an example of how it can be written if you want to change the port, I don't think it is necessary to change it.

iuioiua:
> Since the default port is 8000 and this is an example of how it can be written if you want to change the port, I don't think it is necessary to change it.

Yes, this appears to be the case. I'll close this issue without merging. Thank you either way, @eraykeskinmac.

PR #908, opened 2022-11-18
feat(dev): Re-export collect and generate

### LINKED ISSUE #384: proposal: API for generating manifest manually without using $fresh/dev.ts
How about re-exporting generate and collect on `fresh/dev.ts`?
I want to regenerate the manifest file(fresh.gen.ts) manually without starting fresh server using [$fresh/dev.ts](https://github.com/denoland/fresh/blob/main/src/dev/mod.ts). 
or it will be awesome if there is simple function just collect and generate at once like below:
```typescript
import { collect, generate } from "$fresh/dev.ts";
import { dirname, fromFileUrl } from "https://deno.land/std/path/mod.ts";

async function manifest(entrypoint: string) {
  const dir = dirname(fromFileUrl(entrypoint));
  
  const manifest = await collect(dir);
  await generate(dir, manifest);
}

// this will generate manifest manually.
await manifest("./main.ts");
```

anishkny:
Fixes #384


dev.ts
```diff
@@ -1,2 +1,2 @@
-import { dev } from "./src/dev/mod.ts";
-export default dev;
```
lino-levan, line 2: Why not leave the default export as dev? This seems like an unnecessary breaking change. Other than this, LGTM!

anishkny:
We need multiple named exports. Cannot use single default export. On Jan 3, 2023, at 1:29 PM, Lino Le Van ***@***.***> wrote:﻿
@lino-levan commented on this pull request.



In dev.ts:
> @@ -1,2 +1,2 @@
-import { dev } from "./src/dev/mod.ts";
-export default dev;

Why not leave the default export as dev? This seems like an unnecessary breaking change. Other than this, LGTM!

deer:
@marvinhagemeister, this can get closed now.

PR #910, opened 2022-11-18
feat: encode and decode Uint8Array props

marcushultman:
Adds a shared module (`src/shared/prop_encoding.ts`, location up for debate) that contains 2 entry points to transform the properties for an island into a JSON compatible format and back. It is used in `src/server/render.ts` to convert to JSON serializable version, and in `src/runtime/main.ts` to convert back into the original state. The initial implementation supports encoding and decoding `Uint8Array` values (nested and in arrays). More types may be added.

**Details:**
Each value in the property object tree goes through a conversion step where the value may be transformed into a [enum type number, data string] tuple which is then packed into a single string value on the form `__FRSH_<enum type 00-FF><data>`. The enum type number is declared in the module as such:
```
enum EncodedType {
  UNKNOWN = 0,
  UINT8ARRAY = 1,
  MAX = 256,
}
```
The type is parsed as a single hex byte (as 255 different types is pretty generous). Since the runtime and server dependencies are bumped together, changing the schema on the server will also change what is served to the client, which mean breaking changes would be allowed, if this needs to be expanded in the future.

The PR includes an additional step where a CRC32 for the value is added to the state, and checked at decoding.

### DIFF
--- a/src/shared/prop_encoding.ts
+++ b/src/shared/prop_encoding.ts
@@ -2,6 +2,7 @@ import {
   decode,
   encode,
 } from "https://deno.land/std@0.150.0/encoding/base64.ts";
+import CRC32 from "https://esm.sh/crc-32@1.2.2";
 
 function replaceValues<T>(
   obj: Record<PropertyKey, unknown>,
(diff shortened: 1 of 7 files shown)


src/shared/prop_encoding.ts
```diff
@@ -0,0 +1,73 @@
...
+  const handleObj = (obj: Record<PropertyKey, unknown>) =>
+    Object.fromEntries(
+      Object.entries(obj).map(([key, val]) => [key, handleValue(val)]),
+    );
+
+  return handleObj(obj);
+}
+
+const VERSION = 1;
+const PREFIX = `__FRSH_${encode(new Uint8Array([VERSION]))}`;
```
marcushultman, line 33: Could go with hex encoding/no encoding here as well.

marcushultman:
@lucacasonato

kt3k:
I think it's very rare to pass Uint8Array in component props. What is the typical use case of this feature?

marcushultman:
@kt3k I think Uint8Array is the most portable form of any binary protocol. My use case is to pass a CRDT (Automerge, both documents and changes). Another use case would be protobuf or flatbuffers. Or image data. Sure, all these things could go as a base64 string, and adding a serialize/deserialize step adds 2-lines (per component). But that is essentially what this PR does, so it's mostly a convenience reason. I know embedding it in HTML will more or less force it to be base64 encoded, but from an API standpoint, I don't see a reason Uint8Arrays should behave like they do today (as `number[]`).

iuioiua:
I'm -1 for this change, as I think it'd be nearly never needed. Welcome to hear more opinions.

lino-levan:
I'm not 100% convinced that Uint8Arrays are the reason to build a decoder system like this, but maybe this is a step towards serializing / deserializing vnodes? I'm not entirely sure. If this system were purely for Uint8Arrays (and none of the other typed arrays / vnodes?) then I'd definitely be leaning for -1 on this.

iuioiua:
We appreciate the PR, @marcushultman, but I'll close this without merging. There doesn't appear to be a common use case for this functionality, so it isn't easy to justify adding it. Either way, thank you for your efforts 😄

lucacasonato:
FYI I'm am already adding this feature via https://github.com/denoland/fresh/pull/757

iuioiua:
So, then there is a valid use case for this 🤣

lucacasonato:
I never said there wasn't! 😆

marcushultman:
> valid use case

@iuioiua There is a difference between a _valid use case_ vs something that is _nearly never needed_ - I think I gave a perfectly [valid use case](https://github.com/denoland/fresh/pull/910#issuecomment-1332064655), but I do respect that there might not be general need to efficiently and easily transfer general purpose binary data between the server and the client 😉

lucacasonato:
I have re-implemented this ontop of #1224 in #1225. Thanks @marcushultman.

PR #911, opened 2022-11-19
showcase: webhook-manager

avnigenc:
Hi all, webhook-manager is a simple webhook callback app. You can use created webhook URL for listening to incoming requests. Currently, it is deployed on deno.dev.


hashrock (approved):
Nice! Thank you

PR #912, opened 2022-11-20
Saved temp png in memory instead of writing to fs - screenshot.ts

codesculpture:
Instead of writing in and reading it again through fs, it can be saved in memory and added fresh showcases to be viewed on new tab it will be good.


codesculpture:
Sorry i was new to PRs i didnt make this PR properly

PR #913, opened 2022-11-20
Saved temp png in memory instead of writing to fs - screenshot.ts

codesculpture:
Edit: I actually done some dumb things, I modified something meaning nothing but i forget to remove those and i pushed with this PR but the following statement is done clear. I will explain at bottom what i made mistake

Taking Screenshots, instead of writing in and reading it again through fs, it can be saved in memory. This is retry of PR already i made PR which is mistaked #912 and i closed that.

I decided to add some icon to look great the showcase page. since i was just thinkin and messing with the source slightly. i dint realize i was in main branch which i make PR from on. After some days, i forget this and done this screenshot PR but that icon thing is unfinished and i pushed with that. so that was a terrible mistake so sorry for that. Hope it never happens again. And again sorry for this messy PR :(


hashrock:
Nice! please apply `deno fmt` :)

codesculpture:
> Nice! please apply `deno fmt` :)

Done😀

PR #924, opened 2022-11-29
Allow islands in sub-folders, multiple islands sources, allow URL islands

mikehw:
This PR enhances island bundling to support better island organization. 

Islands in subdirectories of the `./islands/` directory will work as expected.

This change introduces a way to specify multiple island directories, a different island directory, or glob patterns for islands with a new parameter on the `dev` function.

Additionally, you can pass the URL of an island as an island path and fresh will import the island from the remote correctly.

```ts
await dev(import.meta.url, "./main.ts", {
  islandPaths: [
    "./islands",
    "../shared_islands",
    "https://deno.land/x/fresh@1.1.2/tests/fixture/islands/Counter.tsx",
  ],
  islandGlobs: ["./**/*.island.tsx"],
});
```


src/dev/mod.ts
```diff
@@ -33,18 +25,15 @@ interface Manifest {
   islands: string[];
 }
 
-export async function collect(directory: string): Promise<Manifest> {
+export async function collect(
+  directory: string,
+  islandsPaths: string[],
```
lionel-rowe: Should probably be an optional `options` parameter, with sensible defaults:

```ts
const defaultOptions = {
  islandsPaths: ["./islands"],
};

type Options = typeof defaultOptions;

export async function collect(
  directory: string,
  options?: Partial<Options>,
): Promise<Manifest> {
  const resolvedOptions: Options = {
	...defaultOptions,
	...options,
  };
  // ...
```

lionel-rowe: Or maybe the option should be named `islandsPatterns` instead to allow for incorporating similar logic to https://github.com/denoland/fresh/pull/692, either within the same PR or as a future enhancement. So you could have something like this:

```ts
await dev(import.meta.url, "./main.ts", {
  islandsPatterns: [
    "./islands",
    "./**/*.island.tsx",
    "./**/*.island.jsx",
  ],
});
```

mikehw: Updated to have paths and globs as separate options since it's not trivial to differentiate.

lionel-rowe: Couldn't matching be done with similar logic to `find <...> ! -type d`?

```sh
$ mkdir -p islands/a/b/c
$ touch islands/a/b/c/match.tsx match.island.tsx ignore-me.tsx
$ find ./islands ./**/*.island.tsx ! -type d

./islands/a/b/c/match.tsx
./match.island.tsx
```

This would mean the path `./islands` matches against subdirectories, too, which is a bit different from the current behavior. IMO allowing subdirectories is a more sensible default anyway, but if it's preferred not to do so for some reason, the default option could be switched to `["./islands/*.*"]` (assuming all relevant files will have an extension such as `.tsx`).

Concretely, that could be implemented something like this:

```js
import { join } from "https://deno.land/std@0.150.0/path/mod.ts"

const islandPaths = ["./islands", "./**/*.island.tsx"];

const globify = async (dirOrGlob: string) => {
  if (dirOrGlob.includes("*")) {
    return dirOrGlob;
  }

  try {
    const { isDirectory } = await Deno.stat(dirOrGlob);
    
    return isDirectory ? join(dirOrGlob, "**") : dirOrGlob;
  } catch {
    return dirOrGlob;
  }
};

for await (const glob of islandPaths.map(globify)) {
  for await (const entry of expandGlob(glob)) {
    if (entry.isFile) {
      // ...
    }
  }
}
```

mikehw: I don’t have too much of an opinion. I’ll leave it up to @lucacasonato to make the final call on how he’d like the options to work and update the PR with the decided on way. Would love to see this move forward to make sharing islands possible.

Tims777:
#1185 was opened recently, which would implement the islands in sub-folders part.

deer:
There are three enhancements here:
1. islands in sub-directories
2. multiple island sources
3. URL islands

The first one was already completed by 1185, which is great. It seems like this should probably be closed, however, since it's doing multiple features in one PR. Maybe we could get two new PRs for the 2nd and 3rd enhancements? That would probably help to focus the conversation, making the review process easier. But of course @marvinhagemeister has the final word here.

mikehw:
Closing to create more targeted PRs

deer:
Please make sure to include some documentation as well, otherwise people won't know about your great work!

PR #939, opened 2022-12-07
feat: allow custom entrypoints

Danielduel:
Hi,
I want to have ability to customize how revive works by passing optional, custom entrypoint.
For example... we could autoinject Providers on layout level to share client side state between islands... share fetch states or trigger side-effects.
I am aware that I can use browser memory to share data between islands, but I think it would be much cleaner and easier to manage if we could just customize the entrypoint.
If it will be accepted - I will be superhappy to write tests and documentation (+an example in showcase) for this.

### DIFF
--- a/src/server/bundle.ts
+++ b/src/server/bundle.ts
@@ -45,6 +45,7 @@ export class Bundler {
   #plugins: Plugin[];
   #cache: Map<string, Uint8Array> | Promise<void> | undefined = undefined;
   #dev: boolean;
+  #mainEntryPoint: string | undefined = undefined;
 
   constructor(
     islands: Island[],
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Thanks for opening a PR. I'm not sure about he use case. Instead of passing custom entry points it seems more reasonable to set up shared contexts in `_app.tsx` or layout components which we'll be likely adding at some point. I'm hesitant to add support for custom entry points because that can become hard to debug when it's customized to heavily. Even more so for us maintainers, because then we would need to always ask users if they customized this aspect too.

I really appreciate the exploration though 👍

Danielduel:
I am fine with closing this.
Your suggested solution doesn't make much sense in this context, because the entrypoint is (or was?) very different part of the logic than `_app` is.
The greatest superpower that this PR was giving was that it allowed you to modify how the app is hydrated.

PR #946, opened 2022-12-14
feat: add twind v1 plugin

### LINKED ISSUE #834: Support Tailwind v3
It would be great if fresh could support Tailwind v3. Currently, the twind plugin only supports Tailwind v2.

Tailwind v3 supports among other things custom properties, which are needed for example to use pseudo elements with `after:content-['']`. Also built-in colors without configuration is nice.

trescenzi:
**Opening this as a draft as I realize this isn't compete in terms of documentation but I also kinda wanted to get a conversation started around the plans to do this.**

This is adding a _new_ plugin not updating the old one as that would be a breaking change and I assume that isn't desired here.

## Changes in the twind api

Twind did a lot in the v1 to support the behavior that was already being implemented in the plugin. This allows the plugin to become significantly simplified.

- Getting the rules off of the sheet can be done with a provided `stringify` function.
- `virtual` sheets support a prop `includeResumeData` which implements the behavior previously implemented in the plugin if hydration was required
- Because of this the hydration is also simplified significantly. 

An example of twind calling out this resume behavior is in [their changelog](https://github.com/tw-in-js/twind/blob/49cbec823c2092a5d2b80b51a400f470a26fa960/packages/with-sveltekit/CHANGELOG.md#100-next33).

## Example of usage

I've forked the fresh hackernews example and pushed it up to deno deploy. You can see the code [here](https://github.com/trescenzi/temp-deno-test).

- Version using the existing twind plugin: https://trescenzi-temp-deno-test.deno.dev
- Version using the code in this pr: https://trescenzi-temp-deno-test-4wgbkdynxr00.deno.dev

The interesting piece is clicking into comments as they have islands and require hydration. Running `document.querySelector('#__FRSH_TWIND').innerText.length` on a page with comments returns:

- [Existing twind plugin](https://trescenzi-temp-deno-test.deno.dev/item?id=33971377): `4998`
- [Code in this PR](https://trescenzi-temp-deno-test-4wgbkdynxr00.deno.dev/item?id=33971377): `5237`

I haven't yet identified exactly what's accounting for the modest increase in style size but I'm pretty sure there's nothing obviously duplicated so it's possible it's a change in the base css.

## Things I'm totally unsure of

- Should the import map be updated? Not sure the standard here.
- How might we want to test this?
- Should this simply be a separate plugin? 
- Documentation/autogeneration/whatever else must change. I assume there's a lot more to getting this merged as the twind plugin is rather core to fresh.

Thank you for taking the time to look through this. 😄 In the meantime I think I'm going to publish this as a separate plugin. However please feel free to say that I shouldn't if you'd like to publish an "official" one.

closes #834


noxan:
@trescenzi Love the initiative and happy to support - yet I'm not very experienced with fresh myself. I've [tried a few months ago](https://github.com/denoland/fresh/compare/main...noxan:fresh:twind-next) when `twind` was still pre-release. Let me know if there is some way to help and for me current version `twind` would make building with fresh so much easier (not wondering why my components don't work or back-porting tailwind components from v3 to v2) 👍

trescenzi:
I think the biggest question here is what the maintainers would like to do with regards to managing the upgrade. I assume they wouldn't simply swap out the `twind` plugin because that would break people. Even though the fresh side of the api is basically identical the twind isn't.

trescenzi:
I've dug a bit deeper into the length discrepancy in the style tag. It looks like this comes from how twindv1 does the hydration/resuming.

**Old resume example**
```
.pb-3{padding-bottom:0.75rem}/*1r4zlrdu*/
.pb-1{padding-bottom:0.25rem}/*1r4zlrdu*/
```

**New resume example**
```
/*!dbgidc,y,pb-1*/.pb-1{padding-bottom:0.25rem}
/*!dbgidc,y,pb-3*/.pb-3{padding-bottom:0.75rem}
```

So it looks like the default twind behavior is including a tad more information, namely the class name, beyond just the hash. If I strip out all of the resume data the css lengths are almost identical.

tlgimenes:
Hey folks, any news on this one? I'd really love using tailwind 3 features and classes. Thanks!

trescenzi:
I think the big question here is what the maintainers want to do. I'm unsure if this is even desired. Maybe the future of this plug-in is for it to be external to fresh.

mr-mobster:
Given the fact that `twind` is officially integrated into `Fresh` and even part of the project setup, I would be surprised if the maintainers wanted to keep this plugin separate. For the last month or two it seems like the core authors/maintainers are taking a quick break from `Fresh` after all the hard work they have been doing for `1.0` and `1.1`. I hope that the activity will resume in early winter/spring and this issue will be solved.

bendik:
Not sure what is missing here. @trescenzi would you want to just run `deno fmt` and update the branch to fix breaking ci? 

I'm really hoping to see this merged soon. :)

trescenzi:
Any updates here? Happy to do whatever else is necessary to get this over the line.

hashrock:
looks like it is broken with `https://esm.sh/@twind/core@1.0.3`.
changing import url to `https://esm.sh/v106/@twind/core@1.1.3` works for me.

y3km21:
In this implementation of hydrate, there is some overlap between the server and client side styles.

I tried [pr-946-twindv1-test](https://github.com/denoland/fresh/tree/pr-946-twindv1-test/tests/fixture_twind) with some changes to [Islands.tsx](https://github.com/denoland/fresh/blob/pr-946-twindv1-test/tests/fixture_twind/islands/Island.tsx).
`class={"" + (count > 0 ? "text-red-500" : "")}`
to
`class={"" + (count > 0 ? "text-red-500" : "text-lime-500")}`

islands001
This did not occur with the twindv0.16 plugin.

Is this a problem?

It bothers me...
so [I solved this by filtering client-side cssrules by server-side cssrules.
](https://github.com/y3km21/fresh-twindv1-plugin/blob/v1.1.1/twind/main.ts)

nesterow:
@y3km21 There is a hydration function  [`resume`](https://github.com/tw-in-js/twind/blob/d1fe2ce028b904c3f871264db914ac778c059745/packages/core/src/sheets.ts#L206) in `twind`.  Idk why it was left an easter egg, but it works for me. It's a shame this method doesn't have an export, so [my solution](https://github.com/nesterow/tailored/blob/3d2f6d58aa1d19e7d3442bd135bf3ccfa9b18b2f/system/plugins/twind/hydrate.ts) with copy-paste for now. 
```typescript
  const el = document.getElementById(STYLE_ELEMENT_ID) as HTMLStyleElement;
  const sheet = cssom(el);
  sheet.resume = resume.bind(sheet);
  setup(options, sheet);
```

y3km21:
@nesterow  ohhhhh, I missed the [resume](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L206). I tried it right away... and it worked! This one is better. Thanks!

rojvv:
> This is adding a new plugin not updating the old one as that would be a breaking change and I assume that isn't desired here.

@trescenzi I think like let it be a breaking change.

ry (approved):
LGTM - fixes can be done in follow ups

Thank you [trescenzi](https://github.com/trescenzi)

rojvv:
@ry Don't you think this should be named just `twind`, instead of `twindv1`?

hashrock:
Let's update the `twind.ts` in next major update. There is still work to be done on this, such as init scripts, and I think it is a little early to replace them completely.

trescenzi:
Thanks @ry for approving and merging this. Been swamped with other stuff so this had fallen to the back burner.

saolof:
Is it possible to try out the new plugin before the next major update? Or is using the third party plugin the way to go until then?

rimzudev:
What is the step to get tailwindcssv3 support in Fresh now? Is simply changing these lines in the `import_map.json`:
```
    "$fresh/": "https://deno.land/x/fresh@1.1.3/",
    "twind": "https://esm.sh/twind@0.16.17",
    "twind/": "https://esm.sh/twind@0.16.17/",
```
to this
```
    "$fresh/": "https://deno.land/x/fresh@1.1.4/",
    "twind": "https://esm.sh/twind@1.0.7",
    "twind/": "https://esm.sh/twind@1.0.7/",
```
enough?

Sorry, I do not mean to pollute but it could be helpful for others in my eyes and I haven't found anything yet. I only think it is part of the Fresh1.1.4 release, right?

hashrock:
Yes, this is only part of the fresh 1.1.4. When you do an init, the old twind plugin is still used. I feel that simply replacing the version is not enough, and that it might be a breaking change. I think the best time to update this area is at the time of a major version upgrade.

I think the most helpful code so far is in fixtures: 

https://github.com/denoland/fresh/tree/main/tests/fixture_twind_hydrate

To use twind 1.x, you have to: 

- change plugin path to twindv1.ts : https://github.com/denoland/fresh/blob/main/tests/fixture_twind_hydrate/options.ts#L2
- rewrite twind.config.ts : https://github.com/denoland/fresh/blob/main/tests/fixture_twind_hydrate/twind.config.ts#L1
  - use defineConfig
  - use presetTailwind

rimzudev:
Thank you very much! I think I got it with your help. And now I have to adapt to the new ways the plugin works now. x.x
[Here ](https://twind.style/migration#breaking-changes) is some help for migrating from 0.16.

tunnckoCore:
@toastoaster do you have a repo or something?

> To use twind 1.x, you have to


@hashrock so to fork fresh? uh.

I'm waiting that for months, plus catch up with the latest npm support from deno 1.30+

sant123:
> Yes, this is only part of the fresh 1.1.4. When you do an init, the old twind plugin is still used. I feel that simply replacing the version is not enough, and that it might be a breaking change. I think the best time to update this area is at the time of a major version upgrade.
> 
> I think the most helpful code so far is in fixtures:
> 
> https://github.com/denoland/fresh/tree/main/tests/fixture_twind_hydrate
> 
> To use twind 1.x, you have to:
> 
> * change plugin path to twindv1.ts : https://github.com/denoland/fresh/blob/main/tests/fixture_twind_hydrate/options.ts#L2
> * rewrite twind.config.ts : https://github.com/denoland/fresh/blob/main/tests/fixture_twind_hydrate/twind.config.ts#L1
>   
>   * use defineConfig
>   * use presetTailwind

This is my workaround to make it work:

#### main.ts

```ts
/// <reference lib="dom" />
/// <reference lib="dom.iterable" />

import twindPlugin from "$fresh/plugins/twindv1.ts";
import twindConfig from "./twind.config.ts";

await start(manifest, {
  plugins: [twindPlugin(twindConfig as any)],
  port: 35001,
});

```

#### twind.config.ts

```ts
import { defineConfig } from "https://esm.sh/@twind/core@1.1.3";
import presetTailwind from "https://esm.sh/@twind/preset-tailwind@1.1.4";

export default {
  ...defineConfig({
    presets: [presetTailwind()],
  }),
  selfURL: import.meta.url,
};

```

tunnckoCore:
@sant123 seems to not work for me... Fresh new template, exact same code, exact same versions, refreshed my cache too. Nope.

I don't get why there's no update on the template if everything works fine and there's official way.

tunnckoCore:
Ah okay. Just switched to use the official twindv1 plugin in the `main.ts`, and the `@twind/core@1.1.3` on the `twind.config.ts`.

https://github.com/tunnckoCore/deno-fresh-tailwindV3-template

I also like import maps to be inside the `deno.json` so that's the two changes from the bare template.

PR #954, opened 2022-12-23
showcase: grape-chat

yahiro07:
Hi,

I made a sweet chat app as an example for my css-in-js library. It would be helpful for those who want to style the components with usual css syntax.

Thanks!


hashrock (approved):
LGTM!

PR #959, opened 2022-12-30
chore: remove the +x of dev.ts

Decodetalkers:
When I read the source of fresh, I find a `+x` in www/dev.ts, I think it is a mistake, and feel unconfortable, so..


littledivy:
This is intentional, you can call the script directly because of the shebang

https://github.com/denoland/fresh/blob/main/www/dev.ts#L1

Decodetalkers:
I see

PR #970, opened 2023-01-06
fix: The "Made with Fresh" Badge broke

### LINKED ISSUE #969: The "Made with Fresh" Badge broke
image

> The resource at “https://fresh.deno.dev/fresh-badge.svg” was blocked due to its Cross-Origin-Resource-Policy header (or lack thereof). See https://developer.mozilla.org/docs/Web/HTTP/Cross-Origin_Resource_Policy_(CORP)#

Everywhere. For example: https://merch.deno.com/ and https://showcase-chat.deno.dev/

hashrock:
Close #969

I added a CORS header when I added the ga4 middleware, but this applied to the `static/` files as well, and it causes "Made with Fresh" badge broke.

It may be desirable to remove the CORS header from `static/` files only, but for the time being, I have revert the addition of headers including CORS.

### DIFF
--- a/www/main_test.ts
+++ b/www/main_test.ts
@@ -4,7 +4,7 @@ import { delay } from "$std/async/delay.ts";
 
 Deno.test("CORS should not set on GET /fresh-badge.svg", {
   sanitizeResources: false,
-}, async (t) => {
+}, async () => {
   const serverProcess = Deno.run({
     cmd: ["deno", "run", "-A", "./main.ts"],
     stdin: "null",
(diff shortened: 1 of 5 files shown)


kt3k (approved):
LGTM 👍 

> It may be desirable to remove the CORS header from static/ files only

Let's figure out this later

www/main_test.ts
```diff
@@ -0,0 +1,39 @@
...
+  for await (const line of lines) {
+    if (line.includes("Listening on http://")) {
+      started = true;
+      break;
+    }
+  }
+  if (!started) {
+    throw new Error("Server didn't start up");
+  }
+  await delay(100);
```
kt3k: nit: this delay seems unnecessary?

hashrock: yeah you are right!

Issue #825, opened 2022-10-07
Add the ability to configure a base path


NanoSpicer:
Fresh serves a bunch of other files under a fixed url `/_frsh/` something else.

If we want to serve a fresh app alongside others in the same server we need to match `/_frsh/` and proxy-pass that to the original server, which makes running two different fresh apps into the same server impossible.

If we could configure a `baseUrl` parameter into the server that would get us around this issue

cowboyd:
We have a similar situation, but we are not just running in the same server but across different websites, so the `baseURL` would need to be specifiable on a per request basis.

cowboyd:
It appears that this is a duplicate of #697

marvinhagemeister:
Agree, sounds like a duplicate of #697 . If folks feel different I'm happy to reopen this one.

Issue #697, opened 2022-09-05
Add support for custom root paths


RobGarrard:
Howdy,

Short version: 

I'm trying to host my Fresh app at `https://my-domain-name/homepage`. How do I tell Fresh to use the `/homepage` routing extension , analogous to putting the "homepage" address into package.json for a React app?

Longer version:

I have an EC2 set up running Apache that has some Plotly Dash apps, and I'm trying to use Fresh to build a simple home page for it. (I have very little web-dev experience, this is more a learning exercise).  I want this page to be accessible at `https://my-domain-name/homepage`.

In my Apache config I've added the usual reverse proxy statements:

`ProxyPass  /homepage  http://127.0.0.1:8000`
`ProxyPassReverse /homepage  http://127.0.0.1:8000`

(I've got this set up for some Plotly Dash apps and they all work fiine)

I created a new Fresh boilerplate repo (with the Lemon image and the Counter) and ran it with `deno task start` (so running on port 8000 by default).

When I look at `localhost:8000` on this EC2, all is well. App is running normally.

When I look at `http://my-domain-name/homepage`, the html objects show up but the islands and static assets don't load. When I inspct the network traffic, the first request "GET my-domain-name/homepage" returns 200. But all the subsequent requests to fetch the static assets and JS files are like:

"GET my-domain-name/_frsh/refresh.js"
"GET my-domain-name/logo.svg?__frsh_c=927807f6-b7ac-4bb4-b2ab-d55181719490" 
etc

returning 404s because they're all without the /homepage/ extension.  So how do I tell Fresh to put the `/homepage/` extension on these follow up requests?

Cheers.

lucacasonato:
Fresh does not support this yet. We'll need to add a feature to specify a custom root path in the `FreshOptions`.

RobGarrard:
No worries. Thanks!

cowboyd:
It would be nice to configure this on a per-request basis so that we can change the path under which a Fresh application is accessed without having to re-deploy the application. What about adding support for custom root paths via a header in addition to a static option?

cowboyd:
I created a pull request with a version of this strategy working.... Happy to move things forward in this or in a different direction.

groenborg:
What is the status of this @cowboyd @lucacasonato? We are currently building a new app based on Fresh, but we need to add a custom subpath due to the reverse-proxy configuration. 

e.g. `https://mysite.com/web` <- root of the application

[...]

PR #985, opened 2023-01-13
prefix asset urls with a `X-Fresh-Base-Url` header if present.

cowboyd:
## Motivation

See #697 #825 

In order to "mount" a Fresh website onto a path at another website, the subpath of that website needs to be prepended to all assets it serves.

For example, if I want to mount my fresh site at
`https://example.com/my-fresh`, then whenever Fresh loads the source of an island into the DOM, instead of using `/_frsh/.../island-my-island.js` as the `<script src>` of an island, it should use `/my-fresh/_frsh/.../island-my-island.js` or more fully: `https://example.com/my-fresh/_frsh/.../island-my-island.js`

## Approach

This reads a header `X-Fresh-Base-Url`, and if present, prepends it to any asset urls. This allows the mounting site to dynamically instruct Fresh to act as though it were living at the subpath. 

## Todos

- [ ] Arrive at optimal design accounting for various use-cases
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -459,6 +459,10 @@ export class ServerContext {
           }
 
           const preloads: string[] = [];
+          const base = (req.headers.get("X-Fresh-Base-Url") ?? "").replace(
+            /\/$/,
+            "",
+          );
           const resp = await internalRender({
             route,
             islands: this.#islands,
@@ -468,6 +472,7 @@ export class ServerContext {
             preloads,
             renderFn: this.#renderFn,
             url: new URL(req.url),
+            base,
             params,
             data,
             error,
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -15,7 +15,7 @@ import {
 import { HEAD_CONTEXT } from "../runtime/head.ts";
 import { CSP_CONTEXT, nonce, NONE, UNSAFE_INLINE } from "../runtime/csp.ts";
 import { ContentSecurityPolicy } from "../runtime/csp.ts";
-import { bundleAssetUrl } from "./constants.ts";
+import { bundleAssetUrl as bundleAssetUrlStatic } from "./constants.ts";
 import { assetHashingHook } from "../runtime/utils.ts";
 import { htmlEscapeJsonString } from "./htmlescape.ts";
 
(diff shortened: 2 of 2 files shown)


lucacasonato:
I think this should be a server option, not a dynamic option per request.

cowboyd:
We could do that too, but honestly having this available made our transition from Netlify to deploy go a lot smoother. The flexibility it offered allowed us to try out integrating different preview versions of the deploy site from different preview versions.

hmmm... Is there a downside to having both?

lucacasonato:
> Is there a downside to having both?

Yes, this is a possible security bug. If any user can add `X-Fresh-Base-Url` to their requests, they can serve any asset they want at pretty much any base URL.

@cowboyd I don't understand your usecase. If you had a fixed base URL specified in the main.ts, how would that be different to changing it per request.

There is also the question how this works with the `/_frsh` routes, which with this `X-Fresh-Base-Url` change would not be base URL prefixed correctly AFAICT.

cowboyd:
> @cowboyd I don't understand your usecase. If you had a fixed base URL specified in the main.ts, how would that be different to changing it per request.

I'm not saying that it wouldn't work, only that having the capability to set the base url per request made it a lot easier to migrate. This is because we were able to try it out, and adjust urls with a lot fewer deployments whose only purpose was to change the base url. For example, we could to try it out with a local Netlify server using a local fresh server, then a local Netlify server against a deployed preview version of the Fresh server, then preview on preview, preview and production, production on production, and so on. 

> Yes, this is a possible security bug. If any user can add X-Fresh-Base-Url to their requests, they can serve any asset they want at pretty much any base URL.

I'm no security expert, but what extra capabilities would this give them? I.e. Couldn't they just serve it to themselves from the base url directly?

Anyway, now that we have it working, there is no advantage to have it working per request. It just reduced the amount of "config glue" commits, branches and deployments that we had to fiddle with while standing it up. I'm happy to implement it as a server startup parameter. Where would you recommend that?

iuioiua:
Hi @cowboyd, yes, we'd happily merge this as a server option.

iuioiua:
Hi @cowboyd, I'm closing this PR for now, as it's gone stale. Please feel free to re-open once you're ready to continue to work on it, including making it a server option. Either way, thank you.

PR #986, opened 2023-01-13
chore: bump dependencies and license year

### LINKED ISSUE #984: satisfies keyword causes fatal error
Is there a way to use the `satisfies` keyword? because when I try to do it, it causes a fatal error

```ts
const defaults = {
    type: 'page',
    robots: { index: true, follow: true },
} satisfies {
    type: Types
    robots: Robots
}
```

```ts
X [ERROR] Expected ";" but found "satisfies"
      15 │ } satisfies {
```

lino-levan:
blocked by: https://github.com/lucacasonato/esbuild_deno_loader/pull/22
closes #984 
supersedes: https://github.com/denoland/fresh/pull/952


sigmaSd:
[The blocking pr](https://github.com/lucacasonato/esbuild_deno_loader/pull/22) was merged

lino-levan:
@sigmaSd thanks for the reminder, will update this PR when I get the chance.

lino-levan:
There seems to have been a regression between `1.29.2` and `1.30.0` since the tests pass fine (locally) for `1.29.2` not `1.30.0`

lino-levan:
Honestly just finding niche things to change until the tests pass... I have no idea why they are so flaky.

lino-levan:
@kt3k PTAL!

www/utils/markdown.ts
```diff
@@ -1,7 +1,7 @@
...
-import "https://esm.sh/prismjs@1.27.0/components/prism-tsx.js?no-check";
-import "https://esm.sh/prismjs@1.27.0/components/prism-diff.js?no-check";
+export * as gfm from "https://deno.land/x/gfm@0.1.30/mod.ts";
+import "https://esm.sh/prismjs@1.29.0/components/prism-jsx.js?no-check";
+import "https://esm.sh/prismjs@1.29.0/components/prism-typescript.js?no-check";
+import "https://esm.sh/prismjs@1.29.0/components/prism-tsx.js?no-check";
+import "https://esm.sh/prismjs@1.29.0/components/prism-diff.js?no-check";
 
-export { extract as frontMatter } from "$std/encoding/front_matter.ts";
+export { extract as frontMatter } from "$std/encoding/front_matter/any.ts";
```
kt3k: Do we need `front_matter/any.ts`? I guess `front_matter/yaml.ts` is enough for us

src/server/deps.ts
```diff
@@ -3,29 +3,29 @@ export {
...
+// @deno-types="https://deno.land/x/esbuild@v0.17.2/mod.d.ts"
+import * as esbuildWasm from "https://deno.land/x/esbuild@v0.17.2/wasm.js";
+import * as esbuildNative from "https://deno.land/x/esbuild@v0.17.2/mod.js";
 // @ts-ignore trust me
 const esbuild: typeof esbuildWasm = Deno.run === undefined
   ? esbuildWasm
   : esbuildNative;
 export { esbuild, esbuildWasm as esbuildTypes };
-export { denoPlugin } from "https://deno.land/x/esbuild_deno_loader@0.5.2/mod.ts";
+export { denoPlugin } from "https://raw.githubusercontent.com/lucacasonato/esbuild_deno_loader/b4b3ffb43e9380865ad12ab4d1c64c0ae0ca96cd/mod.ts";
```
kt3k, line 32: Maybe let's leave a TODO comment for replacing this to tagged version in the future

src/server/bundle.ts
```diff
@@ -12,7 +12,7 @@ let esbuildInitialized: boolean | Promise<void> = false;
 async function ensureEsbuildInitialized() {
   if (esbuildInitialized === false) {
     if (Deno.run === undefined) {
-      const wasmURL = new URL("./esbuild_v0.14.51.wasm", import.meta.url).href;
+      const wasmURL = new URL("./esbuild_v0.17.2.wasm", import.meta.url).href;
```
kt3k, line 15: I think we need to vendor this wasm at `src/server/esbuild_v0.17.2.wasm` (`src/server/esbuild_v0.14.51.wasm` is removed with this change)

lino-levan: You're right... nice catch! We probably want CI / testing on this. Opening an issue to be addressed in another PR.

lino-levan:
@kt3k PTAL!

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2021 Luca Casonato
+Copyright (c) 2021-2023 Luca Casonato
```
rojvv, line 3: Judging by the title, I feel like this is out of the scope of this PQ.

src/runtime/utils_test.ts
```diff
@@ -1,6 +1,7 @@
 import { assertEquals } from "../../tests/deps.ts";
 import { asset, assetSrcSet } from "./utils.ts";
 
+// @ts-ignore trust me
```
rojvv, line 4:
```suggestion
// @ts-ignore: trust me
```

LICENSE
lino-levan: Agreed. I can revert, I just don't think this is worth a separate PR.

rojvv: You can then update your title.

rojvv: Maybe to `chore: bump dependencies and license year`.

lino-levan:
@kt3k PTAL

rojvv:
The project is in hibernation

kt3k (approved):
LGTM. Thanks, Lino!

kt3k:
This caused the below error in deploy and islands didn't work unfortunately..

Screen Shot 2023-02-24 at 15 10 15

lino-levan:
Will look into it now. Very weird.

lino-levan:
Looks like it's an issue with specifically the wasm version. We should add tests for that. Has caused trouble at least twice before.

kt3k:
> We should add tests for that. Has caused trouble at least twice before.

Yeah, we need a way to test the rendering with wasm in CI

lino-levan:
After scratching my head at this for some time, I think I've narrowed it down to either being a breaking change to esbuild or a bug with esbuild_deno_loader (because of a breaking change with esbuild). I will investigate more tomorrow and hopefully create a PR with the bug fixed.

Issue #901, opened 2022-11-10
Support for TLS, SSL and HTTPS missing from Deno.Serve


Omig12:
https://github.com/denoland/fresh/blob/88171feb6d9f1129b48deb518b7c00b7c43f2389/src/server/mod.ts#L68

As documented in [Deno.serve()](https://deno.land/api@v1.27.2?unstable&s=Deno.serve) this function seems to be capable TSL through  [Deno.serveTls()](https://deno.land/std@0.147.0/http/server.ts?s=serveTls).

lucacasonato:
I don't understand, what is the request here?

Omig12:
Hi Luca,

A bit more context I'm running an Apache Web Server with a simple HTML Landing page in a low-resource server. I'd Like to migrate from Apache Web Server to Deno + Fresh and reuse this landing page which is plain HTML setting it as the index.html file that servers as the entry point from that static file.

ghost:
Should be resolved using deno >= 1.35.0 and fresh >= 1.3.0

Issue #809, opened 2022-09-29
Add support for https.


haodong404:
The `start()` function in fresh is starting an ordinary web server, but I didn't find a function that would start a server with TLS. So I had to define a function myself using `serveTls()` in Deno. And is there a convenient function for that purpose in the future? For example: `startTls()`?

lucacasonato:
`startTls` sounds like a good idea. I welcome a PR :)

AffogatoAlCaffe:
Should be resolved using deno >= 1.35.0

PR #995, opened 2023-01-23
feat: support serveTls & serve with ServeTlsOptions

ghost:
https://github.com/denoland/fresh/issues/901
https://github.com/denoland/fresh/issues/809

### DIFF
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -1,4 +1,3 @@
-import options from "../../tests/fixture/options.ts";
 import { ServerContext } from "./context.ts";
 import { serve, serveTls } from "./deps.ts";
 export { Status } from "./deps.ts";
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -1,5 +1,6 @@
+import options from "../../tests/fixture/options.ts";
 import { ServerContext } from "./context.ts";
-import { serve } from "./deps.ts";
+import { serve, serveTls } from "./deps.ts";
 export { Status } from "./deps.ts";
 import {
   AppModule,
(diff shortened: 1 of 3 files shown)


jiawei397:
I need to use this function in my project now. When will it be merged?

rnbguy:
hey @lucacasonato any update on this PR? Would love to see this merged :slightly_smiling_face:

src/server/mod.ts
```diff
@@ -60,13 +61,43 @@ export interface DenoConfig {
 
 export { ServerContext };
 
-export async function start(routes: Manifest, opts: StartOptions = {}) {
-  const ctx = await ServerContext.fromManifest(routes, opts);
+export async function start(
+  routes: Manifest,
+  opts: StartOptions | TlsStartOptions = {},
```
rnbguy, line 66: I would prefer to have two separate `start(..., opts: StartOptions)` and `startTls(..., opts: StartTlsOptions)`.

I would be happy to create a new PR.

rnbguy: I cooked up a tls branch on my fork with a straight-forward [patch](https://github.com/denoland/fresh/compare/main...rnbguy:denoland-fresh:tls)

iuioiua: I think we should keep the one function, `start()`, that accepts both interfaces. This keeps it in line with `Deno.serve()`, which will soon be made stable.

rnbguy: Ah. I missed that `Deno.serveTls()` doesn't exist.

If we are following `Deno.serve()` pattern, and not `serve()`/`serveTls()` pattern from `std/http/server.ts`, it makes sense to keep one `start()`.

iuioiua: For context, `serve()` and `serveTls()` from `std` look like they'll be removed once `Deno.serve()` is stabilised (PR [here](https://github.com/denoland/deno_std/pull/3381)).

PR #1000, opened 2023-01-27
Remove twind in favor of plain tailwind extension

reggi:
As of 1.1 components invoke tailwind class names directly, the twind extension isn't needed.


thegaryroberts:
isn't the twind and tailwind syntax very different? In particular around grouping?

PR #1007, opened 2023-01-30
docs: add `_` prefix to unused request params

duffn:
- add types to handlers in the Getting Started section to align with the examples in the Concepts section
- prefix unused arguments with `_` as suggested by official VSCode plugin linting.


thegaryroberts:
These types are inferred though, so it is unnecessary to provide these and makes the framework usage appear more verbose/cumbersome than it has to be.

Perhaps the inferred annotations should instead be removed from the Concepts section and type details provided in a framework API section (which I think is currently a documentation omission, do have to dig into types for info quite a bit).

iuioiua (changes requested):
@thegaryroberts is correct - these types are already inferred. Could you please remove the type annotations?

Though, unused arguments with the `_` prefix are a welcome change.

iuioiua (approved):
I've applied the suggested fix. LGTM now. Thank you.

Issue #596, opened 2022-08-07
Doesn't work with vendoring?


jimisaacs:
Not sure, but this doesn't look good?
```sh
❯ deno run -A -r https://fresh.deno.dev my-project

Download https://fresh.deno.dev/
Download https://deno.land/x/fresh@1.0.2/init.ts
Download https://deno.land/x/fresh@1.0.2/src/dev/deps.ts
Download https://deno.land/x/fresh@1.0.2/src/dev/error.ts
Download https://deno.land/x/fresh@1.0.2/src/dev/mod.ts
Download https://deno.land/std@0.150.0/flags/mod.ts
Download https://deno.land/std@0.150.0/fs/walk.ts
Download https://deno.land/std@0.150.0/path/mod.ts
Download https://deno.land/std@0.150.0/semver/mod.ts
Download https://deno.land/std@0.150.0/_util/assert.ts
Download https://deno.land/std@0.150.0/fs/_util.ts
Download https://deno.land/std@0.150.0/_util/os.ts
Download https://deno.land/std@0.150.0/path/_interface.ts
Download https://deno.land/std@0.150.0/path/common.ts
Download https://deno.land/std@0.150.0/path/glob.ts
Download https://deno.land/std@0.150.0/path/posix.ts
Download https://deno.land/std@0.150.0/path/separator.ts
Download https://deno.land/std@0.150.0/path/win32.ts
Download https://deno.land/std@0.150.0/path/_constants.ts
Download https://deno.land/std@0.150.0/path/_util.ts
Do you want to use 'twind' (https://twind.dev/) for styling? [y/N] y
Do you use VS Code? [y/N] y
The manifest has been generated for 3 routes and 1 islands.

Project created!

In order to start the development server, run:

$ cd my-project
$ deno task start

```
```sh
❯ cd my-project
```
```sh
❯ deno vendor main.ts
Vendored 83 modules into vendor/ directory.

Updated your local Deno configuration file with a reference to the new vendored import map at vendor/import_map.json. Invoking Deno subcommands will now automatically resolve using the vendored modules. You may override this by providing the `--import-map <other-import-map>` flag or by manually editing your Deno configuration file.

```
```sh
❯ deno run -A --no-remote main.ts
Listening on http://localhost:8000/
✘ [ERROR] [plugin deno] The module was missing and could not be loaded.

An error occurred during route handling or page rendering. Error: Build failed with 1 error:
error: The module was missing and could not be loaded.
[...]

PR #1008, opened 2023-01-30
fix: fresh vendor dependencies

tlgimenes:
## What's the purpose of this PR?
This PR solves [issues](https://github.com/denoland/fresh/issues/596) related to running `deno vendor` on a fresh project.

## How does it work?
Vendoring in fresh is a little bit more difficult than in a "normal" deno project because it uses `esbuild` and, sometimes, dependency resolution is not accessible to deno's engine. To overcome this issue, I'm using a technique of explicitly configuring deno's engine to vendor dependencies by using the dynamic import tag. The upside of this approach is not actually importing the dependency but only configuring the vendoring engine to include this dependency. 

This PR fixes both fresh's and twind's plugin missing modules.

## How to test it?
To test it, clone this version, link to a newly created fresh project and run:
```
deno vendor main.ts dev.ts
deno run -A --no-remote main.ts
```

The server should be spawned without any issues.

### DIFF
--- a/plugins/twind.ts
+++ b/plugins/twind.ts
@@ -4,6 +4,9 @@ import { Plugin } from "../server.ts";
 import { Options, setup, STYLE_ELEMENT_ID } from "./twind/shared.ts";
 export type { Options };
 
+// Include it on vendor since it's being included on esbuild.
+const _vendor0 = () => import('./twind/main.ts')
+
 export default function twind(options: Options): Plugin {
   const sheet = virtualSheet();
   setup(options, sheet);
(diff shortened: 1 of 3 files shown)


deer:
@marvinhagemeister, this is already working for me with fresh 1.5.4 and deno 1.38.3.

* `deno run -Ar https://fresh.deno.dev`
* `cd vendor_test`
* `deno vendor main.ts dev.ts`
* (delete `imports` from `deno.json` because `"importMap": "./vendor/import_map.json"` was added)
* `deno run -A --no-remote dev.ts`

Then I get:
```
The manifest has been generated for 5 routes and 1 islands.

 🍋 Fresh ready
    Local: http://localhost:8000/
```

PR #1024, opened 2023-02-10
fix: don't change scheme on redirect

### LINKED ISSUE #1015: redirect switches to http (causing a problem behind a https proxy)
When implementing a POST reqeust from an island like

```
const res = await fetch("/person/", {
      method: "POST",
      headers: {
        "Content-Type": "application/json",
      },
      body: JSON.stringify(person),
    });
```

Fresh generates a redirect to "person" (given the handler is implemented in `routes/person.tsx`). But that redirect point to http://... .

When the Fresh app is deployed behind a HTTPS proxy, the redirect to http:// ... causes some trouble.

Workaround/fix is to use the correct endpoint URL, whoch is person  without a trailing /

But given Fresh implements the helper to redirect to the valid andpoint, I would expect it to be flexible to the protocol used.

It might be related to #80 ?

lucacasonato:
Fixes #1015

This scenario can happen when a user has a proxy sitting in front of Fresh, that does TLS termination, while Fresh is running on plain text HTTP.


hashrock (approved):
LGTM!

src/server/context.ts
```diff
@@ -318,13 +318,17 @@ export class ServerContext {
...
-      // Redirect requests that end with a trailing slash
-      // to their non-trailing slash counterpart.
+      // Redirect requests that end with a trailing slash to their non-trailing
+      // slash counterpart.
       // Ex: /about/ -> /about
       const url = new URL(req.url);
       if (url.pathname.length > 1 && url.pathname.endsWith("/")) {
-        url.pathname = url.pathname.slice(0, -1);
-        return Response.redirect(url.href, Status.TemporaryRedirect);
+        const path = url.pathname.slice(0, -1);
```
```suggestion
        // Remove trailing slashes
        const path = url.pathname.replace(/\/+$/, "");
```

iuioiua (changes requested):
`cleanPathname()` from this file can also be removed.

hashrock (approved):
LGTM!

iuioiua (approved):

PR #1025, opened 2023-02-12
feat: Params for middleware

### LINKED ISSUE #903: Why is `params` available in `HandlerContext` but not in `MiddlewareHandlerContext`?
I wanted to add a middleware that takes care of setting something in the `context.state` according to a parameter in the route (I have a folder named `[repository]`), I know that in a page handler I was able to to use `ctx.params.state` but when putting it into a middleware the type checks complains:
```
Property 'params' does not exist on type 'MiddlewareHandlerContext<CustomType>'.
```
Would it be possible to handle such a case in middle ware instead of having the related code in each route handler?

sylc:
Fix https://github.com/denoland/fresh/issues/903 

This PR introduce `middlewareParams` in the `MiddlewareHandlerContext`.  Each middleware have access to the params that are **upstream** of its definition.

For example: 
 With a route '/api/[id1]/[id2]/foo' 
and a middleware located at '/api/[id1]/_middleware', 
middlewareParams will only have a property of 'id1'. 'id2 will' be undefined

**_Why not all the params of the route ?_**
For a middleware to access the params that are downstream from its level would require, I think, to duplicate in fresh the logic that is currently in the Rutt router. This would introduce a speed decrease. One solution could be to integrate the Rutt router in fresh, although I did not want to tackle this.
While this implementation has a limitation, I think it covers most use cases and does not add any significant overhead


sylc:
Note the linter is failing not because of this PR but because of a recent change in deno affecting the whole project. There a 16 instances of Deno.run being deprecated, which should probably be fixed separately

image

deer:
@marvinhagemeister, don't forget that this is no longer necessary due to my 1314.

marvinhagemeister:
Closing in favour of #1314

PR #1026, opened 2023-02-12
Winidicss support as a plugin via context render function

cagixr:
Hi! I prefer windicss than tailwind, so I decided to add it as a plugin like you've done with twind. I will investigate windi aliases and shortcuts features later, but config plugins like aspect-ratio or forms work!

### DIFF
--- a/plugins/windi.ts
+++ b/plugins/windi.ts
@@ -1,0 +1,60 @@
+import { Plugin } from "$fresh/server.ts";
+import { Config } from "windicss";
+import {
+  generateWindicss,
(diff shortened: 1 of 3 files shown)


bartlomieju:
Closing this PR — WindiCSS has been officially deprecated/sunsetted since 2023 in favor of UnoCSS, so adding first-party support for it in Fresh wouldn't be sustainable. For utility-first CSS, Tailwind and UnoCSS are the actively maintained options. Thanks for the contribution\!

PR #1036, opened 2023-02-17
refactor: remove deno.run

sylc:
remove deno.run in favor of Deno.Command


src/server/bundle.ts
```diff
@@ -11,7 +11,7 @@ export interface JSXConfig {
 let esbuildInitialized: boolean | Promise<void> = false;
 async function ensureEsbuildInitialized() {
   if (esbuildInitialized === false) {
-    if (Deno.run === undefined) {
+    if (Deno.Command === undefined) {
```
fdemir: may you consider to check it by `typeof`??

sylc: I think that in Deno, and other recent js runtime, you cannot anymore re-defined `undefined` and so using typeof is not required. tested in repl
```
$ deno
Deno 1.30.3+e0074b5
exit using ctrl+d, ctrl+c, or close()
REPL is running with all permissions allowed.
To specify permissions, run `deno repl` with allow flags.
> globalThis.undefined = 1
Uncaught TypeError: Cannot assign to read only property 'undefined' of object '#<Window>'
    at <anonymous>:2:22
```
Is there any other reasons to use `typeof` ?

jheysaaz:
The actions has failed because `Deno.Command` is unstable, please add `--unstable` flag to Ci workflows.
Failed ci workflow

deer:
`Deno.Command` is going to be stabilized soon. For context: https://github.com/denoland/deno/pull/17628. Seems to be targeted for 1.31 (unless I can't read properly).

lino-levan:
`Deno.Command` has been stabilized. Any chance this could get merged @hashrock?

www/main_test.ts
```diff
@@ -2,16 +2,18 @@ import { assertEquals } from "$std/testing/asserts.ts";
 import { TextLineStream } from "$std/streams/delimiter.ts";
 
 Deno.test("CORS should not set on GET /fresh-badge.svg", {
+  sanitizeOps: false,
```
hashrock: is there any way to avoid using `sanitizeOps: false`?

sylc: @hashrock i have pushed a change. PTAL thanks.

src/server/bundle.ts
lucacasonato: I would not like to change this specific check right now, due to some future proofing work I need to do in `x/esbuild_deno_loader`. Can you revert this specific change?

src/server/deps.ts
```diff
@@ -24,7 +24,7 @@ export * as rutt from "https://deno.land/x/rutt@0.0.14/mod.ts";
 import * as esbuildWasm from "https://deno.land/x/esbuild@v0.14.51/wasm.js";
 import * as esbuildNative from "https://deno.land/x/esbuild@v0.14.51/mod.js";
 // @ts-ignore trust me
-const esbuild: typeof esbuildWasm = Deno.run === undefined
+const esbuild: typeof esbuildWasm = Deno.Command === undefined
```
lucacasonato: This one also.

tests/cli_test.ts
```diff
@@ -332,62 +324,49 @@ Deno.test({
   sanitizeResources: false,
 });
 
-Deno.test("fresh-init error(help)", async function (t) {
+Deno.test("fresh-init error(help)", { sanitizeOps: false }, async function (t) {
```
lucacasonato: Why is the `sanitizeOps: false` necessary? There is a bug here somewhere if this is needed.

sylc: the precedent test had also sanitiseOps false, which was leaking into this test . I  have now cleanup the precedent test.

sylc:
@lucacasonato @hashrock Could this be merged? (the issue blocking the previous auto-merge has now been resolved). thanks

hashrock (approved):

PR #1039, opened 2023-02-20
feat(dev): Ignore common test files in routes and islands

bastilian:
In some cases developers might want to put test files right next to their code, like `...something.something.test.tsx`. For such cases this PR would allow to provide a pattern for files to skip when walking the routes and or islands folders on dev bootup.

Example use:
```ts
import dev from "$fresh/dev.ts";

await dev(import.meta.url, "./main.ts", {
  router: {
    ignoreFilePattern: /[\.|_]cy[\.|_][t|j]s(x)?$/
  }
});
```


crowlKats:
is there any usecase for specifying anything besides tests?

bastilian:
@crowlKats I believe test files would be the only usecase. It's something someone asked about in Discord[0].

[0] https://discord.com/channels/684898665143206084/991511118524715139/1076945536596328539

crowlKats:
If test files are the only usecase, then i think it would make sense for fresh to ignore test files.

bastilian:
@crowlKats Thank you for taking a look at this PR. It's now updated to skip the test files by default.

src/dev/mod.ts
```diff
@@ -10,6 +10,7 @@ import {
 import { error } from "./error.ts";
 
 const MIN_DENO_VERSION = "1.25.0";
+const TEST_FILE_PATTERN = /[\.|_]test[\.|_][t|j]s(x)?$/;
```
lucacasonato:
```suggestion
const TEST_FILE_PATTERN = /[\.|_]test\.[t|j]s(x)?$/;
```

`foo_test_tsx` should not match. Can you add some unit tests for this regexp, or an integration test that ensures test files are ignored?

bastilian: The regex is updated and there is now a test for `collect`, which for now only tests that the test files are not included.

src/dev/mod.ts
```diff
@@ -78,7 +80,10 @@ export async function collect(directory: string): Promise<Manifest> {
       }
       if (entry.isFile) {
         const ext = extname(entry.name);
-        if (![".tsx", ".jsx", ".ts", ".js"].includes(ext)) continue;
+        if (
+          ![".tsx", ".jsx", ".ts", ".js"].includes(ext) ||
+          (entry.name.match(TEST_FILE_PATTERN) !== null)
```
crowlKats: nitpick: use regex.test instead of string.match

bastilian: Not nitpicking at all, `.test` makes much more sense! Thank you for the suggestion @crowlKats!

severindupouy:
Hi,
it would be great to make regex customizable. Having default regex, like the one you did. But having a way to override it. 
Maybe something like this :

```typescript
// main.ts
// default case, testFilesRegex = "/[\.|_]test\.[t|j]s(x)?$/"
await start(manifest);
// override case
await start(manifest, { testFilesRegex: '.spec.ts$' });
```

Just an idea. What do you think ?

bastilian:
@crowlKats @lucacasonato Is there anything I should still do to this PR to get it mergeable? 

I can, if it's wanted added a way for overriding or extending the patterns for ignoring files, as suggested by @severindupouy.

iuioiua:
What does a dev do if they have a route with the word `test` in the name? I can imagine a very confused dev in the future wondering why their route is not being generated.

I suggest adding a `skip: RegExp` (or similar-named) property to `StartOptions`. By default, the value of this property would be the one currently implemented, accompanied by a small comment above explaining the pattern. This also allows the dev to change this behaviour to suit their scenario.

What do you guys think? Either way, the documentation should briefly mention the decided-upon functionality.

iuioiua:
@bastilian, would you be able to resolve the merge conflicts and rebase? It'd be great to get this one landed. It'd be a big win for testing by allowing testing files to live right next to their implementation counterparts.

marvinhagemeister:
+1 I think this is a great feature to have

bastilian:
I finally had a chance to get back to this. The PR is now rebased and the conflicts are resolved, @iuioiua.

It now has also (added back) the ability to configure the pattern (see PR descriptions).

src/dev/mod.ts
```diff
@@ -50,7 +53,10 @@ export interface Manifest {
 }
 
 const GROUP_REG = /[/\\\\]\((_[^/\\\\]+)\)[/\\\\]/;
-export async function collect(directory: string): Promise<Manifest> {
+export async function collect(
+  directory: string,
+  options: FreshOptions,
```
bastilian: Should the `options` here be optional? I went with non-optional, since the `dev` function also has it as non optional.

src/server/types.ts
```diff
@@ -19,6 +19,11 @@ export interface RouterOptions {
    *  @default {false}
    */
   trailingSlash?: boolean;
+  /**
+   *  Allows configuring a pattern of files to ignore in islands and routes
+   *  @default {undefined}
```
iuioiua: This is `/[\.|_]test\.[t|j]s(x)?$/`, by default. Right? Perhaps, it'd be worth having a short explanation. E.g. By default, ignores all test files with `.test` or `_test` as the filename's suffix.

tests/dev_test.ts
iuioiua, line 1: Nice and simple 👍🏾

iuioiua (changes requested):
Mostly LGTM! Just a nit regarding the `ignoreFilePattern` RegExp.

bastilian:
Thank you @iuioiua for reviewing! I updated the text for the config option. I also realised that while I added a fixture for the custom pattern it wasn't used in a test, there is one now for that too.

src/dev/mod.ts
iuioiua: Perhaps this parameter should be optional.

bastilian:
Hm, I'm not 100% sure why the assertion on windows fails, I'd suspect that it's something with windows using `/`... I'll try to see if I can reproduce it on one of my windows machines and fix the issue.

iuioiua:
How about trying something like `globToRegExp("{*_,*.,}test.{ts, tsx, mts, js, mjs, jsx}")`?

References:
* https://deno.land/std@0.201.0/path/mod.ts?s=globToRegExp
* https://deno.land/manual@v1.36.4/basics/testing#running-tests

iuioiua (approved):
LGTM! Thank you, @bastilian.

bastilian:
@iuioiua Thank you! I did try `globToRegExp`, but the Regex it produces is anchored at the start. I did however use the resulting Regex and removed the start anchor to match files.

src/dev/mod.ts
```diff
@@ -1,7 +1,9 @@
 import { gte, join, posix, relative, walk, WalkEntry } from "./deps.ts";
 import { error } from "./error.ts";
-
+import { FreshOptions } from "../server/mod.ts";
 const MIN_DENO_VERSION = "1.31.0";
+const TEST_FILE_PATTERN =
+  /(?:[^/]*_|[^/]*\.|)test\.(?:ts|tsx|mts|js|mjs|jsx|)\/*$/;
```
marvinhagemeister: Nit: this regex can be simplified a bit.

```suggestion
const TEST_FILE_PATTERN = /[._]test\.(?:[tj]sx?|[mc][tj]s)$/
```

bastilian: Updated! Thank you @marvinhagemeister for looking over it as well!

marvinhagemeister (approved):

PR #1040, opened 2023-02-20
docs: Add example of submitting a form using POST method

matthewmorek:
As per title, adds a super simple example of handling a `POST` request within Fresh:
- access `req.formData()`
- access a specific input value submitted in the form (`email`)


docs/getting-started/form-submissions.md
```diff
@@ -71,6 +72,42 @@ query set as the `q` query parameter in the URL. The `GET` handler will then
 filter the names array based on the query, and pass it to the page component for
 rendering.
 
+## `POST` example
+Here's an example processing a simple form submission request:
```
lucacasonato: simple is not a good word to use in technical documentation: https://developers.google.com/style/word-list#simple

docs/getting-started/form-submissions.md
```diff
@@ -71,6 +72,42 @@ query set as the `q` query parameter in the URL. The `GET` handler will then
...
+
+export const handler: Handlers = {
+  async GET(req, ctx) {
+    return await ctx.render();
+  },
+  async POST(req, ctx) {
+    try {
+      const form = await req.formData();
+      const email = form.get("email");
+      return new Response(JSON.stringify({ email }));
```
lucacasonato: There should be a check here that the returned value exists, and is a string (as opposed to being a `File`).

docs/getting-started/form-submissions.md
```diff
@@ -71,6 +72,42 @@ query set as the `q` query parameter in the URL. The `GET` handler will then
...
+```tsx
+// routes/subscribe.tsx
+import { Handlers } from "$fresh/server.ts";
+
+export const handler: Handlers = {
+  async GET(req, ctx) {
+    return await ctx.render();
+  },
+  async POST(req, ctx) {
+    try {
```
lucacasonato: This `try {} catch {}` is not needed. Handler errors are caught already in fresh by default and result in a 500 error page display.

docs/getting-started/form-submissions.md
```diff
@@ -26,6 +26,7 @@ Fresh can handle both GET and POST requests through the
 any necessary processing on the form data, and then pass data to the
 `ctx.render()` call to render a new page.
 
+## `GET` example
```
lucacasonato:
```suggestion
```

docs/getting-started/form-submissions.md
lucacasonato:
```suggestion
## POST request with `multipart/form-data`

Forms can either submit as a GET request with URL search parameter encoding, or as a POST request with `multipart/form-data`.

This example demonstrates how to handle `multipart/form-data` `<form>` submissions:
```

lucacasonato (commented):
I agree this is something useful to add. I feel like this is more something for the "Concepts" page, not "Getting Started" however (getting started can backlink to the concepts page).

I would suggest we create a new concepts page on forms, and include this there. We could also mention how after a `POST` form submission you should likely respond with a `303` redirect and stuff like that.

@lambtron or @hashrock do you want to work on that new docs page?

docs/getting-started/form-submissions.md
matthewmorek: Agreed. Old habits die hard.

docs/getting-started/form-submissions.md
```diff
@@ -71,6 +71,45 @@ query set as the `q` query parameter in the URL. The `GET` handler will then
...
+// routes/subscribe.tsx
+import { Handlers } from "$fresh/server.ts";
+
+export const handler: Handlers = {
+  async GET(req, ctx) {
+    return await ctx.render();
+  },
+  async POST(req, ctx) {
+      const form = await req.formData();
+      const email = form.get("email")?.toString() || null;
```
severindupouy: why `|| null` when `.get` signature is `FormDataEntryValue | null` ?

```suggestion
      const email = form.get("email")?.toString();
```

docs/getting-started/form-submissions.md
```diff
@@ -71,6 +71,45 @@ query set as the `q` query parameter in the URL. The `GET` handler will then
...
+      if (email && typeof email === 'string') {
```
severindupouy: why this check about `typeof email === 'string'` when there is a `.toString()` before ? `email` type is `string | undefined`.

```suggestion
      if (email) {
```

severindupouy (changes requested):

docs/getting-started/form-submissions.md
matthewmorek: I've seen this pattern in a few other examples and picked it up. Not seen anything in docs and I admit, I should've checked the API.

severindupouy: ok, no pb. ;)

severindupouy (approved):

lambtron:
i re-orged the Form docs as per @lucacasonato suggestion in this PR: https://github.com/denoland/fresh/pull/1104

hashrock:
Closing with #1104. thank you @matthewmorek !

PR #1041, opened 2023-02-21
Add padding to docs content

Abogical:
The docs content don't look good on desktop. Text is too close to the edge:

Screenshot_20230220_204106

This patch makes it a bit better:
Screenshot_20230220_204200

### DIFF
--- a/www/routes/docs/[...slug].tsx
+++ b/www/routes/docs/[...slug].tsx
@@ -153,7 +153,7 @@ function Content(props: { page: Page }) {
         {props.page.title}
       </h1>
       <div
-        class="mt-6 markdown-body"
+        class="mt-6 markdown-body p-2 rounded"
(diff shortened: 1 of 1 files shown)


hashrock:
about your screenshot, did you add gray background or is it as-is?

Abogical:
I didn't add it. It was already there.

hashrock:
thanks, hmm strange. I can't see any gray background in my environment (osx chrome with light mode)

Abogical:
Seems to depend on the browser. Opening in chromium instead of Firefox does not give a grey background.

I'm closing this PR as a result. Should I open an issue on this?

hashrock:
Thanks, it fine with be closed for now.

We have similar issue like you mentioned, so If you can think of any possible causes, please let me know.

https://github.com/denoland/dotland/issues/2700

PR #1047, opened 2023-02-24
fix: revert "chore: bump dependencies and license year (#986)"

kt3k:
This PR temporarily reverts commit ed669865003cbcd9b47dd2e4975c0e7b2a3151a9.

#986 caused the error like the below when wasm version of esbuild is used:

```
✘ [ERROR] Cannot read file "www/twind.config.ts": not implemented on js

    data:application/javascript,import hydrate from "file:///src/plugins/twind/main.ts";import options from "file:///src/www/twind.config.ts";export default function(state) { hydrate(options, state); }:1:76:
      1 │ ...rt options from "file:///src/www/twind.config.ts";export default...
        ╵                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

cc @lucacasonato

### DIFF
--- a/.vscode/import_map.json
+++ b/.vscode/import_map.json
@@ -14,6 +14,6 @@
     "@preact/signals": "https://esm.sh/*@preact/signals@1.0.3",
     "@preact/signals-core": "https://esm.sh/@preact/signals-core@1.0.1",
 
-    "$std/": "https://deno.land/std@0.177.0/"
+    "$std/": "https://deno.land/std@0.150.0/"
   }
 }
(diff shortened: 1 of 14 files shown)


lino-levan:
Unfortunate but probably a good call until we figure out what's going on. I think for testing of wasm we might be able to temporarily remove Deno.run (i.e. `Deno.run = undefined`) instead of adding a config option. This better emulates deploy anyways.

crowlKats (approved):

PR #1050, opened 2023-02-26
fix: hydrate in twind-v1-plugin 

y3km21:
This PR is a slightly modified version of hydrate from twind-v1-plugin (#946).

In this plugin, Cssrules of SSR are resumed by [cssom()](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L23) of twind.
There are demos for each of #946 and this PR. (I added a lot of classes to compare the performance.)

- #946 plugin's [Demo](https://y3km21-fresh-twind-pr946.deno.dev/) ([source](https://github.com/y3km21/ex-fresh-twind-pr946))

- This PR plugin's [Demo](https://y3km21-fresh-twind-resume.deno.dev/) ([source](https://github.com/y3km21/ex-fresh-twind-resume-working))

## What is the issue with #946 plugin?

Twind's [getSheet](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L177) returns the sheet created from [cssom](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L23) to which [resume is assigned](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L184), but since SSR's HtmlStyleSheet cannot be targeted to this sheet, [resume](https://github.com/tw-in-js/twind/blob/main/packages/core/src/sheets.ts#L206) at setup doesn't make sense.
As a result, CSSRules of ClientSide that partially overlaps with ServerSide is inserted.
[Demo](https://y3km21-fresh-twind-pr946-j8dzftmr1s90.deno.dev/)
pr946-demo-styles

I think this should be improved.

## What was changed?
I made a sheet created by cssom with SSR's HtmlStyleSheet as the target, and bind the resume of the sheet created by getSheet.
```ts
export default function hydrate(options: TwindConfig) {
  const elem = document.getElementById(STYLE_ELEMENT_ID) as HTMLStyleElement;
  const sheet = cssom(elem);

  sheet.resume = getSheet().resume.bind(sheet);
  document.querySelector('[data-twind="claimed"]')?.remove();

  setup(options, sheet);
}
```
[Demo](https://y3km21-fresh-twind-resume-2f9vnxg7z7d0.deno.dev/)
thisPR-demo-styles

The resume of the sheet is working, so there is no duplication.

## What about performance?
I compared the performance of each plugin on chrome.

- #946 plugin's [Demo](https://y3km21-fresh-twind-pr946.deno.dev/) ([source](https://github.com/y3km21/ex-fresh-twind-pr946))
pr946-demo-per

- This PR plugin's [Demo](https://y3km21-fresh-twind-resume.deno.dev/) ([source](https://github.com/y3km21/ex-fresh-twind-resume-working))
thisPR-demo-per

The Scripting time for each is about the same.

---

Thank you very much for reading through to the end.
Please let me know if my understanding is wrong.🙏


ry (commented):
Thanks!

Is it possible to write a test for this fix?

y3km21:
@ry Yes, I will do it soon!
I have a few questions about creating a test.
- Should I create a test like `fresh/tests/fixture_twind-pr1050`?
- Should the content of the test be such that this PR passes and #946 fails?(For example, compare stylesheets in islands)

hashrock:
Thank you!

> Should I create a test like fresh/tests/fixture_twind-pr1050?

just `fresh/tests/fixture_twind` is also fine!

> Should the content of the test be such that this PR passes and https://github.com/denoland/fresh/pull/946 fails?(For example, compare stylesheets in islands)

Yes, it would be helpful!

y3km21:
@hashrock Thanks!
I'll write a test.

y3km21:
I added test!
The test contents are as follows
1. Ensure that twind compiles class in static page.
1. Ensure that twind does not cause duplication of cssrules in islands.
1. Ensure that twind compiles dynamically inserted classes in islands.

`deno task test` results would look like this
| Test No. | #1050 | #946 | twind v0.16 plugin |
|---|---|---|---|
|1|:white_check_mark:|:white_check_mark:|:white_check_mark:|
|2|:white_check_mark:|:x:|:white_check_mark:|
|3|:white_check_mark:|:x:|:white_check_mark:|

Please let me know if I am missing or wrong!

tests/twind_test.ts
```diff
@@ -0,0 +1,329 @@
...
+    });
+
+    await browser.close();
+
+    await lines.cancel();
+    serverProcess.kill("SIGTERM");
+    serverProcess.close();
+  },
+  sanitizeOps: false,
+  sanitizeResources: false,
```
ry: is this necessary?

tests/twind_test.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+    if (!started) {
+      throw new Error("Server didn't start up");
+    }
+
+    await delay(100);
+
+    const browser = await puppeteer.launch({ args: ["--no-sandbox"] });
+    const page = await browser.newPage();
+
```
ry: lots of duplicate code here - can this be abstracted into a function?

tests/twind_test.ts
```diff
@@ -0,0 +1,329 @@
...
+Deno.test({
+  name: "twind static test",
+  async fn(t) {
+    // Preparation
+    const serverProcess = Deno.run({
+      cmd: ["deno", "run", "-A", "./tests/fixture_twind/main.ts"],
+      stdout: "piped",
+      stderr: "inherit",
+    });
```
ry: We've just stabilized the `Deno.Command` API: https://deno.com/blog/v1.31#api-stabilizations

Can you use that instead?

tests/twind_test.ts
```diff
@@ -0,0 +1,329 @@
...
+    }
+    if (!started) {
+      throw new Error("Server didn't start up");
+    }
+
+    await delay(100);
+
+    const browser = await puppeteer.launch({ args: ["--no-sandbox"] });
+    const page = await browser.newPage();
+
```
ry: deduplicate

tests/fixture_twind/islands/CheckDuplication.tsx
```diff
@@ -0,0 +1,65 @@
...
+                      cssRulesFRSHTWIND,
+                      cssRulesClaimed,
+                    )
+                  }`}
+                </p>
+                <p>cssrules are duplicated</p>
+              </div>
+            );
+          } else if (cssRulesFRSHTWIND != null && cssRulesClaimed == null) {
+            return <p id="okNoDuplicates">Ok : No dupicates</p>;
```
ry: typo "dupicates"

tests/twind_test.ts
```diff
@@ -0,0 +1,329 @@
+import { assert, delay, puppeteer, TextLineStream } from "./deps.ts";
+
+import { cmpStringArray } from "./fixture_twind/utils/utils.ts";
```
ry: maybe rename the directory `fixture_twind` to `fixture_twind_hydrate`

ry (commented):
Thanks for the test. I think this looks good but it could use some clean up still.

y3km21:
@ry Thanks for the review!
I have corrected the areas you reviewed.

But there is one thing I couldn't solve.
The textDecoderStream called here leaks.
https://github.com/denoland/fresh/blob/c4cfb201e72908ae6d0cc9993310a9baa78891dd/tests/twind_test.ts#L23-L25
I have tried calling cancel() of `lines` and  externally TextDecoderStream (ex. `const decoder = TextDecoderStream();`), but could't solve it.


For now, I got the textDecoder's rid from the resourceMap and manually close the resource.
https://github.com/denoland/fresh/blob/c4cfb201e72908ae6d0cc9993310a9baa78891dd/tests/twind_test.ts#L47-L63

Is there a correct cancel() call?
Or should I add `sanitizeResources: false`?

y3km21:
Without `--trace-ops`, a leak occurred, so I added `delay(10)` before `t.step`.

hashrock (approved):

y3km21:
@hashrock Thanks for merging!

PR #1078, opened 2023-03-14
chore: remove stale workaround in `collect()`

gflarity:
The issue in question, https://github.com/denoland/deno_std/issues/1310, has been marked complete. Looks like the underlying issue has been fixed. So this just removes the comment/unnecessary code.


hashrock:
I think the thread comment said we are having problems with Ubuntu, am I wrong?

iuioiua:
@gflarity, there's one more instance of this code block within the codebase. Can you please remove that too?

ayame113:
I think the assertion on line 287 should also be updated to pass the test.
Previously `Deno.readDir()` threw `Deno.errors.NotFound`, now `walk()` throws `WalkError`.

https://github.com/denoland/fresh/blob/366c8afef1d8db1d667abf36c66e2d06d17c451a/src/server/context.ts#L286-L291

According to https://github.com/denoland/deno_std/pull/3054, it seems you can use `error.cause`. Maybe it will be like this?

```diff
-    if (err instanceof Deno.errors.NotFound) {
+    if (err.cause instanceof Deno.errors.NotFound) {
```

-----

Note: Older versions of std's `walk()` had the problem of throwing an `Error` instead of a `Deno.errors.NotFound` (https://github.com/denoland/deno_std/issues/1310). At that time, `Deno.readDir()` was used to detect `Deno.errors.NotFound` as a workaround.
Now that the error thrown from `walk()` has changed to something more meaningful in https://github.com/denoland/deno_std/pull/3054, we can refactor this part.

iuioiua:
Good spot, @ayame113! I've applied the fix. Turns out, we're on an old version of `std` too. I'll merge this once #1201 is landed.

PR #1082, opened 2023-03-21
test: island with wasm-esbuild

hashrock:
To prevent regression with island + deno deploy(wasm env)

https://github.com/denoland/fresh/pull/1047


kt3k (approved):
LGTM

PR #1084, opened 2023-03-22
feat: build step

### LINKED ISSUE #1062: Pre-build islands
I have a couple decently sized islands (https://github.com/ts-why/tswhy/tree/main/islands) where dynamically building them on Deno Deploy takes quite a lot of time, to the point they can regularly exceed the time limit on Deploy, meaning that to gain full functionality an initial user has to reload several times.

While having "no build step" is great in some cases, it would actually be a huge performance benefit and better user experience to statically build these islands.

Also, there doesn't seem to be any easy tooling to manage the size/complexity of islands.  While esbuild underneath does a great job tree shaking, the whole island build process is opaque and you can't really tell if you are dragging in a load of junk into your island which then esbuild then is shaking out (or even worse, unintended unshaken code).

lino-levan:
Most of this code is pretty gross. Obviously not ready to be merged yet, just a demo. Closes #1062


iuioiua:
Hey Lino, if you say that pre-building everything, which I assume means islands, routes and components, is simpler, let's go with that approach first. Couple notes:
1. Please use `INTERNAL_PREFIX` from `src/runtime/utils.ts` instead of hardcoding the use of `/_frsh`.
2. I think this functionality should be set by a `preBuild? = false` property in the server options.

It'd be good to see this function landed.

lino-levan:
Cool, thanks for the review of the PoC. I'll clean this up this weekend if I have time.

lino-levan:
This is being accomplished in #1604.

PR #1088, opened 2023-03-24
Showcase: Add fresh slides to the showcase

digitaldesigndj:
and I will close issue #1085

Issue #1085, opened 2023-03-23
Docs: Slideshow Island Demo


digitaldesigndj:
I have a slideshow Island and I was thinking about adding it to the Components section of the website. My island is on the larger side, so I am not sure if it belongs. Is there a vetting process?

The Demo
https://slides.deno.dev

The Island
https://github.com/Hyprtxt/slides.deno.dev/blob/main/islands/Slideshow.jsx

Also I just have a need to post this somewhere so people can find Fresh Slideshow code. 😸

hashrock:
Great! feel free to add it to fresh components.
I think it might be a bit buried if you just post it on showcase.

PR #1096, opened 2023-03-31
docs: Carousel Component

digitaldesigndj:
Carousel Component added to the site as requested in Pull Request #1088 also related to #1085


digitaldesigndj:
When I do `deno check main.ts dev.ts` locally it seemed to be using more lenient settings... I worked it out using the actions, but why is more strict on actions? (example I can't use type :any on the action)

www/components/gallery/Carousel.tsx
```diff
@@ -0,0 +1,196 @@
...
+          break;
+        case "ArrowRight":
+          event.preventDefault();
+          nextSlide();
+          break;
+        default:
+          break;
+      }
+    };
+    document.addEventListener("keydown", keydownHandler);
```
hashrock: We might want it to respond to arrow keys only when this component has focus

hashrock (changes requested):
Looks great!

> but why is more strict on actions?

I don't know why this is happening. I think it's probably not the intended setting.

digitaldesigndj:
I don't think I know how to do that without adding a whole bunch of code, and I don't think that really fits with the spirit of what I was trying for. How about I just remove the keyboard listeners altogether?

digitaldesigndj:
So the way I see to make this work is to add another signal "focused" then I have to toggle that value with the `onBlur` and `onFocus` events. Then I check "focused" in my keyboard handler and take action from there.

Is there is a simpler way I am missing? `document.activeElement` doesn't seem to do the thing I am thinking of.

hashrock:
how about this?

- add `tabindex=0` to element
- addEventListener to element instead of document

hashrock:
hmm, I think it doesn't need to use signal.
do you think we can use useState instead?

digitaldesigndj:
I think there is a bug in the way I did the useState. My first idea is to make it just a fade transition instead of a Carousel, I have to backtrack a bit to figure out how to use the different state API.

Also, I am trying to understand what the use of the useSignal hook is if you prefer useState in components? I like signals because the API is cleaner IMHO, is there more than that to it?

hashrock:
Oh sorry, I thought that signals was only used for communication between islands. But the preact docs show such as usecase for local state.
https://preactjs.com/guide/v10/signals/

If that's why you prefer the signals API, then I guess you can just use signals here.

hashrock (approved):
LGTM

PR #1100, opened 2023-04-03
Feat change prefix

### LINKED ISSUE #1066: INTERNAL_PREFIX can change it to the relative path [./_frsh]?
INTERNAL_PREFIX is [/_frsh]. Can you change it to the relative path [./_frsh]?
Since I need to use nginx as the reverse proxy, /_frsh is not easy to handle.
thank you very much.

jiawei397:
Fix: https://github.com/denoland/fresh/issues/1066

I tried to use relative path as `./_frsh`, but it may not work when use sub routes. The best and simple way to resolve the proxy problem seems to add an env variable to enable project to change the prefix self.

You can use `INTERNAL_PREFIX=/example/_frsh/ deno task www` start your server. 
Then add nginx config:
```nginx
location ^~ /example/_frsh/ {
      proxy_pass http://xxx:8080;
}
location ^~ /example2/_frsh/ {
      proxy_pass http://xxx:8081;
}
```


src/runtime/utils.ts
```diff
@@ -1,6 +1,8 @@
 import { VNode } from "preact";
 
-export const INTERNAL_PREFIX = "/_frsh";
+export const INTERNAL_PREFIX = typeof Deno !== "undefined"
+  ? (Deno.env.get("INTERNAL_PREFIX") || "/_frsh")
+  : "/_frsh";
```
iuioiua, line 5:
```suggestion
export const INTERNAL_PREFIX = Deno.env.get("INTERNAL_PREFIX") || "/_frsh";
```
Deno is always defined in Fresh as it's built on Deno. The truthy expression of your ternary operator is all that's needed.

iuioiua (changes requested):
Could you please add a test for when the environment variable is defined?

iuioiua:
Hi @jiawei397, I'll close this PR, as it's gone stale. Please feel free to re-open if you're happy to continue working on it. Either way, thank you for your help!

src/runtime/utils.ts
jiawei397: I'm sorry to see this message only now. This file is not only called on the Deno side, but may also be loaded in the browser. I used the following writing method directly at the beginning, until the code reported an error.

PR #1102, opened 2023-04-04
chore: fix ci

kt3k:
This PR fixes CI by specifying the last compatible Deno version in ci.yml, and also skipping the assertion about weak etag. (`Deno.serve` has been replaced by a wrapper of `Deno.serveHttp` and it now compresses the responses automatically and therefore etag becomes weak etag.)


hashrock (approved):
LGTM

PR #1122, opened 2023-04-18
refactor: include router directly in fresh

crowlKats:
This PR changes from depending on https://deno.land/x/rutt to instead include a reduced & simplified version of it.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -398,39 +398,43 @@ export class ServerContext {
   ] {
     const routes: router.Routes<RouterState> = {};
 
-    routes[`${INTERNAL_PREFIX}${JS_PREFIX}/${BUILD_ID}/:path*`].default = this
-      .#bundleAssetRoute();
-
+    routes[`${INTERNAL_PREFIX}${JS_PREFIX}/${BUILD_ID}/:path*`] = {
+      default: this.#bundleAssetRoute(),
+    };
     if (this.#dev) {
-      routes[REFRESH_JS_URL].default = () => {
-        const js =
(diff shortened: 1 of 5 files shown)


src/server/context.ts
```diff
@@ -521,16 +527,23 @@ export class ServerContext {
...
+              ...ctx,
+              params,
+              render: createRender(req, params),
+              renderNotFound: createUnknownRender(req, {}),
+            }),
+        };
       } else {
+        routes[route.pattern] = {};
         for (const [method, handler] of Object.entries(route.handler)) {
-          routes[`${method}@${route.pattern}`] = (req, ctx, params) =>
```
bartlomieju, line 533: Is this a public API? Why change the name here?

src/server/context.ts
```diff
@@ -442,12 +446,14 @@ export class ServerContext {
       const { localUrl, path, size, contentType, etag } of this.#staticFiles
     ) {
       const route = sanitizePathToRegex(path);
-      routes[`GET@${route}`] = this.#staticFileHandler(
```
bartlomieju, line 445: Similarly here, why chance it?

lucacasonato (commented):
This doesn't yet touch the middleware handling itself, right?

src/server/context.ts
lucacasonato: No, not a public API.

src/server/context.ts
lucacasonato: I assume Leo changed these to simplify the router (as we don't have to parse `xx@xx` syntax anymore).

bartlomieju:
> This doesn't yet touch the middleware handling itself, right?

Correct, just upstreaming the router for now. Middleware will be done in a followup.

bartlomieju (approved):
LGTM

PR #1123, opened 2023-04-18
feat: add middleware destination

### LINKED ISSUE #521: Root _middleware is called for every request

crowlKats:
Closes #521

### DIFF
--- a/src/server/router.ts
+++ b/src/server/router.ts
@@ -38,7 +38,7 @@ export interface Routes<T = {}> {
   [key: string]: { [K in KnownMethod | "default"]?: MatchHandler<T> };
 }
 
-export type RouteKind = "internal" | "static" | "route" | "notFound" | "error";
+export type RouteKind = "internal" | "static" | "route" | "notFound";
 
 // deno-lint-ignore ban-types
 export type InternalRoute<T = {}> = {
@@ -182,7 +182,6 @@ export function router<T = unknown>(
       handlers.push(() => otherHandler!(req, ctx));
       return await middlewareCtx.next();
     } catch (err) {
-      middlewareCtx.destination = "error";
       handlers.push(() => errorHandler!(req, ctx, err));
       return middlewareCtx.next();
     }
--- a/src/server/router.ts
+++ b/src/server/router.ts
@@ -10,8 +10,9 @@ export type Handler<T = unknown> = (
 
 export type FinalHandler<T = unknown> = (
   req: Request,
-  ctx: MiddlewareHandlerContext<T>,
+  ctx: HandlerContext<T>,
   handlers: (() => Response | Promise<Response>)[],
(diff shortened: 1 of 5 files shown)


src/server/router.ts
```diff
@@ -30,16 +38,22 @@ export interface Routes<T = {}> {
   [key: string]: { [K in KnownMethod | "default"]?: MatchHandler<T> };
 }
 
+export type RouteKinds = "internal" | "static" | "route" | "notFound" | "error";
```
lucacasonato:
```suggestion
export type RouteKind = "internal" | "static" | "route" | "notFound" | "error";
```

lucacasonato (commented):
Rather than `routeKind` can you make it `destination` (this aligns with fetch destination in browsers).

lucacasonato (commented):
Direction is much better now. Thanks.

I have some suggestions for simplification:

instead of doing:

1. call inner with (req, ctx, handlers, middlewareCtx), perform routing, getting the handler
2. have inner add the handler to `handlers`
3. call `middlewareCtx.next()` in `inner` 

do instead:

1. call `inner` with (req, ctx), perform routing, getting the handler
2. return from `inner` the `handler` and the `targetKind`
3. add the handler to handlers _outside_ of inner
4. add the target to middlewareCtx _outside_ of inner
5. call `middlewareCtx.next()` _outside_ of inner

here `inner` function returned from `router` would return a `{ target: TargetKind, handler: () => {} }` on call. Inner can also be synchronous then.

This is important, because right now the error handling is very strange. There should not be an `"error"` targetKind, because when you call all the middlewares, you don't know if something is going to error yet. The act of routing itself should not be able to result in an error, only ever the execution (the handlers).

tests/main_test.ts
```diff
@@ -494,6 +494,29 @@ Deno.test({
   },
 });
 
+Deno.test("middleware destination", async (t) => {
```
crowlKats, line 497: an `internal` test is missing, but unsure if there is any path that is consistent for an internal asset

lucacasonato: _frsh/refresh.js

crowlKats: done

lucacasonato (approved):
Excellent, thank you!

PR #1127, opened 2023-04-20
Update dynamic-routes docs test url

Mike-Neto:
the local server uses http, this link as it stands does't work


iuioiua (approved):
LGTM! Thanks, @Mike-Neto.

PR #1131, opened 2023-04-25
docs: clarify what props can be passed to islands

lionel-rowe:
I was caught out by trying to pass `Infinity` as a prop to an island, which is a valid IEEE 754 float, but bizarrely JSON serializes to `null`.

### DIFF
--- a/docs/concepts/islands.md
+++ b/docs/concepts/islands.md
@@ -34,14 +34,16 @@ care of automatically re-hydrating the island on the client.
 Passing props to islands is supported, but only if the props are JSON
 serializable. This means that you can only pass:
 
-* Primitive types `string`, `boolean`, and `null`
-* Most `number`s (`Infinity`, `-Infinity`, and `NaN` are silently converted to
+- Primitive types `string`, `boolean`, and `null`
+- Most `number`s (`Infinity`, `-Infinity`, and `NaN` are silently converted to
   `null`, and `bigint`s are not supported)
-* Plain objects with string keys and primitive, plain object, or array values
-* Arrays containing primitives, plain objects, or other arrays
+- Plain objects with string keys and primitive, plain object, or array values
+- Arrays containing primitives, plain objects, or other arrays
 
-It is currently not possible to pass complex objects like `Date`,
-custom classes, or functions. This means that it is not possible to pass
-`children` to an island, as `children` are VNodes, which are not serializable.
+It is currently not possible to pass complex objects like `Date`, custom
+classes, or functions. This means that it is not possible to pass `children` to
+an island, as `children` are VNodes, which are not serializable. To find out
+more about JSON serializable values, check out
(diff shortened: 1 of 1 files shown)


lionel-rowe:
I've also been playing around with replacer/reviver functions that could be used to expand the range of island-compatible props — would PRs with this kind of functionality be welcome? Obviously it'd have a non-zero impact on performance.

```js
const type = '$$_fresh_type'

const replacers = [
    {
        checker: (x) => x === Infinity || x === -Infinity || Number.isNaN(x),
        serializer: (x) => ({ [type]: 'number', value: String(x) }),
    },
    {
        checker: (x) => typeof x === 'bigint',
        serializer: (x) => ({ [type]: 'bigint', value: String(x) }),
    },
    {
        checker: (x) => x instanceof Date,
        serializer: (x) => ({ [type]: 'date', value: x.valueOf() }),
    },
    {
        checker: (x) => x instanceof URL,
        serializer: (x) => ({ [type]: 'url', value: x.href }),
    },
    {
        checker: (x) => x instanceof Error,
        serializer: ({ name, message, stack }) => ({
            [type]: 'error',
            value: { name, message, stack },
        }),
    },
    {
        checker: (x) => x instanceof Uint8Array,
        serializer: (x) => ({
            [type]: 'u8',
            value: btoa([...x].map(x => String.fromCodePoint(x)).join('')),
        }),
    },
]

const deserializers = {
    number: Number,
    bigint: BigInt,
    date: (x) => new Date(x),
    url: (x) => new URL(x),
    error: ({ name, message, stack }) => {
        const E = globalThis[name]?.prototype instanceof Error ? globalThis[name] : Error
        return Object.assign(new E(), { name, message, stack })
    },
    u8: (x) => Uint8Array.from(atob(x), (ch) => ch.codePointAt(0)),
}

function replacer(k) {
    const v = this[k]
    
    for (const { checker, serializer } of replacers) {
        if (checker(v)) return serializer(v)
    }

    return v
}

function reviver(k, v) {
    const t = v[type]
    
    return t ? deserializers[t](v.value) : v
}

const propVals = [
    new Uint8Array([0, 0xff]),
    new TypeError('xyz'),
    new URL('https://deno.land/'),
    Infinity,
    -Infinity,
    NaN,
    99999999999999999999999999999999999999999999999999n,
    new Date(0),
]

const serialized = JSON.stringify(propVals, replacer)

const deserialized = JSON.parse(serialized, reviver)

// all elements of propVals preserved
deserialized
```

iuioiua (commented):
Mostly LGTM! Are you able to reference [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#description) for those seeking more info and run `deno fmt`?

iuioiua (approved):
I've added my suggestions and formatted the document. Now, LGTM! Thanks for this, @lionel-rowe.

PR #1139, opened 2023-04-28
chore: update for deno 1.33

marvinhagemeister:
This PR updates the code for Deno 1.33.x to solve errors on the CI. See https://github.com/denoland/fresh/actions/runs/4834957340/jobs/8616765955?pr=1138

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,8 @@ jobs:
 
     strategy:
       matrix:
-        deno: ["v1.x", "canary"]
+        # FIXME: temporarially disabled 1.x until https://github.com/denoland/deno/commit/64e072e499d36ca824db297a493667415ed67cdf is released
+        deno: ["canary"]
         os: [macOS-latest, windows-latest, ubuntu-latest]
 
     steps:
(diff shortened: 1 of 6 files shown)


marvinhagemeister:
Hm looks like puppeteer has trouble starting in a particular test.

marvinhagemeister:
Investigated a little bit and it seems like `browser.close()` hangs for multiple minutes. The test itself runs fairly quick.

lucacasonato:
@marvinhagemeister The puppeteer issue should bee fixed in 1.33.1 _I think_. @littledivy was working on it yesterday.

lucacasonato:
Oh I see - it only hangs on the close. I wonder if the new WebSocket implementation is not sending the `"close"` event for some reason.

lucacasonato:
It is reproducible in a minimal example: https://github.com/denoland/deno/issues/18912.

Maybe let's lock the tests to 1.32.5 until this is fixed (and disable canary).

lucacasonato:
Since there is now a fix in canary, I have disabled 1.x tests for now. We can re-enable after 1.31.2 is released.

lucacasonato (approved):
LGTM, thanks

PR #1143, opened 2023-05-02
Change BUILD_ID to not use DENO_DEPLOYMENT_ID

### LINKED ISSUE #1142: Change BUILD_ID to not use DENO_DEPLOYMENT_ID
DENO_DEPLOYMENT_ID gives access to old builds on Deno Deploy. It should not be exposed publicly.

Context:

The fresh.deno.dev source code has a JavaScript URL that exposes the deployment ID:
/_frsh/js/xv0bcetb883g/plugin-twind-main.js

This can lead to the discovery of old builds if the page source is cached, such as: https://fresh-xv0bcetb883g.deno.dev/

Using a site like the Internet Archive's Way Back Machine, it is trivial to find builds for any site that has any pages built with Fresh in the archive.

For example: looking at this snapshot from June 2022, I find the deployment id and can pull up that build still:

view-source:https://web.archive.org/web/20220613114408/http://fresh.deno.dev/
https://fresh-0npz6ckppgeg.deno.dev/

Code that causes the issue:

https://github.com/denoland/fresh/blob/f481c1a03fc3e194efa0fe0fb0c5e568cd2994f7/src/server/constants.ts#L5

innovatedev-john-pope:
fixes #1142

### DIFF
--- a/src/server/constants.ts
+++ b/src/server/constants.ts
@@ -2,8 +2,16 @@ import { INTERNAL_PREFIX } from "../runtime/utils.ts";
 
 export const REFRESH_JS_URL = `${INTERNAL_PREFIX}/refresh.js`;
 export const ALIVE_URL = `${INTERNAL_PREFIX}/alive`;
-export const BUILD_ID = Deno.env.get("DENO_DEPLOYMENT_ID") ||
+const deploymentId = Deno.env.get("DENO_DEPLOYMENT_ID") ||
   crypto.randomUUID();
+export const BUILD_ID = await crypto.subtle.digest(
+  "SHA-1",
+  new TextEncoder().encode(deploymentId),
+).then((hash) =>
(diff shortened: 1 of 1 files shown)


lino-levan:
Could we consider bringing in [std/crypto/to_hash_string](https://deno.land/std@0.185.0/crypto/to_hash_string.ts?s=toHashString) instead of trying to do that ourselves? Feels like the code will be more maintainable with that.

iuioiua:
What about `crypto.randomUUID()`?

lino-levan:
Should this be closed since #1168 landed?

kt3k:
Right. Solved in #1168

@innovatedev-john-pope you are attributed in the commit https://github.com/denoland/fresh/commit/6feac0d2d844a205a2f93f10f02128512100232d

PR #1144, opened 2023-05-02
cleanup pathname before processing

### LINKED ISSUE #1114: Too many slashes causes too many redirects
If you try to append `/` multiple times to the end of a URL, for example `http://localhost:8080/////////////////////////////`, the browser will refuse to continue because of too many redirects.

This is because of how [the handler handles (src/server/context.ts#L321)](https://github.com/denoland/fresh/blob/ff9bb0f132bab5b254107569d6146d2b018b92b4/src/server/context.ts#L321) `/`s, one at a time.

innovatedev-john-pope:
fixes #1114

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -323,7 +323,7 @@ export class ServerContext {
       // Ex: /about/ -> /about
       const url = new URL(req.url);
       if (url.pathname.length > 1 && url.pathname.endsWith("/")) {
-        url.pathname = url.pathname.slice(0, -1);
+        url.pathname = url.pathname.replace(/\/+/g, '/').slice(0, -1);
         return Response.redirect(url.href, Status.TemporaryRedirect);
       }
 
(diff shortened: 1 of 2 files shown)


src/server/context.ts
```diff
@@ -323,7 +323,7 @@ export class ServerContext {
       // Ex: /about/ -> /about
       const url = new URL(req.url);
       if (url.pathname.length > 1 && url.pathname.endsWith("/")) {
-        url.pathname = url.pathname.slice(0, -1);
+        url.pathname = url.pathname.replace(/\/+/g, '/').slice(0, -1);
```
lucacasonato: This replaces all slashes (not just the final ones. That behaviour is fine, but it should probably happen _before_ this `if` statement (because right now `/a///b/` normalizes to `/a/b`, but `/a///b` does not.

Can you also add a test case for this scenario?

innovatedev-john-pope: I fixed this, and added a test. Let me know if it needs anything else.

lucacasonato:
I have merged this as part of https://github.com/denoland/fresh/pull/1167.

Thanks!

PR #1146, opened 2023-05-03
fix(dev): server stalls when clicking quickly between pages

### LINKED ISSUE #1079: Server stalls when clicking quickly between pages
Steps to reproduce:

    deno --version
    deno 1.31.1 (release, aarch64-apple-darwin)

    fresh@1.1.4

Start a new Fresh project

```js
deno run -A -r https://fresh.deno.dev my-project
```

Add two simple routes to link between:

```js
// routes/index.tsx:
export default function One(props) {
  return <a class="underline" href="/two">Two</a>;
}
```

```js
// routes/two.tsx
export default function Two(props) {
  return <a class="underline" href="/">One</a>;
}
```

Run: `deno task start`

Then:
- Open http://localhost:8000 in Google Chrome and click the "Two" link
- Click the "One" link
- Click the "Two" link
- Click the "One" link
- Click the "Two" link
- Click the "One" link and wait ~60 seconds for a response

In Chrome Developer Tools > Network > Timing, Connection Start we can see Stalled for ~60s. Seems like we’re hitting the six TCP connection limit for HTTP.

This does not happen when `DENO_DEPLOYMENT_ID` is set, leading me to believe it was a problem with `_frsh/alive`.

So I believe the problem is that the alive stream’s `cancel` is never called, which can be confirmed with a `console.log` here:
https://github.com/denoland/fresh/blob/4bfab6a8a92e6cc62fe9e0e19d1ecdf2454f3487/src/server/context.ts#L421

I started digging into why that is and ran out of time. Possibly related:
- https://www.reddit.com/r/Deno/comments/111xbtw/does_denolistentls_not_support_calling_cancel_in/
- https://github.com/denoland/fresh/pull/995

hashrock:
- Replaced the refresh script based on @audionerd 's [commented code](https://github.com/denoland/fresh/issues/1079#issuecomment-1476826096).
- ~~added a simple minify function for DX.~~

Close #1079


lino-levan:
Other than picking "reflesh" as a name, the actual changes LGTM. Did you mean refresh? I haven't heard of reflesh but maybe it's a technical term I'm not aware of.

hashrock:
Oh, my fault😅
If I make another mistake, please point it out.

lino-levan (approved):
LGTM. Nice!

lucacasonato:
HTTP/1.1's stupid 6 max concurrent connections behaviour is... ahhhh.

LGTM

lucacasonato (approved):

PR #1151, opened 2023-05-04
feat(plugins): make render async

### LINKED ISSUE #728: Asynchronous plugin render functions

crowlKats:
Closes #728

### DIFF
--- a/plugins/twind.ts
+++ b/plugins/twind.ts
@@ -15,9 +15,9 @@ export default function(state) { hydrate(options, state); }`;
   return {
     name: "twind",
     entrypoints: { "main": main },
-    render(ctx) {
+    async render(ctx) {
       sheet.reset(undefined);
-      const res = ctx.render();
+      const res = await ctx.render();
       const cssTexts = [...sheet.target];
       const snapshot = sheet.reset();
       const scripts = [];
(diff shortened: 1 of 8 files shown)


lucacasonato (changes requested):
Render can not be async. It breaks `render` fundamentally, because `render` is meant to give you safe way to wrap the synchronous rendering process (which is needed if the rendering mutates thread local state, like is the case for the Tailwind plugin).

PR #599, opened 2022-08-07
feat: add URL and route to _app

crowlKats:

PR #1153, opened 2023-05-06
Pass page props to application wrapper

shivjm:
I wanted to add a footer whose content changes based on the current route, but the application wrapper is only passed the component. This PR adds the `PageProps` under a new `page` property. I’m a bit confused about whether this is the right approach, though.

`render` currently sets `props` per the docs and [declares it as `Record<string, unknown>` to pass safely to `h`](https://github.com/denoland/fresh/blob/9d13f6ff26181ea6ebcb096043c983f4d485cd9b/src/server/render.ts#L112-133). Passing those `props` alongside `Component` works and passes type checking, but I’m not sure whether there’s a reason for the `unknown` or not, so I can’t tell whether I’ve broken something. (And I can’t run the tests because it keeps failing on account of the port being in use. No idea why. I’m on Windows, and I can see it succeeds in CI.)

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -129,6 +129,7 @@ export async function render<Data>(
     children: h(HEAD_CONTEXT.Provider, {
       value: headComponents,
       children: h(opts.app.default, {
+        page: props,
         Component() {
           return h(opts.route.component! as ComponentType<unknown>, props);
         },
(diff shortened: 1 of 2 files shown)


deer:
Looks like a slightly more generic version of https://github.com/denoland/fresh/pull/599. I have a similar use case, so I'd love for something like this to get accepted.

iuioiua:
Closing in favour of #1229. Either way, thank you @shivjm 🙏🏾

shivjm:
Yup, that sounds good. I’m glad someone was able to implement it in a better way.

PR #1158, opened 2023-05-10
fix(www): docs title align center

ryuapp:
I removed dead class, "block".
The "block" class conflicted with "flex".
That had shifted a docs title to up and down.

Please compare following URL and check the title.
https://fresh.deno.dev/docs/introduction
https://fresh.deno.dev/components


kt3k (approved):
LGTM. Thanks

PR #1160, opened 2023-05-11
ref(init): Update routes/index.tsx to be correctly formatted with deno fmt

kamilogorek:
With current formatting, a fresh project will reformat on save and create an unnecessary diff.


kt3k (approved):
LGTM. Thanks!

PR #1162, opened 2023-05-11
fix(www): cursor on docs search modal

ryuapp:
In the current docs pages, the cursor becomes a pointer on docs search modal.
The bug is caused by twind being styled to be a pointer on `[role="button"]` element.

This PR overwrites twind's css from docsearch.css on the modal.


kt3k (approved):
LGTM

PR #1164, opened 2023-05-12
docs: unify space within section

ryuapp:
Many section tags use `space-y-4` class. Don't have to be `space-y-5` class on that section.
Also, showcase and components page will be the same design and look good.


iuioiua (approved):
This makes sense. Thanks, @ryuapp!

PR #1170, opened 2023-05-13
Async Plugins

pting-me:
Related: #1151 #745 #728

Would like to revisit async plugins.

The main concern around not having async plugins was to guarantee thread safety. The closest solution tried wrapping the sync plugins but didn't seem like it could actually guarantee that either. I believe that having a mix of sync and async plugins is way too difficult to handle.

I've drafted a solution that does the following:
- Users are only allowed to use `FreshOptions.plugins` or `FreshOptions.pluginsAsync`
  - Ensures backwards compatibility
  - Easier to enforce and maintain than trying to make all the current `plugins` polymorphic
  - Could possibly go one step further and make `FreshOptions` a discriminated union (e.g. `FreshOptions = FreshOptionsSync | FreshOptionsAsync`), but that could potentially harm the maintainability of the code
- Async plugins are stored in a queue to guarantee thread safety

If it sounds like a good solution, I will clean it up, add more tests, and fix the current issues. Right now it looks like JS is being injected properly, but not the CSS. PR is obviously not ready, but would like some feedback on this approach. @lucacasonato


lucacasonato:
I don't like this solution. The only solution I'd accept is a form of #745, because:

a) in the presence of async context tracking (for example through promise hooks or AsyncLocalStorage), TLS can be preserved for both sync and async render functions
b) it still allows sync render for when async context tracking is not available

Thanks for the PR, but I'll close in favor of #745

PR #1172, opened 2023-05-14
fix: conditionally rendered element under fragment in island does not render

### LINKED ISSUE #502: Conditionally rendered element under fragment in island does not render
```tsx
/** @jsx h */
/** @jsxFrag Fragment */
import { h, Fragment } from "preact";
import { useState } from "preact/hooks";

export default function Component() {
  const [show, setShow] = useState(false);

  return (
    <>
      Hello
      <div onClick={() => setShow(true)}>
        World
      </div>
      {show && (
        <div>I'm not rendered</div>
      )}
    </>
  );
}
```
This will error with `Uncaught (in promise) DOMException: Node.insertBefore: Child to insert before is not a child of this node` on the client when `World` is clicked.

azerum:
This might be fixing #502, but I am not sure. All tests pass. I've added additional test cases for comments from #502


iuioiua:
Nice work, @azerum. Can you please run `deno fmt` and remove any other changes, such as newlines, that don't directly serve to fix the main issue?

iuioiua (approved):
I've addressed my previous comments myself. Now LGTM! I'll give this one day before merging to allow others to take a look.

azerum:
@iuioiua Thank you! Sorry for delay, wanted to fix formatting today, but you were faster :)

PR #1174, opened 2023-05-15
Allow passing additional ESBuild Targets

mcandeia:
# Description

This pull request introduces a crucial enhancement to the open source repository by enabling the capability to pass additional esbuild targets. This feature proves to be especially valuable for supporting older browsers, such as Safari 11, which lacks support for top-level await.

The inclusion of this enhancement offers developers greater flexibility and compatibility when using fresh. By allowing the specification of additional esbuild targets, it becomes possible to transpile the code to a format that is compatible with Safari 11 and other older browsers that lack support for top-level await.

To achieve this, the following modifications have been made:

1. Added a new configuration option, additionalESBuildTargets, to the existing FreshOptions configuration. This option accepts an array of target environments that need to be supported.

2. Pass the targets to the bundler

By making these changes, developers can now easily extend the support of the repository to encompass browsers like Safari 11 without sacrificing the utilization of top-level await in their codebase.

This pull request aims to empowers developers to build applications that cater to a wider range of browser environments.

Thank you for considering this pull request. I'm excited to hear your feedback and collaborate on further enhancing the repository's functionality.

### DIFF
--- a/src/server/bundle.ts
+++ b/src/server/bundle.ts
@@ -46,6 +46,7 @@ export class Bundler {
   #plugins: Plugin[];
   #cache: Map<string, Uint8Array> | Promise<void> | undefined = undefined;
   #dev: boolean;
+  #additionalTargets: string[];
 
   constructor(
     islands: Island[],
@@ -53,7 +54,9 @@ export class Bundler {
     importMapURL: URL,
     jsxConfig: JSXConfig,
     dev: boolean,
+    additionalTargets?: string[],
   ) {
+    this.#additionalTargets = additionalTargets ?? [];
     this.#islands = islands;
     this.#plugins = plugins;
     this.#importMapURL = importMapURL;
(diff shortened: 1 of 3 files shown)


lucis:
@lucacasonato

igorbrasileiro:
maybe this info could come from deno.json? like package.json `browserlist` https://babeljs.io/docs/babel-preset-env#browserslist-integration

mcandeia:
> maybe this info could come from deno.json? like package.json browserlist https://babeljs.io/docs/babel-preset-env#browserslist-integration

not sure because it seems to be fresh-specific configs

iuioiua:
Hi @mcandeia, I like this. Can you mention this new feature somewhere in the docs? Once done, I think it'll be ready for me to merge.

mcandeia:
Thanks @iuioiua, can you help me to find the right place to drop these docs? I was trying to find anything related to a "configuration" section that seems more suitable to me, what do you think? Another option is to add this as a side note in the `running-locally.md` docs, how that's sound ?

iuioiua:
Oh, hm... Yeah, I think this belongs in a new "3.14 Server Configuration" section.

iuioiua (changes requested):
Had a chat with Luca. Let's change the property name to `targets` with the default value being `["chrome99", "firefox99", "safari15"]`.

mcandeia:
looks good @iuioiua, I can do that til the end of the next week. Thanks!

deer:
@marvinhagemeister, looks like you did this in #1798.

PR #1175, opened 2023-05-16
feat(server): Allow for custom HEAD method implementations

### LINKED ISSUE #1115: HEAD handlers in routes are ignored
As of https://github.com/denoland/fresh/pull/1034, Fresh responds to HTTP HEAD requests by making an internal GET request, omitting the response body, and returning the response headers and status code.

So, for example, given the following route, an HTTP HEAD request would return a 200 status code.
```typescript
export const handler: Handlers<Data> = {
  GET(req, ctx) { return new Response(null, { status: 200 }) }
}
```

The problem is, if a `HEAD` handler function is added to a route, it is ignored.

For example, here an HTTP HEAD request will return a 200 status code:
```typescript
export const handler: Handlers<Data> = {
  HEAD(req, ctx) { return new Response(null, { status: 204 }) },
  GET(req, ctx) { return new Response(null, { status: 200 }) }
}
```

And in this example, an HTTP HEAD request will return a 405 status code:
```typescript
export const handler: Handlers<Data> = {
  HEAD(req, ctx) { return new Response(null, { status: 204 }) }
}
```

### LINKED ISSUE #1165: Instead of rewriting all HEAD requests to GET, add a default HEAD handler to all routes that can be overriden
Right now there is no way to have a custom HEAD implementation in Fresh routes handlers

kamilogorek:
Fixes https://github.com/denoland/fresh/issues/1115
Fixes https://github.com/denoland/fresh/issues/1165


iuioiua (approved):
Clean and simple implementation. Thank you, @kamilogorek!

PR #1181, opened 2023-05-17
chore(ci): dependabot for GitHub actions

iuioiua:
This weekly action ensures other GitHub actions are kept up-to-date. See [here](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot) for more info. It should recommend upgrading a few actions on its initial run.


kt3k (approved):
Ok. Let's try. LGTM

BTW I think this should be labeled as `chore:` because this doesn't provide any visible feature to the users.

iuioiua:
Ah, yes. Good point.

PR #1186, opened 2023-05-18
refactor: disable lock

iuioiua:
This change disables the lock file for tests and the initialisation script. The noise produced in tests by generated lock files is now gone. This was also done because of Fresh's use of [esm.sh](https://esm.sh/) and [this](https://github.com/denoland/deno-gfm/issues/47#issuecomment-1460872019).


hashrock (approved):
LGTM

PR #1189, opened 2023-05-18
feat: preloading JavaScript files

### LINKED ISSUE #1187: Preloading JavaScript files (`link rel="modulepreload"`)
I would like to bring back [JavaScript preloading](https://developer.chrome.com/blog/modulepreload/) which has been disabled since #97.

https://github.com/denoland/fresh/blob/b9db46eccb9eb49fdcc040c044b99407b27e71b0/src/server/bundle.ts#L112-L118

Preloading helps avoid [critical request chains](https://developer.chrome.com/docs/lighthouse/performance/critical-request-chains/), so it should speed up loading.

related: #14, #15, #497, #741

ayame113:
close #1187

## points

- use `<link rel="modulepreload">` tag and `Link: rel="modulepreload"` header to preload dependencies
- No preloading at all for pages that don't use JavaScript at all.
- Information about preloading is delivered only when the JS file has already been built. If the JS file hasn't finished building, it will respond without preload information instead of waiting for the build and then returning the HTML.

## performance comparison

I've deployed this branch to https://fresh-module-preload.deno.dev/ so you can try it out and see how it behaves.

When I accessed [fresh.deno.dev](https://fresh.deno.dev/) and [fresh-module-preload.deno.dev](https://fresh-module-preload.deno.dev/) 20 times each and measured the DOMContentLoaded time, it was 24% faster (157.3ms -> 118.95ms) in my environment (Nagano City, Japan).

<details>
<summary>Actual measured data</summary>

(Don't forget to check `Disable Cache` in devtools)

 | afeter (ms) | before (ms)
-- | -- | --
0 | 180 | 152
1 | 82 | 144
2 | 78 | 186
3 | 156 | 196
4 | 88 | 211
5 | 85 | 123
6 | 162 | 209
7 | 154 | 110
8 | 86 | 175
9 | 135 | 109
10 | 183 | 117
11 | 115 | 159
12 | 99 | 173
13 | 83 | 131
14 | 82 | 142
15 | 135 | 160
16 | 120 | 125
17 | 148 | 152
18 | 118 | 229
19 | 90 | 143
average | 118.95 | 157.3

</details>

## screen shots

image

image


lucacasonato (commented):
This is great - why both in the headers and the body though? I'd prefer sending it in headers only.

ayame113:
Thank you for your review!

> This is great - why both in the headers and the body though? I'd prefer sending it in headers only.

This is for backwards compatibility as the Link header is a new feature.

- The share of browsers that support the Link header is 73%.
    https://caniuse.com/mdn-http_headers_link
    image
- The share of browsers that support `<link rel="modulepreload">` is 96%.
    https://caniuse.com/link-rel-preload
    image

By putting preload information in both the header and body, people whose browsers don't support Link headers can also benefit from preloading.

iuioiua:
I like this! I think we should do one or the other. I prefer using the `<link>` element because it has greater browser support. Outside of that, both perform the same function.

lucacasonato:
I have rebased the patch and have removed the `Link` headers in favour of just using the meta preload tags.

Can you add some tests?

ayame113:
OK, I added a test.

ayame113:
Note: Using the Link header is faster than using the link tag. (For headers, preload can start when response headers arrive. When using tags, response headers arrive -> response body arrives -> parse body -> start preload.)
That said, it's something that can be added later. Merging as is is fine for me!

iuioiua (approved):
LGTM. Nice work, @ayame113.

J053Fabi0:
How can I add the `Link: rel="modulepreload"` header to every request? Is it with a middleware?

Here's my guess:

```ts
  async function (req: Request, ctx: MiddlewareHandlerContext<State>) {
    const response = await ctx.next();
    response.headers.append("Link", 'rel="modulepreload"');
    return response;
  },
```

PR #1193, opened 2023-05-19
Make LinkButton island

oliver-jenny:
Add 'island' tag to LinkButton

### DIFF
--- a/www/islands/ComponentGallery.tsx
+++ b/www/islands/ComponentGallery.tsx
@@ -78,6 +78,7 @@ export default function ComponentGallery(props: ComponentGalleryProps) {
 
       <Section
         title="LinkButton"
+        island={true}
         source={props.sources.LinkButton}
       >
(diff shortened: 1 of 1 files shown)


iuioiua:
`<LinkButton />` is a component. It's in the `components` folder.

oliver-jenny:
> `<LinkButton />` is a component. It's in the `components` folder.

Yes but like all the buttons found on the 'Component' route on the page have a little green tag next to them which sais, they are an island... Correct me if i'm wrong but this should also be marked as island right? I mean its a button which theoraticly in a project would have js code

hashrock:
This is intended. A LinkButton is actually an anchor link that looks like a button and does not require JS.

oliver-jenny:
> 

Well thanks for the explanation!
Guess I still have  lot to learn -> new to Fresh and happy about every advice :)

Issue #1063, opened 2023-03-06
failed to build Docker file with Deno 1.31.1


panov-andy:
Hi,
I created an "empty" fresh project and added a Docker file from the documentation, updated the version to deno 1.31.1, and have got the error:
```
#8 2.742 error: The source code is invalid, as it does not match the expected hash in the lock file.
#8 2.742   Specifier: https://esm.sh/*preact-render-to-string@5.2.4
```

Please help :)

UPDATE: everything works as expected on Linux. The error I describe relates to Windows WSL2.

Details:
```
FROM denoland/deno:1.31.1
ENV DENO_DEPLOYMENT_ID=2023.03.06
WORKDIR /app
COPY . .
RUN deno cache main.ts --import-map=import_map.json
EXPOSE 8000
CMD ["run", "-A", "main.ts"]
```

```
$ docker build  .
[+] Building 126.9s (8/8) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                             0.0s
 => => transferring dockerfile: 226B                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                                0.0s
 => => transferring context: 2B                                                                                                                                                                  0.0s
 => [internal] load metadata for docker.io/denoland/deno:1.31.1                                                                                                                                124.0s
 => [internal] load build context                                                                                                                                                                0.0s
 => => transferring context: 831B                                                                                                                                                                0.0s
 => [1/4] FROM docker.io/denoland/deno:1.31.1@sha256:e09519206406460a6d57380a6553d608570a8a9717f552d90c1d982dcfa67533                                                                            0.0s
 => CACHED [2/4] WORKDIR /app                                                                                                                                                                    0.0s
 => [3/4] COPY . .                                                                                                                                                                               0.0s
 => ERROR [4/4] RUN deno cache main.ts --import-map=import_map.json                                                                                                                              2.8s
------
 > [4/4] RUN deno cache main.ts --import-map=import_map.json:
#8 0.330 Warning the configuration file "file:///app/deno.json" contains an entry for "importMap" that is being ignored.
#8 0.342 Download https://deno.land/x/fresh@1.1.3/server.ts
#8 0.714 Download https://deno.land/x/fresh@1.1.3/src/server/mod.ts
#8 1.772 Download https://deno.land/x/fresh@1.1.3/src/server/context.ts
#8 1.772 Download https://deno.land/x/fresh@1.1.3/src/server/deps.ts
[...]

PR #1195, opened 2023-05-19
Pin preact imports

zaynetro:
Esm.sh might import preact dependencies using different build versions which in turn fails integrity check.

See more details here: https://github.com/denoland/fresh/issues/1063

### DIFF
--- a/src/dev/imports.ts
+++ b/src/dev/imports.ts
@@ -10,11 +10,11 @@ export function freshImports(imports: Record<string, string>) {
   imports["preact"] = `https://esm.sh/preact@${RECOMMENDED_PREACT_VERSION}`;
   imports["preact/"] = `https://esm.sh/preact@${RECOMMENDED_PREACT_VERSION}/`;
   imports["preact-render-to-string"] =
-    `https://esm.sh/*preact-render-to-string@${RECOMMENDED_PREACT_RTS_VERSION}`;
+    `https://esm.sh/*preact-render-to-string@${RECOMMENDED_PREACT_RTS_VERSION}?pin=stable`;
   imports["@preact/signals"] =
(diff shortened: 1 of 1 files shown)


zaynetro:
I am not yet sure but we might need to pin twind imports as well. Let me know what you think.

zaynetro:
OK the problem with twindv1:

[twindv1.ts](https://github.com/denoland/fresh/blob/main/plugins/twindv1.ts#LL1C37-L1C69) imports directly from esm.sh. Most likely you will have twind also in your import map.  
In this case esm.sh might respond with different versions. Example:

```json
    "https://esm.sh/v116/@twind/core@1.1.3/core.d.ts": "64017de2b0dba405dad3041355b6a5c52fe2978c2badd02db010034f331d08a5",
    "https://esm.sh/v122/@twind/core@1.1.3/core.d.ts": "aa5e8d6e77b9ef0968911882ea0fe12418b8108987b059471a876ac8bd45139f",
```

Solution is to pin all twind imports:

```json
    "@twind/core": "https://esm.sh/@twind/core@1.1.3?pin=v122",
    "@twind/preset-tailwind": "https://esm.sh/@twind/preset-tailwind@1.1.4?pin=v122",
    "https://esm.sh/@twind/core@1.1.3": "https://esm.sh/v122/@twind/core@1.1.3",
```

ayame113:
I believe the issue of failing integrity checks was resolved by disabling the lockfile in #1186.

zaynetro:
> I believe the issue of failing integrity checks was resolved by disabling the lockfile in #1186.

Oh wow. I wasn't aware of this. I personally see the benefit of using the lock file so I will just pin all of my dependencies.

Feel free to close this PR.

iuioiua:
Closing as I believe the answer by @ayame113 is right.

PR #1203, opened 2023-05-24
add test case covering islands that return `null`

### LINKED ISSUE #1129: Islands cannot return null
Since the plugin API is rather rudimentary right now, it might be desireable to create some Islands only for their side effects, e.g.:

```
import { useEffect, } from "preact/hooks";

export default function Nothing() {
  useEffect(() => alert("I am only here for the effect"));
  return null;
}
```
However, the given example does currently result in the following error:
```
Uncaught TypeError: replaceNode[(replaceNode.length - 1)] is undefined
    createRootFragment main.ts:9
    walk main.ts:46
    walk main.ts:61
    walk main.ts:61
    walk main.ts:62
    walk main.ts:62
    revive main.ts:64
    <anonymous> (index):43
```

Returning an actual node (e.g. empty `<div/>`)  instead of `null` prevents that error but is not an ideal solution in my opinion, as unneccessary elements will end up in the DOM.

azerum:
Just a small addition to my previous PR. It wasn't obvious for me that PR #1172 also fixed #1129. I guess that issue should be covered by a separate test

### DIFF
--- a/tests/fixture/fresh.gen.ts
+++ b/tests/fixture/fresh.gen.ts
@@ -18,26 +18,28 @@ import * as $11 from "./routes/index.tsx";
 import * as $12 from "./routes/intercept.tsx";
 import * as $13 from "./routes/intercept_args.tsx";
 import * as $14 from "./routes/islands/index.tsx";
-import * as $15 from "./routes/islands/root_fragment.tsx";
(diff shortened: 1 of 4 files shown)


iuioiua (approved):
Great work! Thank you, @azerum.

PR #1206, opened 2023-05-27
fix: better HEAD handling with less special casing

lucacasonato:
We now add a default HEAD handler to the server, rather than special
casing HEAD requests elsewhere. Static files now get their own optimized
HEAD handler that does not read the file contents.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -624,10 +624,9 @@ export class ServerContext {
       if (key !== null && BUILD_ID !== key) {
         url.searchParams.delete(ASSET_CACHE_BUST_KEY);
         const location = url.pathname + url.search;
-        return new Response("", {
+        return new Response(null, {
           status: 307,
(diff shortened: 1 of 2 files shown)


src/server/context.ts
```diff
@@ -604,7 +613,44 @@ export class ServerContext {
...
+      if (key !== null && BUILD_ID !== key) {
+        url.searchParams.delete(ASSET_CACHE_BUST_KEY);
+        const location = url.pathname + url.search;
+        return new Response("", {
+          status: 307,
+          headers: {
+            "content-type": "text/plain",
+            location,
+          },
+        });
```
```suggestion
        return new Response(null, {
          status: 307,
          headers: {
            location,
          },
        });
```

iuioiua (changes requested):
LGTM once that one nit is addressed.

iuioiua (approved):

PR #1208, opened 2023-05-27
feat: remove hardcoded `deno.json` path

### LINKED ISSUE #961: Remove hardcoded dependency on `deno.json`
Currently, fresh has a hardcoded dependency on `./deno.json` for compiler options and throws and error if it doesn’t exist. 

This prevents using fresh in a subdirectory, e.g. `src/`, of the repository, without maintaining duplicate `deno.json` files.

Workaround using a soft or hard link are not optimal. Also they introduce more issues, e.g. a soft link is currently not recognized by `vscode_deno` extension which prevents IntelliSense https://github.com/denoland/vscode_deno/issues/784, while a hard link is not recognized by git which makes it impossible to synchronize it across different computers.

It would be great if the hardcoded dependency on `./deno.json` could be avoided, made optional or configurable.

Seems to be introduced in https://github.com/denoland/fresh/pull/610#pullrequestreview-1095905149

iceghost:
Fixes #961

I needed to put my fresh project inside a subdirectory and this was getting in the way. I fixed it


iceghost:
i was going to add support for `deno.jsonc` as well, but TypeScript is complaining about importing `.jsonc`

tests/cli_test.ts
```diff
@@ -398,3 +398,114 @@ Deno.test("fresh-init .", async function (t) {
...
+        stdout: "piped",
+        stderr: "piped",
+      });
+      const { code } = await cliProcess.output();
+      assertEquals(code, 0);
+    });
+
+    await t.step(
+      "move deno.json and import_map.json one level up",
+      async () => {
```
lucacasonato: This doesn't have to happen in a test step. It's not testing anything. Can you move both this, and the previous "execute init command" out of the `t.step` (one indent up)?

iceghost: I did not notice this. I don't really understand what `t.step` is, but I think I get it now. Thanks!

lucacasonato (approved):
LGTM

Issue #1211, opened 2023-05-29
Docs roadmap


iuioiua:
It can be difficult to quickly find exactly what you're looking for in the documentation. E.g. how can I modify the `<head />` element on my site? There are also opportunities for parts of the documentation to be streamlined. E.g. there's a "Form submissions" section and a "Forms" section. This roadmap aims to address these issues and significantly improve the overall quality and clarity of Fresh's documentation. There are plenty of opportunities for easy contributions here. We're also open to further ideas.

- [x] #1204
- [x] #1212
- [x] #1213
- [x] #1214
- [x] #1215
- [x] #1216
- [x] #1217
- [x] #1218
- [x] #1220

deer:
Hey guys, I've written a lot of documentation over the past few weeks, and it's been really great for learning a lot about Fresh. But I think our current organization could be improved.

Have you heard about https://diataxis.fr/? I first encountered it from reading about [Loopback4](https://loopback.io/doc/en/lb4/) where they write:
```
We use the [Diátaxis documentation authoring framework](https://diataxis.fr/) which organizes technical documentation into a system based on four quadrants:

* Learning-oriented Tutorials provide hands-on lessons where users can learn the framework by doing.

* Problem-oriented How-to Guides provide recipes to solve specific goals you may encounter while building a LoopBack project.

* Understanding-oriented Concepts pages provide the explanation of architecture concepts, wider view and deeper knowledge about the framework.

* Information-oriented Reference guides provide technical description of the machinery and how to use it.
```

Roughly what I'm suggesting is the following:
* Getting Started -> Becomes our first tutorial. Although we'd need to clean it up to really provide a proper walkthrough, with checkpoints in a sample repository.
* Concepts can stay the same, but we'd need to review the pages to make sure they really adhere to the principles Diataxis encourages.
* Examples -> How-to Guides. What we have already perfectly maps to what Diataxis encourages. We just have to rename the section!
* We have no reference guides. This has been the hardest part about jumping in and starting to contribute to Fresh. There's no technical explanation about how the runtime works. Of course I would be happy to start writing these as well.

marvinhagemeister:
@deer These are good points. Agree that the information hierarchy of our docs could be improved and that the diataxis framework works really well. I'm all for restructuring the docs into these categories 👍

iuioiua:
+1

deer:
[...]

PR #1221, opened 2023-05-29
docs: add new "Examples" section

### LINKED ISSUE #1204: proposal(docs): examples section
Like https://deno.com/manual/examples, but for Fresh. I imagine things like https://github.com/denoland/fresh/issues/527#issuecomment-1190379084 to be included.

This will be snippets-based and replace the bringing on of more examples in the `examples` folder. Reason-being is the maintenance incurred by having multiple fully-fledged Fresh apps in the `examples` folder.

Related: #1212

iuioiua:
It's currently empty, but this sets up future Fresh examples.

Closes #1204
Moves towards #1211

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -1,0 +1,9 @@
+---
+description: |
+  In this chapter of the Fresh documentation, you can find examples of features that you may like in your Fresh project.
+---
+
+In this chapter of the Fresh documentation, you can find examples of features
+that you may like in your Fresh project. If there's a specific example you'd
+like to see here, please let us know in
+[a GitHub issue](https://github.com/denoland/fresh/issues/new).
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -5,5 +5,5 @@ description: |
 
 In this chapter of the Fresh documentation, you can find examples of features
 that you may like in your Fresh project. If there's a specific example you'd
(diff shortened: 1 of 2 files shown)


hashrock (approved):
LGTM

docs/examples/index.md
```diff
@@ -0,0 +1,9 @@
+---
+description: |
+  In this chapter of the Fresh documentation, you can find examples of features that you may like in your Fresh project.
+---
+
+In this chapter of the Fresh documentation, you can find examples of features
+that you may like in your Fresh project. If there's a specific example you'd
+like to see here, please let us know in
+[a GitHub issue](https://github.com/denoland/fresh/issues/new).
```
lucacasonato: Discussions instead maybe?

docs/toc.json
```diff
@@ -36,5 +36,8 @@
   },
   "integrations": {
     "title": "Integrations"
+  },
+  "examples": {
+    "title": "Examples"
```
lucacasonato: Let's leave it out of the TOC until we add the first example. A bit confusing otherwise.

docs/examples/index.md
iuioiua: Yeah. Actually, discussions would be better because the example idea would likely achieve greater consensus before becoming an actionable issue.

PR #1225, opened 2023-05-30
feat: support passing Uint8Array as island props

lucacasonato:
This commit adds support for passing Uint8Array's to islands in
component props.


iuioiua (approved):
LGTM! Sidenote: one might look at the codebase in its current form and incorrectly think there are no deserializer tests when there are. They live in `serializer_test.ts`. I did, initially. I think this is fine but worth considering.

PR #1227, opened 2023-05-31
chore: remove JSX pragma tests

iuioiua:
JSX pragma is no longer needed.


lucacasonato (commented):
1.0 was released with this, so we need to keep backwards compat for users using the JSX pragma.

iuioiua:
Closing. Perhaps, we can revisit approaching v1.2 or v2.

PR #1229, opened 2023-05-31
feat: add `PageProps` to _app

huai-jie:
This should same purpose as https://github.com/denoland/fresh/pull/1153 but with more strict `type`

> Could use it as reference for  https://github.com/denoland/fresh/pull/1153 and close this as duplicate

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -116,7 +116,7 @@ export interface Route<Data = any> {
 
 // --- APP ---
 
-export interface AppProps {
+export interface AppProps extends PageProps {
   Component: ComponentType<Record<never, never>>;
 }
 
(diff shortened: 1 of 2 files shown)


src/server/types.ts
```diff
@@ -116,7 +116,7 @@ export interface Route<Data = any> {
 
 // --- APP ---
 
-export interface AppProps {
+export interface AppProps extends PageProps {
```
lucacasonato, line 119: `data` is always `unknown` I guess?

lucacasonato (approved):
LGTM

src/server/types.ts
huai-jie: Yup, `type` parameter is fine as well for `data` property

iuioiua (approved):
LGTM! Thanks, @huai-jie.

src/server/types.ts
J053Fabi0: Is not `unknown`, is `any`.

PR #1231, opened 2023-05-31
chore: clean up esbuild builder logic

lucacasonato:
This will allow us to add various caches and similar around the esbuild
builder easily in follow up PRs.

It also makes doing an AOT build-step + load assets from disk much
simpler.


lino-levan:
Would be super nice to have this. The current (hacky) approach to AOT build-step definitely should not be the long-term solution.

src/build/esbuild.ts
```diff
@@ -0,0 +1,186 @@
+import {
+  denoPlugins,
+  esbuild,
+  esbuildTypes,
+  esbuildWasmURL,
+  fromFileUrl,
+  toFileUrl,
+} from "./deps.ts";
```
ayame113: Adding `export { escape } from "https://deno.land/std@0.189.0/regexp/escape.ts";` to deps.ts and importing `escape` here should fix the error on Windows.

src/build/esbuild.ts
```diff
@@ -0,0 +1,186 @@
...
+        const path = toFileUrl(file.path).href.slice(absWorkingDirLen);
+        files.set(path, file.contents);
+      }
+
+      const metaOutputs = new Map(Object.entries(bundle.metafile.outputs));
+
+      // deno-lint-ignore no-inner-declarations
+      function walk(path: string, seen: Set<string>) {
+        if (seen.has(path)) return;
+        seen.add(path);
```
ayame113, line 96: Is recursion necessary here? From what I've seen, esbuild constructs the dependency tree in such a way that no recursion is required.

src/build/esbuild.ts
lucacasonato: Thanks! Great catch :D

lucacasonato: Thanks! Great catch :D

src/build/esbuild.ts
lucacasonato: It is not guaranteed that the deps are always entrypoint -> chunk. I think entrypoint -> chunk -> chunk is possible

lucacasonato: It is not guaranteed that the deps are always entrypoint -> chunk. I think entrypoint -> chunk -> chunk is possible

lucacasonato: It is not guaranteed that the deps are always entrypoint -> chunk. I think entrypoint -> chunk -> chunk is possible

lucacasonato: It is not guaranteed that the deps are always entrypoint -> chunk. I think entrypoint -> chunk -> chunk is possible

ayame113: It looks like the entry point contains the data for all the chunks it depends on so that no chunk->chunk reads occur.

(For example, `chunk-E7WS2GRN.js` is not used in the entry point like below, but it is imported here because it is imported from `chunk-OYKCTXOR.js`.)
image

That being said, these things aren't documented in esbuild, so I could be wrong!

lucacasonato: I see - I think you're right. We can remove the logic then

PR #1234, opened 2023-06-01
refactor: replace `useState()` with `useSignal()`

iuioiua:
This change replaces `useState()` with the recommended `useSignal()` across the codebase.

Closes #1214.

PR #1236, opened 2023-06-01
chore: upgrade `preact-render-to-string` to 6.0.3

iuioiua:
Depends on #1234.


lino-levan:
#1234 was merged. Should this un-drafted?

lucacasonato (approved):

PR #1242, opened 2023-06-03
fix: preload all the necessary files

### LINKED ISSUE #1241: preloading all dependencies
It seems that the file preloaded before and after #1231 has changed. After #1231, preload information is only partially provided.
I would like to fix this and preload all files that should be preloaded.

(I didn't write a detailed test for this in #1189. I need to add tests to prevent regressions.)

(before #1231)

image

(after #1231)

image

ayame113:
close #1241

According to `console.log(dependencies)`, we should use filenames without slashes to get the dependencies.

image

---

Also, this includes fixes related to https://github.com/denoland/fresh/pull/1231#discussion_r1215491755.

### DIFF
--- a/src/build/esbuild.ts
+++ b/src/build/esbuild.ts
@@ -110,8 +110,6 @@ export class EsbuildBuilder implements Builder {
         dependencies.set(path, [...seen]);
       }
 
-      console.error(dependencies);
-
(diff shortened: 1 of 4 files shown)


lucacasonato (approved):
LGTM!

ayame113:
Thank you for your review!

PR #1243, opened 2023-06-03
chore(www): fix accessibility issues

ayame113:
Address accessibility issues pointed out in lighthouse.


lucacasonato (approved):
LGTM, thank you!

PR #1246, opened 2023-06-04
fix: remove top level <script> imports

### LINKED ISSUE #1177: Revive function gets called before completion of top-level awaits in Safari (with arrow function components)
# Description

When using top-level await with arrow function components in Safari (version 15+), fresh behaves differently compared to Chrome (version 112+). In Safari, the revive function is called before all top-level awaits have finished, resulting in silent failures of the preact/revive library. This behavior causes a bug that can be reproduced consistently.

Steps to Reproduce

1. Clone [the repository](https://github.com/mcandeia/fresh-bug-safari/pull/1) (notice that this repository is exactly the quickstart from `deno run -A -r https://fresh.deno.dev` with few changes)
2. `deno task start`
3. Go to `http://localhost:8000` on safari
4. Observe that the revive function in fresh starts before the top-level await operation has finished.

Also, notice that I've created my own forked fresh version to add some logs on revive, that's the only way that I could find the issue because the revive silently fails without any message (and the message can vary), you can see my changes [here](https://github.com/mcandeia/fresh/commit/c1f7c1c23cd3fe24e4de62ba892c8251bc36cd55).

Hydration silently failed.
image

image

The message does not make sense since you can remove the `await sleep()` and it will start to work again.
image

Same code running on chrome:

image

lucacasonato:
Instead, rely only on modulepreload for script preloading.

Fixes #1177

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -208,7 +208,19 @@ export async function render<Data>(
 
   bodyHtml = bodyHtml as string;
 
-  const imports: [string, string][] = [];
+  const moduleScripts: [string, string][] = [];
+
+  for (const url of opts.imports) {
+    const randomNonce = crypto.randomUUID().replace(/-/g, "");
+    if (csp) {
+      csp.directives.scriptSrc = [
+        ...csp.directives.scriptSrc ?? [],
+        nonce(randomNonce),
+      ];
+    }
+    moduleScripts.push([url, randomNonce]);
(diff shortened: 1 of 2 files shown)


marvinhagemeister (commented):
Ignore this, just a duplicate comment

marvinhagemeister (approved):
LGTM 👍

PR #1250, opened 2023-06-05
chore(test): extract shared start server code

marvinhagemeister:
Noticed that the tests have lots of repeated code for starting up a fresh server and waiting for it to be ready. To me that seemed common enough that it makes sense to extract it in a function to be reused in test cases.

### DIFF
--- a/tests/cli_test.ts
+++ b/tests/cli_test.ts
@@ -8,8 +8,8 @@ import {
   delay,
   puppeteer,
   retry,
-  TextLineStream,
 } from "./deps.ts";
+import { startFreshServer } from "./test_utils.ts";
 
 type FileTree = {
   type: "file";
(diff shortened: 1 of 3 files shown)


lino-levan (approved):
LGTM. Nice cleanup!

lucacasonato (approved):
Awesome!! :D

I wanted to do this for a long time - ideally I'd like to refactor to a HOF for `Deno.test` to handle cleanup using `try finally` correctly.

e.g `Deno.test(withFresh([args], (t, addr) => { ... }))`

marvinhagemeister:
@lucacasonato yeah cleanup logic for such things in tests is a always bit tricky. I'm not sure if a HOF approach or one where we can register cleanup logic in the test framework is easier to grasp. Let's flesh that out in the future.

lucacasonato:
Just need something to tide us over until https://github.com/tc39/proposal-explicit-resource-management ships

Issue #134, opened 2022-03-13
`h('!--frsh')` generates `<!--frsh></!--frsh>` which is invalid closing tag


kuboon:
https://github.com/lucacasonato/fresh/blob/67b3eb470e35abe50b0b9838ad05e5ac1a743e86/src/server/render.tsx#L313-L317

Passing `voidElements` option to `renderToString` can fix this behavior.
https://github.com/preactjs/preact-render-to-string/blob/936f71d60a67336bc2639de15dd1c323aa4ff669/src/index.js#L320-L322

kuboon:
Oh, `<!--frsh-- />` is still invalid..

lucacasonato:
@kuboon This is actually a valid HTML comment. When a compliant HTML parser encounters a `<`, followed by a `/`, followed by something other than an ASCII alpha, `/`, or EOF, it goes into a ["bogus comment"](https://html.spec.whatwg.org/multipage/parsing.html#bogus-comment-state) parsing state that will parse the bytes up to the next `>` as a HTML comment. The relevant soft parse error is https://html.spec.whatwg.org/multipage/parsing.html#parse-error-invalid-first-character-of-tag-name.

This is why the comment text of `</--frsh-blabla-->` is `--frsh-blabla--` and not `frsh-blabla`.

PR #1251, opened 2023-06-05
feat: Support JSX island children

### LINKED ISSUE #750: Support passing JSX components to island props
We should support passing JSX elements to island props. Example:

```tsx
import AddToCart from "../islands/AddToCart.tsx";

function IndexPage() {
  return (
    <div>
      <AddToCart>
        Cream Cheese <b>Bagel</b>
      </AddToCart>
    </div>
  );
}
```

The JSX elements passed to island props would be serialized to static markup on the server (just like the surrounding components in the route). These components would **not** be re-rendered on the client. Instead on the client the prop would be replaced with a proxy component that represents the underlying static markup, rather than a real VDOM tree.

marvinhagemeister:
This PR adds support for JSX children being passed to islands. It's limited to non-components though as serialising functions would rely on eval'ing the code in the browser which is not cool from a security perspective.

- `<></>` works
- `<div />` works
- `foo` works
- `<Foo />` doesn't work, cannot serialise function

Note that this is a first draft of how this feature could work as I need to get more familiar with fresh itself. The final implementation might change substantially.

Fixes #750

TODO:
- [ ] Might make sense to drop `Fragments` and only support DOM nodes. Benefit would be that we could skip serializing these and reconstruct the vnodes from the DOM itself. I kinda want to explore that.
- [ ] Add more tests

### DIFF
--- a/demo/fresh.gen.ts
+++ b/demo/fresh.gen.ts
@@ -4,6 +4,7 @@
 
 import * as $0 from "./routes/index.tsx";
 import * as $$0 from "./islands/Counter.tsx";
+import * as $$1 from "./islands/Passthrough.tsx";
 
 const manifest = {
   routes: {
@@ -11,6 +12,7 @@ const manifest = {
   },
   islands: {
     "./islands/Counter.tsx": $$0,
+    "./islands/Passthrough.tsx": $$1,
   },
   baseUrl: import.meta.url,
 };
--- a/demo/islands/Passthrough.tsx
+++ b/demo/islands/Passthrough.tsx
@@ -1,0 +1,9 @@
+import { type ComponentChildren } from "preact";
+
+interface PassThroughProps {
+  children: ComponentChildren;
+}
+
+export default function PassThrough({ children }: PassThroughProps) {
+  return <div>children: {children}</div>;
+}
--- a/demo/routes/index.tsx
+++ b/demo/routes/index.tsx
@@ -1,5 +1,6 @@
 import { useSignal } from "@preact/signals";
 import Counter from "../islands/Counter.tsx";
+import PassThrough from "../islands/Passthrough.tsx";
 
 export default function Home() {
   const count = useSignal(3);
@@ -12,6 +13,9 @@ export default function Home() {
       <Counter count={count} />
       <Counter count={count} />
       <Counter count={count} />
+      <PassThrough>
+        <h1 key="abc">asd test</h1>
+      </PassThrough>
     </div>
   );
 }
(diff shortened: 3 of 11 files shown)


lucacasonato (commented):
Cool! Have you considered doing a server side render of the passed VNodes, and passing the serialized value to the client? In this case we could also do an optimization where if the node is rendered as part of the "regular" render (ie is part of the outputted HTML), we could just annotate that node in HTML and retrieve it using that annotation on the client rather than reserializing it into JSON.

I think there are these two use cases:

- `<AddToCart>Add <b>this</b> item to the cart</AddToCart>`: here the children are unconditionally rendered, so the passed children are always present in HTML.
- `<Button activeLabel={<b>Enabled</b>} disabledLabel={<i>Disabled</i>} />`: here the children are conditionally rendered. The server render may only include `disabledLabel` in the HTML. The other child needs to be serialised in some way, out of band from the rendered HTML.

WDYT?

marvinhagemeister:
Yeah I feel similar in that passing server side rendered nodes is the way to go. I'm leaning towards picking those up from the final rendered HTML as well, for the same reasons you outlined. That would also avoid sending content twice to the client, e.g. one time as HTML and a second time as json for the props.

I'm currently evaluating which approach would be appropriate for the client to pick up that subtree. Right now I can think of two approaches:

1. Add some markers in the HTML to find which parts where the static children. Gets a little complex when there isn't a single root child of that subtree or the root child is a text node. `<Island>foo</Island>` or `<Island><p>A</p><p>B</p></Island>`
2. Inject a component around island children that bails out of rendering completely. That way the renderer doesn't have to care about the static subtree at all

The second approach seems much more pragmatic so far, but need to explore how feasible that is.

lucacasonato:
Great! Either approach sounds fine.

For reference, we use the comment marker approach for islands already, opposed to HTML Custom Elements to avoid changing the tree structure (to not disturb CSS in any way): https://github.com/denoland/fresh/blob/main/src/server/render.ts#L408-L412

(on an unrelated note, we should consider making `preact-render-to-string` not output "invalid" HTML comments - see https://github.com/denoland/fresh/issues/134)

marvinhagemeister:
Yeah the comment marker vs custom element is an interesting discussion. There are pros and cons to both approaches. The custom element approach has a few niceties in that it moves elements into the right place automatically for via built-in browser mechanisms when it comes to streaming rendering. Having a built-in island primitive is very appealing too. Now that I can actually time to work on those things, I can push these forward.

marvinhagemeister:
Closing in favour of https://github.com/denoland/fresh/pull/1285 which is a much better approach which doesn't require to send the serialized JSX tree down to the browser like the approach in this PR does.

PR #1255, opened 2023-06-07
fix(server): don't crash if no islands folder is present

marvinhagemeister:
I feel like the island folder should be optional and not lead to an error when it's missing.


marvinhagemeister:
Need to take another look. The test is passing although it shouldn't.

iuioiua:
I wasn't able to reproduce this issue with the latest init script. What was the error that you were getting?

marvinhagemeister:
This is the error I was getting:

Screenshot 2023-06-08 at 01 19 38

iuioiua:
Ok, I can reproduce the issue. I previously used the init script from the website, which must've been cached. My bad.

deer:
For anyone curious: this was solved by https://github.com/denoland/fresh/pull/1327.

Issue #761, opened 2022-09-12
Import `preact/debug` server side during dev mode


lucacasonato:
According to Marvin this will fix #753. We should be able to dynamically import it during startup if we are in dev mode.

aadamsx:
I can't figure out how to debug my fresh project with VS code -- are there instructions somewhere?  Seems like a basic thing for starters.

marvinhagemeister:
Resolved by #1256 .

PR #1256, opened 2023-06-07
feat(debug): show rendering issues during SSR

### LINKED ISSUE #753: Object as children does not show any error or warning
When an object is passed as children to be rendered there's no warning or error and it doesn't render anything.
I'm not sure if the issue is related to preact or fresh.

Example code
```ts
const activities = [
  { id: "1", name: "Wake up", date: new Date() },
  { id: "2", name: "Eat breakfast", date: new Date() },
  { id: "3", name: "Go to work", date: new Date() },
  { id: "4", name: "Go to sleep", date: new Date() },
];

  return (
    <ul>
      {activities.map((activity) => (
        <li key={activity.id}>{activity}</li>
      ))}
    </ul>
  );
}
```

marvinhagemeister:
We were already importing `preact/debug` in the client bundle, but not on the server. So any rendering issues on the server wouldn't show up.

Screenshot 2023-06-08 at 01 44 41

Fixes #753, #761 .


src/server/context.ts
```diff
@@ -56,6 +56,11 @@ function isObject(value: unknown) {
     value !== null;
 }
 
+function isDevMode() {
+  // Env var is only set in prod (on Deploy).
+  return typeof Deno.env.get("DENO_DEPLOYMENT_ID") !== "string";
```
```suggestion
  return Deno.env.get("DENO_DEPLOYMENT_ID") === undefined;
```
Nit. Reference: https://deno.land/api@v1.34.1?s=Deno.Env#method_get_0

iuioiua (approved):
LGTM! This also closes #761.

PR #1258, opened 2023-06-09
chore: correct misspelled variable

garbles:
Renames `entires` to `entries`.


iuioiua (approved):
Nice catch. Thank you!

PR #1260, opened 2023-06-09
fix(ssr): invalid HTML comment tags

### LINKED ISSUE #838: HTML validation issues with island tags
Check any fresh-generated HTML with islands and you are likely to find a couple of issues with the [w3 validator](https://validator.w3.org). It seems fresh is wrapping islands in comment tags. The opening tag is a valid HTML comment, however the closing tags for the `<!--fresh-` tags are invalid HTML.

Current versions of Chrome and Firefox seem to be able to deal with this, but this doesn't mean other clients will too, or that it won't cause any issues in the future or in weird edge cases.

image

marvinhagemeister:
I just landed a [preview feature](https://github.com/preactjs/preact-render-to-string/releases/tag/v6.1.0) in `preact-render-to-string` that allows you to render comments during SSR. That solves the invalid comment around islands that was reported in #838. It's marked as unstable mostly because we don't want folks other than framework authors to rely on that for now.

Fixes #838


marvinhagemeister:
Looks like something changed with twind styles. Currently investigating...

EDIT: Turns out one test was referencing an older version of `preact-render-to-string`. I've changed the PR so that both newer and older versions are supported. When v6.1.0 is detected which has the comments support we'll use that and if there is an older version, we'll print a warning and fall back to the previous behavior.

Here is what the warning looks like: 

Screenshot 2023-06-09 at 11 01 39

lucacasonato (approved):
LGTM

PR #1262, opened 2023-06-09
chore(tests): don't cancel child process stream

marvinhagemeister:
This makes debugging tests a lot easier with console logging. Before this it would often error with a "Broken Pipe" error because the stream was closed.

See: https://github.com/denoland/deno/issues/19401


bartlomieju (approved):
LGTM

PR #1145, opened 2023-05-03
add state to page props automatically

innovatedev-john-pope:
poc for #1128

PR #833, opened 2022-10-12
Add context state as an app prop

lytedev:
This allows you to share global state with the App component for doing things like rendering a shared header differently based on whether or not the user is logged in.

PR #1264, opened 2023-06-09
feat: state in props

### LINKED ISSUE #663: [Help Wanted] Is that possible for App Component have access to ctx.state
Hello, everyone!

I am currently working on a project using `Fresh` (love it so for),  and I am wondering "Is that possible for App Component has access to `ctx.state`"

 ```javascript
  // somehow get data from ctx.state
   const { user } = ctx.state
   return (
    <div class={tw`h-screen bg-white dark:bg-black`}>
      <Head>
        <title></title>
        <style>
        {// I want to set some custom css variables based on user}
        {`:root { --color-who: '#fff'}`}
       </style>
      </Head>
      <Component />
    </div>
  )
}
 ```

### LINKED ISSUE #1128: proposal: pass `ctx.state` to page props by default
Say, I have the middleware:
```ts
// routes/_middleware.ts
export interface MiddlewareState {
  isAuthenticated: boolean;
}
```
Currently, for the `MiddlewareState` context to be available in a page component's `props.data`, one must have the following handler:
```ts
// routes/index.tsx
  GET(_req, ctx) {
      return ctx.render(ctx.state);
  },
};
```
I propose that `ctx.state` be passed to `props.data` by default so that a handler is no longer required for each route with folder-level middleware. It seems as though some expect this to be the default behaviour, but it isn't. I'd be keen to hear if there are any reasons why this might not be technically feasible or a poor design decision to implement. Either way, I'll have a crack at this because I think it'd be useful.

See:
* https://discord.com/channels/684898665143206084/1063376560062730282
* https://discord.com/channels/684898665143206084/991511118524715139/1078729688937144400

deer:
Standing on the shoulders of giants here.
supersedes: https://github.com/denoland/fresh/pull/1145 & https://github.com/denoland/fresh/pull/833
closes: https://github.com/denoland/fresh/issues/1128, closes https://github.com/denoland/fresh/issues/663

Now that `AppProps` extends `PageProps` (thanks to https://github.com/denoland/fresh/pull/1229), there seems to be some duplication of issues and PRs. https://github.com/denoland/fresh/issues/663 (state in `_app`) is a special case of https://github.com/denoland/fresh/issues/1128 (state in props). Likewise, https://github.com/denoland/fresh/pull/833 (add state to `AppProps`) is a special case of https://github.com/denoland/fresh/pull/1145 (add state to `PageProps`).

Both the PRs were incredibly instructive to me, but both seemed to get stuck at that classic last hurdle: tests and docs. I've tried my hand at modifying the docs to reference state being available (I think this is ok) and at writing tests (not so sure about this part). Of course I'm happy to receive feedback on all parts (especially including the right way to dust off someone's old PR for something I'm interested in moving forward).


tests/main_test.ts
```diff
@@ -55,7 +64,7 @@ Deno.test("/props/123 page prerender", async () => {
   const body = await resp.text();
   assertStringIncludes(
     body,
-    `{&quot;params&quot;:{&quot;id&quot;:&quot;123&quot;},&quot;url&quot;:&quot;https://fresh.deno.dev/props/123&quot;,&quot;route&quot;:&quot;/props/:id&quot;}`,
+    `{&quot;params&quot;:{&quot;id&quot;:&quot;123&quot;},&quot;url&quot;:&quot;https://fresh.deno.dev/props/123&quot;,&quot;route&quot;:&quot;/props/:id&quot;,&quot;state&quot;:{&quot;root&quot;:&quot;root_mw&quot;}}`,
```
deer, line 52: Now that `props` includes state, `JSON.stringify(props)` is going to be different.

tests/main_test.ts
```diff
@@ -44,6 +45,14 @@ Deno.test("/ page prerender", async () => {
     body,
     '<meta name="description" content="Hello world!"/>',
   );
+  assertStringIncludes(
+    body,
+    '<meta name="generator" content="The freshest framework!"/>',
+  );
+  if (body.includes("specialTag")) {
```
deer: I didn't see a negation version of `assertStringIncludes` (something like `assertStringDoesntInclude`), so I did it like this. Suggestions?

docs/concepts/app-wrapper.md
```diff
@@ -7,18 +7,20 @@ An application wrapper is defined in an `_app.tsx` file in `routes/` folder. It
...
 // routes/_app.tsx
 
 import { AppProps } from "$fresh/server.ts";
 
-export default function App({ Component }: AppProps) {
+export default function App(props: AppProps) {
   return (
     <div class="wrapper">
-      <Component />
+      <props.Component />
```
marvinhagemeister: Super tiny nit: I prefer the previous variant because newcomers learn that components in JSX start with a capital letter.

tests/main_test.ts
marvinhagemeister: You can use `assert()` for plain boolean checks. Example:

```js
assert(foo === bar, "My error message when the check fails")

// your example
assert(!body.includes("specialTag"), `Expected actual: "${body}" to not contain: "specialTag"`);
```

tests/main_test.ts
```diff
@@ -634,6 +659,14 @@ Deno.test("experimental Deno.serve", {
...
       '<meta name="description" content="Hello world!"/>',
     );
+    assertStringIncludes(
+      body,
+      '<meta name="generator" content="The freshest framework!"/>',
+    );
+    if (body.includes("specialTag")) {
+      const msg = `Expected actual: "${body}" to not contain: "specialTag"`;
+      throw new AssertionError(msg);
+    }
```
marvinhagemeister:
```suggestion
    assert(!body.includes("specialTag"), `Expected actual: "${body}" to not contain: "specialTag"`);
```

marvinhagemeister (commented):
This is a great PR! I really love the thoroughness of it!

A thought that crossed my mind is that with the introduction of `state`, we now have `pageProps.data` and `pageProps.state`. Like when is something `data` and when should it be `state`? Or should this be merged into one? I'm unsure myself and curious about what others think about this.

docs/concepts/app-wrapper.md
deer: Yes, that seems like a fine suggestion. But as a newcomer coming from java 8 (don't laugh!), destructuring assignments are crazy, so having `App(props: AppProps)` seems much more understandable than `App({ Component }: AppProps)`. But I suppose if you can't beat 'em, join 'em. I guess the desired format is then:
```
export default function App({ Component, state }: AppProps) {
  //do something with state here
  return (
    <div class="wrapper">
      <Component />
    </div>
  );
}
```

deer:
> A thought that crossed my mind is that with the introduction of `state`, we now have `pageProps.data` and `pageProps.state`. Like when is something `data` and when should it be `state`? Or should this be merged into one? I'm unsure myself and curious about what others think about this.

My understanding is that `state` comes from middleware and `data` comes from the `Handler` in the same file. This is based on https://fresh.deno.dev/docs/concepts/middleware and https://fresh.deno.dev/docs/concepts/data-fetching.

lino-levan (approved):
This is a super nice PR. I think that the `state`/`data` difference makes sense to me. It's not super clear how to resolve conflicts if we decided to merge them into one property. Would the data override the state or the other way around? What about nested properties?

LGTM.

deer:
Thanks Lino. If I'm honest, the thing I'm most excited for here is that it closes two issues and we can close another PR as no longer necessary 😅

marvinhagemeister (approved):
LGTM, not sure about the `data` vs `state` distinction, but I don't want to block this. Might be a case where I haven't worked enough with that part of the code base.

lytedev:
Glad my PR was helpful to you! Thanks for carrying this through to the end! :heart:

PR #1270, opened 2023-06-11
Feat/vscode twind intellisense

ybabts:
# What does this feature aim to accomplish?
This feature aims to make the VSCode Intellisense compatible with Twind. Currently, the Twind extension for VSCode is broken, so I have been using the Tailwind CSS extension as a replacement.

# What does it do?
This feature is a plugin for Fresh that clones the Twind config to the Tailwind config, enabling proper auto-completion of custom values. It runs during the initialization of `dev.ts`. With the start task, it automatically regenerates the Tailwind config for VSCode whenever it changes.

# How does it work?
First, it reads the Twind config file and applies a regex pattern to capture the inner Twind config. It checks if the config was successfully captured; if not, the promise is rejected. It then removes the `selfURL: import.meta.url,` line and wraps the Twind config with `module.exports = {$1}`. Finally, it writes the file to the location where the Tailwind CSS extension expects the `tailwind.config.js` file.

# When does this plugin apply?
This plugin is applied during the project initialization process. If the user selects Tailwind and VSCode, this plugin will be automatically applied. Otherwise, it will have no effect.


Norfeldt:
Does this also fix the [issue with typescript 5](https://github.com/tw-in-js/vscode-twind-intellisense/issues/22)?

deer:
@marvinhagemeister, another tailwind victim here.

PR #1272, opened 2023-06-12
fix(www): show missing GA env warning once

marvinhagemeister:
This makes sure that the google analytics middleware only warns once when it's not configured vs warning on every request.


hashrock (approved):
LGTM!

PR #1273, opened 2023-06-12
feat(server): make startup message easier to see

marvinhagemeister:
Noticed that it takes me a tiny amount of time to find the address in our output. It looked a bit too similar to the other output we log, and I think it's a good opportunity to make that more visually distinct. Best explained via pictures.

Before:

Screenshot 2023-06-12 at 08 14 09

After:

Screenshot 2023-06-12 at 08 12 37


hashrock (approved):
LGTM

marvinhagemeister:
Amended a commit that adds the trailing slash like before.

PR #1274, opened 2023-06-12
feat(server): pick free port if none specified

marvinhagemeister:
Noticed that it's a little annoying running multiple fresh projects at the same time because of trying to use the same port by default. When running locally during development I don't really care what port it's on, so we can just pick a free one ourselves. The logic is the same as found in vite or wmr.

This should also allow us to run tests in parallel.

### DIFF
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -106,7 +106,7 @@ export async function start(routes: Manifest, opts: StartOptions = {}) {
         firstError = undefined;
         break;
       } catch (err) {
-        if (err.code === "EADDRINUSE") {
+        if (err instanceof Deno.errors.AddrInUse) {
           // Throw first EADDRINUSE error
           // if no port is free
           if (!firstError) {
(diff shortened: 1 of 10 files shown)


marvinhagemeister:
We'll definitely want to enable parallel testing at some point. It's much faster. Will PR that later today.

| Mode | Time |
|---|--:|
| Sequential | 43s |
| Parallel | 12s |

lucacasonato (commented):
Can we compute the ctx.handler() once instead of multiple times?

marvinhagemeister:
@lucacasonato good idea! Pushed a commit that hoists the handler creation 👍

src/server/mod.ts
```diff
@@ -86,10 +85,51 @@ export async function start(routes: Manifest, opts: StartOptions = {}) {
...
+    // That way the user only needs to increment a number when running
+    // multiple apps vs having to remember completely different ports.
+    let firstError;
+    for (let port = 8000; port < 8020; port++) {
+      try {
+        await bootServer(handler, { ...opts, port });
+        firstError = undefined;
+        break;
+      } catch (err) {
+        if (err.code === "EADDRINUSE") {
```
lucacasonato: This is correct - but `err instanceof Deno.errors.AddrInUse` is more Deno idiomatic.

lucacasonato (approved):

src/server/mod.ts
marvinhagemeister: Good point, I'll change that.

PR #1275, opened 2023-06-12
feat(init): make project name as default header title

huai-jie:
default header title should be same as project name


iuioiua (approved):
LGTM!

marvinhagemeister (approved):
Good idea🎉

PR #1277, opened 2023-06-12
www: fix anchor links not visible

marvinhagemeister:
Before:

Screenshot 2023-06-12 at 13 06 50

After:
Screenshot 2023-06-12 at 13 06 47

### DIFF
--- a/www/routes/docs/[...slug].tsx
+++ b/www/routes/docs/[...slug].tsx
@@ -140,6 +140,21 @@ function MobileSidebar(props: { path: string }) {
           </nav>
         </div>
       </div>
+      <script
+        dangerouslySetInnerHTML={{
+          __html: `
+          var input = document.querySelector('#docs_sidebar');
+          if (input) {
+            input.addEventListener("change", function (ev) {
+              var value = ev.target.checked;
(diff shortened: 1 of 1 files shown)


lucacasonato:
Docs pages with code blocks now overflow on mobile: 
image

www/routes/docs/[...slug].tsx
```diff
@@ -140,6 +140,21 @@ function MobileSidebar(props: { path: string }) {
...
+      <script
+        dangerouslySetInnerHTML={{
+          __html: `
+          var input = document.querySelector('#docs_sidebar');
+          if (input) {
+            input.addEventListener("change", function (ev) {
+              var value = ev.target.checked;
+              var main = document.querySelector("main");
+              if (!main) return;
+              main.classList.toggle("overflow-hidden");
```
Comment (lucacasonato) [inline www/routes/docs/[...slug].tsx:152]:
```suggestion
              main.classList.toggle("${tw`overflow-hidden`}");
```

Otherwise this class may not be generated if we happen to remove `overflow-hidden` from the other components that use it.

crowlKats:
this could be done with just twind rules instead of adding inline js

PR #1278, opened 2023-06-12
feat: don't require connInfo in request handler

lucacasonato:
This makes testing easier, because users now do not need to pass a
fake `ConnInfo` themselves anymore.


marvinhagemeister (approved):
Great idea! LGTM

PR #1279, opened 2023-06-12
www: landing page redesign

marvinhagemeister:
Scroll down for a screenshot.


marvinhagemeister:
There was a bit of an internal discussion about keeping the lemon drop as is internally. So I've removed those changes from the PR. I think it's ready to land now.

localhost_8000_ (3) 1

hashrock (approved):
LGTM!

PR #1283, opened 2023-06-13
fix: update script to work with non-embedded import maps

lucacasonato:
The script now updates non-embedded import maps to embedded ones.


lucacasonato:
Needed this to run update.ts on merch.deno.com

marvinhagemeister (approved):

PR #1284, opened 2023-06-13
Invoke getTime() on the Date value and not on the date signal

najumancheril:
Fix typo in example code

### DIFF
--- a/docs/getting-started/adding-interactivity.md
+++ b/docs/getting-started/adding-interactivity.md
@@ -61,7 +61,9 @@ export default function Countdown(props: { target: string }) {
 
   // Otherwise, we format the remaining time using `Intl.RelativeTimeFormat` and
   // render it.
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good catch! Thanks for filing a PR 🎉

marvinhagemeister:
@najumancheril  Looks like the changes need a format to pass CI. You can do that by running `deno fmt`

najumancheril:
I have made the following changes:
- Fixed a second location where code was using now rather than now.value
- Fixed format
- Reordered the code a bit so we compute seconds once up front. This avoids the annoying "The big event is happening in 0 seconds." message due to timer jitter. Instead, we show "...1 second" and then party. This also reduces the number of now.value references from 2 to 1. Unlike the other edits, this change is not a typo and totally optional. Let me know if you don't want it at all or if you want it in a separate pull request.

marvinhagemeister (approved):
Awesome, thanks for the PR 👍

PR #1285, opened 2023-06-13
feat(islands): allow passing JSX children to island + nested islands

### LINKED ISSUE #168: Support nesting island components

### LINKED ISSUE #395: Support props.children in island components
Such as Dialog component. These components need hooks and props.children

### LINKED ISSUE #750: Support passing JSX components to island props

### LINKED ISSUE #1209: useState not working in _app.tsx
useState not working in _app.tsx, just to show or hide menu on clicking a button. So moved my menu handling code to islands. 

Including islands and passing <Component /> as children to the island causing below error.

TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property '__c' -> object with constructor 'Object'
    --- property '__v' closes the circle
    at JSON.stringify (<anonymous>)
    at render (https://deno.land/x/fresh@1.1.6/src/server/render.ts:280:33)
    at eventLoopTick (ext:core/01_core.js:182:11)
    at async https://deno.land/x/fresh@1.1.6/src/server/context.ts:501:24

Am I missing anything here , please advise.

marvinhagemeister:
This PR allows you to basically mix and match static and dynamic content in whatever way you like. An island can receive JSX children that in turn spawns another island for example. All possible.

Just because we can: Here is a demonstration of the flexibility of that. The red boxes are the JSX subtrees that were purely rendered on the server and passed to an island as JSX children. There is no code sent for these over the wire. They only exists in HTML.

https://github.com/denoland/fresh/assets/1062408/b8d7653d-5da3-4c55-bd40-47d395b87b03

The beauty of this approach is that it works well with DevTools out of the box. We could explore different ways to visualize that these are server components in the future.

Screenshot 2023-06-13 at 18 50 04

## Technical details

The way this works is by applying the same strategy of using comment markers to denote island boundaries for "slots" like `props.children`. During the initial bootup process when we're reviving the islands we now know with the slot markers which HTML nodes in a subtree are children of an island or of a server rendered JSX subtree that was passed as a slot. From that we stitch together the final virtual-dom tree that is rendered in the end.

```html
<div>
  <!--frsh-island:0-->
  <div class="island">
    <!--frsh-slot-island:children-->
    <p>it works</p>
    <!--/frsh-slot-island:children-->
  </div>
  <!--/frsh-island:0-->
</div>
```

This even works when the island renders no HTML and causes markers to be in the same level in the HTML hierarchy.

```html
<div>
  <!--frsh-island:0-->
  <!--frsh-slot-island:children-->
  <p>it works</p>
  <!--/frsh-slot-island:children-->
  <!--/frsh-island:0-->
</div>
```

So when an island receives a JSX child, the `children` prop is marked as a slot. That's the only bit that's sent to the client.

```jsx
<Island> // <-- Client only
  <h1>...</h1> // <-- server
  <p>...</p>  // <-- server
</Island>
```

Island's props sent to the client:

```js
const props = {
  children: { __slot: children}
}
```

As you can see right now only `children` slots are supported. In theory it should be possible to extend this to other props too:

```jsx
<Island foo={<ServerComponent />} /> // <-- future
```

Islands are pretty straight forward in the sense that we have the code to create them form scratch on the client anyway. Turns out that reconstructing what the server rendered is actually pretty easy. Thing is that any server rendered JSX passed to the client (like not islands), must be plain HTML. And you know we _have_ the plain HTML right in front of us. So we can easily reconstruct the virtual tree structure from the HTML to construct the fully fledged out virtual-dom tree. From Preact's point of view it thinks that it rendered everything itself. This means from that point on it's the usual rendering and allows you to conditionally destroy the server rendered bits if needed.

When islands are nested and regardless of whether some server supplied JSX is in between them, the nested islands will be revived in the same hydration call as the parent ones.

```jsx
<Island>
  <Island> // <-will wake up with parent island
    <p>hello from server</p>
  </Island>
</Island>
```

In the future it would be even cooler to bail out of hydration somewhere in the middle of the tree when an island inside another island is detected. But for that to work elegantly some changes to Preact itself are needed as we currently expect you to hydrate the whole tree at once. Need to put some more thought to that.

## Conclusion

I do think the strategy laid out here is pretty strong after playing around with it intensively the past days. It's a good starting point to even optimize this further. We could skip creating the virtual sub tree from the server JSX once Preact gains the ability to skip over DOM nodes and lazily hydrate sections of the tree. Either way, I think it's already useful enough in its current form that it's worth it to land imo.

The list of changes is unfortunately a bit messy due to the number of tests. Basically this PR rewrites all the revival code.

Fixes #168, fixes #750, fixes #395, fixes #1209


src/server/render.ts
```diff
@@ -420,6 +420,31 @@ if (!supportsUnstableComments) {
...
+      }),
+    );
+  } else {
+    return h(
+      `!--${markerText}--`,
+      null,
+      vnode,
+    );
+  }
+}
```
marvinhagemeister, line 446: Extracted this to a function because there are now two places which need this.

deer:
I guess this deals with the confusingly titled https://github.com/denoland/fresh/issues/1209 as well? And looks like this supersedes https://github.com/denoland/fresh/pull/517.

marvinhagemeister:
Good point, I haven't looked into #1209 yet and the title made it sound like the issue is with hooks. Reading it know it seems like it's about passing JSX children to islands too.

src/runtime/entrypoints/main.ts
```diff
@@ -24,49 +32,266 @@ function createRootFragment(
...
+          // Remove markers
+          marker.startNode.remove();
+          sib = sib.nextSibling;
+          marker.endNode.remove();
+          continue;
+        } else if (marker.kind === MarkerKind.Island) {
+          // We're ready to revive this island if it has
+          // no roots of its own. Otherwise we'll treat it
+          // as a standard component
+          if (markerStack.length === 0) {
```
lucacasonato, line 193: Essentially this means that in this configuration, only the outermost island is explicitly revived, right? The inner islands are revived because it is part of the outer islands' subgraph.

```
// route/index.tsx

<Island n={1}>
  <div>
    <Island n={2}>
      <div>Hello</div>
    </Island>
  </div>
</Island>
```

So that means that for both the "inner" and "outer" island, we need to serialize and re-inject props.

---

This is interesting, because you can already "nest" islands inside of each other in a slightly different way:

```
// routes/index.tsx

<OuterIsland />
```

// islands/OuterIsland.tsx

<div><InnerIsland /></div>

// islands/InnerIsland.tsx
<div>Inner!</div>
```

Here, the `InnerIsland` is really just a regular component child of the original island. However, in the current island serializer logic, we still serialize the props for the inner island (but I don't think it gets revived).

Right now the docs say that this is a totally unsupported configuration (because no tests), but clarifying / fixing this is probably also a good follow up.

Anyway - nothing for this PR :sweat_smile:

File: src/server/serializer.ts [hunk 3]
```diff
@@ -108,6 +116,11 @@ export function serialize(data: unknown): SerializeResult {
           referenceArr.push(currentPath);
         }
         return 0;
+      } else if (isVNode(value)) {
+        requiresDeserializer = true;
+        const res = { __slot: key };
```
lucacasonato: This `__slot` isn't used for anything right?

How about `null`, `0`, or an empty object literal instead?

tests/fixture_island_nesting/main.ts
```diff
@@ -0,0 +1,13 @@
+/// <reference no-default-lib="true" />
```
lucacasonato, line 1: I feel like we'll need to improve the test fixture situation here soon. Unfortunate that adding new tests adds so much bloat

tests/fixture_island_nesting/options.ts
```diff
@@ -0,0 +1,11 @@
+import { FreshOptions } from "$fresh/server.ts";
+
+export default {
+  async render(_ctx, render) {
```
lucacasonato: Is this `render` function useful for anything, or just copy paste?

lucacasonato (approved):
LGTM!

lucacasonato:
Can you add something to the docs about this? Probably warrants a whole section in the getting started, and in the reference docs.

src/runtime/entrypoints/main.ts
marvinhagemeister: Ohh that's a good point, that should be doable without much effort to keep serializing props and revive from that. I'll look into that.

What I wanted to say with this section that we cannot stop hydration somewhere in the middle of the tree during hydration. That's something we're hoping to add to Preact at some point in the future. Therefore hydration can only be started by the outermost island.

tests/fixture_island_nesting/main.ts
marvinhagemeister: Yeah true :S On the hand it's kinda neat to just go in a directory and boot up a test fixture to hack on without the test runner around it though. The verbosity of creating these fixtures though is a bit much nonetheless :S

tests/fixture_island_nesting/options.ts
marvinhagemeister: err... just copy & paste 😅

src/server/serializer.ts
marvinhagemeister: Good point, that can be dropped. It's a leftover from an earlier iteration when I used this to infer the slot name. It's not needed anymore.

src/runtime/entrypoints/main.ts
marvinhagemeister: Nevermind! Turns out this already works as expected in the PR. Reason is that as soon as we encounter the opening marker for an island we instantiate the `vnode` and deserialize the props. This path is hit for every island regardless if it's nested or not. Only once we exit the recursion and hit the end marker we'll overwrite the vnode's children.

Added a test for that though. That was indeed missing. Thanks for pointing that out 👍

marvinhagemeister:
@lucacasonato added some docs to the existing island page. I checked the getting started guide, but I feel like no changes are needed there. The page there explains the idea behind islands and how to use them and I think it will be natural for folks to pass childrens to them.

PR #1286, opened 2023-06-13
docs: add "Changing the source directory" example

deer:
closes https://github.com/denoland/fresh/issues/257
supersedes https://github.com/denoland/fresh/pull/456

I reworked this to include the latest project structure and results. @iuioiua, please take a look, since you requested 456 to be updated.

PR #1287, opened 2023-06-13
fix: use `await Deno.serve().finished`

iuioiua:
[A recent change in the runtime](https://github.com/denoland/deno/pull/19485) has changed the behaviour of `Deno.serve()` such that it throws when called a promise. The correct way to do it is instead to call `Deno.serve().finished` as a promise. This PR does that.

Fixes https://github.com/denoland/fresh/pull/1286#issuecomment-1589987387


marvinhagemeister (approved):
Ohh good catch!

PR #1290, opened 2023-06-14
feat(init): add update task

### LINKED ISSUE #1289: Add `deno task update` to generated projects
Just an alias for `deno run -A -r https://fresh.deno.dev/update`

marvinhagemeister:
Fixes #1289

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -304,6 +304,7 @@ const config = {
   lock: false,
   tasks: {
     start: "deno run -A --watch=static/,routes/ dev.ts",
+    update: "deno run -A -r https://fresh.deno.dev/update .",
   },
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
LGTM

nicolo-ribaudo:
Could https://github.com/denoland/fresh/blob/main/update.ts be updated to add this task to my existing `deno.json` file, so that I have to run `deno run -A -r https://fresh.deno.dev/update` once and then since the next update I will be able to just run `deno task update` without needing to manually update my tasks list?

marvinhagemeister:
@nicolo-ribaudo I like this idea! Would only do this though if no `update` task is present, which seems like a fair tradeoff.

PR #433, opened 2022-07-05
feat: trailingSlash

fdemir:
I didn't write a test for this feature, yet. I've created a new property in StartOptions called `route`. We can provide more functionality for routing in the future. 

Please leave some preview. Thanks in advance.

PR #1291, opened 2023-06-14
feat(server): trailing slash configuration

### LINKED ISSUE #359: Add optional trailing slash `/` redirect behavior configuration to manifest
Requests that end with a trailing slash are redirected to their non-trailing slash counterpart ([source](https://github.com/denoland/fresh/blob/main/src/server/context.ts#L275-L277)).

The Fresh manifest could support a new, optional `trailingSlash` config to revert this behavior, where urls without trailing slashes were redirected to their counterparts _with_ trailing slashes `/about -> /about/`. This would exclude static assets.

This flexibility could be desirable in some cases with specific SEO needs.

Happy to look into this and open a PR if the core team greenlights the proposal!

deer:
closes https://github.com/denoland/fresh/issues/359
supersedes https://github.com/denoland/fresh/pull/433

Once again standing on the shoulders of giants here (i.e. I took the solution from 433). I tried adding some documentation for configuring the server as well.


deer:
Is there any way for me to retrigger certain checks?

src/server/types.ts
```diff
@@ -19,6 +19,14 @@ export interface FreshOptions {
   render?: RenderFunction;
   plugins?: Plugin[];
   staticDir?: string;
+  router?: RouterOptions;
+}
+
+export interface RouterOptions {
+  /**
+   *  @default [false]
```
iuioiua:
```suggestion
   *  @default {false}
```
The `@default` tag takes curly braces. Example [here](https://deno.land/std@0.191.0/async/mod.ts?s=DelayOptions).

Could you please add a small description for this property too?

iuioiua (changes requested):
Thanks for adding the test!

Keen to see the server configuration added. But can you please create a separate PR for it?

deer:
> Thanks for adding the test!
> 
> Keen to see the server configuration added. But can you please create a separate PR for it?

Sorry, what do you mean? Which stuff should stay here then?

docs/concepts/server-configuration.md
iuioiua, line 1: To clarify, this file should be part of a separate PR.

deer: Sure, I thought it would be good to introduce documentation for this new feature at the time I was introducing it, but if you want a separate PR I can do that.

iuioiua: Generally, we like each PR to take care of a single thing. E.g.
1. docs: add "Server configuration" concept explanation
2. feat(server): trailing slash configuration

Having both changes in separate PRs makes changes easier to navigate.

deer: if only this were in a "Contributing" section 🙃

iuioiua (approved):
LGTM! Thanks you, @deer.

PR #1292, opened 2023-06-14
docs: add "initializing the server" example

### LINKED ISSUE #235: document server initialization step
It turns out something that is [fairly difficult to set up in sveltekit](https://github.com/sveltejs/kit/issues/927) its pretty easy to add in fresh. I am referring to having an explicit server initialization step (e.g. connect to databases, pull down assets, etc). It looks like this:

#### main.ts:
```ts
/// <reference lib="dom" />
/// <reference lib="deno.unstable" />

import { Context } from './routes/_middleware.ts'

await Context.init()
await start(manifest);
```

#### routes/_middleware.ts:
```ts
import { MiddlewareHandlerContext } from "$fresh/server.ts";
import { Database } from '../database.ts'

export interface State {
  context: Context
}

export class Context {
  private static context: Context

  public constructor() {
    this.database = new Database('sqlite.db')
  }

  public static async init() {
    Context.context = new Context()
  }

  public static instance() {
    if (this.context) return this.context
    else throw new Error('Context is not initialized!')
  }
}

  req: Request,
) {
  ctx.state.context = Context.instance()
  const resp = await ctx.next();
  return resp;
}
```

I can see this being useful for a lot of devs, since setup steps are pretty universal to fullstack apps. Imo its probably worth documenting.

deer:
closes https://github.com/denoland/fresh/issues/235

Turns out there's tons of gold hanging around in issues, just waiting for someone to put it out in the open for people to read. I certainly didn't think of this pattern on my own, but it's very useful.


docs/examples/index.md
```diff
@@ -11,4 +11,5 @@ like to see here, please open
 - [Modifying the `<head>`](./examples/modifying-the-head)
 - [Setting the language](./examples/setting-the-language)
 - [Writing tests](./examples/writing-tests)
-- [Changing the source directory](./changing-the-src-dir.md)
+- [Changing the source directory](./examples/changing-the-src-dir.md)
+- [Initializing the server](./examples/init-the-server.md)
```
iuioiua:
```suggestion
- [Changing the source directory](./examples/changing-the-src-dir)
- [Initializing the server](./examples/init-the-server)
```

iuioiua (approved):
LGTM!

marvinhagemeister (approved):

PR #1294, opened 2023-06-15
feat: Add `basePath` config to host from a directory

### LINKED ISSUE #697: Add support for custom root paths

marvinhagemeister:
Mostly as an exercise for me to get familiar with the routing code in fresh. This PR adds the ability to host a fresh app from a sub path `http://localhost:8000/` -> `http://localhost:8000/foo/bar/` 

The PR is a bit larger mostly due to the bunch of new tests.

Fixes #697


src/server/context.ts
```diff
@@ -180,8 +197,9 @@ export class ServerContext {
         throw new TypeError("Page is not a child of the basepath.");
       }
       const path = url.substring(baseUrl.length).substring("routes".length);
-      const baseRoute = path.substring(1, path.length - extname(path).length);
-      const name = baseRoute.replace("/", "-");
+      let baseRoute = path.substring(1, path.length - extname(path).length);
+      const name = (basePath + baseRoute).replace(/\//g, "-");
+      baseRoute = basePath + baseRoute;
```
marvinhagemeister: A bit unrelated: I feel like the whole code in this file is a bit difficult to follow. Hoping to refactor it sometime in the future.

docs/concepts/server-configuration.md
```diff
@@ -93,3 +99,30 @@ you can configure this to `https://www.example.com/about/` by using the
 ```ts
 await start(manifest, { router: { trailingSlash: true } });
 ```
+
+### basePath
```
iuioiua, line 103:
```suggestion
### `basePath`
```

src/server/context.ts
```diff
@@ -801,6 +834,7 @@ const DEFAULT_RENDER_FN: RenderFunction = (_ctx, render) => {
 
 const DEFAULT_ROUTER_OPTIONS: RouterOptions = {
   trailingSlash: false,
+  basePath: "",
```
iuioiua, line 837: Would things be made easier if `basePath` had a default value of `/`?

iuioiua (approved):
LGTM! Only a couple of nits.

iuioiua:
Also, IMO, having `basePath` configurable as a server option is sufficient. I can't think of a use case for controlling this via an environment variable. Even if there is a use case, I think it's fine for the user to feed the value of the environment variable to the `basePath` property themselves.

docs/concepts/server-configuration.md
```diff
@@ -93,3 +99,30 @@ you can configure this to `https://www.example.com/about/` by using the
...
+
+This setting allows you to serve a Fresh app from sub-path of a domain. A value
+of `/foo/bar` would serve the app from `http://localhost:8000/foo/bar` instead
+of `http://localhost:8000/` for example.
+
+This setting can also be set via the `FRESH_BASE_PATH` environment variable. It
```
lucacasonato, line 109: I don't think we need to add an env var for this OOTB. If someone really wants to do this with env var config, they can set `basePath: Deno.env.get("BASE_PATH")`.

docs/concepts/server-configuration.md
```diff
@@ -93,3 +99,30 @@ you can configure this to `https://www.example.com/about/` by using the
...
+will take precedence over passing the `basePath` option to `start()`.
+
+The `basePath` will be automatically applied to links in your app. For example,
```
lucacasonato, line 112:
```suggestion
The `basePath` will be automatically applied to absolute links in your app. For example,
```

OldManFroggy:
This is a bad idea as a default feature, as mentioned before you can
implement this yourself easily if needed.

Out of the Box, this is yet another security attack surface an
enterprise has to check for every instance of a Fresh APP (or to be more
explicit an attack surface of an Fresh APP that users will not think to
protect.)

Env Vars should be used for things that do not affect where an app is
running from period, when comes to security, as they are trivial to
abuse in a system breach.  This is why most enterprise only use env vars
for keys to secure vaults of settings that are obtained at runtime
(Hashicorp vault, K8s/K3s Secure keys, and others.)

It would also mean, that for my enterprise clients I would have to get
all my Apps security approved again, which is extremely painful from a
personal and company point of view not to mention costly, and likely
would mean I would have to drop Fresh from 10 or so clients as a result.

While I do understand most using Fresh are not as security concerned,
but this PR is not a good move for the default build of Fresh if you
expect large enterprise to use it seriously in a secure environment.

Or of course I am unlucky working security related and concious
companies, just this week we had to tell a vendor to not use env var for
database and api keys not relalted to secure key store in their app,
before it could be used for example.

Sorry for the wall of text all ;)

marvinhagemeister:
> This is a bad idea as a default feature, as mentioned before you can
implement this yourself easily if needed.

@scott-ling Do you think serving fresh from a directory in general is a bad idea or are you referring to being able to set that via environment variables here? It seems like the rest of your post is mostly directed at being against setting that via environment variable. I'll remove that bit from the PR and agree with the others here that setting it from the server options is sufficient.

OldManFroggy:
Yes, you are correct I am mainly concerned on the setting of base
directory from env var, since an easy attack surface, settings in server
options are fine.  

Also, having a place to set it in 2 places seems a tad redundant, just
my 2c.  


On June 26, 2023, SleepyShay ***@***.***> wrote:
> > This is a bad idea as a default feature, as mentioned before you can
> >  implement this yourself easily if needed.
> >
>
> @scott-ling <https://github.com/scott-ling> Do you think serving fresh
> from a directory in general is a bad idea or are you referring to
> being able to set that via environment variables here? It seems like
> the rest of your post is mostly directed at being against setting that
> via environment variable. I'll remove that bit from the PR and agree
> with the others here that setting it from the server options is
> sufficient.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/denoland/fresh/pull/1294#issuecomment-1607311017>,
or unsubscribe <https://github.com/notifications/unsubscribe-
auth/ATXWUNSRLMV7DYN7WSGRCKLXNF2DLANCNFSM6AAAAAAZHSHRR4>.
You are receiving this because you were mentioned.Message ID:
***@***.***>

mcgear:
Is there a timeline of release for this? i am wondering if there may be a more generic way of handling this, similar to how it is handled by react in setting the homepage: '.'

Perhaps your implementation takes this into account?  

Is there a reason we wouldn't want to tackle this using the same approach as most other web frameworks?  Utilizing the html <base href />?

This provides plenty of flexibility to host an application on multiple paths (any number of diverse paths).

mcgear:
To elaborate a bit more, we have an applications that is built in Deno that provides our users the ability to view their real-time IoT device data.  The way our tech works, each user deploys their own instance of our open source Image into their own environments.  This is done with a Revers Proxy configuration to some sub path on their domain... say '/dashboard', but realistically it could be anything they choose (/devices/data or /joe-schmo/stuff).  

In other frameworks like react, we set the homepage to '.' in the package.json and it changes the paths to certain system references (like main.ts, vendor.ts) to be prepended with . making them all paths that honor the base href './{path-here}'.

image

With Fresh, i am seeing a couple of things:
1) When adding the base href into the <Head /> element, the base href is injected after all the system references, and ideally would be first
2) Even with a base href set, the deno server doesn't seem to handle requests correctly because the server is obviously still seeing the request with the full path, and nothing matches that. This would be handled (i think) by this PR, but would require all of our users to rebuild the docker container and deploy the image themselves for any custom sub paths.  The environment variable would work for us in theory in this case, however with the security concerns, wondering how welse we could handle it.

marvinhagemeister:
> Is there a timeline of release for this? i am wondering if there may be a more generic way of handling this, similar to how it is handled by react in setting the homepage: '.'
> 
> Perhaps your implementation takes this into account?
> 
> Is there a reason we wouldn't want to tackle this using the same approach as most other web frameworks? Utilizing the html ?

I'm confused by this. The whole idea of this PR is to mirror the behavior common in other frameworks. Can you elaborate on the differences you found or why you're assuming that the PR deviates from that?

> In other frameworks like react, we set the homepage to '.' in the package.json and it changes the paths to certain system references (like main.ts, vendor.ts) to be prepended with . making them all paths that honor the base href './{path-here}'.

This is exactly what this PR aims to do. Just that we have no `package.json` (it's Deno after all) and that the setting is called `basePath`.

vjdv:
I was hoping this in fresh 1.5

tkovs:
I am trying to serve a Fresh website by rewrites from [NextJS](https://nextjs.org/docs/pages/api-reference/next-config-js/rewrites), but I need to setup a custom path to the build folder (statics, _frsh, etc) in order to avoid conflicts. Since this PR is freezed, is there a way to do this? The goal is to render a Fresh website when you access a page X from my NextJS website.

deer:
I'm working on rebasing this, since lots of things (everything!?) have changed since this was created. All of the tests are passing again, but I'll add a few more to cover some of the new features. Look for an updated PR next week.

deer:
@marvinhagemeister, thanks for starting me off, but I don't think we need this one anymore. 🙃

PR #1295, opened 2023-06-15
fix: rendering islands inside other islands

marvinhagemeister:
Noticed that one use case of nesting islands wasn't supported and this PR addresses that.

```jsx
function Island() {
  return <OtherIsland />
}
```

The previous code had trouble with this too, but it often wasn't a problem because the outer island removed the incorrect HTML.

The approach taken here relies on keeping track of `vnode` "owners". An owner of a `vnode` is the component whose render call lead to the creation of said `vnode`. The same logic is used in lots of tooling related to Preact (DevTools, prefresh, debug) and can be found in other frameworks like React too. That said this makes the fix look a little more involved than it actually is.

This is a follow up to #1285 .

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -1,5 +1,14 @@
 import { renderToString } from "preact-render-to-string";
-import { ComponentChildren, ComponentType, Fragment, h, options } from "preact";
+import {
+  Component,
+  ComponentChildren,
+  ComponentType,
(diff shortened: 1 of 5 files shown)


lucacasonato (approved):
Great, LGTM

Would be good to expose the `__c` and `__r` hooks publically properly at some point :)

PR #1299, opened 2023-06-15
fix(www): incorrect `changing-the-src-dir` path on  `examples` docs page

huai-jie:


marvinhagemeister (approved):
Good catch! Thanks for creating a PR ❤️

Issue #290, opened 2022-06-26
Styling integrations


dragonwocky:
Hey! I've been using UnoCSS with Fresh privately for a while, updating it as Fresh has evolved, and it's seemed to be pretty reliable/fast/capable. I've made that integration available here and documented how to use it: https://github.com/dragonwocky/deno_tools#fresh_unocss

What might be the best way for that to be made available for Fresh users? Just leaving it there as-is, or publishing a module on https://deno.land/x/, or opening a PR to add it to Fresh itself? Is there a plan for any sort of plugin system with Fresh? I expect if that is planned, it might be complex to do right and wouldn't be part of v1.0.0.

I've noticed that the styling page of the docs is still empty, and I'd be happy to write that up, referencing how to use UnoCSS, Twind, post-processed stylesheets and normal stylesheets.

Side question: the `_app.tsx` route isn't documented yet either, and it doesn't really look like there's a gap where it would fit into the current docs? I'd be curious to know the plan for/intended uses of that.

lucacasonato:
I'd love to add plugins for styling integrations. I have unfortunately not found the right way to do this yet, and it won't be part of 1.0.0.

Once I have more time to work on plugins (hopefully in a couple of weeks), we can figure out how to integrate UnoCSS as a CSS engine. Ideally it would just be a built in plugin in the Fresh repo :)

miguelrk:
**FYI:** just opened PR https://github.com/denoland/fresh/pull/1303 which adds a `unocss` plugin based on the new `renderAsync()` method 👍🏼

marvinhagemeister:
We've since moved Fresh to vite. In vite you can pick any styling option you want.

PR #1303, opened 2023-06-15
feat(plugins): add unocss plugin (WIP)

miguelrk:
**TLDR:** [`unocss`](https://unocss.dev/) is an instant on-demand atomic CSS engine and a joy to use. I believe it can (and should) replace `twind` as the default atomic CSS engine for fresh.

### Context

`fresh@v1.2.0` introduced the [`renderAsync`](https://fresh.deno.dev/docs/concepts/plugins#hook-renderasync) method, which allows calling the asynchronous `uno.generate()` method as first [proposed](https://github.com/denoland/fresh/issues/728#issuecomment-1413024482) by @Hladikes (thanks!) but without the workaround. I've tested it for routes, components and islands, and it's working for all (tests are pending). 

### Checklist

This PR is a WIP... I will continue tomorrow with the following:

- [x] enhance defaultConfig
- [x] add tests
- [x] add docs

### Related Issues

https://github.com/denoland/fresh/issues/728
https://github.com/denoland/fresh/issues/290


miguelrk:
Hey @marvinhagemeister! From your comment in the [August 2023 iteration plan issue](https://github.com/denoland/fresh/issues/1485) (didn't want to pollute that conversation), I'm curious to ask what's your stance on [`unocss`](https://unocss.dev/). I find `unocss` to be a better alternative to `twind`, as it is not only NOT tied to `tailwindcss`, but rather CSS framework agnostic. It is also much more flexible for building your own custom CSS framework. The plugin here uses `const { css } = await uno.generate(htmlText)` at the render step, so that aligns with fresh, but it even has the ability of using it at runtime by [loading the `unocss` engine from a CDN](https://unocss.dev/integrations/runtime). IMO, the most important advantage of  `unocss` over tailwind-based alternatives is NOT being tied PostCSS, `unocss` is "pure javascript".

That being said, I haven't had the time to continue with the to-do items in this PR, but I've been using this plugin extensively and can confirm "it just works".

Eager to hear your opinon, I wonder why most people still try to work with `twind` in fresh despite the known issues, when unocss could very much be a plug-in replacement (by adding the `@unocss/preset-uno` which includes TailwindCSS support). I could then take this PR forward 👍🏼

lino-levan:
Super excited to have UnoCSS in Fresh. I have some decently sized apps to test this on for regressions (which I will do later on in the review process for this PR). This is much more "fresh"-y in nature and kind of avoids a whole bunch of twind-related issues. I haven't had the chance to look at code yet but huge +1

marvinhagemeister:
I'm all for adding support for `unocss`. I don't have a horse in the race and don't mind which way to create CSS folks use, as long as it works for them and maintenance is low for me 👍

plugins/unocss.ts
```diff
@@ -0,0 +1,29 @@
...
+`
+
+const defaultConfig: UserConfig = {
+  presets: [presetWind() as unknown as Preset],
```
lino-levan: I wonder if we should use `presetUno()` instead of `presetWind()` here.

miguelrk: I agree, just pushed an update to use `presetUno` instead 👍🏼

plugins/unocss.ts
```diff
@@ -0,0 +1,29 @@
+import type { Preset, UserConfig } from 'https://esm.sh/@unocss/core@0.53.1'
+import { UnoGenerator } from 'https://esm.sh/@unocss/core@0.53.1'
+import presetUno from 'https://esm.sh/@unocss/preset-uno@0.46.3?bundle&no-check'
+import { Plugin } from '$fresh/server.ts'
+
+const unoResetCSS = `/* reset */
```
lino-levan, line 15: Could you add a comment where you're getting this from?

miguelrk: It's originally from [here](https://github.com/Hladikes/fresh/blob/main/plugins/unocss.ts) but not sure which exact reset is being applied from the [official unocss resets](https://unocss.dev/guide/style-reset). Happy to add a comment...

Hladikes: Not gonna lie, it feels great to have my nickname in comments since this is my first sort of a contribution. However I'm not the original author. Here is the [UnoCSS plugin](https://github.com/ije/html/blob/main/plugins/unocss.ts) by [ije](https://github.com/ije), which I took and slightly modified, and in which he commented the name of the used preset:) Take a look at [here](https://github.com/ije/html/blob/6c5612fb81e5159c1a598eb6b37a3f4c72a25fe5/plugins/unocss.ts#L6C52-L6C52)

miguelrk: Ha! Glad to have you around @Hladikes ! Thanks for the pointer, I will update the comment to point to `https://esm.sh/@unocss/reset@0.51.2/tailwind.css` instead 👍🏼

lino-levan:
Any chance you could push this forward @miguelrk? I'd love to pick this up if not.

miguelrk:
> Any chance you could push this forward @miguelrk? I'd love to pick this up if not.

Thanks for the interest and reminder @lino-levan ! I added some basic docs based on the existing `using-twind-v1.md`. Not sure how to proceed with testing though, happy if you could take that over.

In any case, do you think we should be defaulting to using the `presetUno` if no config is provided? This allows simply using the plugin like `await start(manifest, { plugins: [unocssPlugin()] })`, without having to provide your own config or having a `uno.config.ts` altogether. 

I'm leaning towards removing the defaults, since the defaults are pretty simple an can be manually configured inline in a single line:
```ts
import unocssPlugin from "$fresh/plugins/unocss.ts";
import presetUno from "@unocss/preset-uno";

await start(manifest, { plugins: [unocssPlugin({ presets: [presetUno()] })] });```
```
and also very easily imported from another file:
```ts
import unocssPlugin from "$fresh/plugins/unocss.ts";
import unocssConfig from "./uno.config.ts";

await start(manifest, { plugins: [unocssPlugin(unocssConfig)] });```
```
What do you think @lino-levan ?

docs/latest/examples/using-unocss.md
```diff
@@ -0,0 +1,60 @@
...
+import { start } from "$fresh/server.ts";
+import manifest from "./fresh.gen.ts";
+
+import unocssPlugin from "$fresh/plugins/unocss.ts";
+import unocssConfig from "./uno.config.ts";
+
+await start(manifest, { plugins: [unocssPlugin(unocssConfig)] });
+```
+
+The template generates a Twind v0 project by default. If you want to use UnoCSS
```
miguelrk, line 6: I think lines 6-23 are redundant (coming from `using-twind-v1.md`) and could be removed.

lino-levan: Agreed, I think the docs are pretty clear.

docs/latest/examples/using-unocss.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+The unocss config object at `uno.config.ts` can be customized to your liking.
+Refer to the [unocss docs](https://unocss.dev/guide/config-file) for more
+information. If no config is provided, the default config is used, which
+defaults to the following:
+
+```ts
+import type { UserConfig } from "https://esm.sh/@unocss/core@0.53.1";
+import presetUno from "https://esm.sh/@unocss/preset-uno@0.46.3?bundle&no-check";
```
lino-levan: Is there a reason you're using `?bundle&no-check` here?

miguelrk: Hmm good catch, don't remember where that came from, just pushed a fix 2a64f85735d44e935261995b161d72ab5c549499 removing the `?bundle&no-check`. Tested it and works as well 👍🏼

docs/latest/examples/using-unocss.md
miguelrk: Done, removed redundant text in 86c82278775bfe9cbba7d8fb1dcef7b92b247863

lino-levan:
Hey @miguelrk! I can't seem to get the VSCode extension to work, so my conclusion is that it likely doesn't matter. I think we should have a default config to make migration easy, but we should encourage people to use their own config by making a `uno.config.ts`. This could mean that they get intellisense on classes in the future which would be nice.

miguelrk:
Hmm now that you mention it @lino-levan, the `antfu.unocss` vscode extension is also not working for me. In any case, if `unocss` is to replace `twind` as the default for the `fresh init` script, then I would agree we should scaffold an initial `uno.config.ts`. While that's not the case (while `unocss` is not the default for `fresh init`), I think its best to indeed remove the "sensible" default to use `presetUno`, since setup is literally one-line more to set it up yourself, encouraging users to do so, instead of "easing migration" for them.

lino-levan:
> Hmm now that you mention it @lino-levan, the antfu.unocss vscode extension is also not working for me. 

Yeah, we should look into that. I'm sure they'd be happy to accept a PR.

>  I think its best to indeed remove the "sensible" default to use presetUno, since setup is literally one-line more to set it up yourself, encouraging users to do so, instead of "easing migration" for them.

Sure. Let's do that instead. Having no default doesn't seem unreasonable.

miguelrk:
> Sure. Let's do that instead. Having no default doesn't seem unreasonable.

Great! I've done that in b397d6f89dcdf9f5133d6daf78a2ae8d34fb634d and updated the docs. Something else that comes to mind?

docs/latest/examples/using-unocss.md
```diff
@@ -0,0 +1,42 @@
...
+```
+
+The unocss config object at `uno.config.ts` can be customized to your liking.
+Refer to the [unocss docs](https://unocss.dev/guide/config-file) for more
+information. If no config is provided, the default config is used, which
+defaults to the following:
+
+```ts
+import type { UserConfig } from "https://esm.sh/@unocss/core@0.53.1";
+import presetUno from "https://esm.sh/@unocss/preset-uno@0.46.3";
```
lino-levan: Can we keep all of the versions in this at the same number? It's kind of weird that we have all these different versions throughout the codebase
```suggestion
import type { UserConfig } from "https://esm.sh/@unocss/core@0.54.2";
import presetUno from "https://esm.sh/@unocss/preset-uno@0.54.2";
```

plugins/unocss.ts
```diff
@@ -0,0 +1,25 @@
+import type { UserConfig } from "https://esm.sh/@unocss/core@0.53.1";
+import { UnoGenerator } from "https://esm.sh/@unocss/core@0.53.1";
```
lino-levan:
```suggestion
import type { UserConfig } from "https://esm.sh/@unocss/core@0.54.2";
import { UnoGenerator } from "https://esm.sh/@unocss/core@0.54.2";
```
see above

plugins/unocss.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Plugin } from "$fresh/server.ts";
+
+// inline reset from https://esm.sh/@unocss/reset@0.51.2/tailwind.css
```
lino-levan:
```suggestion
// inline reset from https://esm.sh/@unocss/reset@0.54.2/tailwind.css
```
see above

lino-levan (commented):
Outside of tests, this is looking pretty close to ready to me.

lino-levan:
Is there any reason you can't write a fixture test like we have for `twindv1`? It should be almost copy-paste.

miguelrk:
Oh! Hadn't seen the existing twind fixture. That'll help. I'll add this and fixes to your last comments tomorrow morning 👍🏼

miguelrk:
> Outside of tests, this is looking pretty close to ready to me.

Fixture test `fixture_unocss_hydrate` added in c6e81c3551edf6318602cce373401d3f99b1715e based on twind's and pending comments resolved to bump unocss-related dependencies 👌🏼

miguelrk:
@lino-levan let me know if something else comes to mind. In case you missed it, Marvin [agreed](https://discord.com/channels/684898665143206084/991511118524715139/1138794442485145663) to update the `init` script to setup `unocss` instead of `twind` once this is merged.

marvinhagemeister:
FYI: The CI is failing due to formatting. Running `deno fmt` locally should resolve that.

tests/fixture_unocss_hydrate/islands/InsertCssrules.tsx
adamgreg, line 1: I'm sorry if this is a stupid question, but does this work? I can't see how the plugin delivers the unocss JS to the client. Does this just add the `text-green-600` class to the element without defining it in the stylesheet?

I hope review comments from random people aren't too annoying! I just wanted to read through the PR to improve my own understanding, and got stuck on this part.

miguelrk:
> FYI: The CI is failing due to formatting. Running `deno fmt` locally should resolve that.

Thanks for the heads-up @marvinhagemeister , CI should now pass 👍🏼 (it's been in queue for about an hour now). I also added `antfu.unocss` to `.vscode/extensions.json`. NOTE: if this indeed replaces `twind`, then the `sastan.twind-intellisense` should be removed from that list once that happens.

tests/fixture_unocss_hydrate/islands/InsertCssrules.tsx
miguelrk: No prob! There's no such thing as stupid questions 😉 and actually, you just pointed me to an issue with this fixture I think 😆. It appears that the css rule for `text-green-600` is being generated since unocss picks it up from the html "Add \`text-green-600\` to Current Number Class" (=37 css rules in total). If I replace that with e.g. "Add a class to Current Number Class", then unocss won't pick up `text-green-600` from anywhere (=36 css rules in total), so when clicking the button, it won't turn green (since no css rule for `.text-green-600` was generated in the render step).

This plugin only runs on each render (via `renderAsync`), so dynamically updating classes in the client won't cause `unocss` to re-evaluate the html and re-inject css rules (except if you register the [runtime script](https://unocss.dev/integrations/runtime#runtime) via `<script src="https://cdn.jsdelivr.net/npm/@unocss/runtime"></script>`, which might not be desired for this plugin, at least not as a default?). What do you think @lino-levan ? Should we remove this fixture or add an option to inject the unocss runtime script and test it with that?

lino-levan: Yeah, I'm not too sure. Thoughts @marvinhagemeister?

adamgreg: Would it work to just add a boolean option to the plugin config to enable UnoCSS for the client, and add UnoCSS to the `scripts` it returns if the option is set?

Again please excuse my ignorance if I'm missing the whole point. This is how I learn!

miguelrk: Okay, so I've been trying to use some components from `shadcn/ui` and most of them work with the `unocss` plugin, except for dynamically mounted components like dropdowns/dialogs/menus. In these cases, once the dynamic component mounts (e.g. dropdown is activated/opened on click) its html is injected to the DOM client-side (with proper tailwindcss css classes), but these are not picked up by the server-side `pluginUnocss`. Adding the `unocss` runtime script fixes this, since styles are generated client-side dynamically whenever DOM changes are detected. 

Taking that into account, I now think that this `pluginUnocss` should include the client-side runtime by default (to avoid headaches) or at least have the option to include it. The good thing is, that since its a mere javascript engine, its lightweight at **167.61KB** for `https://cdn.jsdelivr.net/npm/@unocss/runtime`.

I hope this discovery sheds some light. How does twind handle this if at all?

lino-levan: I believe twind ships the engine to the client side. That being said, I'm in favor of including the engine by default, with an option to disable it.

lino-levan: Especially with the engine being that small.

miguelrk: Just pushed a7243b15e1bf98b4e6bf5ade503adda5dc9209ce to include the `unocss` runtime script by default (I agree this is best), but with the option to disable it. The plugin signature is now 

```ts
function unocss(config: UserConfig, runtime = true): Plugin
```

With this changes, the `insert-cssrules` fixture test is now working properly:

Screenshot 2023-08-10 8 07 55 AM

One last issue is [preventing FOUC](https://unocss.dev/integrations/runtime#preventing-fouc). For smaller components like that of this test, there is no perceivable "flash of unstyled content", however, with larger components like the `shadcn/ui` dropdown, this flash is very apparent. 

That being said, since this requires users to manually add a custom selector e.g. an `un-cloak` attribute and a CSS rule `[un-cloak] { display: none; }` to circumvent this (as suggested by unocss docs), I think this is out of scope for this plugin. Open to further discuss/improve this.

adamgreg:
Am I right in thinking the client doesn't get access to the config you set in `uno.config.ts`?

miguelrk:
> Am I right in thinking the client doesn't get access to the config you set in `uno.config.ts`?

That's true, good catch! This plugin should [set the provided config directly](https://unocss.dev/integrations/runtime#usage) in `window.__unocss` as well. I've been on holidays but will try to get back to this as soon as I can. Otherwise this is looking close to complete 👍🏼 

```html
<script>
// pass unocss options
window.__unocss = {
  rules: [
    // custom rules...
  ],
  presets: [
    // custom presets...
  ],
  // ...
}
</script>
```

adamgreg:
It looks like CI is failing because the import map in `tests/fixture_unocss_hydrate/deno.json` isn't used as expected when type checking. The plugin itself imports `@unocss/core` (with a different version) directly from an esm.sh url.

I'd suggest it would be good for usability anyway for `plugins/unocss.ts` to re-export `UserConfig`. Then the test fixture and user's `main.ts` can import it from there at the same time as the plugin itself.

Does the import of `@unocss/preset-uno` in the test fixture's `uno.config.ts` need to be made absolute? I don't know how the test fixtures are supposed to be used though.

@miguelrk I'd be happy to make a PR to your PR branch if that's not too weird to do :smile: . I'm very keen to start using this!

miguelrk:
> @miguelrk I'd be happy to make a PR to your PR branch if that's not too weird to do :smile: . I'm very keen to start using this!

Go ahead! It's perfectly fine by me, otherwise I could get to this again maybe later today or tomorrow 👍🏼

adamgreg:
> > Am I right in thinking the client doesn't get access to the config you set in `uno.config.ts`?
> 
> That's true, good catch! This plugin should [set the provided config directly](https://unocss.dev/integrations/runtime#usage) in `window.__unocss` as well. I've been on holidays but will try to get back to this as soon as I can. Otherwise this is looking close to complete 👍🏼
> 
> ```
> <script>
> // pass unocss options
> window.__unocss = {
>   rules: [
>     // custom rules...
>   ],
>   presets: [
>     // custom presets...
>   ],
>   // ...
> }
> </script>
> ```

I tried passing the config in to the plugin's runtime entrypoint using the `state` argument, but it didn't work because the config is full of functions, which can't be serialised to JSON.

Maybe we can transpile/bundle the config file to a JS string using `esbuild.transform()` and pass that as the state argument? Does that sound sensible @marvinhagemeister ?

marvinhagemeister:
@adamgreg yes, that sounds sensible. Plugins can add scripts that are fed to `esbuild` internally. Both twind plugins do this and can be used as an example on how to do that.

FYI: I've added this PR to the next release cycle. Would really like to replace twind with unocss for the next release.

miguelrk:
Great @adamgreg! Thanks for this, I'll get back to my computer on monday, do keep us posted!

adamgreg:
> @adamgreg yes, that sounds sensible. Plugins can add scripts that are fed to `esbuild` internally. Both twind plugins do this and can be used as an example on how to do that.
> 
> FYI: I've added this PR to the next release cycle. Would really like to replace twind with unocss for the next release.

Ah, I see Luca dealt with this same issue for the twind plugin in pull request https://github.com/denoland/fresh/pull/705 . It relies upon the config/options object containing a URL for its own defining module in a `selfURL` property. I'll quickly try doing the same in https://github.com/miguelrk/fresh/pull/1 and see if it works out.

This could be a bit non-obvious for someone setting up a new Fresh project, but the twind plugin benefits from the bootstrapping in `init.ts`. Should UnoCSS be offered as an alternative choice in the init script, as part of this PR?

marvinhagemeister:
Let's skip changing the init script for now. That can happen later in another PR. I'd rather get this in for a while, so that people can test drive it for a week or so. And then we replace twind with unocss in the init script rather than keeping both options around.

My experience is that most folks aren't attached to twind in particular. They just want tailwind syntax to work and don't care how that's done. As long as we provide that in the init script we're good.

adamgreg:
I've added a commit to https://github.com/miguelrk/fresh/pull/1 to make the plugin bundle and send the config file. It works, but it's relatively easy to break things by sending Node/Deno-specific code to the browser. In particular, it's possible to end up importing `node:process` (I think via the `unconfig` package).

For example, you can manually specify the `presetUno` preset in your config file, if you import it from "https://esm.sh/@unocss/preset-uno". If you import it from "https://esm.sh/unocss" though, you hit an import of `node:process` in the browser.

I also don't think it's a great DX to have to extend the config object to add `selfURL`. Are there any other approaches we could try to get the config into the browser with less footguns? I like the idea of serialising the config object itself to JS, instead of bundling the module containing it, so that the URL is not needed. Maybe that's not reasonable though? I'm very much at the limit of my understanding! Probably time for me to take a step back and leave it to the experts :smile:

miguelrk:
> I've added a commit to [miguelrk#1](https://github.com/miguelrk/fresh/pull/1) to make the plugin bundle and send the config file. It works, but it's relatively easy to break things by sending Node/Deno-specific code to the browser. In particular, it's possible to end up importing `node:process` (I think via the `unconfig` package).

Thanks for this @adamgreg ! Nice additions, just merged them 👍🏼 

> For example, you can manually specify the `presetUno` preset in your config file, if you import it from "https://esm.sh/@unocss/preset-uno". If you import it from "https://esm.sh/unocss" though, you hit an import of `node:process` in the browser.

I have had this same issue of `node:process` before when importing from `"https://esm.sh/unocss"`, but the root cause of this will soon be resolved with native `npm:` specifier support in Deploy. So maybe a note/warning in the docs for users to use each preset's URL e.g. `"https://esm.sh/@unocss/preset-uno"` instead of `"https://esm.sh/unocss"` is sufficient, at least for now...

> I also don't think it's a great DX to have to extend the config object to add `selfURL`. Are there any other approaches we could try to get the config into the browser with less footguns? I like the idea of serialising the config object itself to JS, instead of bundling the module containing it, so that the URL is not needed. 

I agree, it's not the best DX extending the unocss config object... Any ideas @marvinhagemeister, @lino-levan ?

miguelrk:
> For example, you can manually specify the `presetUno` preset in your config file, if you import it from "https://esm.sh/@unocss/preset-uno". If you import it from "https://esm.sh/unocss" though, you hit an import of `node:process` in the browser.

I'm hitting the `node:process` import when importing anything from `https://esm.sh/unocss`, so it basically crashes with a separate `uno.config.ts`. Now I see what you meant with "It works, but it's relatively easy to break things", but for me, it's not even working in my own project 😆... I wonder why directly passing-in the config during plugin registration didn't cause this same issue despite using same imports... In any case, this is a blocker if we want to allow auto-loading config from `uno.config.ts` (which we should do IMO) instead of inlining it.

Good idea to open https://github.com/unocss/unocss/issues/3014 💯

adamgreg:
The `unocss` package just seems to collect together exports from various `@unocss/xx` packages. I haven't tested then all, but I know that importing  `@unocss/preset-icons` brings in a `node:process` import, but the individual packages for `presetUno` and `presetMini` are ok. I think it's something to do with an icon loader, so maybe there's a way around it.

I still think serialising the config object itself on the server (including functions) would be ideal, but I don't know if it's feasible!

adamgreg:
I've got a working setup with `presetUno` and `shadcn-ui`, including the runtime, I just had to be careful with the import and add the `selfURL` property, which are not ideal

plugins/unocss.ts
```diff
@@ -0,0 +1,45 @@
...
+    name: "unocss",
+    entrypoints: {
+      "main": `
+        data:application/javascript,
+        import config from "${config.selfURL}";
+        export default function() {
+          window.__unocss = config;
+          const script = document.createElement("script");
+          script.src = "https://cdn.jsdelivr.net/npm/@unocss/runtime";
+          document.head.appendChild(script);
```
zhmushan:
```suggestion
        import init from "https://esm.sh/@unocss/runtime@0.55.1";
        export default function() {
          window.__unocss = config;
          init();
```

docs/latest/examples/using-unocss.md
```diff
@@ -0,0 +1,43 @@
...
+information. If no config is provided, the default config is used, which
+defaults to the following:
+
+```ts
+import type { UserConfig } from "https://esm.sh/@unocss/core@0.55.1";
+import presetUno from "https://esm.sh/@unocss/preset-uno@0.55.1";
+
+export default {
+  presets: [presetUno()],
+} satisfies UserConfig;
```
zhmushan:
```suggestion
import { defineConfig } from "$fresh/plugins/unocss.ts";
import presetUno from "https://esm.sh/@unocss/preset-uno@0.55.1";

export default defineConfig({
  presets: [presetUno()],
  selfURL: import.meta.url,
});
```

miguelrk:
> I've got a working setup with `presetUno` and `shadcn-ui`, including the runtime, I just had to be careful with the import and add the `selfURL` property, which are not ideal

@adamgreg mind sharing how you achieved this? I keep getting `node:` import errors with the new setup using `uno.config.ts`. I'm also testing by trying to use `shadcn-ui` components but can't get past this issue...

**Update:** the issue with `node:` dependencies was causes soleley by `@unocss/preset-icons` as you mentioned. Importing from `@unocss/preset-icons/browser` instead resolves this! 🎉 See https://github.com/unocss/unocss/issues/1946#issuecomment-1336128549

adamgreg:
> > I've got a working setup with `presetUno` and `shadcn-ui`, including the runtime, I just had to be careful with the import and add the `selfURL` property, which are not ideal
> 
> @adamgreg mind sharing how you achieved this? I keep getting `node:` import errors with the new setup using `uno.config.ts`. I'm also testing by trying to use `shadcn-ui` components but can't get past this issue...

@miguelrk  I think to avoid the `node:` import error in the client you just have to avoid certain imports including `unocss` and `@unocss/preset-icons`. Can you get a stack trace to see where the `node:` import is happening?

I've attached a zip file with my uno config, a shadcn preset, and my utils module, all based on this: https://github.com/fisand/unocss-preset-shadcn . I've been manually pasting Shadcn components into my project, and replacing React references with Preact. The `React.ElementRef` and `React.ComponentPropsWithoutRef` types don't seem to have equivalents in Preact, so I've added my own to the utils module. I can't claim to understand them very well though. I've also been replacing the `className` props of shadcn components with `class`. Feel free to DM me in Discord if you want to discuss more, or share more code.

[uno_config.zip](https://github.com/denoland/fresh/files/12419051/uno_config.zip)

miguelrk:
> **Update:** the issue with `node:` dependencies was causes soleley by `@unocss/preset-icons` as you mentioned. Importing from `@unocss/preset-icons/browser` instead resolves this! 🎉 See [unocss/unocss#1946 (comment)](https://github.com/unocss/unocss/issues/1946#issuecomment-1336128549)

@adamgreg thanks for the thorough response! Don't know if you read my update above, but that fixed it 👍🏼 . Thanks for the zip! I'll dm you later today to further discuss how we can move this PR forward!

adamgreg:
I realised that the plugin doesn't support UnoCSS transformers, which need to rewrite the HTML. I wrote and tested a function we could use to apply them to `htmlText`... before realising that the plugin has no way to change the HTML sent to the client! 🤦. @marvinhagemeister would it be at all possible to add this capability to the plugin hooks?

adamgreg:
Raised https://github.com/denoland/fresh/issues/1701

zhmushan:
https://github.com/unocss/unocss/pull/3040 will make Fresh and UnoCSS happy together on VSCode.

miguelrk:
> [unocss/unocss#3040](https://github.com/unocss/unocss/pull/3040) will make Fresh and UnoCSS happy together on VSCode.

Huge thanks for this @zhmushan ! From our conversation earlier:

> I have completed a quick adaptation of the UnoCSS VSCode plugin with Fresh by directly serializing the Fresh configuration file. For the initial version of the UnoCSS plugin for Fresh, let's temporarily abandon support for uno.config.ts because it would require significant changes to the code of the UnoCSS VSCode plugin.

I still don't get why adding support for `uno.config.ts` would require significant changes. What specifically is preventing the `ContextLoader` class supporting `uno.config` like it already supports `fresh.config`?

zhmushan:
Because we cannot be certain that uno.config is specifically designed for Fresh, separating its original logic from Deno's special handling requires some additional work (although it may not necessarily be complex, there might be boundary conditions to consider). On the other hand, for fresh.config, we can confidently determine that it can be handled use Deno, and the UnoCSS plugin has a designated position for handling framework configuration files. Therefore, only minimal effort is needed to achieve quick adaptation.

miguelrk:
@zhmushan I see... I'm still uncomfortable with removing support for `uno.config` just to get the vscode extension to work, but we can maybe find an alternative/solution...

**Alternative 1**

If I understand correctly, what prevents us from using `uno.config` is that the unocss `ContextLoader` requires a separate (config) file e.g. `fresh.config` in order to detect it should use the custom `transform` right? Wouldn't it be safe to assume that ALL deno fresh projects using the plugin will have a `fresh.config` file (as now [required for ALL plugins](https://fresh.deno.dev/docs/concepts/ahead-of-time-builds#migrating-existing-projects-with-plugins))? Because if that's the case, the unocss `ContextLoader` would pick up the `fresh.config` file, but we could read the config from `uno.config` instead of from `fresh.config`. I think its safe to assume that deno fresh projects will have deno-compatible `uno.config` files, so [serialization](https://github.com/unocss/unocss/blob/0844ec429114dfa0d104b64d480e1d8d8c90e636/packages/vscode/deno/fresh.ts) should still work (not sure what you meant with possible edge cases). 

If the above is true, we could import from `uno.config` by default instead of `fresh.config`:

```diff
// unocss/packages/vscode/deno/fresh.ts
- const { default: freshConfig } = await import(Deno.args[0])
- const unoConfig = freshConfig.plugins.filter(it => it.name === 'unocss')
+ const { default: unoConfig } = await import(Deno.args[0])
console.log(JSON.stringify(unoConfig))
```

**Alternative 2**

A simpler workaround (for now) could be declaring the unocss config in `uno.config`  and simply re-exporting from `deno.fresh` for unocss to import from.

```ts
// uno.config.ts
export default defineConfig({...})

// fresh.config.ts
export { default as unoConfig } from "./uno.config.ts"
export default defineConfig({...})

// unocss/packages/vscode/deno/fresh.ts
const { unoConfig } = await import(Deno.args[0])
const unoConfig = freshConfig.plugins.filter(it => it.name === 'unocss')
console.log(JSON.stringify(unoConfig))
```

zhmushan:
@miguelrk 

My PR does not mean permanently giving up on adapting `uno.config` in the Deno project. It is aimed at quickly enabling UnoCSS and Fresh to work together in the most efficient way possible, and to validate feasibility.

Ideally, `uno.config` should work in Deno projects. But it is not as straightforward as imagined. It involves the overall architecture of UnoCSS. [The crucial code resides here](https://github.com/unocss/unocss/blob/3a909d3e4597fa31480fd3bad160eadc85b59cbd/packages/config/src/index.ts#L49-L50).

The prerequisites for adapting Deno are as follows:
1. `@unocss/config` and `@unocss/shared-integration` should not concern themselves with the runtime.
2. Recognition of the Deno environment should only be done in `@unocss/vscode`.
3. `@unocss/vscode` should still create the Context using `@unocss/shared-integration`.

adamgreg:
For sending the config to the runtime, could we maybe avoid esbuild and the need for a `selfUrl` property by using a custom replacer argument to `JSON.stringify()` to handle functions? Something like this:

```js
const configStr = JSON.stringify(config, (key, value) => (typeof value === "function")? value.toString() : value);
```

I believe that would allow standard Uno config files without the extra `selfUrl` property, or allow the config to be defined directly in a `fresh.config.ts` file or similar.

adamgreg:
> For sending the config to the runtime, could we maybe avoid esbuild and the need for a `selfUrl` property by using a custom replacer argument to `JSON.stringify()` to handle functions? Something like this:
> 
> ```js
> const configStr = JSON.stringify(config, (key, value) => (typeof value === "function")? value.toString() : value);
> ```
> 
> I believe that would allow standard Uno config files without the extra `selfUrl` property, or allow the config to be defined directly in a `fresh.config.ts` file or similar.

I've gone off this idea because of the possibility of functions including closures or referencing module-level variables, which would not be serialised.

Instead, I think a better alternative to `selfURL` would be to expect the config to be found in a `uno.config.ts` file at the root of the project. This reflects the recommended usage of UnoCS, and I can't think of a compelling reason for a project not to stick to it. Also, the VS Code plugin makes a similar assumption. I'm working on implementing this now. The config could still be passed into the plugin as an object, but without browser runtime support.

miguelrk:
Hey @marvinhagemeister ! I see you've been working on [build-hooks](https://github.com/denoland/fresh/tree/build-hooks) in relation maybe to https://github.com/denoland/fresh/issues/1700 and https://github.com/denoland/fresh/issues/1701 (for which there's also @adamgreg's PR https://github.com/denoland/fresh/pull/1769). This might just be what we were missing to better integrate UnoCSS to the plugin. However, there will still be some scenarios where shipping the unocss runtime to the client is inevitable, for which it might be best that this plugin provides the following 3 independent options (proposed by @adamgreg):
-  "generate CSS on SSR"
- "transform HTML on SSR"
- "send browser runtime"

Other than that, I've merged @adamgreg's latest changes enforcing the use of the standard `uno.config` file (avoiding having to pass `selfURL` to the plugin options). I think the related [unocss transformers](https://github.com/denoland/fresh/pull/1303#issuecomment-1690352702) mentioned above could be tracked in another issue, in order to move this PR forward, or how do you suggest proceeding?

adamgreg:
I've found a problem with running `UnoGenerator` over the rendered HTML (other than performance) - any `&` in the source have been replaced by `&amp;`. This means that UnoCSS can not pick up classes like that contain arbitrary variants such as `[&>div]:bg-red`.

Any generation of CSS during SSR should be done via a Preact hook, like the twind plugin does.

adamgreg:
> I've found a problem with running `UnoGenerator` over the rendered HTML (other than performance) - any `&` in the source have been replaced by `&amp;`. This means that UnoCSS can not pick up classes like that contain arbitrary variants such as `[&>div]:bg-red`.
> 
> Any generation of CSS during SSR should be done via a Preact hook, like the twind plugin does.

I've made a PR to improve SSR into this branch: https://github.com/miguelrk/fresh/pull/4

adamgreg:
Based on the latest discussion of this with @marvinhagemeister on the Discord channel, it sounds like we should generally be targeting AOT usage, not JIT. I think we just need a single JIT/AOT option:

* AOT mode (enabled by default): Run UnoCSS over the codebase in an after-build hook. In the render hook the plugin can insert a link to the generated CSS file (is this possible currently?). For a decent developer experience, the plugin would also have to detect during initialisation when the server is running in "dev" mode, and run UnoCSS over the codebase immediately - hopefully that is not too slow.
* JIT mode (disabled by default): Extract classes during SSR _and_ send runtime to the client (basically the default mode that currently exists). Use cases:
    1. During development to allow experimentation with the classes in the browser DevTools
    2. Possibly to save time when the server restarts, if UnoCSS extraction from source code takes more than a couple of seconds.

@marvinhagemeister @miguelrk thoughts?

adamgreg:
Transformer support is perhaps not essential... however... I believe the only way to support transformers would be by plugging into the Preact render, JIT-style. Before running class extraction over the source code, we could run transformers to make sure that all the necessary classes are generated. However, the HTML/DOM rendered JIT by Preact would itself still need to be transformed in order to actually _use_ those generated classes. A disabled-by-default plugin option could enable transformer support, with the added render performance cost.

Maybe at some point UnoCSS will add an extractor which works with un-transformed variant group classes, as an alternative to the variant group transformer. That seems to me the only really compelling use for transformers at the moment.

marvinhagemeister:
@adamgreg Oh does UnoCSS only do variants through transpilation?

adamgreg:
> @adamgreg Oh does UnoCSS only do variants through transpilation?

Afraid so: https://unocss.dev/transformers/variant-group. It seems to me like it should be possible to extract directly, but maybe the parentheses or spaces would be a problem for class names

adamgreg:
I've just checked, and WindiCSS also transpiles the HTML to make variant groups work.

I think it's inevitable - there's no way to escape spaces in CSS class names. The parentheses aren't a problem though, and it would be possible to implement a variant 🙄 of this feature in a pure extractor, using a different delimiter instead of spaces. I've experimented a little here: https://jsfiddle.net/kypjLdc8/ . I think commas are very understandable. It's more of a UnoCSS feature request than a Fresh one though.

In any case, I think it would be good if the plugin allowed opt-in to transformers by patching into Preact. AOT support is the priority though.

miguelrk:
As discussed with @adamgreg, closing this in favor of the following with full support for AOT builds:
- https://github.com/denoland/fresh/pull/1966

Linking these other relevant PRs as well:
- https://github.com/denoland/fresh/pull/1963
- https://github.com/denoland/fresh/pull/1999

PR #1306, opened 2023-06-16
docs: add "Using Fresh canary version" example

deer:
I've done all of these things at one point or another, and I guess it would be useful if these techniques were documented.

Please help with some of the wording, in particular "nonstandard" and "latest commit". Maybe there are better words / phrases available.

### DIFF
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -13,3 +13,4 @@ like to see here, please open
 - [Writing tests](./examples/writing-tests)
 - [Changing the source directory](./examples/changing-the-src-dir)
 - [Initializing the server](./examples/init-the-server)
+- [Using nonstandard Fresh](./examples/using-nonstandard-fresh)
--- a/docs/examples/index.md
+++ b/docs/examples/index.md
@@ -13,4 +13,4 @@ like to see here, please open
 - [Writing tests](./examples/writing-tests)
 - [Changing the source directory](./examples/changing-the-src-dir)
 - [Initializing the server](./examples/init-the-server)
-- [Using nonstandard Fresh](./examples/using-nonstandard-fresh)
+- [Using Fresh canary version](./examples/using-fresh-canary-version)
--- a/docs/examples/using-nonstandard-fresh.md
+++ b/docs/examples/using-nonstandard-fresh.md
@@ -1,0 +1,78 @@
+---
+description: |
+  For cases where the latest release doesn't fit your needs.
+---
+
+Pretend you have a use case where you need to modify your project to use a
+nonstandard version of Fresh. Or you want to use a slightly different
+initialization script. This page has you covered.
+
+## Nonstandard Fresh in `deno.json`
+
+### Latest commit
+
+Let's say you like living life in the fast lane, and want the latest and
+greatest from `main`. How can you modify your project to no longer use the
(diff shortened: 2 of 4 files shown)


docs/examples/index.md
```diff
@@ -13,3 +13,4 @@ like to see here, please open
 - [Writing tests](./examples/writing-tests)
 - [Changing the source directory](./examples/changing-the-src-dir)
 - [Initializing the server](./examples/init-the-server)
+- [Using nonstandard Fresh](./examples/using-nonstandard-fresh)
```
marvinhagemeister: "nonstandard" sounds a little weird to me. What if we rephrase this document as "Fresh canary version" instead? That's likely the most popular reason to go down this route anyway. I prefer this over "nonstandard" as that kinda conveys that APIs we provide aren't good enough to hook into. If such a case pops up I'd rather address any shortcomings in our API instead.

docs/examples/using-nonstandard-fresh.md
```diff
@@ -0,0 +1,78 @@
...
+Instead of doing it like this:
+
+```sh
+deno run -A -r https://fresh.deno.dev/
+```
+
+do it like this:
+
+```sh
+deno run -A -r ~/code/fresh/init.ts
```
marvinhagemeister: How do you feel about changing this to `path/to/fresh/init.ts`? Just as an additional hint that this is an example path.

marvinhagemeister (commented):
This is a great idea! I really like the document and only commented on two minor nits. Let me know what you think 👍

docs/examples/using-nonstandard-fresh.md
deer: sounds good

docs/examples/using-nonstandard-fresh.md
```diff
@@ -0,0 +1,78 @@
...
+Pretend you have a use case where you need to modify your project to use a
+nonstandard version of Fresh. Or you want to use a slightly different
+initialization script. This page has you covered.
+
+## Nonstandard Fresh in `deno.json`
+
+### Latest commit
+
+Let's say you like living life in the fast lane, and want the latest and
+greatest from `main`. How can you modify your project to no longer use the
```
lucacasonato: We should not encourage people to import directly from `main`, but rather from a specific commit.

docs/examples/index.md
deer: I specifically wanted help with this, so I'm happy to change it. But what exactly are you suggesting? The document should be titled `Using Fresh canary version`? And then the following:
* ... a nonstandard version of Fresh. -> ... a canary version of Fresh.
* Nonstandard Fresh in `deno.json` -> Canary Fresh in `deno.json`

marvinhagemeister: Yup, those wording changes sound much better imo 👍

docs/examples/using-nonstandard-fresh.md
deer: This also improves the quality of the document as well, since I hadn't covered using a particular commit. And of course the first time I tried doing it (just now) I got the format wrong. Best to have this documented for future users. Thanks.

marvinhagemeister (approved):
LGTM thanks for the PR 🙌

PR #1311, opened 2023-06-16
docs: add "Sharing state between islands" example

### LINKED ISSUE #1248: State sharing between islands using signals
Needs a docs page, maybe also a Getting Started entry

deer:
closes https://github.com/denoland/fresh/issues/1248

I'm definitely not sure if this is what Luca was looking for when he opened 1248. I've basically just lifted the code from https://fresh-with-signals.deno.dev/ with a minimal amount of explanation. Breaking it down like this was super helpful for my personal understanding; not sure if it's good for others.


docs/examples/sharing-state-between-islands.md
```diff
@@ -0,0 +1,186 @@
...
+here, yet.
+
+## Multiple Sibling Islands with Shared State
+
+But we can switch things up by looking at a `Slider.tsx` like this:
+
+```ts
+import { signal } from "@preact/signals";
+
+const slider = signal(50);
```
marvinhagemeister: Whilst this is a valid pattern, let's push folks to use `useSignal` instead. Like a good example would be to instantiate `useSignal` inside a route file and then pass that to two different islands.

The signal here is not just shared between islands, but also for every request, which is likely unintended.

deer: Sadly I don't follow. This is in an island. How is the signal shared across every request?

I can rework this to use `useSignal` and then pass it in instead.

marvinhagemeister: Right, I missed that this is in an island. Still showing how signals are shared outside of that is more useful which means they'll be passed as props to the island.

docs/examples/sharing-state-between-islands.md
```diff
@@ -0,0 +1,195 @@
...
+  const sliderSignal = useSignal(50);
+  return (
+    <div>
+      <SynchronizedSlider slider={sliderSignal} />
+      <SynchronizedSlider slider={sliderSignal} />
+      <SynchronizedSlider slider={sliderSignal} />
+    </div>
+  );
+}
+```
```
marvinhagemeister, line 87: This is great!!

flaky_test_tester.ts
```diff
@@ -0,0 +1,29 @@
+async function main() {
+  const runs = 100;
+  let failures = 0;
+
+  for (let i = 0; i < runs; i++) {
```
marvinhagemeister: Was this file committed accidentally?

deer: !!!!! 🙈

deer: This is what multitasking does :(

deer:
Is there a way to retrigger tests without a new commit?

marvinhagemeister (approved):

PR #1314, opened 2023-06-17
feat(server): middleware params

### LINKED ISSUE #903: Why is `params` available in `HandlerContext` but not in `MiddlewareHandlerContext`?

deer:
closes https://github.com/denoland/fresh/issues/903
supersedes https://github.com/denoland/fresh/pull/1025

I drew inspiration from the deficiency in the comment of 1025. Sylc says:
> For a middleware to access the params that are downstream from its level would require, I think, to duplicate in fresh the logic that is currently in the Rutt router.

Fortunately some problems get solved by just sitting around. The router was inlined into Fresh in April via https://github.com/denoland/fresh/pull/1122. So now, as my test confirms, middleware at all layers have access to all the params.

I did this by refactoring the router into two steps:
1. compute the params and route, based on the request
2. do the actual routing

If you look at the existing [router](https://github.com/denoland/fresh/blob/1d606416a8f87016fc08ae5dba3d82c2571f1314/src/server/router.ts#L128) you can see these two parts are there, but they're all lumped together in one method. Everything for the computation of `groups` (i.e. params) is done as we're looping over `processedRoutes`. Presumably this is for speed, so we don't have to loop twice.

What I've done is pull this code out to compute the params and route ahead of time. Then we can set up the `MiddlewareHandlerContext` with the params. Later on, we can reuse the pre-computed params and route when invoking the `FinalHandler` like so:
```ts
inner(
  req,
  ctx,
  paramsAndRouteResult.params,
  paramsAndRouteResult.route,
)
```

Obviously some of the naming here is horrible and I'm happy to receive suggestions. Problems: `paramsAndRouteResult` and `getParamsAndRoute`.


deer:
I see Luca has a `handler_2` branch, so maybe this can get closed in lieu of that. Or I can rework this to build on that new version, once it's merged.

marvinhagemeister (approved):
This is great! Love the tests 🙌 

EDIT: Oh no things my merge just before this lead to a conflict here with the `fresh.gen.ts` 🙈  We really need to get rid of that file in the future.

PR #1318, opened 2023-06-17
fix: Exclude Twitter/Slack bots from root page redirects

### LINKED ISSUE #1313: The social media card of our website isn't branded
The social media cards when you share a link to `fresh.deno.dev` on twitter, slack or somewhere else don't match the fresh brand.

kamilogorek:
It's caused by 307 redirects that most Open Graph protocols happily follow.
Unfortunately, there's no single trivial way to "catch" them all, so we'll need to go the "manual" route for now and add them for every more widely used vendor.

Fixes https://github.com/denoland/fresh/issues/1313

---

Before Twitter:
image

After Twitter:
image

Before Slack:
image

After Slack:
image

### DIFF
--- a/www/routes/index.tsx
+++ b/www/routes/index.tsx
@@ -11,10 +11,18 @@ import Projects from "../components/Projects.tsx";
 import projects from "../data/showcase.json" assert { type: "json" };
 import Header from "../components/Header.tsx";
 
+function isOpenGraphUA(header: string | null): boolean {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Awesome, this looks much better! A bit of a bummer that we need to do UA sniffing, but oh well... Anyhow thank you so much for the PR ❤️

PR #1323, opened 2023-06-18
fix: incorrect coloring in ansi256 terminals

marvinhagemeister:
Looks like the built in `Terminal.app` app on macOS is the only one which doesn't support 24bit colors yet. I've attempted to work on a more generalised solution for converting color spaces in deno itself in https://github.com/denoland/deno/pull/18549 , but that takes a little more time to get ready. For now just using the closest match for fresh's brand color in `ansi256` space is a good enough.

Before:

Screenshot 2023-06-18 at 20 23 43

After:
Screenshot 2023-06-18 at 20 23 53

Here is a comparison on the original color value from a terminal that supports 24bit colors and the closest ansi256 color.

Original:

Screenshot 2023-06-18 at 20 28 38

Ansi256:

Screenshot 2023-06-18 at 20 28 18

### DIFF
--- a/src/server/mod.ts
+++ b/src/server/mod.ts
@@ -80,11 +80,12 @@ export async function start(routes: Manifest, opts: StartOptions = {}) {
   if (!opts.onListen) {
     opts.onListen = (params) => {
       console.log(
-        `\n%c 🍋 Fresh ready %c`,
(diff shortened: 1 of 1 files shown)


lucacasonato (approved):
LGTM - @bartlomieju would be good to get the upstream fix moving

Issue #1322, opened 2023-06-18
Server-side component rendered based on client-side condition doesn't work


nicolo-ribaudo:
```tsx
// islands/Togglable.tsx

export default function Togglable({ visibleSignal, children },) {
  return visibleSignal.value ? <>AAA: {children}</> : null;
}
```

```tsx
// islands/Button.tsx

export default function Button({ signal }) {
  return (
    <button
      onClick={() => { signal.value = !signal.value }}
    >Toggle</button>
  );
}
```

```js
// routes/page.tsx

import Togglable from "@/islands/Togglable.tsx";
import Button from "@/islands/Button.tsx";
import { signal } from "@preact/signals";

const show = signal(false);

export default function Page({ data }) {
  return (
    <>
      <Togglable visibleSignal={show}>
        BBB
      </Togglable>
      <Button signal={show}>
    </>
  );
}
```

When clicking on "Toggle", it only renders `AAA:` and not `AAA: BBB`. Additionally, `AAA:` is rendered after the button but it should be before.

Versions:
```json
    "$fresh/": "https://deno.land/x/fresh@1.2.0/",
    "preact": "https://esm.sh/preact@10.15.1",
    "preact/": "https://esm.sh/preact@10.15.1/",
    "preact-render-to-string": "https://esm.sh/*preact-render-to-string@6.1.0",
    "@preact/signals": "https://esm.sh/*@preact/signals@1.1.3",
    "@preact/signals-core": "https://esm.sh/*@preact/signals-core@1.2.3",
```

marvinhagemeister:
The ordering is an obvious bug and I think I have an idea what's going on there. I'm less sure what to do with conditional children. The only solution I can come up with right now is to detect if the passed `children` were rendered and serialize them in case they were not.

nicolo-ribaudo:
Thanks for the fix! :)

PR #1324, opened 2023-06-18
fix: conditional sibling islands DOM order

marvinhagemeister:
This happened when both share the same parent and one or both render nothing into the DOM. Once an island starts rendering something into our root fragment that we wrap around each island, it was guaranteed to call `.appendChild()`. But when islands share the same parent this leads to children being added to the very end of the parent DOM node instead of just of the end of the current island tree.

For that reason we need to keep at least the end comment marker of an island as a reference point of where `appendChild` should insert children.

Fixes the first issue mentioned in #1322

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -12,6 +12,7 @@ import { assetHashingHook } from "../utils.ts";
 function createRootFragment(
   parent: Element,
   replaceNode: Node | Node[],
+  endMarker: Text,
 ) {
   replaceNode = ([] as Node[]).concat(replaceNode);
   // @ts-ignore this is fine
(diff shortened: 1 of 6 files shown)


marvinhagemeister:
Random thought: Another alternative could be to insert an empty text node if we want to get rid of the comment.

lucacasonato (commented):
Great! Can you try the empty text node approach? If too complicate, this is fine too.

This doesn't fix the issue if the conditional element is `props.children`, right?

marvinhagemeister:
Switched to using empty text nodes as markers. Seems to work too and haven't run into CSS issues so far. If something pops up in the future we can always fall back to keeping the comment marker easily.

lucacasonato (approved):
LGTM

PR #1330, opened 2023-06-20
Feat prefix

### LINKED ISSUE #1066: INTERNAL_PREFIX can change it to the relative path [./_frsh]?

jiawei397:
Fix: https://github.com/denoland/fresh/issues/1066

I tried to use relative path as ./_frsh, but it may not work when use sub routes. The best and simple way to resolve the proxy problem seems to add an env variable to enable project to change the prefix self.

You can use `FRESH_INTERNAL_PREFIX=/example/_frsh deno task www` start your server. 

The requirement is that `FRESH_INTERNAL_PREFIX` must start with `/` and not ends with `/`, as this function is not commonly used, compatibility processing will not be performed.

Then add nginx config:
```nginx
location ^~ /example/_frsh/ {
      proxy_pass http://xxx:8080;
}
location ^~ /example2/_frsh/ {
      proxy_pass http://xxx:8081;
}
```
PS: The reason why I moved `INTERNAL_PREFIX` from `src/runtime/utils.ts` to `src/server/constants.ts` is that the runtime code is called in the browser and cannot directly use environment variables. My previous solution was to add a `typeof Deno` check, but later I found out that this variable was only used in the server and nowhere else, so I moved it directly.


marvinhagemeister:
Have you seen #1294 ? That would allow you to serve the whole app from a sub-path.

jiawei397:
Great, if the request for 1294 is merged, I don't need it anymore.

PR #1333, opened 2023-06-21
Oldest Deno version should change to 1.31.0

benebsiny:
Fresh uses `Deno.Command` since [this commit](https://github.com/denoland/fresh/commit/7e72fa236c1abc183aae31e5532af914ed060ec5) (release as [1.1.5](https://github.com/denoland/fresh/compare/1.1.4...1.1.5)) to run command. However, `Deno.Command` was first introduced in [Deno 1.28.0](https://github.com/denoland/deno/releases/tag/v1.28.0) as an unstable function and became stable at [Deno 1.31.0](https://github.com/denoland/deno/releases/tag/v1.31.0).

If you're running latest Fresh with older version of Deno, an error will be raised as below:

```
error: Uncaught (in promise) TypeError: Deno.Command is not a constructor
  const proc = new Deno.Command(Deno.execPath(), {
               ^
    at generate (https://deno.land/x/fresh@1.2.0/src/dev/mod.ts:91:16)
    at https://deno.land/x/fresh@1.2.0/init.ts:378:7

```

Therefore, the oldest version of Deno should change to 1.31.0.

If there's any problem of this PR, please let me know. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The [documentation](https://fresh.deno.dev/docs/) is available on
 
 ## 🚀 Getting started
 
-Install [Deno CLI](https://deno.land/) version 1.25.0 or higher.
+Install [Deno CLI](https://deno.land/) version 1.31.0 or higher.
 
 You can scaffold a new project by running the Fresh init script. To scaffold a
 project run the following:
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):
Oh good catch! You're absolutely right. Thank you so much for making a PR to fix this ❤️

PR #1340, opened 2023-06-21
chore: add index layout to island tests

marvinhagemeister:
This PR adds links to the index route of the island fixture for nested islands. Makes it much easier to click through them.


marvinhagemeister:
Merging myself because this only affects testing fixtures.

PR #1343, opened 2023-06-22
docs: add section about "Middleware Destination" feature

### LINKED ISSUE #1341: Docs: Explain how middleware can execute only for routes
The Fresh documentation seems slightly misleading with regards to when Middleware code is executed. [The docs state the following](https://fresh.deno.dev/docs/concepts/middleware):

> _**Route**_ middleware
> A middleware is defined [...] to perform custom logic before or after the _**route**_ handler.

But when generating a new Fresh project and creating a middleware that logs each `Request.url`, we see URLs logged from files that are not "routes" (i.e. under the `routes/*` folder):

<details><summary><code>GET</code> requests for <code>/</code></summary>
<p>

```
http://localhost:8001/
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/deserializer.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/signals.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/chunk-PDMKJVJ5.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/chunk-UGFDDSOV.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/plugin-twind-main.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/main.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/chunk-FJBXTIC3.js
http://localhost:8001/_frsh/js/edcc08a7371927e996ddee1a948a0cbd0df68a05/island-counter.js
http://localhost:8001/_frsh/refresh.js
http://localhost:8001/logo.svg?__frsh_c=edcc08a7371927e996ddee1a948a0cbd0df68a05
http://localhost:8001/_frsh/alive
http://localhost:8001/favicon.ico
````

</p>
</details> 

**Non-route requests**
- `/_frsh/*` for client Javascript and Dev socket (x10)
- Files from the `/static` folder (`favicon.ico` and `logo.svg`)

## Motivation

I think there are great cases for both cases; running middleware only for `/routes` and running for all requests (including static files and files served by the Fresh runtime).

**Running for all requests**
- Logging
- Custom headers
- Access control (hiding access to images for anonymous users etc)

**Running only for `/routes/*`**
- Setting (session) cookie(s)
- Notifications/Other UI status (from e.g. Deno KV)
- Retrieving navigation-tree (often loaded from other systems in E-commerce)

For the latter scenarios, it's common to add data to `ctx.state` so that it's available to routes when rendering the app. But it would be very bad if these run for e.g. Javascript files.

```ts
export async function handler(req: Request, ctx: MiddlewareHandlerContext<State>) {
  const [notifications, navTree] = await Promise.all([
    getUserNotifications(req.headers),
    getNavigationTree()
  ];
  ctx.state.notifications = notifications;
  ctx.state.navTree = navTree;
}
```

## Feature request

It would be great if it was possible to somehow configure that `/routes/_middleware.ts` will run only for all routes in the `/routes` folder, and ignore static/fresh files. I think the most intuitive thing could be that "global" middleware is configured in `main.ts` as `StartOptions` and `/routes/_middleware.ts` only handles routes, but that would be a breaking change.

If it's out of scope to separate global/route middleware, then it would be nice to have some utility/helper method that can help filter out requests for static/fresh files (see workaround below).

## Current workarounds

The best I've come up with so far to identify "route" requests for middleware is:

- Ignore URLs that start with `/_frsh/`
- If HTTP method is `GET`, check headers to determine if top-level navigation, using one of the below options
  - `Sec-Fetch-Mode` header is `navigation`
  - `Sec-Fetch-Dest` header is `document`
  - `Accept` header has `text/html`

The above is not perfect, but at least it seems to mostly work okay, and avoids extra network requests for static and fresh-generated files.

deer:
closes https://github.com/denoland/fresh/issues/1341

The issue seems to be the exact case introduced by the "middleware destination" concept, but there isn't any documentation for it! Here's a small contribution to kick things off.

### DIFF
--- a/docs/concepts/middleware.md
+++ b/docs/concepts/middleware.md
@@ -90,3 +90,29 @@ export const handler = [
   },
 ];
 ```
+
+## Middleware Destination
+
+See these [two](https://github.com/denoland/fresh/issues/521)
+[issues](https://github.com/denoland/fresh/issues/1341) which led to this
+[PR](https://github.com/denoland/fresh/pull/1123), but there is now a way to
+determine the request's destination. `MiddlewareHandlerContext` looks like this:
+
+```ts
+export interface MiddlewareHandlerContext<State = Record<string, unknown>>
+  extends ConnInfo {
+  next: () => Promise<Response>;
+  state: State;
+  destination: router.DestinationKind;
+}
+```
+
+and `router.DestinationKind` is defined like this:
+
+```ts
+export type DestinationKind = "internal" | "static" | "route" | "notFound";
+```
+
+This is useful for if you want your middleware to only run when a request is
+headed for a `route`, as opposed to something like
+`http://localhost:8001/favicon.ico`.
--- a/docs/concepts/middleware.md
+++ b/docs/concepts/middleware.md
@@ -173,8 +173,3 @@ That first `route` request is for when `Fresh` responds with the root level
 `index.tsx` route. The rest, as you can see, are either `internal` or `static`
 requests. You can use `ctx.destination` to filter these out if your middleware
 is only supposed to deal with routes.
-
-For additional context, these
-[two](https://github.com/denoland/fresh/issues/521)
-[issues](https://github.com/denoland/fresh/issues/1341) led to this
-[PR](https://github.com/denoland/fresh/pull/1123) which introduced this feature.
--- a/docs/concepts/middleware.md
+++ b/docs/concepts/middleware.md
@@ -93,10 +93,7 @@ export const handler = [
 
 ## Middleware Destination
 
-See these [two](https://github.com/denoland/fresh/issues/521)
-[issues](https://github.com/denoland/fresh/issues/1341) which led to this
-[PR](https://github.com/denoland/fresh/pull/1123), but there is now a way to
-determine the request's destination. `MiddlewareHandlerContext` looks like this:
+To set the stage for this section, `MiddlewareHandlerContext` looks like this:
 
 ```ts
 export interface MiddlewareHandlerContext<State = Record<string, unknown>>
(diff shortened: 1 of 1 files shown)


docs/concepts/middleware.md
```diff
@@ -90,3 +90,29 @@ export const handler = [
   },
 ];
 ```
+
+## Middleware Destination
+
+See these [two](https://github.com/denoland/fresh/issues/521)
+[issues](https://github.com/denoland/fresh/issues/1341) which led to this
+[PR](https://github.com/denoland/fresh/pull/1123), but there is now a way to
```
marvinhagemeister: I feel like this section should provide a small blurb about what you can do with middleware destinations instead of pointing at issues in the tracker. What do you think?

deer: I tried to communicate this with the following section:
```
This is useful for if you want your middleware to only run when a request is
headed for a `route`, as opposed to something like
`http://localhost:8001/favicon.ico`.
```
Perhaps I should expand upon this? And maybe I should put this at the top and put the issues and PR at the bottom, for reference.

deer: I've completely redone it. Thanks for holding a high bar of quality, @marvinhagemeister.

Perhaps we need a followup example that actually has useful code (as opposed to what I have here, which is just logging values to the console). If you want, I can add the example here. Something like "Running middleware only for routes". Thoughts?

docs/concepts/middleware.md
```diff
@@ -90,3 +90,91 @@ export const handler = [
...
+
+That first `route` request is for when `Fresh` responds with the root level
+`index.tsx` route. The rest, as you can see, are either `internal` or `static`
+requests. You can use `ctx.destination` to filter these out if your middleware
+is only supposed to deal with routes.
+
+For additional context, these
+[two](https://github.com/denoland/fresh/issues/521)
+[issues](https://github.com/denoland/fresh/issues/1341) led to this
+[PR](https://github.com/denoland/fresh/pull/1123) which introduced this feature.
```
marvinhagemeister: Let's drop this part. It's not important for a manual to point to all the historical bits on how the feature came to be, unless the explanation of the feature is lacking.

marvinhagemeister (commented):
Love this! This is are good docs!!

Left one minor comment about dropping the passage about linking to closed issues/PRs. Once that's been removed it's ready to be merged. Let me know what you think.

marvinhagemeister (approved):

docs/concepts/middleware.md
deer: Ok, I like the extra context, but maybe I'm weird 🤷

marvinhagemeister:
Thanks again for this PR. Really appreciate it and sorry if I come off as nitpicky at times.

PR #1347, opened 2023-06-23
docs: add "Dealing with CORS" example

### LINKED ISSUE #1247: Add docs on how to enable CORS for a route
IE how to handle OPTION requests in a route, or maybe using a middleware.

deer:
closes https://github.com/denoland/fresh/issues/1247

I'll let someone who has spent more than one day thinking about CORS have the final say on how this should be phrased. But this is at least a starting point to critique / improve upon.

Technically it does even solve the problem, as evidenced by the links [here](https://github.com/denoland/fresh/issues/1247#issuecomment-1603079202). Maybe these should be included in the documentation as an example?

### DIFF
--- a/docs/examples/dealing-with-cors.md
+++ b/docs/examples/dealing-with-cors.md
@@ -98,4 +98,4 @@ requests.
 
 Of course there's no reason why you need to use middleware in order to solve
 this. The headers can be set directly in the
-[handler](https://fresh.deno.dev/docs/getting-started/custom-handlers) as well.
+[handler](/docs/getting-started/custom-handlers) as well.
--- a/docs/examples/dealing-with-cors.md
+++ b/docs/examples/dealing-with-cors.md
@@ -98,4 +98,4 @@ requests.
 
 Of course there's no reason why you need to use middleware in order to solve
 this. The headers can be set directly in the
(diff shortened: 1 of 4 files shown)


docs/examples/dealing-with-cors.md
```diff
@@ -0,0 +1,101 @@
...
+2. the browser makes the actual request
+
+So you can see the middleware has some special handling to deal with `OPTIONS`
+requests.
+
+## CORS in Routes
+
+Of course there's no reason why you need to use middleware in order to solve
+this. The headers can be set directly in the
+[handler](http://localhost:8000/docs/getting-started/custom-handlers) as well.
```
marvinhagemeister: Nit: a `http://localhost:8000` url sneaked into the doc here.

deer: I looked closely at our urls and realized I was doing it wrong in the "server configuration" document. I also saw that the way we do urls [here](https://raw.githubusercontent.com/denoland/fresh/76d838cf105621a2bd86d3b04259d7bcc1e37941/docs/concepts/deployment.md) is different than in all the other places. Perhaps we should harmonize our url strategy at some point...

marvinhagemeister (approved):
This is great!

PR #1348, opened 2023-06-23
fix: signals not serialized with newer versions

### LINKED ISSUE #1342: the sample code in fresh 1.2 release blog will not work
The sample code below at https://deno.com/blog/fresh-1.2 will not work.

```
// routes/index.tsx
export function Page() {
  const cart = useSignal<string[]>([]);
  return (
    <div>
      <Header cart={cart} />
      <div>
        <h1>Lemon</h1>
        <p>A very fresh fruit.</p>
        <AddToCart cart={cart} id="lemon" />
      </div>
    </div>
  );
}

// islands/Header.tsx
export default function Header(props: { cart: Signal<string[]> }) {
  return (
    <header>
      <span>Fruit Store</span>
      <button>Open cart ({props.cart.value.length})</button>
    </header>
  );
}

// islands/AddToCart.tsx
export default AddToCart(props: { cart: Signal<string[]>, id: string }) {
  function add() {
    props.cart.value = [...props.cart.value, id];
  }
  return <button onClick={add}>Add to cart</button>
}
```

At least, 
```
export default AddToCart(props: { cart: Signal<string[]>, id: string }) {
```
should accompany `function`:

```
export default function AddToCart(props: { cart: Signal<string[]>, id: string }) {
```

After adding it, the code still not work in my environment.
When I press "Add to cart" button, the browser console says "Uncaught TypeError: props.cart.value is not iterable".
It looks like `const cart = useSignal<string[]>([])` is not correctly passed to the AddToCart island, which is initialized on the server, and then (expected to be) passed to the client.

I wrote the following code, which is successful.
```
// routes/cart.tsx
import Cart from "~/islands/Cart/cart.tsx";

export default function Page(props: PageProps) {
  return (
    <div>
      <Cart />
    </div>
  );
}

// islands/Cart/cart.tsx
import { useSignal } from "@preact/signals";
import AddToCart from "~/islands/Cart/add.tsx";
import Header from "~/islands/Cart/header.tsx";

export default function Cart() {
  const cart = useSignal<string[]>([]);

  return (
    <div>
      <Header cart={cart} />
      <div>
        <h1>Lemon</h1>
        <p>A very fresh fruit.</p>
        <AddToCart cart={cart} id="lemon" />
      </div>
    </div>
  );
}

// islands/Cart/header.tsx
import { Signal } from "@preact/signals";

export default function Header(props: { cart: Signal<string[]> }) {
  return (
    <header>
      <span>Fruit Store</span>
      <button>Open cart ({props.cart.value.length})</button>
    </header>
  );
}

// islands/Cart/add.tsx
import { Signal } from "@preact/signals";

export default function AddToCart(props: { cart: Signal<string[]>, id: string }) {
  function add() {
    props.cart.value = [...props.cart.value, props.id];
  }
  return (<button onClick={add}>Add to cart</button>)
}
```

The above code works as I expect. Pressing the "Add to cart" button refreshes the "Open cart (0)" to "(1)".
I assume that I miss something that needs to be done for the original code to run as expected, or, the original code is only intended within a large island (as in my sample code).

Question: What is the expected behavior of useSignal within routes, then passed to an island?

marvinhagemeister:
Newer versions of @preact/signals-core have a `toJSON` method which is called by `JSON.stringify` automatically. This means we miss signals during serialization. This PR adds a check if the value we received was from an unwrapped signal.

Fixes #1342

### DIFF
--- a/src/server/serializer_test.ts
+++ b/src/server/serializer_test.ts
@@ -3,7 +3,9 @@
 import { serialize } from "./serializer.ts";
 import { assert, assertEquals, assertSnapshot } from "../../tests/deps.ts";
 import { deserialize, KEY } from "../runtime/deserializer.ts";
-import { signal } from "@preact/signals";
+import { signal } from "@preact/signals-core";
+import { signal as signal130 } from "@preact/signals-core@1.3.0";
+import { signal as signal123 } from "@preact/signals-core@1.2.3";
 
 Deno.test("serializer - primitives & plain objects", async (t) => {
   const data = {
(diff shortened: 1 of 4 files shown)


src/server/serializer_test.ts
```diff
@@ -58,6 +60,60 @@ Deno.test("serializer - signals", async (t) => {
...
+  assertEquals(typeof deserialized, "object");
+  assertEquals(deserialized.a, 1);
+  assertEquals(deserialized.b.value, 2);
+  assertEquals(deserialized.b.peek(), 2);
+});
+
+Deno.test("serializer - multiple versions of @preact/signals-core", async (t) => {
+  const data = {
+    a: 1,
+    b: signal(2),
```
marvinhagemeister, line 98: Right now this aliases to `1.3.0` too, but I left it here in case we update the version in the future. That way we'll automatically cover that together with `1.2.3` and `1.3.0` too.

lucacasonato (approved):
LGTM

osddeitf:
@marvinhagemeister could you have a look at [preactjs/signals#389](https://github.com/preactjs/signals/issues/389)?
I tried with Fresh v1.3, the signal seems to be deserialized correctly, but it becomes 2 different signals client side.
I also tried with Fresh v1.2, the signal is deserialized incorrectly into primitive value, that is the same as the issue this PR fixed.

PR #1349, opened 2023-06-23
fix: update script supports deno.jsonc files

### LINKED ISSUE #800: Upgrade script doesn't work with deno.jsonc
[The script to upgrade to Fresh v1.1 doesn't support `deno.jsonc` configuration files.](https://github.com/denoland/fresh/blob/b19c4bb0415b4b2741a59e9b33bdfb3bed3519aa/update.ts#L45)

Right now I get:

> error: Uncaught (in promise) NotFound: No such file or directory (os error 2)
> let denoJsonText = await Deno.readTextFile(DENO_JSON_PATH);
>                    ^
>     at async Object.readTextFile (deno:runtime/js/40_read_file.js:56:20)
>     at async https://deno.land/x/fresh@1.1.1/update.ts:46:20

[Deno docs](https://deno.land/manual@v1.25.4/getting_started/configuration_file) say both `deno.json` and `deno.jsonc` are supported, so I think this script should support both.

deer:
closes https://github.com/denoland/fresh/issues/800

I refactored the test a bit because it seemed to be getting a bit repetitive. There's now an `updateAndVerify` function which is called from the three cases:
1. execute update command
2. execute update command deno.jsonc support
3. execute update command (no islands directory)

### DIFF
--- a/src/dev/deps.ts
+++ b/src/dev/deps.ts
@@ -13,6 +13,7 @@ export {
 export { walk, WalkError } from "https://deno.land/std@0.190.0/fs/walk.ts";
 export { parse } from "https://deno.land/std@0.190.0/flags/mod.ts";
 export { gte } from "https://deno.land/std@0.190.0/semver/mod.ts";
+export { existsSync } from "https://deno.land/std@0.190.0/fs/mod.ts";
 
 // ts-morph
 export { Node, Project } from "https://deno.land/x/ts_morph@17.0.1/mod.ts";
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
This is great! Thank you for making a PR to make this better 💯

frou:
It's debatable whether this "supports" `deno.jsonc` because although it can find the file, it still dies if the file actually uses comments? 😛

marvinhagemeister:
@frou can you open an issue for that so that we can track it and make sure we resolve that for the next release?

PR #1352, opened 2023-06-24
fix link

digitaldesigndj:
real simple ❤️


deer:
Great, but there's a second one on the page as well. I saw this in my https://github.com/denoland/fresh/pull/1347/files, but maybe yours can sneak in first, since it's smaller.

marvinhagemeister (approved):
Whopps, good catch! Thanks for making a PR ❤️

PR #1355, opened 2023-06-24
feat: support conditional island JSX children

### LINKED ISSUE #1322: Server-side component rendered based on client-side condition doesn't work

marvinhagemeister:
This PR address the scenario where a server rendered JSX tree is passed to an island but isn't rendered initially in the client. In that case we cannot reconstruct it from the DOM alone and some form of serialization is needed. This PR stores the not rendered children part in a `<template>` tag so that it can be revived in the client.

This PR is based on #1324.

Fixes #1322

### DIFF
--- a/tests/islands_test.ts
+++ b/tests/islands_test.ts
@@ -478,7 +478,7 @@ Deno.test({
 Deno.test({
   name: "render nested islands with server children conditionally",
 
-  async fn(_t) {
+  async fn() {
     await withPageName(
       "./tests/fixture_island_nesting/main.ts",
       async (page, address) => {
@@ -490,15 +490,12 @@ Deno.test({
       },
     );
   },
-
-  sanitizeOps: false,
-  sanitizeResources: false,
 });
 
 Deno.test({
   name: "revive island in lazy server rendered children conditionally",
 
-  async fn(_t) {
+  async fn() {
     await withPageName(
       "./tests/fixture_island_nesting/main.ts",
       async (page, address) => {
@@ -513,7 +510,4 @@ Deno.test({
       },
     );
   },
-
-  sanitizeOps: false,
-  sanitizeResources: false,
 });
--- a/tests/islands_test.ts
+++ b/tests/islands_test.ts
@@ -478,3 +478,46 @@ Deno.test({
   sanitizeOps: false,
   sanitizeResources: false,
 });
+
+Deno.test({
+  name: "render nested islands with server children conditionally",
+
+  async fn(_t) {
+    await withPageName(
+      "./tests/fixture_island_nesting/main.ts",
+      async (page, address) => {
+        await page.goto(`${address}/island_conditional_lazy`);
+        await waitForText(page, ".island p", "island content");
+
+        await page.click("button");
+        await waitForText(page, ".island p", "server rendered");
+      },
+    );
+  },
+
+  sanitizeOps: false,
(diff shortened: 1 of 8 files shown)


deer:
I really appreciate the comments in the new code.

tests/islands_test.ts
```diff
@@ -478,3 +474,46 @@ Deno.test({
   sanitizeOps: false,
   sanitizeResources: false,
 });
+
+Deno.test({
+  name: "render nested islands with server children conditionally",
+
+  async fn(_t) {
```
iuioiua:
```suggestion
  async fn() {
```
Same with other test.

tests/islands_test.ts
```diff
@@ -478,3 +474,46 @@ Deno.test({
...
+        await waitForText(page, ".island p", "island content");
+
+        await page.click("button");
+        await waitForText(page, ".island p", "server rendered");
+      },
+    );
+  },
+
+  sanitizeOps: false,
+  sanitizeResources: false,
```
iuioiua:
```suggestion
```
Tests on my local machine pass when these options are removed. Same with the other test.

tests/islands_test.ts
```diff
@@ -478,3 +474,46 @@ Deno.test({
...
+        await waitForText(page, ".island .server", "server rendered");
+
+        await page.click("button.counter");
+        await waitForText(page, ".island .count", "1");
+      },
+    );
+  },
+
+  sanitizeOps: false,
+  sanitizeResources: false,
```
iuioiua:
```suggestion
```
Ditto

tests/islands_test.ts
```diff
@@ -478,3 +474,46 @@ Deno.test({
...
+});
+
+Deno.test({
+  name: "revive island in lazy server rendered children conditionally",
+
+  async fn(_t) {
```
iuioiua:
```suggestion
  async fn() {
```
Ditto

iuioiua (approved):
LGTM! Just nits.

Issue #1350, opened 2023-06-23
Confusing naming `handler` when types is plural `Handlers


marvinhagemeister:
This is tripping folks up (for good reason). We currently expect the export to be named `handler` but the type name is the same word in plural: `Handlers`. See https://discord.com/channels/684898665143206084/991511118524715139/1121692086886993970

```ts
// Export is singular "handler", type is plural
export const handler: Handlers = {
  // ...
}
```

aushwin:
Hey @marvinhagemeister, I would like to work on this issue. (Please note sometimes would need some guidance as am new into open source contribution)

marvinhagemeister:
The solution to this problem very likely looks like something along the lines of this:

1. Rename `Handlers` type to `MultiHandler`
2. Ensure that we still export a `Handlers` type nonetheless, to ensure we don't break existing apps. Can be done via `export type Handlers<T> = MultiHandler<T>` or something similar
3. To encourage folks to migrate to `MultiHandler`, mark the `Handlers` type as deprecated via the `@deprecated` JSDoc comment.
4. If one feels extra fancy, we could add a check to detect when a `handler` export is missing but a `handlers` export is present here: https://github.com/denoland/fresh/blob/7a7d0e77d0dd1d67371b5efff8d55bd2cf6aa85e/src/server/context.ts#L194 . Add a test case for this by adding a new route to `tests/fixture/`. Everything in that directory is basically a bog standard fresh app, so you might need to run it's `dev.ts` command to refresh the contents of the `fresh.gen.ts` file.

aushwin:
Could you please assign this to me? And how can I setup the project locally, so that I make changes and test it?

aushwin:
@marvinhagemeister please review 
https://github.com/denoland/fresh/pull/1360

lausuarez02:
Hey I see that you are missing the test case that @marvinhagemeister recommended to you, i would be happy to add it :)

aushwin:
@lausuarez02 Hey am working on it. Will ping you if am unable to do it 😄

marvinhagemeister:
FWIW: I've opened a PR for a lint rule to catch the incorrect name directly in the editor https://github.com/denoland/deno_lint/pull/1173

aushwin:
Cool 🔥.

frou:
[...]

PR #1360, opened 2023-06-26
refactor: changed Handlers type to MultiHandler

aushwin:
This PR is for issue #1350 

Hey @marvinhagemeister, I did not get what you meant in the 4th step. Kindly help me out here 
https://github.com/denoland/fresh/issues/1350#issuecomment-1605394508


choucavalier:
@aushwin i think what @marvinhagemeister meant was to check that the user exported a function called `handlers` (with an "s") instead of `handler` (without an "s"). this would help detect confused users who try to create a `handler` but naming the function incorrectly might lead them to think their `handler` is not working, bang their head against their desk for 3 days, and then plunge into depression as they realise they had an extra "s" in their function name. but maybe i misunderstood

marvinhagemeister (commented):
@aushwin  This looks great!! Let's run a quick `deno fmt` and `deno lint` to ensure the changes match the expected formatting. After that it's ready to be merged 👍

marvinhagemeister:
@tgy Yeah exactly. I think the type hints in the editor are good enough and gets us 80% there already. Like you said, I was referring to checking the actual exported values from the modules which might be neat. It's more for bonus points and can happen in another PR too.

aushwin:
@marvinhagemeister @tgy  is it something like this 
`MODULE_EXPORT_HANDLERS` is a constant for the string `handlers`
image

And am getting this error in my terminal.
image

aushwin:
But I guess we need to show this in the error container right ?

aushwin:
Hi @marvinhagemeister Any updates on this ?

marvinhagemeister (approved):
LGTM, thanks for the PR 🎉

PR #853, opened 2022-10-19
(documentation): Document HTTP Methods

gutenfries:
Hi all, 

I added documentation for the HTTP Methods, as mentioned in #768 

I also optimized the static images, via [Imgbot](https://imgbot.net/), which optionally may be cherry-picked past.

Let me know if there is any feedback, concerns, or comments.

Thanks!

<details>
<summary>Imgbot optimization</summary>
*Total -- 1,260.84kb -> 1,021.16kb (19.01%)

/www/static/home-og.png -- 23.32kb -> 10.37kb (55.51%)
/www/static/showcase/optimem2x.jpg -- 177.82kb -> 128.77kb (27.58%)
/www/static/showcase/imasparql-dd2x.jpg -- 52.76kb -> 38.95kb (26.18%)
/tests/fixture/static/image.png -- 16.82kb -> 12.48kb (25.77%)
/www/static/showcase/videopoker-academy2x.jpg -- 69.50kb -> 52.97kb (23.78%)
/www/static/showcase/ycrm2x.jpg -- 197.07kb -> 150.97kb (23.39%)
/www/static/showcase/ycrm1x.jpg -- 80.94kb -> 64.87kb (19.86%)
/www/static/showcase/videopoker-academy1x.jpg -- 24.71kb -> 19.94kb (19.29%)
/www/static/showcase/imasparql-dd1x.jpg -- 21.48kb -> 17.58kb (18.16%)
/www/static/showcase/optimem1x.jpg -- 76.59kb -> 63.64kb (16.91%)
/www/static/showcase/balello2x.jpg -- 147.97kb -> 123.00kb (16.87%)
/www/static/showcase/hn2x.jpg -- 104.61kb -> 89.08kb (14.84%)
/www/static/showcase/balello1x.jpg -- 68.58kb -> 58.85kb (14.19%)
/www/static/showcase/hn1x.jpg -- 41.43kb -> 36.26kb (12.49%)
/www/static/showcase/adam-portfolio2x.jpg -- 90.00kb -> 86.69kb (3.67%)
/www/static/showcase/adam-portfolio1x.jpg -- 36.95kb -> 36.43kb (1.43%)
/www/static/showcase/fresh-strapi1x.jpg -- 30.32kb -> 30.31kb (0.03%)

Signed-off-by: ImgBotApp <ImgBotHelp@gmail.com>
</details>

PR #1361, opened 2023-06-26
docs: add "Creating a CRUD API" example

### LINKED ISSUE #768: Document post requests

deer:
closes https://github.com/denoland/fresh/issues/768
supersedes https://github.com/denoland/fresh/pull/853

I'm not sure if this is what Ryan had in mind when opened 768. But this PR includes some working code, references Deno KV, and briefly touches on `OPTIONS` as a bonus.

Should I link to https://deno.com/blog/build-crud-api-oak-denokv where I borrowed some code from? I must admit I'm a bit confused about why Andy chose to write about Oak instead of Fresh in this post...

### DIFF
--- a/docs/examples/creating-a-crud-api.md
+++ b/docs/examples/creating-a-crud-api.md
@@ -1,0 +1,269 @@
+---
+description: |
+  Use HTTP CRUD methods to perform operations on resources. Learn how to use HTTP handlers to create a RESTful API.
+---
+
+The MDN [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods) are a
+great resource to learn more about HTTP methods. We'll touch on the four
+fundamental methods necessary to create a basic CRUD (create, read, update,
+delete) API here. Additionally, we'll briefly mention CORS requests and how
+`OPTIONS` comes into play.
+
+Using HTTP methods is a common way to create a REST API. Fresh supports common
+HTTP methods in handlers out of the box. Async HTTP requests are also supported.
+Read more about custom handlers [here](/docs/getting-started/custom-handlers).
+
+In this example we'll be creating a small API that uses
+[Deno KV](https://deno.com/kv) to store users in a database.
+
+Our project structure will look like this (in addition to the rest of the Fresh
+code from a new project):
+
+```
+├── routes
+│   └── api
+│       └── users
+│           ├── [id].ts
+│           └── index.ts
+```
+
+In each section about a method, only the relevant handler will be shown. The
+full files are available at the bottom for reference.
+
+## POST
+
+`POST` (create) is used to create a resource.
+
+```tsx
+// /routes/api/index.ts
+async POST(req, _ctx) {
+  const body = await req.text();
+  const user = JSON.parse(body) as User;
+  await upsertUser(user);
+  return new Response(JSON.stringify(user));
+},
+```
+
+Test this with Postman (or your favorite client) with a URL like
+`http://localhost:8000/api/users` and a method of `POST`. Make sure to have a
+payload like:
+
+```json
+{
+  "id": "2",
+  "name": "TestUserName"
+}
+```
+
+You should receive the same thing back:
+
+```json
+{ "id": "2", "name": "TestUserName" }
+```
+
+## GET
+
+`GET` (read) is used to retrieve a resource and is by far the most common HTTP
+method. You can use `GET` to fetch database content, markdown, or static files.
+
+```tsx
+// /routes/api/[id].ts
+async GET(_req, ctx) {
+  const id = ctx.params.id;
+  const user = await getUserById(id);
+  return new Response(JSON.stringify(user));
+},
+```
+
+Let's practice retrieving our user! A `GET` request to
+`http://localhost:8000/api/users/2` should return:
+
+```json
+{ "id": "2", "name": "TestUserName" }
+```
+
+## PUT (and PATCH)
+
+`PUT` (update) and `PATCH` are used to update a resource. While similar, there
+are differences and you should use the one that best suits your use case. Read
+more about HTTP methods on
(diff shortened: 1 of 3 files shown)


docs/examples/creating-a-crud-api.md
```diff
@@ -0,0 +1,269 @@
...
+`POST` (create) is used to create a resource.
+
+```tsx
+// /routes/api/index.ts
+async POST(req, _ctx) {
+  const body = await req.text();
+  const user = JSON.parse(body) as User;
+  await upsertUser(user);
+  return new Response(JSON.stringify(user));
+},
```
marvinhagemeister: 1. Let's display the full handler object here to allow users to copy & paste it. The snippet right now is invalid JS.

2. Let's use the [`request.json()`-method](https://developer.mozilla.org/en-US/docs/Web/API/Request/json) instead of parsing to JSON ourselves.

3. We parse the user as JSON only to serialize it back to a string again when we send it.

4. Let's inline the `upsertUser` function here. That way we see directly what's going on.

docs/examples/creating-a-crud-api.md
```diff
@@ -0,0 +1,269 @@
...
+      .set(userKey, user)
+      .commit();
+    if (!ok) throw new Error("Something went wrong.");
+  } else {
+    const ok = await kv.atomic()
+      .check(oldUser)
+      .set(userKey, user)
+      .commit();
+    if (!ok) throw new Error("Something went wrong.");
+  }
```
marvinhagemeister: Both branches are identical if I'm not mistaken.

docs/examples/creating-a-crud-api.md
```diff
@@ -0,0 +1,269 @@
...
+```json
+{ "id": "2", "name": "TestUserName" }
+```
+
+## PUT (and PATCH)
+
+`PUT` (update) and `PATCH` are used to update a resource. While similar, there
+are differences and you should use the one that best suits your use case. Read
+more about HTTP methods on
+[MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods).
```
marvinhagemeister, line 98: Let's add a sentence here explaining the difference. I had to look it up myself, because I've only worked on GraphQL-based projects in the past few years where only GET + POST is used.

docs/examples/creating-a-crud-api.md
```diff
@@ -0,0 +1,269 @@
...
+async PUT(req, ctx) {
+  const id = ctx.params.id;
+  const body = await req.text();
+  const userBody = JSON.parse(body) as User;
+  updateUserById(id, userBody);
+  return new Response(JSON.stringify(userBody));
+},
+```
+
+Time to change their name. We'll now `PUT` a request to
```
marvinhagemeister, line 124: Might be me not being a native speaker, but I got confused for a second to what "their" is referring to here.

```suggestion
Time to change a user's name. We'll now `PUT` a request to
```

marvinhagemeister (commented):
This is great! I really like the idea of having a page dedicated for CRUD stuff in our docs. 

Left a couple of minor comments. I think it's best if we try to make code snippets copy & paste-able if possible and ensure the syntax is valid JavaScript. Let me know what you think.

docs/examples/creating-a-crud-api.md
deer: I'll push back 💪  here as a native speaker. We're still talking about user 2, but we don't know whether user 2 is male or female (or if they identify differently, and it's entirely irrelevant as well), so we can just use "their". See https://en.wikipedia.org/wiki/Singular_they for details. (I was just talking about this on Sunday with a friend from Berlin, so you're not alone!)

If you still find it confusing, I would go for something like `Time to change their name from "TestUserName" to something else` or `Time to change user 2's name`.

docs/examples/creating-a-crud-api.md
deer: Sure, I've added a sentence and will resolve this when I push the new commit.

docs/examples/creating-a-crud-api.md
deer: 1. I'll assume you mean the trimmed version just concerning the `POST` method, since the full object might have other methods as well. (In my test project, for example, I have `GET` implemented at the index as well, in order to get all users.) So something like this:
```ts
export const handler: Handlers<User | null> = {
  async POST(req, _ctx) {
    const user = await req.json() as User;
    await upsertUser(user);
    return new Response(JSON.stringify(user));
  },
};
```
2. Yes, that's what I wanted! Thanks.
3. I thought we were doing it to store the user. How else should I respond here?
4. Ok, and I'll try to clean this function up a bit.

docs/examples/creating-a-crud-api.md
deer: 🧐  good catch. This is a result of not fully simplifying Andy's code, now that I'm not worried about email. Thanks.

deer:
I've boldly resolved all the conversations, since I think I responded to all of your concerns. The code is definitely better now. Thanks!

deer:
Looks like my change about updating when using a source directory has introduced a flaky test. This is the second time I've seen something like the following:
```
error: NotFound: No such file or directory (os error 2): rename '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/e8b1d7f8/src/fresh.gen.ts' -> '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/e8b1d7f8/fresh.gen.ts'
```

docs/examples/creating-a-crud-api.md
marvinhagemeister: Right that makes sense. Thanks for the explanation 👍

marvinhagemeister (approved):
This is great! With the improved snippets it's much more clear what's going on. Thanks again for another fantastic addition to the docs! Much appreciated 🙌

PR #1362, opened 2023-06-27
fix: Use custom 500 page for *all* handlers if there is one present

### LINKED ISSUE #1357: _500.tsx not rendering if error is thrown inside GET handler
Having set `routes/_500.tsx` like this:

```ts
import { ErrorPageProps } from "$fresh/server.ts";

export default function Error500Page({ error }: ErrorPageProps) {
  return <p>500 internal error: {(error as Error).message}</p>;
}
```

It's not rendering when the error is thrown inside the GET handler.

```ts
import { Handlers } from "$fresh/server.ts";

export const handler: Handlers<undefined, undefined> = {
  GET(req, ctx) {
    if ([].length === 0) throw new Error("This is a test error");

    return ctx.render();
  },
};

  return <p>All right</p>;
}
```

image

Only when it's thrown inside the component itself.

```ts
  if ([].length === 0) {
    throw new Error("This is a test error");
  }
  return <p>All right</p>;
}
```
image

I'm using fresh `1.2.0`.

kamilogorek:
That was a really nasty one to track down 👀 This issue is _not_ possible to reproduce if your method handlers are async, as they will correctly produce a promise wrapper around themselves, catching and effectively rejecting thrown errors.

The same will happen if you use middlewares (cc @deer - https://github.com/denoland/fresh/issues/1357#issuecomment-1606246219), as the `composeMiddlewares` method will call any attached middlewares first, producing a valid promise chain.

Closes https://github.com/denoland/fresh/issues/1357


deer:
Thanks @kamilogorek, I'm glad someone could get this sorted out. I guess I still have some work to do on my end, since I clearly don't understand promise chains, async/await, and how catch plays with the previous two.

kamilogorek:
I believe it's easier to describe it using some small snippets:

```js
// This will not be captured
function foo () {
    throw new Error('wat');
};

Promise.resolve(foo()).catch((e) => console.log('where are you!?'));
```

```js
// This will be captured, as function producing result is already inside the promise chain
async function bar() {
    return 'next please';
}

function foo () {
    throw new Error('wat');
};

Promise.resolve(bar()).then(() => foo()).catch((e) => console.log('where are you!?'));

// same goes for this one, despite function producing value for `Promise.resolve` throws
Promise.resolve(bar()).then(() => Promise.resolve(foo())).catch((e) => console.log('where are you!?'));
```

Basically `Promise.resolve` and `Promise.reject` expect a _value_ to be passed to them. And function throwing an error will not produce one, blowing everything up 😅

marvinhagemeister (approved):
Some fantastic debugging skills here! I can only imagine how long it took to track this one down. Thank you so much for opening a PR 💯

marvinhagemeister:
Yup like @kamilogorek described: The error is thrown before the whole Promise stuff becomes active. That's why the `.catch()` has no effect.

PR #1365, opened 2023-06-27
docs: add "Handling complex routes" example

deer:
In my quest for understanding, I wanted to understand how `RouteConfig` works. I noticed there wasn't an example about using `routeOverride`, so I decided to document what I learned. I've included two sample routes in increasing order of complexity. Let me know if I'm missing something or if there are any changes. Nothing suggestion is too nitpicky!

### DIFF
--- a/docs/examples/handling-complex-routes.md
+++ b/docs/examples/handling-complex-routes.md
@@ -36,15 +36,16 @@ Now if we hit the server with a request like
 show the following:
 
 ```
-module: 'bestModule'
-version: '1.33.7'
-path: 'qqqqqqq'
+{
+  module: "bestModule",
+  version: "1.33.7",
+  path: "asdf"
+}
 ```
 
 ## Complex Route Config
 
-Let's look at something a bit more complex. Issue 587 brings up the following
-example:
+Let's look at something a bit more complex:
 
 ```ts
 import { HandlerContext, RouteConfig } from "$fresh/server.ts";
--- a/docs/examples/handling-complex-routes.md
+++ b/docs/examples/handling-complex-routes.md
@@ -1,0 +1,86 @@
+---
+description: |
+  Sometimes URL based routing isn't enough.
+---
+
+The page on [routing](/concepts/routing) hints at complex routing based on URL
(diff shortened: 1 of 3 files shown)


docs/examples/handling-complex-routes.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+Now if we hit the server with a request like
+`http://localhost:8000/x/bestModule@1.33.7/asdf`, then logging the params will
+show the following:
+
+```
+module: 'bestModule'
+version: '1.33.7'
+path: 'qqqqqqq'
```
marvinhagemeister: Should this be "x" instead of "qqqqqqq"?

docs/examples/handling-complex-routes.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## Complex Route Config
+
+Let's look at something a bit more complex. Issue 587 brings up the following
+example:
```
marvinhagemeister: Let's remove the "issue" part.

marvinhagemeister (commented):
This great! Left to very minor nits. Let me know what you think 👍

docs/examples/handling-complex-routes.md
deer: No, but it should be asdf, as per the request I have in the docs: `http://localhost:8000/x/bestModule@1.33.7/asdf`
Good catch, thanks.

docs/examples/handling-complex-routes.md
deer: Eagle eyed Marvin catches all references to issues 😅  Never again, don't worry!

deer:
It strikes again:
```
error: PermissionDenied: Access is denied. (os error 5): rename 'C:\Users\RUNNER~1\AppData\Local\Temp\7a5dd5d2\src\fresh.gen.ts' -> 'C:\Users\RUNNER~1\AppData\Local\Temp\7a5dd5d2\fresh.gen.ts'
```

I'll create an issue for this.

marvinhagemeister (approved):

marvinhagemeister:
Another great PR 🎉

Issue #1367, opened 2023-06-27
Refactor: Refactor code from if/else statement


lausuarez02:
### Description
I've found in a raw.ts file a if/else statement that could be probably be changed by a switch statement if that's the case i would be put a pr to this specific issue to be considered.
<img width="664" alt="Screenshot 2023-06-27 at 21 21 38" src="https://github.com/denoland/fresh/assets/67484166/bc9b762c-839e-4f37-b0e3-58ad5b1fcc4d">

PR #1368, opened 2023-06-27
Add: Refactor code on raw file to improve readability

lausuarez02:
### Description
As i already describe on the issue i opened this is for code readability and to improve the quality of the code

Issue: #1367


marvinhagemeister:
Thanks for opening a PR! Always great to see new contributors. I'm a little confused as to how the proposed changes would make the code better. What does the `switch`-statement with the additional `true` check improve over the `if`-statement?

lausuarez02:
Hey @marvinhagemeister,
- Readibility: I found that making a if else statement make the code a little less readable
- Scalabilty: this is not the main option but in case we want to add more cases it would be easier to scale it.
This is basically it, i hope you find this helpful. 
Thanks for the reply.

marvinhagemeister:
> - Readibility: I found that making a if else statement make the code a little less readable
> - Scalabilty: this is not the main option but in case we want to add more cases it would be easier to scale it.

I feel like this is less a fact but rather highly subjective and comes down to personal preference. The switch statement adds an unnecessary condition here and has the potential issue of the fallthrough semantics when forgetting a `break` statement.

I'm with you that there is lots of repeated calls in both approaches. What do you think of using a table instead? Something like:

```js
import { extname } from "$std/path/mod.ts";

const contentTypes = new Map([
  [".html", "text/plain"],
  [".ts", "application/typescript"],
  [".js", "application/javascript"],
  [".tsx", "text/tsx"],
  [".jsx", "text/jsx"],
  [".json", "application/json"],
  [".wasm", "application/wasm"],
]);

const value = contentTypes.get(extname(path)) ?? "text/plain";
response.headers.set("Content-Type", value);
```

lausuarez02:
Well that indeed looks like a better solution, it improves the code readability and scalability. 
I found this to be a good alternative.
If you wish i can add a new commit for this new alternative :).
Thanks so much for the discussion!

marvinhagemeister:
Go for it. Feel free to add this to this PR.

lausuarez02:
I've add a the isHtml statement that i found that is needed for my understanding. If it is not let me know and would be happy to remove isHtml variable and accept variable. 
Thanks.

www/routes/raw.ts
```diff
@@ -37,23 +38,20 @@ export const handler: Handlers = {
...
+    const contentTypes = new Map([
+      [".html", "text/plain"],
+      [".ts", "application/typescript"],
+      [".js", "application/javascript"],
+      [".tsx", "text/tsx"],
+      [".jsx", "text/jsx"],
+      [".json", "application/json"],
+      [".wasm", "application/wasm"],
+    ]);
+    if(isHTML){
```
marvinhagemeister: This doesn't look correct. We should always check if we have a content type value for a given extension like we did before, not only when we're dealing with HTML files.

www/routes/raw.ts
```diff
@@ -37,23 +38,20 @@ export const handler: Handlers = {
...
-      response.headers.set("Content-Type", "text/tsx");
-    } else if (path.endsWith(".jsx")) {
-      response.headers.set("Content-Type", "text/jsx");
-    } else if (path.endsWith(".json")) {
-      response.headers.set("Content-Type", "application/json");
-    } else if (path.endsWith(".wasm")) {
-      response.headers.set("Content-Type", "application/wasm");
-    } else {
-      response.headers.set("Content-Type", "text/plain");
+    const contentTypes = new Map([
```
marvinhagemeister: Nit: This `Map` doesn't need to be recreated on every request. We can move it outside the handler.

lausuarez02:
I've reviewed your comments and fix them :)

www/routes/raw.ts
```diff
@@ -2,13 +2,22 @@ import { RouteConfig } from "$fresh/server.ts";
...
+  [".tsx", "text/tsx"],
+  [".jsx", "text/jsx"],
+  [".json", "application/json"],
+  [".wasm", "application/wasm"],
+]);
+
 export const handler: Handlers = {
-  async GET(req, ctx) {
-    const accept = req.headers.get("Accept");
-    const isHTML = accept?.includes("text/html");
```
marvinhagemeister: Let's keep this bit here. It's used to install fresh. There can be multiple values in the `Accept` header which I think the code was trying to account for before.

lausuarez02: Done!

www/routes/raw.ts
```diff
@@ -37,23 +48,9 @@ export const handler: Handlers = {
...
-    } else if (path.endsWith(".jsx")) {
-      response.headers.set("Content-Type", "text/jsx");
-    } else if (path.endsWith(".json")) {
-      response.headers.set("Content-Type", "application/json");
-    } else if (path.endsWith(".wasm")) {
-      response.headers.set("Content-Type", "application/wasm");
-    } else {
-      response.headers.set("Content-Type", "text/plain");
-    }
+      const value = contentTypes.get(extname(path)) ?? "text/plain";
```
marvinhagemeister: Very close! Now if `isHTML` is `true` we can bypass the `extname(path)` and directly assume a value of `.html`.

lausuarez02: Do you mean something like this? 
let value = isHTML ? "text/html" : contentTypes.get(extname(path));

marvinhagemeister: Close, if we read the existing code carefully, it returns `text/plain` for html files, not `text/html`.

lausuarez02:
I could just add a ! at the end
const value = isHTML ? "text/plain" : contentTypes.get(extname(path))!
but since we want a default value in case of nullish or undefined i've add the ?? "text/plain"
const value = isHTML ? "text/plain" : contentTypes.get(extname(path)) ?? "text/plain";

marvinhagemeister (approved):
LGTM

lausuarez02:
I've to remove a white space while formatting

Issue #1369, opened 2023-06-27
Refactor: Refactor ts in utils


lausuarez02:
### Description
Refactor of typescript in the utils file where i've found that we can add more types to be precise about what we are waiting for the function to return

marvinhagemeister:
There is no need to open up issues for cosmetic PRs. Adding a explicit TS return types instead of letting TS infer them automatically doesn't change behaviour of the code in any observable way.

marvinhagemeister:
Closing this issue, because the mentioned changes have no visible effect for users or folks working on fresh.

PR #1370, opened 2023-06-27
Add: Add a string type to know that our function will return a string

lausuarez02:
### Description
Add a type for the function to know that it will return a string

Issue: #1369


marvinhagemeister (commented):
Looks like the code is not formatted correctly. A quick `deno fmt` should resolve that.

deer:
Do we actually need this? It seems like typescript can always infer that this returns a string, so I'm not sure what this does. That being said, I'm still new to typescript, so maybe I'm missing something.

If anything, it seems like we should _remove_ the static typing from the 2nd method in this file, `assetSrcSet`. It's currently declared like this:
```ts
export function assetSrcSet(srcset: string): string {
```
and returns like this:
```ts
  return constructed.join(",");
```
so we know (and so does typescript) it always returns a string.

marvinhagemeister:
@deer It's more of a cosmetic change, since like you said TS can infer those types. It would help in cases where changes suddenly change the inferred return type though. But that said in all the years working with TS my experience is that this is more of a theoretical risk that I haven't encountered in practice. But I don't mind if somebody wants to add these return types to ramp up on the code base. So I'm happy to merge such PRs when CI passes.

lausuarez02:
It's just like @marvinhagemeister said. It's more about a theoretical risk rather than an actual risk and also for me is to continue with the typing just like i've seen in other functions just like the next one which has a :string as a return. 
Thank you both!

marvinhagemeister (approved):

PR #1376, opened 2023-06-30
docs: add "Using CSP" example

### LINKED ISSUE #981: docs: write docs of 'csp' config and 'useCSP' helper

deer:
closes https://github.com/denoland/fresh/issues/981

This took a while, but was good to learn about how CSP works (and, if I'm honest, what it is). There aren't any tests for this stuff. Maybe that can be a followup issue if we're happy with the direction here.


marvinhagemeister (approved):
This is great! 💯

PR #1320, opened 2023-06-17
feat: improve starter template design

marvinhagemeister:
Made it so that the same styles apply for the twind preset and the non-twind one. When twind is not selected we'll create a plain old style sheet directly and insert it into the head via a generated `_app.tsx`.

Before:

Screenshot 2023-06-17 at 19 03 02

After:

Screenshot 2023-06-17 at 18 59 34

Fixes #1308 .

PR #1380, opened 2023-07-01
feat(starter): always initialize _app.tsx

deer:
https://github.com/denoland/fresh/pull/1320 introduced `_app.tsx` when not using twind. This PR makes it so that `_app.tsx` is always initialized. There was recently a [discussion](https://discord.com/channels/684898665143206084/991511118524715139/1124360575074455604) in discord about this, and there's some confusion about layouts/templates. I thought I would take a look, and was surprised to see that we _do_ initialize `_app.tsx`, but not in all cases.

When using twind, we now produce a nice boring `_app.tsx` for people like this:
```tsx

  return (
    <>
      <Component />
    </>
  );
}
```

Additionally, I noticed that there was a blank line at the top of `_app.tsx` when not using twind; I removed that line here.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -401,8 +401,18 @@ html {
 }
 `;
 
-const NO_TWIND_APP_WRAPPER = `
-import { AppProps } from "$fresh/server.ts";
+const APP_WRAPPER = useTwind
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Good idea!

PR #1381, opened 2023-07-01
feat(docs): convert toc to .ts

deer:
To make conflicts easier to deal with, we should have a blank line at the end to essentially have a trailing comma. Now we can add a new entry _before_ the blank row without touching any other rows. Borrowed the idea from the showcase.


deer:
or should this be `feat(www)`? 🤔

lino-levan:
What worked for [denobyexample](https://github.com/denoland/denobyexample/blob/main/toc.ts) is just making the file a typescript file and exporting an object. JS allows trailing commas. I feel like that would be a cleaner solution.

deer:
> What worked for [denobyexample](https://github.com/denoland/denobyexample/blob/main/toc.ts) is just making the file a typescript file and exporting an object. JS allows trailing commas. I feel like that would be a cleaner solution.

Good idea, that does sound better. Will update when I get a chance.

marvinhagemeister:
+1 on turning it into `.ts` file. Makes typing easier too.

docs/toc.ts
```diff
@@ -1,6 +1,8 @@
-{
+import { RawTableOfContents } from "../www/data/docs.ts";
```
marvinhagemeister: Nit: Let's move these types into this file to avoid circular imports.

marvinhagemeister (approved):
Great work! I like that this makes working with the TOC a little simpler 👍

PR #1383, opened 2023-07-01
fix(showcase): reorder to have blank entry at the bottom

deer:
Looks like the battleship entry (which is very cool!) displaced the special blank entry at the bottom. This reorders things correctly. I also snuck in a very small change to the blank entry to improve the grammar:
`... for prevent conflicting by ...` -> `... for preventing conflicts by ...`

### DIFF
--- a/www/data/showcase.json
+++ b/www/data/showcase.json
@@ -342,15 +342,15 @@
     "image": "uspb"
   },
   {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good catch! Thanks for fixing this 💯

Issue #1339, opened 2023-06-21
`useContext` inside a handler.


osvald0:
Is it posible to use `useContext` inside a handler?

marvinhagemeister:
Which kind of handler do you mean? Can you share more information about where you want to use the `useContext` hook?

osvald0:
Yes, sure. sorry for the missing information:

Something like this:

```
export const handler: Handlers = {
  async POST(req, ctx) {
    const { login } = useAuth();
    const mycontext = useContext(MyContext); // Context inside handler here.

    const form = await req.formData();
    const email = form.get("email")?.toString() || "";
    const password = form.get("password")?.toString() || "";

    const response = await login({ email, password });

    if (response) {
      // Do something with mycontext here!

      const url = new URL(req.url);
      url.pathname = "/";
      return Response.redirect(url);
    }
    return await ctx.render(response);
  },
};

export default function Login({ data }: PageProps<AuthResponse>) {
  return (
    <MyFormHere/>
  )
}
```

marvinhagemeister:
I see, thanks for providing further information. It looks like you want to be able to access from a request middleware. Where would you expect the context to be set?

The current flow is:

1. Browser sends request
2. Request middlewares are executed (if any)
3. If it was a `GET` request -> render route.

osvald0:
My idea (I'm not sure if it's ok or not) is to assign to the context the user information from the success login in the handler.

deer:
But isn't that what https://fresh.deno.dev/docs/concepts/data-fetching is all about? Do something asynchronous in the handler, and then (optionally) pass that information to the render function.

If you want to receive information from a previous middleware, then look into `State` which is briefly discussed here https://fresh.deno.dev/docs/concepts/middleware. In particular, each middleware has a ctx, and `ctx.state` can be mutated. The mutated value gets passed along the chain of middleware handlers.

If you really need some heavy duty initialization (the kind of thing that should only get done once, during server startup) then look at https://fresh.deno.dev/docs/examples/init-the-server.

Do these help to clarify? If not, please provide more details.

CAYdenberg:
@osvald0 When you say `useContext` are you asking about the (p)react hook?

osvald0:
@CAYdenberg Yes.

@deer I finally replaced the context by a signal variable without the context. I know how the handler works, the idea behind the context was to to put some information available there (context) without need to fetch again the same thing later in other handler.

CAYdenberg:
[...]

PR #1385, opened 2023-07-01
feat(server): useProps context

### LINKED ISSUE #755: Feature Request: Add in a route match object to the runtime.ts files.
Other solutions offer a way to programmatically get the current matched route, such as something like `useRoute`. Would be nice to have something similar for Fresh, without needing to pass it down through page props.

deer:
closes https://github.com/denoland/fresh/issues/755

This change allows things like this:
```tsx
export default function Grandchild() {
  const props = useProps();
  if (props.data === "computing this is really hard") {
    return <div>{props.url.toString()}</div>;
  }
  return <div></div>;
}
```
So components can get access to the props from the root component by using the newly introduced `useProps` context.

This is a generalization of the PR that was originally opened to close this issue. I thought it might be convenient to have not just the route available, but the entire props. I'm opening this as a draft since I didn't write any documentation yet, and I'm concerned about performance implications.

@marvinhagemeister, would you be so kind to clarify what sort of performance problems this will introduce? I see https://stackoverflow.com/questions/75060633/react-context-performance-and-suggestions and they talk about when context changes, the components re-render. Since we're just rendering this all on the server and the context is only created once and never changes, maybe it doesn't matter.

Mildly related to https://github.com/denoland/fresh/issues/1339.


marvinhagemeister:
@deer Performance of context providers is not a concern on the server. It's mostly just a property access which costs nothing. It matters much more on the client where things are stateful and context updates can cause an app to re-render.

deer:
Thanks for the clarification Marvin. I've tried my hand at writing some documentation for this and have therefore marked it as ready.

deer:
Another FYI: we'll need to update the ToC to include this new example when this feature gets released in the next version. Currently the doc is just in the canary ToC. (Assuming this gets approved!)

PR #1390, opened 2023-07-02
docs: add "Server Components" concept page

### LINKED ISSUE #1384: docs: better explain island architecture and server components
Seems like our docs don't communicate that aspect good enough. Got a tweet where somebody didn't know that fresh had this: https://twitter.com/CodeMonument/status/1675084963295764486 

Here is the conversation:

> **kasvith:** Waiting for PSC, Preact Server Components
> **marvinhagemeist:** A form of that is already available in fresh  [https://fresh.deno.dev](https://t.co/IxuBinVhhQ). The Deno sites have been running this in production for a good while.
> **CodeMonument**: Can you link the exact docs for these „preact server components“ like things in fresh? I think I know nearly all of the docs, but I know nothing in there which resembles RSCs to me :O
> **marvinhagemeist**: We don't mention the term explicitly in the docs. Basically every route is treated internally as a server component and islands are client components. When you pass server side rendered children to an Island, then these children are treated internally like a server component.

deer:
closes https://github.com/denoland/fresh/issues/1384

Here's a first pass at making this more clear. I'm sure many improvements can be made to the wording here, but I always find it easier to improve something someone else has done than starting from scratch, so I will happily put this up as a starting point.

At https://github.com/preactjs/preact/discussions/2879 the people seem to be asking for some clarification of this as well.


marvinhagemeister (approved):
LGTM. Thanks for adding this 🙌

marvinhagemeister (approved):

PR #1393, opened 2023-07-02
fix(server): properly sanitize island names

### LINKED ISSUE #1391: [bug] Islands cannot be nested more than 1 folder deep
I've recently upgraded my website to Fresh v1.2, made a bunch of folders and subfolders for my islands and was confused by why my forms were submitting all of a sudden.
image

Turns out, Fresh can't properly render those islands if they are nested more than 1 folder deep.
image

If I move the `HighlightedCode.tsx` island inside another subfolder, the forms will start submitting instead of having their default prevented and the browser console will display the error from the first screenshot.

Let me know if you want to guide me on how to fix this as I would love to contribute fixing something that I'm going to be basically using daily.

deer:
closes https://github.com/denoland/fresh/issues/1391

This was interesting to debug. I was able to reproduce the case and saw the following in the `sources` tab:
```js
const ST = document.getElementById("__FRSH_STATE").textContent;const STATE = JSON.parse(ST).v;import { revive } from "/_frsh/js/546c8cc610e5f6749bbb353f6c3518c35ac0ff22/main.js";import Firstsecond/DoubleNested from "/_frsh/js/546c8cc610e5f6749bbb353f6c3518c35ac0ff22/island-firstsecond/doublenested.js";revive({firstsecond/doublenested:Firstsecond/DoubleNested,}, STATE[0]);
```

That's not good, variables shouldn't have slashes in them!

After a bit of hunting, eventually found this block:
```ts
const path = url.substring(baseUrl.length).substring("islands".length);
const baseRoute = path.substring(1, path.length - extname(path).length);
const name = sanitizeIslandName(baseRoute);
const id = name.toLowerCase();
if (typeof module.default !== "function") {
  throw new TypeError(
    `Islands must default export a component ('${self}').`,
  );
}
islands.push({ id, name, url, component: module.default });
```
which is making the actual islands.

Turns out we're not sanitizing correctly.
```ts
function sanitizeIslandName(name: string): string {
  const fileName = name.replace("/", "");
  return toPascalCase(fileName);
}
```
This is only going to strip the first slash. After 1185 was merged, we might have many slashes. I applied a fix, and then I got the following:
```js
const ST = document.getElementById("__FRSH_STATE").textContent;const STATE = JSON.parse(ST).v;import { revive } from "/_frsh/js/2e0bfa95a39161dcd4b0c6dd1bb7d0c02fadf6a5/main.js";import First_second_DoubleNested from "/_frsh/js/2e0bfa95a39161dcd4b0c6dd1bb7d0c02fadf6a5/island-first_second_doublenested.js";revive({first_second_doublenested:First_second_DoubleNested,}, STATE[0]);
```
This looks better and makes the errors go away in the browser's console. Adding a test case and reverting my fix causes the test to fail. Reapplying the fix causes the new test to pass. 🙌

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -931,7 +931,7 @@ function toPascalCase(text: string): string {
 }
 
 function sanitizeIslandName(name: string): string {
-  const fileName = name.replace("/", "");
+  const fileName = name.replace(/\//g, "_");
   return toPascalCase(fileName);
 }
 
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -931,7 +931,7 @@ function toPascalCase(text: string): string {
 }
 
 function sanitizeIslandName(name: string): string {
-  const fileName = name.replace(/\//g, "_");
+  const fileName = name.replaceAll("/", "_");
   return toPascalCase(fileName);
 }
 
--- a/tests/fixture/fresh.gen.ts
+++ b/tests/fixture/fresh.gen.ts
@@ -43,7 +43,8 @@ import * as $$2 from "./islands/RootFragment.tsx";
 import * as $$3 from "./islands/RootFragmentWithConditionalFirst.tsx";
 import * as $$4 from "./islands/Test.tsx";
 import * as $$5 from "./islands/folder/Counter.tsx";
-import * as $$6 from "./islands/kebab-case-counter-test.tsx";
+import * as $$6 from "./islands/folder/subfolder/Counter.tsx";
+import * as $$7 from "./islands/kebab-case-counter-test.tsx";
 
 const manifest = {
   routes: {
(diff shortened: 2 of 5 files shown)


src/server/context.ts
```diff
@@ -931,7 +931,7 @@ function toPascalCase(text: string): string {
 }
 
 function sanitizeIslandName(name: string): string {
-  const fileName = name.replace("/", "");
+  const fileName = name.replace(/\//g, "_");
```
lino-levan: I'm personally against using regex when not necessary. I feel like is a lot easier to read.
```suggestion
  const fileName = name.replaceAll("/", "_");
```

marvinhagemeister: I'm neutral on this one. Either way works.

marvinhagemeister (approved):
Great catch! That's quite a nasty one to debug. Love the fix and addition to the tests. Thank you for another great PR 💯

PR #1396, opened 2023-07-03
fix: change MultiHandler back to Handlers

### LINKED ISSUE #1350: Confusing naming `handler` when types is plural `Handlers

marvinhagemeister:
This PR reverts #1360 . The deprecation of the `Handlers` type caused a bit of churn in our docs. After a bit of back and forth internally we felt that keeping it `Handlers` and throwing an error instead, when a `handlers` export detected is a better approach.

An additional benefit is that the object notation is more widely used and it makes sense to make that as ergonomic as possible.

Fixes #1350


iuioiua:
What about using a `handlers` variable name instead of `handler`? That makes the most sense to me.

marvinhagemeister:
@iuioiua that in itself would be a breaking change if we just changed that. We could allow both, but that's also weird.

iuioiua:
True. How much would it complicate things if we supported both during a deprecation period, _if_ we were to do so? To elaborate, I'd assume that having a `handlers` variable name match its type name `Handlers` would be most apparent to a newcomer.

iuioiua (approved):
Revert LGTM! I think we should consider moving to a `handlers` variables name in another issue or something.

PR #1397, opened 2023-07-03
feat(server): multiple island exports

### LINKED ISSUE #1394: suggestion: allow non-default exports and multiple exports from islands
I think it would be really nice if non-default exports from islands were islands as well (instead of the components they are treated as now). This would also allow fresh component libraries to do something like this:

```ts
// Mid component library example
// ./islands/mid.tsx

export * from "https://deno.land/x/mid/mod.ts";
```

Consumers could then do something like

```tsx
import { Button } from "@/islands/mid.tsx";

  return (
    <>
      <Head>
        <title>Demo</title>
      </Head>
      <Button>Hey!</Button>
    </>
  );
}
```

deer:
closes https://github.com/denoland/fresh/issues/1394

I'm not sure if there's a good reason we had this code:
```ts
throw new TypeError(
  `Islands must default export a component ('${self}').`,
);
```
other than the fact that we just didn't get around to improving it. If so, then consider this the improvement 😄 

This stuff was all a bit confusing at first, but I suppose it makes sense now. The basic approach is to start keeping track of the `exportName` for the island, and looping over all the exports of the module. Then we need to pass the `exportName` out to the client as well.

This change has the breaking effect of changing the client code we emit. This is visible in the diff to `main_test` where the asserted name goes from `test.js` to `test_default.js`. But this isn't part of the API, so people shouldn't be doing anything with this. If they are, they'll have to switch 😅 

Here's an example of some islands (from the test case):
```ts
{
  id: "multiplecounters_default",
  name: "MultipleCounters",
  url: "file:///Users/reed/code/fresh/tests/fixture/islands/MultipleCounters.tsx",
  component: [Function: CounterZero],
  exportName: "default"
}
{
  id: "multiplecounters_counterone",
  name: "MultipleCounters",
  url: "file:///Users/reed/code/fresh/tests/fixture/islands/MultipleCounters.tsx",
  component: [Function: CounterOne],
  exportName: "CounterOne"
}
{
  id: "multiplecounters_countertwo",
  name: "MultipleCounters",
  url: "file:///Users/reed/code/fresh/tests/fixture/islands/MultipleCounters.tsx",
  component: [Function: CounterTwo],
  exportName: "CounterTwo"
}
{
  id: "folder_subfolder_counter_default",
  name: "Folder_subfolder_Counter",
  url: "file:///Users/reed/code/fresh/tests/fixture/islands/folder/subfolder/Counter.tsx",
  component: [Function: Counter],
  exportName: "default"
}
```

With the above islands, we now emit html (what's the right way to say this?) like this:
```html
<!--frsh-multiplecounters_default:default:0-->
<div id="counter0">
    <p>4</p>
    <button id="b-counter0" disabled>+1</button>
</div>
<!--/frsh-multiplecounters_default:default:0-->
<!--frsh-multiplecounters_counterone:CounterOne:1-->
<div id="counter1">
    <p>3</p>
    <button id="b-counter1" disabled>+1</button>
</div>
<!--/frsh-multiplecounters_counterone:CounterOne:1-->
<!--frsh-multiplecounters_countertwo:CounterTwo:2-->
<div id="counter2">
    <p>10</p>
    <button id="b-counter2" disabled>+1</button>
</div>
<!--/frsh-multiplecounters_countertwo:CounterTwo:2-->
<!--frsh-folder_subfolder_counter_default:default:3-->
<div id="counter3">
    <p>4</p>
    <button id="b-counter3" disabled>+1</button>
</div>
<!--/frsh-folder_subfolder_counter_default:default:3-->
```
Note this is coming from the change to the comment generation:
```ts
`frsh-${island.id}:${island.exportName}:${ISLAND_PROPS.length - 1}`
```
We now pass out the `exportName` to enable the lookup within the module. It looks a bit silly for default exports, because the `id` also has `default` in it as well. Open to suggestions here. I went around in circles trying to simplify this, but eventually just left it.

Finally we get a script like this:
```js
import * as MultipleCounters_default from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_default.js";import * as MultipleCounters_CounterOne from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_counterone.js";import * as MultipleCounters_CounterTwo from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_countertwo.js";import * as Folder_subfolder_Counter_default from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-folder_subfolder_counter_default.js";revive({multiplecounters_default:MultipleCounters_default,multiplecounters_counterone:MultipleCounters_CounterOne,multiplecounters_countertwo:MultipleCounters_CounterTwo,folder_subfolder_counter_default:Folder_subfolder_Counter_default,}, STATE[0]);
```
(Oh, you don't find that easy to read? Me too. Please help. When testing this I manually went and inserted `\n` to the necessary places in `render.ts` so the client code looks nice. Seems like a good use case for the dev tools issue, where Bartek writes: `Currently Fresh's server doesn't know if it's running in dev mode or not, the dev script is a thin wrapper that generates the manifest and watches for file changes.`)

So the formatted version is like this:
```js
import * as MultipleCounters_default from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_default.js";
import * as MultipleCounters_CounterOne from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_counterone.js";
import * as MultipleCounters_CounterTwo from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-multiplecounters_countertwo.js";
import * as Folder_subfolder_Counter_default from "/_frsh/js/48d9737696c04a606a9f9aaddd5b8b17411852dd/island-folder_subfolder_counter_default.js";
revive({multiplecounters_default:MultipleCounters_default,multiplecounters_counterone:MultipleCounters_CounterOne,multiplecounters_countertwo:MultipleCounters_CounterTwo,folder_subfolder_counter_default:Folder_subfolder_Counter_default,}, STATE[0]);
```

And with that I think everything is covered, and there should be enough examples to understand what's going on. I am very happy to receive suggestions on how to simplify this, since I spent far too long trying various things.

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -47,8 +47,11 @@ function isElementNode(node: Node): node is HTMLElement {
   return node.nodeType === Node.ELEMENT_NODE;
 }
 
-// deno-lint-ignore no-explicit-any
-export function revive(islands: Record<string, ComponentType>, props: any[]) {
+export function revive(
+  islands: Record<string, Record<string, ComponentType>>,
+  // deno-lint-ignore no-explicit-any
+  props: any[],
+) {
   _walkInner(
     islands,
     props,
@@ -127,7 +130,7 @@ interface Marker {
  * fashion over an HTMLElement's children list.
  */
 function _walkInner(
-  islands: Record<string, ComponentType>,
+  islands: Record<string, Record<string, ComponentType>>,
   // deno-lint-ignore no-explicit-any
   props: any[],
   markerStack: Marker[],
@@ -253,7 +256,7 @@ function _walkInner(
         }
       } else if (comment.startsWith("frsh")) {
         // We're opening a new island
-        const [id, n] = comment.slice(5).split(":");
+        const [id, exportName, n] = comment.slice(5).split(":");
         const islandProps = props[Number(n)];
 
         markerStack.push({
@@ -262,7 +265,7 @@ function _walkInner(
           text: comment,
           kind: MarkerKind.Island,
         });
-        const vnode = h(islands[id], islandProps);
+        const vnode = h(islands[id][exportName], islandProps);
         vnodeStack.push(vnode);
       }
     } else if (isTextNode(sib)) {
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -281,14 +281,21 @@ export class ServerContext {
       }
       const path = url.substring(baseUrl.length).substring("islands".length);
       const baseRoute = path.substring(1, path.length - extname(path).length);
-      const name = sanitizeIslandName(baseRoute);
-      const id = name.toLowerCase();
-      if (typeof module.default !== "function") {
-        throw new TypeError(
-          `Islands must default export a component ('${self}').`,
-        );
+
+      for (const [exportName, exportedFunction] of Object.entries(module)) {
+        if (typeof exportedFunction !== "function") {
+          continue;
+        }
+        const name = sanitizeIslandName(baseRoute);
+        const id = `${name}_${exportName}`.toLowerCase();
+        islands.push({
+          id,
+          name,
+          url,
+          component: exportedFunction,
+          exportName,
+        });
       }
-      islands.push({ id, name, url, component: module.default });
     }
 
     const staticFiles: StaticFile[] = [];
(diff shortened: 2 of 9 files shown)


marvinhagemeister (approved):
I like this a lot because it also means more code can be bundled together. Great work!

PR #1400, opened 2023-07-03
www: add versioning to docs

### LINKED ISSUE #1392: Add versioning to docs
With recent PRs merged the docs have been updated to the canary version, which leads to them not working for the stable release line. In other words: We need a versioned docs system.

Some learnings from doing that on preactjs.com:

- The stable docs should always have a `lastest` tag in the URL, rather then the version. The stable docs are most likely to be shared and thus most likely recommended by google. In Preact we only went with `v8.x` etc and that lead to search engines over-indexing old sites.
- A version selector is not enough, there needs to be a visual indictor and hint that the user is looking at old docs.

marvinhagemeister:
This PR adds versioning to our documentation. We ran into the scenario where we're documenting a feature that landed in `main`, but wasn't released in a stable version. This led to confusion because the docs didn't match the released version.

I've went a bit back and forth on the various approaches to versioning and I think I landed on a way that's best for our project and least maintenance intensive. In particular I wanted to avoid a situation where we need to copy all content files and have to backport updates between them. That should only be necessary when we do breaking changes like release Fresh 2.0 or something.

To resolve that the versioning follows the one chosen by https://eslint.org/docs/latest/ . Their strategy is to have one current version and one canary. There are no other versions. They implemented it in a way that the repo only contains the canary version and each release copies the canary version to a release version that's then pulled from some storage during the CI step. I very much like the overall approach of having one current and one canary version, but I wanted to avoid the CI logic and keep things simple.

Instead we have simply two version folders: One current and for canary. The canary folder can inherit pages from the current folder, which avoids us having to sync content.

tl;dr:
- Links are all the same as without this PR. Canary versions have `/canary` suffix
- Canary docs can inherit pages from current docs to not have to copy everything over

Screenshot 2023-07-03 at 16 40 51
Screenshot 2023-07-03 at 16 40 55

TODO:
- Add tests
- Rename `1.2` to something else but show the version in the version selector.

Fixes #1392

### DIFF
--- a/docs/canary/the-canary-version/index.md
+++ b/docs/canary/the-canary-version/index.md
@@ -1,0 +1,15 @@
+---
+description: |
+  Learn more about Fresh canary releases
+---
+
+The canary version represents the current development state of Fresh. It's
+intended for testing work in progress features before it lands in a stable
+release. Whenever new code is merged into the `main` branch on
+[GitHub](https://github.com/denoland/fresh) a new canary version is created that
+matches the hash of the git commit.
+
+With the addition of new features that are not yet in a stable release, we need
+a place to update our documentation. That's what this section of the
+documentation is about. It contains all information about unreleased features
+and changes that will land in a stable release.
--- a/docs/toc.ts
+++ b/docs/toc.ts
@@ -1,66 +1,161 @@
-type RawTableOfContents = Record<string, RawTableOfContentsEntry>;
+import FRESH_VERSIONS from "../versions.json" assert { type: "json" };
+
+type RawTableOfContents = Record<
+  string,
+  {
+    label: string;
+    content: Record<string, RawTableOfContentsEntry>;
+  }
+>;
 
 interface RawTableOfContentsEntry {
   title: string;
-  pages?: [string, string][];
+  link?: string;
+  pages?: [string, string, string?][];
 }
 
 const toc: RawTableOfContents = {
-  "introduction": {
-    "title": "Introduction",
+  canary: {
+    label: "canary",
+    content: {
+      "the-canary-version": {
+        title: "The canary version",
+      },
+      introduction: {
+        title: "Introduction",
+        link: "latest",
+      },
+      "getting-started": {
+        title: "Getting Started",
+        link: "latest",
+        pages: [
+          ["create-a-project", "Create a project", "link:latest"],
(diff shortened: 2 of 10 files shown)


deer:
This seems to strike the right balance between ease of implementation and functionality, given that it's just you on the team! Let me know once this is fully ready and I will check it out and try poking around with making some doc changes, to test the experience.

marvinhagemeister:
These changes are a little more than I had hoped, but couldn't find a way to make the PR smaller, without breaking the feature. It's best to review this on the deploy preview instance.

bartlomieju (approved):
LGTM

deer:
I made a fictitious change to the `Architecture` page and updated the `toc` and got the following diff:
```diff
diff --git a/docs/toc.ts b/docs/toc.ts
index 2019f77..2b98e07 100644
--- a/docs/toc.ts
+++ b/docs/toc.ts
@@ -44,7 +44,7 @@ const toc: RawTableOfContents = {
         title: "Concepts",
         link: "latest",
         pages: [
-          ["architecture", "Architecture", "link:latest"],
+          ["architecture", "Architecture", "link:canary"],
```
Everything works as expected -- i.e. I can swap between the pages using your spiffy dropdown and the content changes accordingly. This makes it easy enough to add new pages. But what does the release strategy look like? Seems like we'll need to manually go through and move the canary pages back to the default locations, and then maybe update the canary toc.

I suppose we could just leave the canary toc, with the strategy being that all updates go into these docs, and then once per release the release version is overwritten. Am I missing something obvious?

marvinhagemeister:
Current plan is to add docs for new features in fresh to the canary version and improving documentation for existing stuff can happen in the latest version. It's true that making those new pages available in current is a bit of manual work. I'm hoping that it won't be too much and if it becomes an issue we can change it.

My thinking behind it was that otherwise improvements to the current docs would only come out on a monthly basis or would need to be done both for the latest and the canary version. But then again, maybe that's not so bad? I don't know, happy to change if the current workflow doesn't work out.

@deer what do you think would make it easiest for you? I'm curious about your thoughts on that.

deer:
I'm happy to try out the newest approach and see what happens. As you said, it can always be changed. And, given that I don't have to worry about releasing new versions, I don't think I'm impacted by the worst part 😂

PR #1402, opened 2023-07-03
www: hopefully fix agolia search exceptions

marvinhagemeister:
Using the search on the production instance throws a couple of exceptions when using it.

Screenshot 2023-07-03 at 22 31 19

Took a while to investigate, because I couldn't reproduce it locally at first. So what's happening is that we use the native `esbuild` binary locally, but use the `wasm` build on deploy. The wasm build of the `esbuild` version we used seems to have a bug and messes up a variable somewhere. Forcing that to be used locally allowed me to reproduce the error. I couldn't pinpoint to exactly where it was going wrong, but simply updating `esbuild` resolves this.

### DIFF
--- a/src/build/deps.ts
+++ b/src/build/deps.ts
@@ -6,15 +6,15 @@ export {
 export { escape as regexpEscape } from "https://deno.land/std@0.190.0/regexp/escape.ts";
 
 // -- esbuild --
-// @deno-types="https://deno.land/x/esbuild@v0.17.19/mod.d.ts"
-import * as esbuildWasm from "https://deno.land/x/esbuild@v0.17.19/wasm.js";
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
CI error makes me thing that there is an issue with wasm in deno. I can reproduce this outside of testing too.

```sh
TypeError: NetworkError when attempting to fetch resource.
    at async mainFetch (ext:deno_fetch/26_fetch.js:266:12)
    at async fetch (ext:deno_fetch/26_fetch.js:490:7)
    at async tryToInstantiateModule (https://deno.land/x/esbuild@v0.18.11/wasm.js:1765:682)
```

marvinhagemeister:
hm running into another issue. Marking as draft again.

Issue #1321, opened 2023-06-17
Fresh Devtools


bartlomieju:
This is something @marvinhagemeister and I discussed loosely this week, opening an issue for we can keep track of it.

As Fresh projects get more complex a need for better tooling becomes more obvious. While Fresh already works with [Preact Devtools](https://preactjs.github.io/preact-devtools/), they are some parts of the framework that are not covered by Preact Devtools (eg. routes).

Marvin suggested that instead of creating a browser extension we could use a great idea introduced in [`Nuxt Devtools`](https://nuxt.com/blog/introducing-nuxt-devtools) that the devtools are part of the rendered Fresh application. It has a benefit of simpler development as we would be in complete control over the devtools (not constrained by the browser extensions API). That should also help with iteration speed, as development of the devtools itself would not require much additional setup compared to the current development of Fresh itself.

One interesting idea that Marvin raised is that in addition to Fresh specific tools, we could embed Preact Devtools into Fresh's devtools panel directly. I'm not entirely sure if that's for the better or worse, but the idea is certainly interesting!

Adding devtools would be quite a big project, fortunately it could be split into several phases where people could collaborate without stepping on each others' toes:
- [ ] add a devtools panel that we can render various tools into - reiterating here that what [Nuxt did](https://nuxt.com/blog/introducing-nuxt-devtools) is quite awesome solution. It could be incorporated into the `./demo` project first just to get initial design going. 
- [ ] Devtools should only be enabled in "dev" mode - ie. only when running `dev.ts` script. They should be completely disabled when running `main.ts`. Currently Fresh's server doesn't know if it's running in dev mode or not, the dev script is a thin wrapper that generates the manifest and watches for file changes.
- [ ] Create some actual tools
   - [ ] A "routing" panel is the obvious first choice. Having an interactive list of all routes is certainly useful, especially if we could present visually things like pattern matches (eg. `demo/routes/[name].tsx`). Again I think Nuxt Devtools did a brilliant job here and taking inspiration from it would result in a great outcome.
   - [ ] Another possibly useful panel would be "plugins" panel, though I'm not exactly sure what could be controlled from it at this point
- [ ] Test, iterate and ship it 🚀 

Curious to hear people's thoughts on this idea as well as other tools that might be useful.

marvinhagemeister:
Other random ideas:
- bundle analyzer
- something like vite inspector

cbinzer:
[...]

PR #1404, opened 2023-07-04
feat(server): Add dev mode env var

wtchen:
I'm hoping this small feature will be useful for people to develop against dev and production environments.

This will also help kick off https://github.com/denoland/fresh/issues/1321, which needs this distinction.


bartlomieju:
I believe I was wrong in my original comment in #1321 - Fresh already decides if it is in devmode or not based on `DENO_DEPLOYMENT_ID` env var.

wtchen:
Ah, good catch. I think having something explicit would still be useful; I can turn this into an abstraction.

wtchen:
Ok, I made some changes, let me know what you think 🙂

marvinhagemeister:
I'm not sure I understand what we gain from exposing if Fresh is in dev mode via an environment variable. Can you share more about your use case?

wtchen:
Of course. The idea is if the user wants to have different code for dev vs not dev, they can easily check against that env variable.

```ts
if (Deno.env.get("FRSH_DEVMODE") === 'true') {
  return new Response('Welcome to my dev site');
} else {
  return new Response('Welcome to my production site');
}
```

lucacasonato:
Could that usecase be resolved by a per-project env var?

jamesbradlee:
Isn't this something you could solve by just adding a `Deno.env.set("FRESH_DEVMODE", "true")` in your `dev.ts` file?

bartlomieju:
Thanks for the contribution! This PR is from the Fresh 1.x era and the codebase has since been completely rewritten for Fresh 2.x. The files modified here (`src/server/context.ts`, `src/server/mod.ts`) no longer exist.

Fresh 2.x has this built in via `config.mode` which is `"development"` or `"production"`, so the use case is already covered.

Closing as stale.

PR #1407, opened 2023-07-04
feat(server): workers in islands

### LINKED ISSUE #1398: Fresh should bundle worker entrypoints
Right now using workers in Fresh is kinda painful. For example, I want to load a worker in an island. I would expect to be able to do this with:

```js
const worker = new Worker(import.meta.resolve("./worker.ts"), { type: "module" });
```

However, Fresh does not pick up on this import and thus does not transpile / bundle the worker and replace `import.meta.resolve` with a static string. 

At some point we'll have a JS standard way of doing this, but for now we should use the commonly accepted stupid way of doing this.

```ts
import module myWorker from "./worker.ts";
myWorker instanceof Module; // true

const worker = new Worker(myWorker);

// or

const worker = new Worker();
worker.addModule(myWorker);
```

deer:
closes https://github.com/denoland/fresh/issues/1398

Or, at least it attempts to. Not entirely sure what the desired functionality is here. I'd appreciate guidance on some sensible tests and clarification around how we intend this to be used. I improved my test case from the comment I posted last night on the issue. You can see it [here](https://github.com/denoland/fresh/commit/e4b6576021ff74f0d34a0cf1fc77d678bb6f6c38). I support two forms of worker import rewriting right now.
1.  do it manually, e.g. `new URL("../workers/myFirstWorker.ts", import.meta.url),`
2. use the new hook, e.g. `useWorker("../../workers/doStuff.ts");`

This is still a bit hacky, since I had to declare a global state in order to share the current request url with the esbuild plugin.  You can see that in `bundleAssetRoute` where I do this:
```tsx
sharedState = {
  reqUrl: req.url,
};
```
I then use it later like:
```tsx
const url = sharedState.reqUrl;
```
Otherwise I don't know how to rewrite the `new URL(.....)` and `useWorker(....)` calls to point to the correct value in the server, since we append that randomNonce to our URLs. Maybe I'm misunderstanding something.

As always looking into a new section of code was very illuminating, although I think I need a bit of help here in order to get this over the line.

### DIFF
--- a/src/build/esbuild.ts
+++ b/src/build/esbuild.ts
@@ -5,6 +5,7 @@ import {
 } from "https://deno.land/x/esbuild@v0.19.4/mod.js";
 import { denoPlugins, fromFileUrl, regexpEscape, relative } from "./deps.ts";
 import { Builder, BuildSnapshot } from "./mod.ts";
+import { bundleAssetUrl } from "../server/constants.ts";
 
 export interface EsbuildBuilderOptions {
   /** The build ID. */
@@ -95,6 +96,7 @@ export class EsbuildBuilder implements Builder {
         metafile: true,
 
         plugins: [
+          workerUrlReplacePlugin,
           devClientUrlPlugin(opts.basePath),
           buildIdPlugin(opts.buildID),
           ...denoPlugins({ configPath: opts.configPath }),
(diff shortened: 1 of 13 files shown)


KnorpelSenf:
I need to use web workers in my site and I'm happy to help, but I have never contributed to fresh before, so I'm not sure how useful I can be here :D

For starters, I have tried to test your changes. I swapped out the import URL for `$fresh/` in `deno.json` to `https://raw.githubusercontent.com/deer/fresh/feat_workers_in_islands/`, and while I could run
```ts
const url = new URL("../workers/worker.ts", import.meta.url);
const worker = new Worker(url.href, { type: "module" });
worker.onmessage = (event) => {
  console.log("rcv", event);
};
worker.postMessage("from parent");
console.log('sent')
```
successfully, I don't see `worker.ts` actually being evaluated. Putting a `console.log` on the top level of `worker.ts` does not print anything.

Could you provide more exact instructions on how I can test this?

deer:
Whatever I'm doing here clearly isn't working on windows (as per the tests), so if you're on windows, please let me know. If you're on mac or linux, did you try looking at the tests I wrote? Look at those, and then see if they work for you, and then what the difference between the tests and your usage is.

PR #1408, opened 2023-07-04
Deprecate "handler" in favor of "handlers"

### LINKED ISSUE #1406: proposal: rename `handler` to `handlers` for route handlers
Continuing the discussion from #1396, the object in question has multiple handlers. However, it's called `handler`. Its type is also `Handlers`. I'd imagine this would make the average newcomer wonder why. I know I did. For these reasons, I think it'd make the most sense to rename `handler` to `handlers`.

wtchen:
This update makes "handlers" the canonical way of overriding the handlers function, and deprecates "handler".

Fixes https://github.com/denoland/fresh/issues/1406


marvinhagemeister (commented):
FYI: The new deno version that will likely be released later today or tomorrow contains a lint rule that warns when `handlers` is used. See https://github.com/denoland/deno_lint/pull/1173 . Before we merge this PR we'd likely need to drop the linting rule first.

I'm not sure if it's a good idea to deprecate `handler` right now timing wise. It causes churn, we need to update our docs and a valid handler can also be a function in which case the plural version is not fitting.

```js
export const handler = () => new Response("hello world");
```

wtchen:
Closing this PR, can resurrect at a future date

PR #1414, opened 2023-07-06
feat(server): warning for conflicting dynamic routes

### LINKED ISSUE #475: Behavior of `/[dynamic].ext` routes together with `/static.ext`

deer:
closes https://github.com/denoland/fresh/issues/475
supersedes https://github.com/denoland/fresh/pull/539

This is a slightly modified approach from what 539 does. Here we do the check after the `ServerConfig` has been created, so we only do this when `isDevMode` returns true. Additionally, this gives us access to the `Routes[]` array, which means we've already processed any `RouteOverride`s. What's nice about this is that we can then filter out any routes that a filename based approach wouldn't catch.

So when we do `deno task fixture` we receive the following message:
```
Potential routing conflict: The dynamic route '/:name' will sort below the following static routes:
  /connInfo
  /evil
  /failure
  /intercept
  /intercept_args
  /middleware_root
  /not_found
  /static
  /status_overwrite
```

I think this is useful, because it might not be obvious to new users what the behavior of having a dynamic and static route in the same folder is.

I noticed there are many other areas where we should be warning developers. The docs repeatedly use the word "must" and we should be checking these assumptions and giving sensible errors if they're violated. I will create a followup issue.

I had written a test for this functionality which asserts the message in the console, but I couldn't figure out how to kill the server after a period. So I commented out the test and refactored the changes to return the error instead.


deer:
Perhaps there should be a way to opt out of this check (and any future checks). At minimum such an opt-out would be useful for the tests, because when you now run `deno task test` the console is spammed with these warnings.

marvinhagemeister:
I'm admittedly not sure if a warning is warranted in this case. Warning when to files with dynamic parts like `[first].tsx` and `[second].tsx` exist in the same folder is definitely a good thing to do. But when it comes to having a bunch of static routes and one dynamic in the same folder seems fine to me. 

A use case I can come up with is a blog site where you want articles to be accessible at the root like `https://example.com/my-cool-post` but also want to have `https://example.com/about` to point to a specific about page.

> I noticed there are many other areas where we should be warning developers. The docs repeatedly use the word "must" and we should be checking these assumptions and giving sensible errors if they're violated. I will create a followup issue.

This is a very good point. Using the word "must" is odd and we should fix the underlying issue instead.

deer:
I agree it's fine, but I still strongly think this is a foot gun. If people don't fully understand dynamic routes or have misconfigured their route overrides, it should be made clear what's going on. Otherwise debugging the issue is a big pain. (e.g. ```I have `routes/myfolder/mypage.tsx` but when I go to `localhost:8000/myfolder/mypage` I'm not getting that page!``` And then someone more experienced would have to ask them if they have dynamic routes or route overrides. We could save lots of time by just letting people know when they start the server) Do you think there's somewhere more appropriate for these sorts of checks? Some sort of "doctor" task to analyze the health of the project.

Also this test can benefit from the knowledge you shared in my other PR. I will change this as well, if you're convinced that this functionality should exist somewhere.

deer:
I've created https://github.com/denoland/fresh/issues/1416 with my giant list of things the developer can do to cause unexpected behavior. The issue this PR is solving belongs on the list as well, but I hope this one will be accepted 😄

lino-levan:
I'm pretty strongly against warning for "conflicting" routes. I haven't checked it against this PR, but a pretty common pattern I've seen throughout fresh apps is having `/nested/index.ts` and `/nested/[id].ts`.

Additionally, "conflicting" routes are a pretty popular in general (ex. twitter having `twitter.com/[user]` but also having `twitter.com/explore`).

What this PR could do that I would approve of is two wildcard routes in one directory.

marvinhagemeister:
+1 for changing it to only check for multiple wildcard routes.

deer:
I still think you guys are approaching this from expert perspective, or at least "experienced web dev" perspective. It wasn't at all obvious to me that having a `/nested/index.ts` and `/nested/[dynamic].ts` is fine, or that we could also have `/nested/static.ts` as well and that would have priority over the dynamic.

But I'm a man of the people. If this is what you want, I will change it. I think any hint is better than nothing.

marvinhagemeister:
I get where you're coming from but I think it's more an issue of making Fresh more debuggable vs limiting its capabilities. In the future, I'm hoping we have some sort of `--debug` flag where you can see which middleware and which route was hit when Fresh handles a request. Something that makes it more visible what Fresh is doing and how it operates.

deer:
It now just checks for dynamic route conflicts, as requested. I will start thinking about the `--debug` flag.

marvinhagemeister:
Check out vite's `--debug` flag. I think theirs is pretty cool as they show you everything vite is doing under the hood.

src/server/context_test.ts
```diff
@@ -22,3 +26,26 @@ Deno.test("selectMiddlewares", () => {
...
+  fn() {
+    const routes = [
+      { name: "control-index", pattern: "/control" },
+      { name: "control-normal_route", pattern: "/control/normal_route" },
+      {
+        name: "two_dynamic-[second_dynamic]",
+        pattern: "/two_dynamic/:second_dynamic",
+      },
+      { name: "two_dynamic-[dynamic]", pattern: "/two_dynamic/:dynamic" },
+      { name: "override-[dynamic]", pattern: "/override/:dynamic" },
```
marvinhagemeister: Let's add a test case for when the dynamic parts of the route pattern are at different levels:

- `/foo/:bar` vs `/:bar/foo`

In that case it shouldn't warn. If I read the regex in `isDynamicRoute` correctly this would be treated as a conflict.

deer: I didn't add this exact one since I can't see how this could possibly cause an issue. I settled on this instead:
```ts
{ name: "nested2-[foo]/bar", pattern: "/nested2/:foo/bar" }, //routes/nested2/[foo]/bar.tsx
{ name: "nested2-foo/[bar]", pattern: "/nested2/foo/:bar" }, //routes/nested2/foo/[bar].tsx
```

Let me know if you want your exact case added.

deer:
I redid this to match the style from #1489.

deer:
Closing based on https://github.com/denoland/fresh/issues/1558#issuecomment-1698526674

PR #1415, opened 2023-07-06
fix: twind including unused CSS rules

marvinhagemeister:
While working on #1388 I noticed that our twind plugins add CSS classes based on vnode creation instead of when a vnode is actually used during rendering. Just because a vnode was created, doesn't mean that it will be used. Common scenarios where that's not the case is conditional rendering.

```jsx
let vnode = <div class="text-red-600" />
if (someCondition) {
  vnode = <div class="text-blue-600" />
}
```

This will make #1388 easier as the rendering there is still synchronous under the hood and we can keep track of local state synchronously.

Whilst `options.__b` (=`options._diff`) was initially not intended to be a public render hook for Preact, it has been stable since the original 10.x release a couple of years ago. Problem is that changing that now in itself would be a breaking change. We should expose it through typings or something else though at some point 👀

### DIFF
--- a/plugins/twind/shared.ts
+++ b/plugins/twind/shared.ts
@@ -1,6 +1,8 @@
 import { JSX, options as preactOptions, VNode } from "preact";
 import { Configuration, setup as twSetup, Sheet, tw } from "twind";
 
+type PreactOptions = typeof preactOptions & { __b?: (vnode: VNode) => void };
+
 export const STYLE_ELEMENT_ID = "__FRSH_TWIND";
 
 export interface Options extends Omit<Configuration, "mode" | "sheet"> {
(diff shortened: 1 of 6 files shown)


marvinhagemeister:
Well, looks like some other test is relying on that behavior. Need to look into that.

lucacasonato:
Won't this break the ``<div class={tw`text-blue-500 ${xyz}`}`` kind of use cases (where `tw` is called manually)? I am not sure if it's important to keep them working, but if they are not intended to work we should figure out if we can issue a warning when a user tries to do this anyway.

This is mainly used in conjunction with the `css` template literal tag / function.

marvinhagemeister:
@lucacasonato That keeps working because the `tw` helper writes into the global stylesheet directly, regardless of what Preact is doing. Added a test case for that to ensure that this works.

lucacasonato:
But to clarify, that would _not_ work in an async component right, because there is no single "global" style sheet because of thread local storage?

This PR makes sense to me in any case, just want to clarify that behaviour.

marvinhagemeister:
@lucacasonato The secret is that the actual rendering is still synchronous, even if it looks async from the outside. The work in #1388 treats the function as a handler that happens to return JSX, and then we render the returned JSX. That rendering is still synchronous. If we decide at some point to do some form of streaming rendering (could see this happening) where we throw promises to delay subtree rendering, it's all still synchronous under the hood. The work is just split up in batches.

That means we can pass some sort of object as the render context around and always restore it for the currently rendering subtree. That would allow us to stuff a virtual subtree in that sorta "thread local" storage and always write into that compared to a truly global one. All this works because it's very unlikely that we'll add truly async components to `preact-render-to-string`. Can you pass in a `Promise` that halts rendering until that resolves - Sure. But that will be pretty much it.

lucacasonato:
Right, but what about:

```tsx
export default async function Route() {  
  const buttonStyle = tw`text-red-500`;
  await fetchSomething();

  return (
    <div>
      <button class={buttonStyle} />
      <button class={buttonStyle} />
      <button class={buttonStyle} />
    </div>
  );
}

marvinhagemeister:
Merging this as the discussion here is more about #1388 and not about the changes in this PR itself.

PR #1422, opened 2023-07-08
feat(init): task to generate manifest file

### LINKED ISSUE #384: proposal: API for generating manifest manually without using $fresh/dev.ts

deer:
closes https://github.com/denoland/fresh/issues/384
supersedes https://github.com/denoland/fresh/pull/908

I think this is the last of the PRs where I'm refreshing an old PR and closing an existing issue! Here I've taken (in my opinion) a slightly cleaner approach than 908. I'm just exporting a new function.

That's not terribly useful on its own, so I've wrapped it in a small CLI to be easily invokable. Now users can do `deno task manifest` to regenerate their manifest file.

This approach could be expanded to something like `deno task doctor` to do a bunch of static analysis checks on the health of the user's project, but that's for a separate discussion.

Opening as a draft, since I suspect the command needs to be modified to account for windows machines, and I guess the CI will fail there.


deer:
This still does the same thing as before, but in a slightly better approach. I'm not convinced this is absolutely necessary, but I still think it would be interesting to consider what other sort of tasks could fit into this pattern. As mentioned, a `deno task doctor` to launch the health checks would be nice.

Regardless, if you're not interested in this, please just close this PR along with #384 and #908.

marvinhagemeister (approved):

deer:
@marvinhagemeister, do you know if there are talks to have the task runner respect import maps? It would of course be more direct to have:
```
manifest: "deno task $fresh/src/dev/cli.ts manifest $(pwd)",
```

marvinhagemeister:
I'm not aware of anything in that direction.

PR #1423, opened 2023-07-08
www: update docs design

marvinhagemeister:
This updates the doc page design on larger screen. Best explained via screenshots.

Before:

Screenshot 2023-07-08 at 10 47 40

After:

Screenshot 2023-07-08 at 10 47 47


deer (approved):
Great, I'm glad we can use more of the screen now.

PR #1426, opened 2023-07-08
fix: trailingSlash option

### LINKED ISSUE #1425: BUG: trailing slash support is not working
Hi, I activated the trailing slash option in the start function:

`await start(manifest, { plugins: [], router: { trailingSlash: true } });`

But that leads to that the JS and static files could not be downloaded anymore. It seems like Fresh is doing a redirect of the static and JS files to the trailing slash version and that files can't be found.

Bildschirmfoto 2023-07-08 um 15 37 16

Fresh Version: 1.2.0
Deno Version: 1.35.0

lino-levan:
Closes #1425 

The current trailingSlash option is fundamentally broken and I'm not really sure how it ever worked:
- It just added a "/" to the end of the href instead of the pathname. If your URL had a hash or query parameter it would crash instantly.
- It added a trailing slash to files which causes problems for the router
- It added a trailing slash to internal paths which causes problems for the router

These should be fixed in this PR.


marvinhagemeister (commented):
The fix looks great! Should we add some test for this to ensure that we don't accidentally break it in the future?

lino-levan:
Tests added!

lino-levan:
I always forget about windows 😅

marvinhagemeister (approved):

marvinhagemeister:
LGTM, thanks for the PR 👍

PR #1427, opened 2023-07-08
feat: move to Deno.serve

### LINKED ISSUE #1424: Switch to `Deno.serve` by default

lino-levan:
Closes #1424


www/routes/_middleware.ts
```diff
@@ -61,6 +61,7 @@ function ga4(
 
     // Create basic report.
     const measurementId = GA4_MEASUREMENT_ID;
+    // @ts-ignore GA4Report doesn't even use the localAddress parameter
```
lino-levan, line 64: Pretty unfortunate but the typing of this module is just wrong.

docs/concepts/updating.md
```diff
@@ -23,6 +23,7 @@ below.
 | 1.0.0-1.0.2   | >=10.8.1 <11.0.0  | >=5.2.0 <6.0.0          | >= 1.23.0 |
 | 1.1.0-1.1.5   | >=10.8.1 <11.0.0  | >=5.2.0 <6.0.0          | >= 1.25.0 |
 | 1.2.0         | >=10.15.0 <11.0.0 | >=6.1.0                 | >= 1.25.0 |
+| 1.2.1         | >=10.15.0 <11.0.0 | >=6.1.0                 | >= 1.35.0 |
```
marvinhagemeister: Super tiny nit: We have some unreleased features in `main`, so the next version will be very likely `1.3.0`.

```suggestion
| 1.3.0         | >=10.15.0 <11.0.0 | >=6.1.0                 | >= 1.35.0 |
```

marvinhagemeister (commented):
Reading through the PR I'm wondering if we can make it a non-breaking change. We could keep the "old" way of doing things around a little longer. Like check if `Deno.serve` exists, and if not fall back to the server module form std.

www/routes/_middleware.ts
marvinhagemeister: All good, happens from time to time with third party libraries.

lino-levan:
> Like check if Deno.serve exists, and if not fall back to the server module form std.

The thing is, the serve module from std is going to be deprecated like, the next update. We'd need to keep around an old version of std which feels wrong to me. Considering the friendly error message users would get ("Please upgrade to at least Deno version 1.35.0!"), I'm not too concerned about trying to maintain backwards compatibility on that front. If someone needed to use an older Deno version for whatever reason, they could just use the `--unstable` flag.

Other than that, the only breaking change is moving from `ConnInfo` to `Deno.ServeHandlerInfo` which removes the `localAddress` property which should only matter for people relying on that which I have not seen.

marvinhagemeister:
We had a call today and I got confirmation that the server module in std will not be deprecated soon. The current plan is to only deprecated it once the std library hits `1.0`. I really want this to land as part of the next release which I'm hoping to cut on Thursday. I don't want this to be a breaking release as that's already planned for next month.

lino-levan:
Ah sorry, I should have clarified. It will be deprecated as in "please don't use this" (I made the PR for this!). It will be removed in `1.0.0`. What do you propose to make this not a breaking change?

marvinhagemeister (approved):
Took the liberty of updating the PR to make it a non-breaking change.

PR #1431, opened 2023-07-08
feat: make vscode-deno default formatter

marvinhagemeister:
The Prettier extension setting are language specific and would take preference over the global default formatter setting. This lead to the extensions constantly fighting on who gets to format the code on my machine.


bartlomieju (approved):
LGTM

PR #1433, opened 2023-07-08
fix: manifest file encoding non-latin chars

### LINKED ISSUE #1432: Incorrectly handles routes that have (non-ASCII?) non-Latin characters
Screenshot 2023-07-08 at 10 49 21 PM

e.g., تایبەتێتی.tsx

marvinhagemeister:
This PR ensures that we don't parse file paths as URLs when traversing the file system. Fixes #1432 on the fresh side, but it seems like something in deploy needs to be changed too.

### DIFF
--- a/src/dev/deps.ts
+++ b/src/dev/deps.ts
@@ -5,6 +5,7 @@ export {
   extname,
   fromFileUrl,
   join,
+  posix,
   relative,
   resolve,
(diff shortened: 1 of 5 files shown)


rojvv (approved):

iuioiua (approved):
LGTM!

PR #1434, opened 2023-07-09
feat(init.ts): enable lint rules for fresh apps

uki00a:
`deno lint` now supports rules for fresh apps, such as `fresh-handler-export` and `fresh-server-event-handlers`.  This change enables those rules in the project generated by `init.ts`.


marvinhagemeister (approved):
Awesome, thanks for making a PR for this 🎉 

Just checked with an older version of Deno and it seems like it just ignores unknown tags. so this is not a breaking change.

PR #1327, opened 2023-06-19
fix: `update.ts` fails when there is no islands directory

arrow2nd:
Fix #1326 

Fixed to ignore the error if the directory did not exist, referring to the change diff in #1185.

PR #1435, opened 2023-07-09
refactor: optional directory check

marvinhagemeister:
Just a small refactor because we only need to check if it's a directory before we start the search.

Follow up to #1327

### DIFF
--- a/src/dev/mod.ts
+++ b/src/dev/mod.ts
@@ -1,12 +1,4 @@
-import {
-  dirname,
-  fromFileUrl,
-  gte,
-  join,
-  toFileUrl,
-  walk,
-  WalkError,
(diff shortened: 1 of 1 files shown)


deer (approved):
A reasonable simplification. 👍

lucacasonato (approved):
LGTM

Can you report this to std? This looks like a bug there (the catch should have caught it). I reported this already months ago, but then was assured it's working correctly now. Seems not yet unfortunately

marvinhagemeister:
@lucacasonato It looks to be fixed in https://github.com/denoland/deno_std/issues/1310 but then again I find it easier to check it first, rather than wrapping the whole loop in a `try/catch`.

PR #1438, opened 2023-07-10
docs: fix file locations in comments of crud api

### LINKED ISSUE #1436: Wrong comment
Hi everyone, everything fine?

Browsing through the documentation, some code comments on the "https://fresh.deno.dev/docs/examples/creating-a-crud-api" page are inconsistent with the page structure presented by the beginning of the explanation. For example, the comment "//routes/api/index.ts" should be //routes/api/users/index.ts (given the structure shown).

I'm learning the framework and maybe some newbies might end up getting confused.

deer:
closes https://github.com/denoland/fresh/issues/1436

This wasn't a regression from my recent cleanup when I standardized filenames in examples, but was wrong when I introduced the example. Oops 🙈

### DIFF
--- a/docs/examples/creating-a-crud-api.md
+++ b/docs/examples/creating-a-crud-api.md
@@ -35,7 +35,7 @@ full files are available at the bottom for reference.
 `POST` (create) is used to create a resource.
 
 ```tsx
-// routes/api/index.ts
+// routes/api/users/index.ts
 export const handler: Handlers<User | null> = {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
LGTM 👍

deer:
One thing to keep in mind when doing this: I forgot to check to see if we've branched this doc for canary. Fortunately we hadn't, but just a heads up in case you forgot to check as well.

PR #1442, opened 2023-07-10
feat: support Preact error boundaries

marvinhagemeister:
This PR adds support for Preact error boundaries on the server. They allow you to catch errors during rendering and render a fallback UI. In the future we could think of a more native way to do that other than creating a custom `<ErrorBoundary>` component. Anyhow this PRs lays the foundation for all of that.

Example:

```jsx
class ErrorBoundary extends Component {
  state = { error: null };

  static getDerivedStateFromError(error) {
    return { error };
  }

  render() {
    return this.state.error
      ? this.props.fallback(this.state.error)
      : this.props.children;
  }
}

// later:
<ErrorBoundary
  fallback={error => <p>Error happened: {error.message}</p>}
>
  <SomeComponentThatThrows />
</ErrorBoundary>
```


lucacasonato (approved):
LGTM!

PR #1444, opened 2023-07-10
feat: add upgrade check

marvinhagemeister:
This PR adds an upgrade check to notify users of a Fresh version. The check is done once per day.

This is how it looks like.

Screenshot 2023-07-07 at 16 55 28

### DIFF
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -1,0 +1,29 @@
+name: post_publish
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  update-dl-version:
+    name: update dl.deno.land version
+    runs-on: ubuntu-22.04
+    if: github.repository == 'denoland/deno'
+    steps:
+      - name: Authenticate with Google Cloud
+        uses: google-github-actions/auth@v1
+        with:
+          project_id: denoland
+          credentials_json: ${{ secrets.GCP_SA_KEY }}
+          export_environment_variables: true
+          create_credentials_file: true
+
+      - name: Setup gcloud
+        uses: google-github-actions/setup-gcloud@v1
+        with:
+          project_id: denoland
+
+      - name: Upload version file to dl.deno.land
+        run: |
+          echo ${GITHUB_REF#refs/*/} > release-latest.txt
+          gsutil -h "Cache-Control: no-cache" cp release-latest.txt gs://dl.deno.land/fresh/release-latest.txt
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -30,5 +30,5 @@ jobs:
 
       - name: Upload version file to dl.deno.land
         run: |
-          echo ${GITHUB_REF#refs/*/} > release-latest.txt
+          cat versions.json | jq -r ".[0]" > release-latest.txt
           gsutil -h "Cache-Control: no-cache" cp release-latest.txt gs://dl.deno.land/fresh/release-latest.txt
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -15,6 +15,9 @@ jobs:
     runs-on: ubuntu-22.04
     if: github.repository == 'denoland/fresh'
     steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
       - name: Authenticate with Google Cloud
         uses: google-github-actions/auth@v1
         with:
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -1,13 +1,8 @@
 name: post_publish
 
-# on:
-#   release:
-#     types: [published]
 on:
-  push:
-    branches: [main]
-  pull_request:
-    branches: [main]
+  release:
+    types: [published]
 
 jobs:
   update-dl-version:
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -1,8 +1,13 @@
 name: post_publish
 
+# on:
+#   release:
+#     types: [published]
 on:
-  release:
-    types: [published]
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
 
 jobs:
   update-dl-version:
--- a/.github/workflows/post_publish.yml
+++ b/.github/workflows/post_publish.yml
@@ -13,7 +13,7 @@ jobs:
   update-dl-version:
     name: update dl.deno.land version
     runs-on: ubuntu-22.04
-    if: github.repository == 'denoland/deno'
+    if: github.repository == 'denoland/fresh'
     steps:
       - name: Authenticate with Google Cloud
         uses: google-github-actions/auth@v1
--- a/docs/canary/concepts/updating.md
+++ b/docs/canary/concepts/updating.md
@@ -76,7 +76,9 @@ Fresh 1.0.2 to 1.1.3, and update Preact to the latest version:
 
 ## Automatic update checks
 
-Fresh will periodically check if a new Fresh version is available if it's running outside of a CI. This happens once per day and can be disabled by setting the `FRESH_NO_UPDATE_CHECK=true` environment variable.
+Fresh will periodically check if a new Fresh version is available if it's
+running outside of a CI. This happens once per day and can be disabled by
+setting the `FRESH_NO_UPDATE_CHECK=true` environment variable.
 
 ## Code mods
 
--- a/docs/canary/concepts/updating.md
+++ b/docs/canary/concepts/updating.md
@@ -77,8 +77,8 @@ Fresh 1.0.2 to 1.1.3, and update Preact to the latest version:
 ## Automatic update checks
 
 Fresh will periodically check if a new Fresh version is available if it's
-running outside of CI. This happens once per day and can be disabled by
-setting the `FRESH_NO_UPDATE_CHECK=true` environment variable.
+running outside of CI. This happens once per day and can be disabled by setting
+the `FRESH_NO_UPDATE_CHECK=true` environment variable.
 
 ## Code mods
 
--- a/docs/canary/concepts/updating.md
+++ b/docs/canary/concepts/updating.md
@@ -1,0 +1,126 @@
+---
+description: |
+  Fresh regularly releases new versions of the framework. This page explains how to update your project.
+---
+
+Fresh constists of multiple pieces which are independently versioned and
+released.
+
+- Fresh (https://deno.land/x/fresh)
+- Preact (https://esm.sh/preact)
+- preact-render-to-string (https://esm.sh/preact-render-to-string)
+
+Some plugins also have their own dependencies that can be updated independently.
+
+- Twind (https://esm.sh/twind) (for the twind plugin)
+
+For the most part these pieces can be updated independently. Certain versions of
+Fresh may require a minimum version of a given dependency. This is documented
+below.
+
+| Fresh version | Preact            | preact-render-to-string | Deno      |
+| ------------- | ----------------- | ----------------------- | --------- |
+| 1.0.0-1.0.2   | >=10.8.1 <11.0.0  | >=5.2.0 <6.0.0          | >= 1.23.0 |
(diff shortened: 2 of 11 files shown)


src/server/update_check.ts
```diff
@@ -0,0 +1,86 @@
+import * as path from "https://deno.land/std@0.190.0/path/mod.ts";
```
deer: Let's not import them like this, if possible. Lino just updated everything to 0.193.0 anyway. What about importing from deps.ts?

src/server/update_check.ts
```diff
@@ -0,0 +1,101 @@
...
+  const output = await new Deno.Command("deno", { args: ["info"] }).output();
+  const denoDir = colors.stripColor(new TextDecoder().decode(output.stdout))
+    .split(
+      /\n/g,
+    ).find(
+      (line) => line.startsWith("DENO_DIR"),
+    )?.replace("DENO_DIR location: ", "");
+
+  return denoDir;
+}
```
lino-levan: This feels really bad. Spawning a subprocess to read the output text? I was looking at `/x/plug` since they also write stuff to the cache directory but they hardcode the path calculations depending on the OS.

Could we possibly use https://github.com/denoland/deno_cache instead?

tests/cli_update_check_test.ts
```diff
@@ -0,0 +1,156 @@
+import { colors, join } from "../src/server/deps.ts";
+import {
+  assertEquals,
+  assertMatch,
+  assertNotMatch,
+} from "$std/testing/asserts.ts";
+
+Deno.test({
```
lino-levan, line 9: praise: thanks for writing tests

tests/fixture_update_check/mod.ts
```diff
@@ -0,0 +1,10 @@
+import { updateCheck } from "$fresh/src/server/update_check.ts";
+
+// deno-lint-ignore require-await
+async function getLatestVersion() {
+  console.log("fetching latest version");
+  return Deno.env.get("LATEST_VERSION") ?? "1.2.0";
```
lino-levan: Are we going to have to update this every time fresh releases a new version? Feels super unfortunate.

src/server/update_check.ts
```diff
@@ -0,0 +1,101 @@
...
+  throw new Error(`Could not fetch latest version.`);
+}
+
+export async function updateCheck(
+  interval: number,
+  getLatestVersion = fetchLatestVersion,
+) {
+  // Skip update checks on CI or Deno Deploy
+  if (
+    Deno.env.get("CI") === "true" || Deno.env.get("FRESH_NO_UPDATE_CHECK") ||
```
lino-levan: Should we add documentation for `FRESH_NO_UPDATE_CHECK`?

src/server/update_check.ts
```diff
@@ -0,0 +1,101 @@
...
+    return;
+  }
+
+  // Abort if we couldn't find a deno_dir
+  const denoDir = await getDenoDir();
+  if (!denoDir) {
+    return;
+  }
+
+  const versions = (await import("../../versions.json", {
```
lino-levan: Is this portable?

tests/fixture_update_check/mod.ts
marvinhagemeister: Nah, this file is only used for the tests. The tests don't care in particular what the version is and are independent of the real Fresh version number. Could've used `99.99.999` here too.

src/server/update_check.ts
marvinhagemeister: What do you mean with "portable"?

tests/fixture_update_check/mod.ts
lino-levan: I wonder if `99.99.999` isn't clearer in terms of intent. Up to you.

src/server/update_check.ts
lino-levan: Ah sorry, I misread this as a fetch. Closing this.

src/server/update_check.ts
marvinhagemeister: Agree, this was mostly the quickest way to prototype the feature and mostly why this is still marked as draft. But yeah it's not ideal and needs to be changed before the PR is ready for review. Will switch to storing something in the user's home dir instead.

tests/fixture_update_check/mod.ts
marvinhagemeister: Good point, that avoids any confusion in the first place, will change that.

src/server/mod.ts
```diff
@@ -83,6 +85,9 @@ export async function createHandler(
 export async function start(routes: Manifest, opts: StartOptions = {}) {
   const ctx = await ServerContext.fromManifest(routes, opts);
 
+  // Run update check in background
+  updateCheck(UPGRADE_CHECK_INTERVAL);
```
lucacasonato: Should we only run this in the dev script maybe?

lucacasonato: Also consider adding a .catch here

marvinhagemeister: Good idea about moving the check to `dev`. Updated the PR

.github/workflows/post_publish.yml
```diff
@@ -0,0 +1,37 @@
+name: post_publish
+
+# on:
+#   release:
+#     types: [published]
```
iuioiua: What's happening here?

docs/canary/concepts/updating.md
```diff
@@ -0,0 +1,128 @@
+---
+description: |
+  Fresh regularly releases new versions of the framework. This page explains how to update your project.
```
iuioiua:
```suggestion
  New versions of Fresh are regularly released. This page explains how to update your project.
```
Nit: it may sound like "Fresh" and "the framework" are two separate entities.

docs/canary/concepts/updating.md
```diff
@@ -0,0 +1,128 @@
...
+      "twind": "https://esm.sh/twind@0.16.17",
+      "twind/": "https://esm.sh/twind@0.16.17/"
+    }
+  }
+```
+
+## Automatic update checks
+
+Fresh will periodically check if a new Fresh version is available if it's
+running outside of a CI. This happens once per day and can be disabled by
```
iuioiua:
```suggestion
running outside of CI. This happens once per day and can be disabled by
```
Nit

src/dev/mod.ts
```diff
@@ -164,6 +170,9 @@ export default manifest;
 export async function dev(base: string, entrypoint: string) {
   ensureMinDenoVersion();
 
+  // Run update check in background
+  updateCheck(UPGRADE_CHECK_INTERVAL).catch(() => {});
```
iuioiua: You could use https://deno.land/std@0.194.0/datetime/constants.ts?s=DAY

src/dev/update_check.ts
```diff
@@ -0,0 +1,126 @@
...
+export async function updateCheck(
+  interval: number,
+  getCacheDir = getFreshCacheDir,
+  getLatestVersion = fetchLatestVersion,
+) {
+  // Skip update checks on CI or Deno Deploy
+  const FRESH_NO_UPDATE_CHECK = Deno.env.get("FRESH_NO_UPDATE_CHECK");
+  if (
+    Deno.env.get("CI") === "true" || FRESH_NO_UPDATE_CHECK === "true" ||
+    FRESH_NO_UPDATE_CHECK === "1" ||
```
iuioiua: Is the `FRESH_NO_UPDATE_CHECK === "1"` check needed?

tests/cli_update_check_test.ts
```diff
@@ -0,0 +1,157 @@
...
+  name: "stores update check file in $HOME/fresh",
+  async fn() {
+    const tmpDirName = await Deno.makeTempDir();
+    const filePath = join(tmpDirName, "latest.json");
+
+    await new Deno.Command(Deno.execPath(), {
+      args: ["run", "-A", "./tests/fixture_update_check/mod.ts"],
+      env: {
+        CI: "false",
```
iuioiua, line 18:
```suggestion
```
Nit

src/dev/update_check.ts
```diff
@@ -0,0 +1,126 @@
...
+  const home = getCacheDir();
+  if (!home) return;
+  const filePath = join(home, "latest.json");
+  try {
+    await Deno.mkdir(dirname(filePath), { recursive: true });
+  } catch (err) {
+    if (!(err instanceof Deno.errors.AlreadyExists)) {
+      throw err;
+    }
+  }
```
iuioiua: Is this code block needed? If so, this creates a directory called `latest.json`. Is that intentional?

src/dev/update_check.ts
```diff
@@ -0,0 +1,126 @@
+import { colors, dirname, join } from "./deps.ts";
+
+interface CheckFile {
+  last_checked: string;
```
iuioiua, line 4: How about this being a `Date` or even just a `number`?

src/dev/update_check.ts
```diff
@@ -0,0 +1,126 @@
...
+      throw err;
+    }
+  }
+
+  const versions = (await import("../../versions.json", {
+    "assert": { type: "json" },
+  })).default as string[];
+  if (!versions.length) {
+    return;
+  }
```
iuioiua:
```suggestion
```
`versions.length` is always greater than zero.

tests/cli_update_check_test.ts
```diff
@@ -0,0 +1,157 @@
...
+  async fn(t) {
+    const envs = ["FRESH_NO_UPDATE_CHECK", "CI", "DENO_DEPLOYMENT_ID"];
+
+    for (const env of envs) {
+      await t.step(`checking ${env}`, async () => {
+        const tmpDirName = await Deno.makeTempDir();
+        const out = await new Deno.Command(Deno.execPath(), {
+          args: ["run", "-A", "./tests/fixture_update_check/mod.ts"],
+          env: {
+            CI: "false",
```
iuioiua:
```suggestion
```
Nit

iuioiua (dismissed):
Looking good! Few nits.

src/dev/update_check.ts
```diff
@@ -0,0 +1,126 @@
...
+  const FRESH_NO_UPDATE_CHECK = Deno.env.get("FRESH_NO_UPDATE_CHECK");
+  if (
+    Deno.env.get("CI") === "true" || FRESH_NO_UPDATE_CHECK === "true" ||
+    FRESH_NO_UPDATE_CHECK === "1" ||
+    Deno.env.get("DENO_DEPLOYMENT_ID")
+  ) {
+    return;
+  }
+
+  // Abort if we couldn't find a deno_dir
```
kt3k: This comment looks outdated?
```suggestion
  // Abort if we couldn't find fresh cache dir
```

src/dev/update_check.ts
kt3k: `dirname(filePath)` is `<OS's cache dir>/fresh`. This looks correct to me.

BTW we could avoid `dirname` call if we do `Deno.mkdir(home, { recursive: true })` at line 68

iuioiua: Ah, yes. I see.

.github/workflows/post_publish.yml
marvinhagemeister: That's a leftover from debugging. Goal is to only update the release file when a new release is published on GH. So the commented code is what we want.

src/dev/update_check.ts
marvinhagemeister: Good point about calling `mkdir` directly on `home`.

src/dev/update_check.ts
marvinhagemeister: Yeah, I guess since we control the contents of `versions.json` ourselves we don't have to be that defensive here.

src/dev/update_check.ts
marvinhagemeister: What would we gain by changing it? This is an update check and this file isn't used for anything else. Benefits of keeping this an ISO string is that it's human readable.

lucacasonato (approved):
LGTM

PR #1449, opened 2023-07-11
fix(server): make AppProps generic

### LINKED ISSUE #1446: AppProps needs to be generic type
Right now, the AppProps type in a _app.tsx file doesn't support types for the data and state variables.

Here is a workaround I am currently using that seems to work for my needs:

```
// TODO: remove this once AppProps is generic type
// import { AppProps } from "$fresh/server.ts";
import { PageProps } from "$fresh/server.ts";
import { ComponentType } from "preact";
interface AppProps<Data, State> extends PageProps<Data, State> {
  Component: ComponentType<Record<never, never>>;
}
```

deer:
closes https://github.com/denoland/fresh/issues/1446

Since I haven't done anything purely working on types, I thought this would be a good way to jump in. I just copied over the generics from `PageProps` -- not sure if this is the right thing to do.

I noticed we don't have any tests for this stuff. Thoughts on using something like https://github.com/SamVerschueren/tsd here?


lino-levan:
> I noticed we don't have any tests for this stuff. Thoughts on using something like https://github.com/SamVerschueren/tsd here?

The standard library has [a module](https://deno.land/std@0.193.0/testing/types.ts?doc=) for testing types. I'd recommend using that instead.

deer:
Thanks for pointing that out. I tried writing a test, although I'm not sure how useful it is. Thoughts?

deer:
@iuioiua, you're my types guy. Thoughts about this one?

tests/types_test.ts
iuioiua, line 1: Dare I say that this test may not be necessary based on the following:
1. These types have been added to the tests included in this PR.
2. These same tests are type-checked in CI and pass. This can be validated once #1454 is merged.

Thoughts, @kt3k?

iuioiua (approved):
LGTM! Just questioning the type test is all.

tests/types_test.ts
deer: Well, I'm not kt3k, but I agree. The type test has always seemed a bit strange, but I thought I would try it. Interesting, and now time to revert 😄

iuioiua (approved):

src/server/types.ts
```diff
@@ -161,7 +161,9 @@ export interface RouterState {
 
 // --- APP ---
 
-export interface AppProps extends PageProps {
+// deno-lint-ignore no-explicit-any
+export interface AppProps<T = any, S = Record<string, unknown>>
```
lino-levan, line 165: Do we need to say `T = any` here?

deer: We have `export interface PageProps<T = any, S = Record<string, unknown>>` so don't we want to mimic this functionality here? What do you suggest?

mcgear:
Based on the completion of code for this task, we may not need to include any additional development at this point. However, I believe it’s important to address LayoutProps concurrently with this task to maintain consistency between the APIs, as discussed in issue #1580.

Therefore, I would like to directly address the following question within this PR: Should AppProps and LayoutProps be the same type, rather than separate? Should AppProps extend LayoutProps?

By making this decision now, we can improve the maintainability of the code base for future changes that require synchronizing these wrapper pattern props. In my opinion, AppProps should extend LayoutProps. Since AppWrapper itself is a Layout, extending LayoutProps allows us to keep a hook for potential additional functionality that may not be available in every Layout.

lino-levan:
I think that medium-term we should move towards deprecating `_app.tsx`. Layouts should be able to do everything that the `_app.tsx` can do now. I'm in favor of them being separate types.

marvinhagemeister:
> I think that medium-term we should move towards deprecating _app.tsx. Layouts should be able to do everything that the _app.tsx can do now. I'm in favor of them being separate types.

I've been thinking about this and initially I was in favor of deprecating `_app.tsx` too. But thinking some more about it you likely want to be able to also overwrite layouts to opt out of inheriting them. That's currently a feature we don't have but if we had, than you kinda likely still want to have all the meta tags etc from `_app.tsx` instead of having to declare everything from scratch again. So `_app.tsx` is somewhat of a special layout.

marvinhagemeister (approved):
Nice!

PR #1453, opened 2023-07-12
chore: add test for middleware error handling

### LINKED ISSUE #586: Error middleware
Hi,
I would like to plug an error middleware. I have this in `routes/_middleware.ts` :

```js
  request: Request,
) {
  try {
    return await ctx.next();
  } catch (error) {
    console.debug(error);
  }
}
```

But when i throw an error from another route, it is not catched here

deer:
closes https://github.com/denoland/fresh/issues/586

Leo did all the hard work; I'm just adding a test to confirm we don't regress this functionality in the future. See my [comment](https://github.com/denoland/fresh/issues/586#issuecomment-1630175078) for details. Happy to add documentation, if we think that's necessary.


marvinhagemeister (approved):
LGTM

PR #1009, opened 2023-02-01
Enable Plugins to create custom routes and middlewares

RoeHH:

PR #1455, opened 2023-07-12
feat(server): plugins can add middlewares and routes

### LINKED ISSUE #516: Extensibility — how can a dependency inject a route in a project's manifest?
Hi! First, thank you all for the great work in fresh and deno. It makes me happy to code again :)

I'm building a live site editor based on fresh and so I would like to configure some features for my framework-powered projects. Think of it as a meta-framework for fresh. 

For example, I have built an utility to inspect DOM elements in the browser and open VSCode directly on correct source/line/column. https://github.com/deco-cx/inspect-vscode

I would like a way to inject the `/inspect-vscode` in a fresh manifest without having to ask the user to copy a file into `routes/inspect-vscode.ts`. My point is that this is not authored by my user and they shouldn't have it "clutter" their repo. 

@lucacasonato any thoughts on this usecase? I'd like a way to let a user "install a lib" and have it inject some routes, islands, etc. Perhaps something a la express `app.use(myFeature())`, in spirit?

deer:
closes https://github.com/denoland/fresh/issues/516
supersedes https://github.com/denoland/fresh/pull/1009

This is basically the same as 1009, except I've added a nice test showing how this is used. It's also updated to the latest changes in Fresh, mainly relevant because https://github.com/denoland/fresh/pull/745 introduced a filter in the form of `const syncPlugins = opts.plugins.filter((p) => p.render);`, so the filter that 1009 does in `render.ts` is no longer necessary.

As a followup I can create a new example page called "Writing a plugin which adds routes" or something like that.


deer:
This is useful if we want to provide a first party GA4 or CORS middleware plugin, which would probably save people a lot of time.

marvinhagemeister:
This looks good to me. Let's add @RoeHH as a co-author of the commit since the changes here are most of his work. This SO answer has an explanation on how to do that in git https://stackoverflow.com/questions/7442112/how-to-attribute-a-single-commit-to-multiple-developers

deer:
Thanks, good tip. I'll do that in future as well.

marvinhagemeister (approved):
LGTM 🎉

PR #1456, opened 2023-07-12
fix: init banner not colored in ansi256 terminals

marvinhagemeister:
This basically uses the same fix as #1377 and #1323 for the init script.

Before:

Screenshot 2023-07-12 at 22 18 29

After:

Screenshot 2023-07-12 at 22 18 38


iuioiua (approved):
LGTM

PR #1459, opened 2023-07-13
feat: add `fresh` lint preset on update

marvinhagemeister:
Running the update script should add the `fresh` linting preset if not present.

### DIFF
--- a/update.ts
+++ b/update.ts
@@ -60,6 +60,20 @@ if (denoJson.importMap) {
   await Deno.remove(IMPORT_MAP_PATH);
 }
 
+// Add fresh lint preset
+if (!denoJson.lint) {
+  denoJson.lint = {};
+}
+if (!denoJson.lint.rules) {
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
LGTM

update.ts
```diff
@@ -60,6 +60,20 @@ if (denoJson.importMap) {
...
+}
+if (!denoJson.lint.rules) {
+  denoJson.lint.rules = {};
+}
+if (!denoJson.lint.rules.tags) {
+  denoJson.lint.rules.tags = [];
+}
+if (!denoJson.lint.rules.tags.includes("fresh")) {
+  denoJson.lint.rules.tags.push("fresh");
+}
```
bartlomieju, line 75: I suggest this should include `recommended` as well

marvinhagemeister: Good idea! Added that too when it isn't already present.

PR #1467, opened 2023-07-17
docs: update twindv1 example to resolve type issue

deer:
The docs currently recommend a `twind.config.ts` like this:
```ts
import { defineConfig } from "https://esm.sh/@twind/core@1.1.3";
import presetTailwind from "https://esm.sh/@twind/preset-tailwind@1.1.4";
import presetAutoprefix from "https://esm.sh/@twind/preset-autoprefix@1.0.7";

export default {
  ...defineConfig({
    presets: [presetTailwind(), presetAutoprefix()],
  }),
  selfURL: import.meta.url,
};
```
But this leads to an unfriendly error in `main.ts` like this:
```
Argument of type '{ selfURL: string; darkMode?: DarkModeConfig; darkColor?: DarkColor<BaseTheme & TailwindTheme & { screens: Record<string, MaybeArray<ScreenValue>>; colors: Record<...>; }> | undefined; ... 7 more ...; finalize: Finalize<...>[]; }' is not assignable to parameter of type 'Options'.
  Types of property 'darkColor' are incompatible.
    Type 'DarkColor<BaseTheme & TailwindTheme & { screens: Record<string, MaybeArray<ScreenValue>>; colors: Record<string, MaybeColorValue>; }> | undefined' is not assignable to type 'DarkColor<BaseTheme> | undefined'.
      Type 'DarkColor<BaseTheme & TailwindTheme & { screens: Record<string, MaybeArray<ScreenValue>>; colors: Record<string, MaybeColorValue>; }>' is not assignable to type 'DarkColor<BaseTheme>'.
        Types of parameters 'context' and 'context' are incompatible.
          Type 'Context<BaseTheme>' is not assignable to type 'Context<BaseTheme & TailwindTheme & { screens: Record<string, MaybeArray<ScreenValue>>; colors: Record<string, MaybeColorValue>; }>'.
            The types returned by 'theme()' are incompatible between these types.
              Type 'BaseTheme' is not assignable to type 'BaseTheme & TailwindTheme & { screens: Record<string, MaybeArray<ScreenValue>>; colors: Record<string, MaybeColorValue>; }'.deno-ts(2345)
```
I don't find this particularly useful, and the issue has been raised a few times in discord. This [discussion](https://github.com/denoland/fresh/discussions/1447) links to a workaround (towards the end of the comment) that makes the red squiggle go away. Simply cast it like this: `presetTailwind() as Preset`.

Given the number of complaints about this, I think it's reasonable to add this to the example until someone finds what the real issue is.


docs/examples/using-twind-v1.md
```diff
@@ -52,11 +52,13 @@ import presetAutoprefix from "https://esm.sh/@twind/preset-autoprefix@1.0.7";
 
 export default {
   ...defineConfig({
-    presets: [presetTailwind(), presetAutoprefix()],
+    presets: [presetTailwind() as Preset, presetAutoprefix()],
```
marvinhagemeister, line 55: Nit, where is the `Preset` type coming from?

deer: the price of not working with real examples in the docs 🤦 
thanks

lino-levan:
Should we perhaps raise an issue upstream? I know twind is essentially a dead project as far as I know but making a PR to fix typing doesn't seem unreasonable. I'd be willing to give it a look at some point in the near future.

deer:
Agree, seems inactive. The last serious commit was on Jan 24th. I suspect not much will come from your PR, if you choose to create it.

Therefore I'm still in favor of updating what we can control here.

lino-levan:
I think this raises a valid concern about relying on a dead library. Perhaps someone could fork it into another project that is perhaps more active?

marvinhagemeister (approved):
LGTM

marvinhagemeister:
Agree, twind seems unfortunately unmaintained and it seems to make sense to look for an alternative. For now I think it's best to keep the docs up to date and switch to something else once we decided on something.

PR #1468, opened 2023-07-17
feat: improve Hero responsiveness

TheYuriG:
The [Hero component example](https://fresh.deno.dev/components) doesn't respond well to smaller display view widths. This PR aims to improve that behavior by forcing a minimum height and enforcing minimum vertical padding.

Additionally, the buttons will start as a column and will extend to a row on the device width of `md` or larger, just like the other components on the Fresh Components page.


marvinhagemeister (approved):
Awesome, thanks a bunch for the PR 🙌

PR #1469, opened 2023-07-17
www: fix spelling error on homepage

mostlyinvisible:
At the bottom of the homepage, the call-to-action prompts the user to `Lean everything you need to know in seconds.`

This PR corrects the spelling so that it reads: `Learn everything you need to know in seconds.`


marvinhagemeister (approved):
Ohh good catch! Thanks for correcting that 💯

PR #1472, opened 2023-07-18
feat(server): plugins can create islands

### LINKED ISSUE #1471: Plugins don't support islands
When I added route and middleware support for plugins, I forgot that islands are handled completely differently. We need to add this functionality in order for plugins to be truly useful. I am working on a fix.

Thoughts?

deer:
closes https://github.com/denoland/fresh/issues/1471

I'm not super excited about the implementation, but it seems to work. ~~Perhaps we shouldn't rush this and let it go into 1.40; there would then be time to make sure it's really implemented how we want.~~

If anyone is really wanting this feature, they can be excited it's already up for review. 😄


deer:
Looks like the island tests are failing on windows. Of course this works on my machine 😅  I will look into this later.

lino-levan (commented):
Could you add a section in the docs?

cbinzer:
Hi, I tested the islands support and it works! Nice work!

src/server/types.ts
```diff
@@ -408,3 +412,9 @@ export interface PluginRoute {
   // deno-lint-ignore no-explicit-any
   handler?: Handler<any, any> | Handlers<any, any>;
 }
+
+export interface PluginIsland {
```
cbinzer: Are name and path necessary? In my perfect world I see something like this:

`
{
  islands: [
    IslandComponent1,
    IslandComponent2
  ]
}
`

Is the location field in the Plugin type also necessary?

marvinhagemeister: Currently the location is necessary to be able to feed it into esbuild. We could explore creating a virtual entry point for these though.

cbinzer: Ok, I understand :-) Maybe it would be better to transform the islands field to an object:

```typescript
{
  ...
  islands: {
    location: '...',
    entries: [
      {
        name: '...',
        path: '...',
        component: MyIslandComponent
      }
    ]
  }
  ...
}
```

Or something similar. With this model it is more clear that location belongs to the islands.

lino-levan: > Currently the location is necessary to be able to feed it into esbuild. We could explore creating a virtual entry point for these though.

I don't think the plugin writer should care about whether we use esbuild under the hood or not. I think @cbinzer's proposed API makes a lot more sense.

Additionally, I imagine this could lead to undebuggable naming conflicts. You'd have to know to look in the source code of a plugin to determine which names you are allowed to use?

deer: Thanks for the feedback. My free time was recently spent trying to get [coverage working in a subprocess](https://discord.com/channels/684898665143206084/1131193353665523762). tl;dr: success. I will get back to this PR tomorrow.

marvinhagemeister: @lino-levan How do you bundle those islands for the browser when all you have is a reference to the function provided by the plugin? That's something I'm having trouble figuring out.

lino-levan: I see the issue. Definitely non-trivial. 

This feels pretty hacky but we could just
```tsx
function test() {
  // contents
}

export const plugin = {
    islands: [test]
}
```
and then just do
```tsx
test.toString()
```
on the other side to extract the actual function definition.

marvinhagemeister: The `island.toString()` only works when it's pure and not accessing variables or identifiers outside of the function body like import statements etc. When I look at the code bases here at Deno, most islands are not pure and access identifiers outside of the function body.

lino-levan: Ah you're right. I don't see a way out of this one. Is there any reason we're even specifying the component at all? It seems like we need to bundle it regardless?

deer: What do we think about this?
```tsx
    islands: {
      location: import.meta.url,
      names: ["./sample_islands/IslandFromPlugin.tsx"],
    },
```

This tries to line it up with how we import things from the manifest. Just a filename and the module.

cbinzer: It is fine for me. I would rename the name field to paths and the location field to baseLocation. That is for me a bit clearer but it is only my opinion.

Many thanks for your work. With this approach we can create cool plugins now and that is also a first good approach to structure big Fresh applications better.

deer: Seems like a reasonable suggestion. I renamed both, given that there were no other comments on the topic.

cbinzer (approved):

deer:
> error: Import 'https://deno.land/std@0.192.0/path/glob.ts' failed: 500 Internal Server Error
    at https://deno.land/std@0.192.0/path/mod.ts:53:15

cbinzer:
Hi, me again :-) I am wondering how to add static files like styles or images to a plugin? Is there a way to specify this in a plugin?

deer:
Should this be closed? The core code hasn't been changed in more than two weeks, but it didn't make it into 1.4. I guess the preferred approach is to do something different, e.g. #1602. Still a bit disappointing to have to wait at least another month until plugins can create islands.

marvinhagemeister:
Oh looks like I misunderstood it. I wasn't sure if you wanted to see this merged since you opened #1602 and wrote this in your initial post "I'm not super excited about the implementation, but it seems to work."

I'm having the next week off and I'll take a look at this PR after that.

deer:
It's that tough balancing act of releasing functionality that will definitely be deprecated in the future. What maintenance burden does releasing this impose, if we know we're going to do something better in the future? But on the other hand, there's a need for this now, and no concrete implementation for the "grand refactor" of projects as plugins.

Fortunately I don't have to make this decision; I can just complain! 😂 

That being said, I do volunteer to fix bugs with this, just like I did with the route/middleware plugin stuff.

Edit: my point about not being super excited is that it's not going to scale to all the other things plugins might want. But it at least seems to solve this one issue.

mcgear:
I could use this feature for sure, as i have a number of component libraries that need to register certain controls as islands...  Are there plans to merge this as a temp solution while we wait for projects as plugins (#1602 )?

deer:
Hi @marvinhagemeister, I hope you had a relaxing vacation last week. Will you be able to look at this PR this week?

Additionally it looks like a change to deno canary is causing `fmt` to look at the format of code within markdown files. It seems like they're not properly formatted. I'll open a PR for this.

#1715 does fix the formatting issues, but fails for some unrelated flaky test issues.

marvinhagemeister (approved):
Nice one! Sorry it took so long to review.

PR #1473, opened 2023-07-18
chore: make test more timing resilient

marvinhagemeister:
This test case is causing flakyness on windows. The previous `delay(100)` is too sensitive to timing differences.

There is a longer backstory to this and something `puppeteer` hasn't addressed in the past years. For context: I worked on a [custom e2e test runner](https://github.com/boxine/pentf) based on `puppeteer` for a client a couple of years back (around 2019-ish). At the time playwright wasn't really a thing so `puppeteer` was the only option.

A common source of problems in e2e testing is that it's not obvious when an element _can_ be safely clicked. Sure, you can always simulate a click on a DOM node, but that doesn't account for the framework potentially not having attached an event listener yet. It's an internal race condition as to when the click is fired. The [cypress blog](https://www.cypress.io/blog/2019/01/22/when-can-the-test-click/) has a good post about the problem. To work around that I've added a `clickWhenListenerReady` helper function that waits until Preact has added a listener before firing the click.

Then there is a very related race condition in that firing the click doesn't mean that the changes have been immediately applied to the DOM. The framework (here Preact) might do some scheduling or maybe something inside puppeteer or even Chrome took a bit longer. It's hard to pinpoint where exactly the issue is, but this leads to cases where the click has fired but the new value has not been written to the DOM yet. A much more reliable way is to instead wait until the expected string is present in the DOM.

I did a couple of local runs on windows and with the changes in this PR I get way less flukes.

### DIFF
--- a/tests/cli_test.ts
+++ b/tests/cli_test.ts
@@ -90,7 +90,10 @@ Deno.test({
       );
       await buttonPlus?.click();
 
-      await delay(100);
+      await page.waitForFunction(() => {
+        return document.querySelector("body > div > div > div > p")!
+          .textContent === "4";
+      }, {});
 
       counterValue = await counter?.evaluate((el) => el.textContent);
       assert(counterValue === "4");
@@ -183,7 +186,10 @@ Deno.test({
       );
       await buttonPlus?.click();
 
-      await delay(100);
+      await page.waitForFunction(() => {
+        return document.querySelector("body > div > div > div > p")!
+          .textContent === "4";
+      }, {});
 
       counterValue = await counter?.evaluate((el) => el.textContent);
       assert(counterValue === "4");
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
Well, looks like I celebrated too early. The CI failed again 😅

marvinhagemeister:
Re-ran the CI and all is green again. I had hoped to eliminate any flukes completely, but looks like some are still there. Nonetheless I feel like the changes here are worth merging as the flukes seem to happen less frequently.

bartlomieju (approved):
LGTM - one idea here is that we could think about supporting `flaky` or `retries` attribute on the `Deno.test()` - we do similar thing in some of our Rust tests in `deno` and it helped greatly reduce flakes. That way you could mark tests that they should be tried 3 times and only fail when all 3 tries failed. Happy to discuss it offline.

marvinhagemeister:
Adding support for retries is a great idea. We should look into that 👍

PR #1479, opened 2023-07-18
Added FAQ Page

PhoneDroid:
<br>

> As with any bigger open source project I'm quite 
> concerned about the count of issues / discussions.
>
> While Fresh is still on the 'lower' side, 
> I can easily see it getting worse quickly.
>
> To this end I would like to propose - for one -
> to extract questions and answers from both
> places into a dedicated FAQ and with it,
> close the source tickets / discussions.

### Steps

- [x] Add FAQ route
- [x] Collect questions & answers
- [x] Process and write them out
- [ ] Close source issues / discussions 

### To Be Closed

Discussions that can be closed once the FAQ is merged.

- #190 
- #191
- #203 
- #325 
- #341

<br>

image

<br>

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -14,5 +14,8 @@
   },
   "test": {
     "exclude": ["www/"]
+  },
+  "fmt": {
+    "exclude": ["docs/faq.md"]
   }
 }
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,6 +1,6 @@
 ---
 description: |
-    We have colleted various questions you asked 
+    We have collected various questions you asked 
     and combined them with our answers this FAQ.
 ---
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,0 +1,6 @@
+---
+description: |
+    We have colleted various questions you asked 
+    and combined them with our answers this FAQ.
+---
+
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -4,19 +4,10 @@ description: |
     and combined them with our answers this FAQ.
 ---
 
-
 [Contributing]: https://deno.land/manual@v1.35.1/references/contributing#submitting-a-pr-to-fresh
 [Discussions]: https://github.com/denoland/fresh/discussions
 [Roadmap]: https://github.com/denoland/fresh/issues/563
 [SaaS]: https://deno.com/saaskit
-
-<!-- 
-https://github.com/denoland/fresh/discussions/190
-https://github.com/denoland/fresh/discussions/191
-https://github.com/denoland/fresh/discussions/203
-https://github.com/denoland/fresh/discussions/325
-https://github.com/denoland/fresh/discussions/341
- -->
 
 <br>
 
@@ -66,6 +57,4 @@ Not yet, it's on the **[Roadmap]**, however not as a top priority.
 #### Is there a **SaaS Template**
 
 Deno has a whole **[Page][SaaS]** dedicate to it.
-
-
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,6 +2,8 @@
 description: |
     We have collected various questions you asked 
     and combined them with our answers this FAQ.
+title : |
+    FAQ
 ---
 
 [Contributing]: https://deno.land/manual@v1.35.1/references/contributing#submitting-a-pr-to-fresh
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -4,3 +4,68 @@ description: |
     and combined them with our answers this FAQ.
 ---
 
+
+[Contributing]: https://deno.land/manual@v1.35.1/references/contributing#submitting-a-pr-to-fresh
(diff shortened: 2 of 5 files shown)


deno.json
```diff
@@ -13,5 +13,8 @@
   },
   "test": {
     "exclude": ["www/"]
+  },
+  "fmt": {
+    "exclude": ["docs/faq.md"]
```
NuroDev, line 19: Why does this file need to be excluded from formatting?

docs/faq.md
```diff
@@ -0,0 +1,60 @@
+---
+description: |
+    We have colleted various questions you asked
```
NuroDev:
```suggestion
    We have collected various questions you asked 
```

www/routes/faq.tsx
```diff
@@ -0,0 +1,92 @@
...
+        markdown: body,
+        data: attrs ?? {},
+      },
+    } satisfies Data;
+
+    return ctx.render(data);
+  },
+} satisfies Handlers;
+
+function FAQ(props: PageProps<Data>) {
```
NuroDev: Better to keep things consistent with the current codebase formatting.
If applied, this suggestion will also require you to remove the `export default` at the top of the file.
```suggestion
export default function FAQ(props: PageProps<Data>) {
```

www/routes/faq.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const data = {
+      page: {
+        markdown: body,
+        data: attrs ?? {},
+      },
+    } satisfies Data;
+
+    return ctx.render(data);
+  },
+} satisfies Handlers;
```
NuroDev: If applied, this suggestion will also require you to remove the `export { handler }` at the top of the file.
```suggestion
export const handler: Handlers<Data> = {
  async GET(_req, ctx) {
    const url = new URL(`../../docs/faq.md`, import.meta.url);
    const fileContent = await Deno.readTextFile(url);
    const { body, attrs } = frontMatter<Record<string, unknown>>(fileContent);

    const data = {
      page: {
        markdown: body,
        data: attrs ?? {},
      },
    } satisfies Data;

    return ctx.render(data);
  },
};
```

www/routes/faq.tsx
```diff
@@ -0,0 +1,92 @@
...
+  let description = "Fresh Document";
+
+  if (props.data.page.data.description) {
+    description = String(props.data.page.data.description);
+  }
+
+  return (
+    <>
+      <Head>
+        <title>FAQ</title>
```
NuroDev: Would it be better to have this title set in the markdown file? We already do it for the description.

docs/faq.md
```diff
@@ -0,0 +1,60 @@
...
+We would like to in the future, however it will need to be properly  
+coordinated to keep the translations maintained and up-to-date.
+
+<br>
+
+## Features
+
+#### Do you offer **i18** support
+
+Not yet, it's on the **[Roadmap]**, however not as a top priority.
```
marvinhagemeister, line 57: This makes it sound `i18n` wouldn't be possible with Fresh. Sure, we don't have some helpers that make it nicer, but any website can add `i18n` support without helper functions from the framework.

docs/faq.md
```diff
@@ -0,0 +1,60 @@
...
+
+Currently the roadmap is tracked as a **[GitHub Issue][Roadmap]**.
+
+#### Can I talk to the **Core Maintainer**
+
+You can find **Marvin** on the following platforms:
+
+-   Discord **@marvin_h** for quick questions
+
+-   **[GitHub][Discussions]** for longer form discussions
```
marvinhagemeister, line 36: I don't think this is a good idea. Other users can help other users too. Pointing everyone in my direction means I'll be likely bombarded with support requests that others can help with too.

deno.json
PhoneDroid: The autoformatting breaks markdown syntax.
I'll happily take the role of janitor for that file ;3

docs/faq.md
PhoneDroid: How about this instead?
image

www/routes/faq.tsx
PhoneDroid: I'll adjust that.

www/routes/faq.tsx
PhoneDroid: I'll adjust that.

docs/faq.md
PhoneDroid: How about this?
image

www/routes/faq.tsx
PhoneDroid: Sure, I'll add that.

bartlomieju:
Thanks for putting this together! The idea of an FAQ is nice, but the codebase has been completely rewritten for Fresh 2.x since this PR was opened — the www routes, manifest system, and navigation are all different now. The FAQ content and links would also need updating for the current state of the project.

Closing as stale.

PR #1480, opened 2023-07-18
fix: generate only one random Content-Security-Policy nonce per render

jviide:
This pull request modifies the server rendering step to generate only one random Content-Security-Policy (CSP) nonce per each `render(...)` call (when CSP headers are enabled).

Previously a separate nonce was generated during rendering for each module script and the inline script. This was not strictly necessary: [one nonce for every response](https://web.dev/strict-csp/#step-1-decide-if-you-need-a-nonce-or-hash-based-csp) is enough. Having multiple nonces ends up making things (_very theoretically_) easier for an attacker's standpoint, as just one of the multiple nonces has to be guessed to circumvent the CSP.

This pull request also modifies the internal `addImport` helper to skip nonce generation, as those nonces didn't end up being used anywhere.

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -309,28 +309,27 @@ export async function render<Data>(
 
   bodyHtml = bodyHtml as unknown as string;
 
-  const moduleScripts: [string, string][] = [];
-
-  for (const url of opts.imports) {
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Awesome, thanks a bunch for the PR ❤️

PR #1482, opened 2023-07-19
fix: current version not updated in update check

marvinhagemeister:
Discovered an issue with the update check logic which caused the current version never to be updated once it was written into the cache file.

@lino-levan I think this is the same issue you were seeing yesterday.

### DIFF
--- a/src/dev/deps.ts
+++ b/src/dev/deps.ts
@@ -11,7 +11,7 @@ export {
   SEP,
   toFileUrl,
 } from "https://deno.land/std@0.193.0/path/mod.ts";
-export { DAY } from "https://deno.land/std@0.193.0/datetime/constants.ts";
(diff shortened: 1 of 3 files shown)


lino-levan:
Ah okay, given no other reports I thought I might just have another broken install 😅. Let me take a look.

lino-levan:
On this branch:
image

Nice! Looks like the fix works. I still don't understand *where* it's reading the fresh version from if not from the import map, but it seems to work well now. LGTM.

marvinhagemeister:
Fresh creates a very small JSON file to keep track of when it last did the update check in the user's home dir. The default current version was read from `version.json`, but immediately after that we read the cache file which overwrote that version with the one in the cache file.

bartlomieju (approved):

PR #1483, opened 2023-07-19
fix: Prevent deserialize break on safari < 15.4

igorbrasileiro:
## The problem
The `Array.prototype.at` is only available at safari >= 15.4 [caniuse](https://caniuse.com/?search=Array.prototype.at)
image. So, the hydration breaks in older browsers.

Preview: https://deco-sites-fashion-nfcj5n1zeycg.deno.dev/

## The solution
This PR changes the `.at(-1)` to `[array.length - 1]`.

### DIFF
--- a/src/runtime/deserializer.ts
+++ b/src/runtime/deserializer.ts
@@ -56,7 +56,7 @@ export function deserialize(
         (o, k) => k === null ? o : o[k],
         v,
       );
-      parent[refPath.slice(-1)[0]!] = target;
+      parent[refPath[refPath.length - 1]!] = target;
     }
   }
   return v;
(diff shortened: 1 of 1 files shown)


src/runtime/deserializer.ts
```diff
@@ -56,7 +56,7 @@ export function deserialize(
         (o, k) => k === null ? o : o[k],
         v,
       );
-      parent[refPath.at(-1)!] = target;
+      parent[refPath.slice(-1)[0]!] = target;
```
marvinhagemeister: Cloning the array just to get the last value seems a bit much. Let's skip the temporary array allocation here by accessing the last item directly.

```suggestion
      parent[refPath[refParent.length - 1]!] = target;
```

igorbrasileiro: Oh, sure! Thank you =D

marvinhagemeister (approved):
Awesome, thanks a bunch for taking the time to make this PR ❤️

Issue #1125, opened 2023-04-20
Should the target of libraries loaded from islands be es2022 by default?


ayame113:
By default, esm.sh checks the User-Agent header to determine the build target. ([src](https://esm.sh/#docs:~:text=in%20development%20mode.-,ESBuild%20Options,-By%20default%2C%20esm))

So if people import esm.sh from islands, the bundle target will be Deno. (ref: #988)
Maybe a proper User-Agent needs to be specified when bundling with esbuild? Then people shouldn't have to manually specify `?target=es2022` in their dependency URLs.

adamgreg:
I'd be very interested in any solution to this. I'm new to Deno, and trying to import a component in an Island from esm.sh. It seems that even if I add `?target=es2022` to the import URL the browser ends up importing dependencies of the imported module from `deno` URLs instead of `es2022` ones.

EDIT: I got there in the end, but only once I'd added every dependency (and dependency of a dependency, etc.) to my import map using esm.sh URLs with `?target=es2022` and the `*` to make all dependencies external, as I couldn't find a way to make the `target` option propagate through the import graph.

adamgreg:
I'd be very interested in any solution to this. I'm new to Fresh, and trying to use a component library with lots of dependencies in my islands. I eventually got it working, but only once I'd added every dependency (and dependency of a dependency, etc.) to my import map using esm.sh URLs with `?target=es2022` and the `*` to make all dependencies external, as I couldn't find a way to make the target option propagate through the import graph.

lucacasonato:
`npm:` specifiers will mean native NPM bundling rather than using esm.sh for this. Let's wait for that :)

ayame113:
Can this be reconsidered?

For me, it's a bit annoying to ask the library author "sorry, please add `?target=es2020` to the import statements in your library" every time you run into a problem.

- https://github.com/denoland/fresh/pull/1490/files
- https://github.com/denoland/saaskit/pull/357/files
- https://github.com/denoland/deno-gfm/pull/35/files

marvinhagemeister:
Good call. Maybe we can solve this by adding an esbuild plugin that appends the target parameter during resolution automatically.

marvinhagemeister:
Since Fresh 1.5 you can set the esbuild target option in the Fresh configuration via `build.target`. This allows you to down transpile the code for the specific browser requirements of the project.

adamgreg:
[...]

PR #1490, opened 2023-07-19
fix(search): add explicit target for search button deps

### LINKED ISSUE #1441: Documentation search input throws setTimeout error
Currently if you try to search for anything on the documentation site/page(s) you will receive the following error in the console upon every key press:
```
Uncaught TypeError: o.environment.setTimeout is not a function
```

Seeing as the documentation primarily just renders a button to mount the Algolia search modal I presume this is some kind of issue with their script getting mounted?
The current `SearchButton` component [is pinned to version 3](https://github.com/denoland/fresh/blob/main/www/islands/SearchButton.tsx#L3C24-L3C52) but it seems the latest version is `3.5.1` so maybe a dependency upgrade could fix this?

mbhrznr:
closes #1441 (hopefully 🤞 ).

following the discussion in the aforementioned issue, it just might be the case that esm.sh is spitting out the wrong target for our island.

this has been also been a re-appearing issue on `fresh_charts` and `saaskit` when introducing a client-side rendered `chart.js` chart.

for further reference there's also the following issue:
https://github.com/denoland/fresh/issues/1125

### DIFF
--- a/www/islands/SearchButton.tsx
+++ b/www/islands/SearchButton.tsx
@@ -1,6 +1,6 @@
 import { Head } from "$fresh/runtime.ts";
 import { useEffect, useRef } from "preact/hooks";
-import docsearch from "https://esm.sh/@docsearch/js@3";
+import docsearch from "https://esm.sh/@docsearch/js@3?target=es2020";
 
 export default function SearchButton(props: { class?: string }) {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Awesome, thank you for looking into this ❤️

PR #1492, opened 2023-07-20
chore: don't run tests in parallel on CI

marvinhagemeister:
Trying out if the windows CI gets more stable when we don't run our tests in parallel.


marvinhagemeister:
Well looks like it does solve the issue.

marvinhagemeister:
Merging, because it unblocks the other PRs where the windows CI is failing pretty often.

Issue #1491, opened 2023-07-19
Trouble trying to use react libraries in client islands even with aliases


iivvaannxx:
## The problem
I can understand this is still quite a "new" framework, but I have to say that is a little bit painful trying to integrate some external React libraries with Fresh. Especially when we don't have any piece of documentation on how to do it properly anywhere, at least outside these issues, where everyone seems to have a different working solutions regarding the parameters used in `esm.sh` import URLs. It would be certainly helpful to have it clearly documented somewhere, even if it's to say that most packages won't work. 

 ## My attempts

I tried to integrate 4 different libraries without success, first, it was MUI and then Headless UI, even though I was able to bring some components into my app they wouldn't work as intended or throw some weird errors.

Recently I tried to import 2 different libraries to work with forms, the first one being [Modular Forms](https://modularforms.dev/) (which has a preact compatible package) and the second being [react-hook-form](https://react-hook-form.com/). For some reason even though I use each and every component or function of these packages in islands exclusively, the JS seems not to work, maybe for some hydration issue?

To be clear, with Modular Forms for some reason my island was being rendered twice, and the error checking/value updating performed by the `useForm` hook provided by the package wasn't working at all. Then with React Hook Form, even though the values were updating correctly, the submit handlers didn't work, and neither did the error checking with Zod (integrated).

For Modular Forms I had this import map (which was added through `deno task esm:add @modular-forms/preact`

```json
"@modular-forms/preact": "https://esm.sh/v128/*@modular-forms/preact@0.6.0"
```

Which I really was looking forward to get it working, but I couldn't, even without the need of aliasing react dependencies.

Then for React Hook Form I had this (also added through `deno task esm:add react-hook-form`)

```json
"react-hook-form": "https://esm.sh/v128/*react-hook-form@7.45.2?alias=react:preact/compat&external=react"
```
I don't know if I misread something in [esm.sh](https://esm.sh) about how to alias dependencies or I misunderstood how to do it from different issues here but it would certainly be helpful to have a little section in the docs showing how to do it. I may be doing something wrong myself, which could perfectly be the problem, but after trying with many different libraries I am starting to think that is some error during the hydration process.

[...]

PR #1495, opened 2023-07-20
Fix island ordering bug

marvinhagemeister:
Rendering an island that was instantiated inside an island, but was rendered outside of said island in another component, triggered an edge case which would lead to duplicated HTML. We were missing a check for when an island ends up being nested in another one to keep the current vnode stack in sync.

Fixes the duplicate HTML issue mentioned in #1491 .

Before:

Screenshot 2023-07-20 at 21 02 41

After:

Screenshot 2023-07-20 at 21 02 57

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -305,7 +305,14 @@ function _walkInner(
         _walkInner(islands, props, markerStack, vnodeStack, sib.firstChild);
       }
 
-      if (marker !== null && marker.kind === MarkerKind.Slot) {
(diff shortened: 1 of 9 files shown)


marvinhagemeister:
Added a commit that fixes a very related issue where the ordering got wrong in Preact `10.16.0`. It has to do with it calling `insertBefore` instead of `appendChild` like in Preact `10.15.1`. That uncovered a bug in our `insertBefore` shim.

marvinhagemeister:
Merging myself, since both users of the two related issues confirmed that the changes resolve it. The changes are harmless too and have tests in place as well.

PR #1499, opened 2023-07-21
fix: Only show update message if current < latest

marvinhagemeister:
Noticed that we still had a case where the cache has the old value latest and users may have just upgraded to a newer version than in the cache, that we'd still show the upgrade message. This was because we only checked if `current !== latest`, which doesn't account for the `current > latest`.

Note that this only during the time interval between a new release and the next update check. Once the update check updates the cache we don't run in the situation where `current > latest`. 

To address that I've added a basic semver comparison check.

### DIFF
--- a/src/dev/update_check.ts
+++ b/src/dev/update_check.ts
@@ -36,6 +36,20 @@ function getFreshCacheDir(): string | null {
   return null;
 }
 
+const SEMVER_REG = /^(\d+)\.(\d+)\.(\d+)$/;
+function sortSemver(current: string, latest: string): number {
+  if (current === latest) return 0;
+
+  const currentParts = current.match(SEMVER_REG);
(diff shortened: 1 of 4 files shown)


src/dev/update_check.ts
```diff
@@ -36,6 +36,20 @@ function getFreshCacheDir(): string | null {
   return null;
 }
 
+const SEMVER_REG = /^(\d+)\.(\d+)\.(\d+)$/;
+function sortSemver(current: string, latest: string): number {
```
iuioiua: Perhaps, you could instead use https://deno.land/std@0.195.0/semver/gt.ts?s=gt

marvinhagemeister: Oh nice! Didn't know that we have a semver library in our standard library

bartlomieju (approved):
LGTM!

PR #1500, opened 2023-07-21
fix: middleware state reset between middlewares

### LINKED ISSUE #1497: `ctx.state` is reset between middlewares
### Discussed in https://github.com/denoland/fresh/discussions/1496

<div type='discussions-op-text'>

<sup>Originally posted by **bfkss** July 21, 2023</sup>
I used to assign my state-object directly to `ctx.state` via
```js
ctx.state = {
  foo: 'bar'
}
```
but since 1.2 or 1.3 I can only assign to properties of `ctx.state`, e.g.:
```js
ctx.state.foo = 'bar'
// or
ctx.state.data = {
  foo: 'bar'
}
```

Assigning an object leads to `ctx.state = {}` in the next handler after calling `ctx.next()`.

I haven't seen this documented and am not sure whether it is intended?</div>

marvinhagemeister:
Fixes #1497 .

### DIFF
--- a/tests/main_test.ts
+++ b/tests/main_test.ts
@@ -564,6 +564,21 @@ Deno.test({
 });
 
 Deno.test({
+  name: "/middleware - should pass state through all middlewares",
+  fn: async () => {
+    await withPageName("./tests/fixture/main.ts", async (page, address) => {
+      await page.goto(`${address}/state-middleware/foo`);
+      const text = await page.$eval("pre", (el) => el.textContent);
+      assertEquals(JSON.parse(text), {
+        handler1: "it works",
+        handler2: "it works",
+        handler3: "it works",
+      });
+    });
+  },
+});
+
+Deno.test({
   name: "/middleware - middlewareParams",
   fn: async () => {
     const resp = await handler(
--- a/tests/main_test.ts
+++ b/tests/main_test.ts
@@ -9,7 +9,7 @@ import {
 import manifest from "./fixture/fresh.gen.ts";
 import options from "./fixture/options.ts";
 import { BUILD_ID } from "../src/server/build_id.ts";
-import { startFreshServer, withPageName } from "./test_utils.ts";
+import { parseHtml, startFreshServer, withPageName } from "./test_utils.ts";
 import { assertMatch } from "https://deno.land/std@0.193.0/testing/asserts.ts";
 
 const ctx = await ServerContext.fromManifest(manifest, options);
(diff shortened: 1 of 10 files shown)


tests/main_test.ts
```diff
@@ -563,6 +563,21 @@ Deno.test({
...
+      const text = await page.$eval("pre", (el) => el.textContent);
+      assertEquals(JSON.parse(text), {
+        handler1: "it works",
+        handler2: "it works",
+        handler3: "it works",
+      });
+    });
+  },
+});
+
```
deer: I like avoiding `withPageName` since using `handler` makes it way easier to debug. I think we also need to assert that `root_mw` isn't included, since that's overridden when you directly assign state with `ctx.state = { handler1: "it works" }`.
```suggestion
Deno.test({
  name: "/middleware - should pass state through all middlewares",
  fn: async () => {
    const resp = await handler(
      new Request("https://fresh.deno.dev/state-middleware/foo"),
    );
    assert(resp);
    assertEquals(resp.status, Status.OK);
    const body = await resp.text();
    assert(
      !body.includes("root_mw"),
      "shouldn't include root_mw, because we override it at the first layer",
    );
    assertStringIncludes(
      body,
      "&quot;handler1&quot;: &quot;it works&quot;,\n  &quot;handler2&quot;: &quot;it works&quot;,\n  &quot;handler3&quot;: &quot;it works&quot;",
    );
  },
});
```

deer (changes requested):
Lovely fix! I'm surprised there weren't tests for this already.

Perhaps you should add a small note in the documentation? Right after ```The `ctx` also has a `state` property that can
be used to pass arbitrary data to downstream (or upstream) handlers.``` you could add:
``` `State` is normally set by modifying its properties, e.g. `ctx.state.loggedIn = true`, but you can also modify the entire object like `ctx.state = { loggedIn = true }`.``` (Or something like that. Just to let people know that both options are available.)

File: tests/main_test.ts (hunk 1 above)
Comment (iuioiua) [reply tests/main_test.ts]: +1

**Review by iuioiua — APPROVED** [review]
LGTM

File: tests/main_test.ts (hunk 1 above)
Comment (marvinhagemeister) [reply tests/main_test.ts]: Good point about using the handler way to test things.

FYI: `assertEquals` already tests for an exact match of keys and would fail if it included `"root_mw"`.

File: docs/concepts/middleware.md [hunk 2]
```diff
@@ -13,7 +13,9 @@ used to trigger child handlers. The `ctx` also has a `state` property that can
 be used to pass arbitrary data to downstream (or upstream) handlers. This
 `state` is included in `PageProps` by default, which is available to both the
 special [\_app](/docs/concepts/app-wrapper.md) wrapper and normal
-[routes](/docs/concepts/routes.md).
+[routes](/docs/concepts/routes.md). `ctx.state` is normally set by modifying its
+properties, e.g. `ctx.state.loggedIn = true`, but you can also replace the
+entire object like `ctx.state = { loggedIn = true }`.
```
deer: Canary docs? This won't work until the next release.

marvinhagemeister: whoops, I messed up my own system. You're absolutely right 👍

deer (approved):

PR #1502, opened 2023-07-22
www: use smart quotes

rojvv:


marvinhagemeister (approved):
Good idea! Thanks for making a PR 👍

PR #1506, opened 2023-07-22
Route layout

mcgear:
This is an initial implementation of a _layout file implementation that allows for applying layouts to routes. It supports cascading _layout files in route sub directories and treats them all as optional.

There are a good number of tests that were broken when I forked, so could use some pointers on where to add additional tests for this.


mcgear:
I was able to make some additions to the fixture.  So if you run deno task fixture, you should be able to navigate to the following paths to see it in action:
/route-layout
/route-layout/sub
/route-layout/skip
/route-layout/skip/sub

Those were some of the use cases i was checking out. You can see that it has no impact on existing pages:
/
/islands
/books

mcgear:
My bad on that, forgot to rerun deno fmt after adding the fixture tests. Hopefully that resolves what i saw in the workflow error.

deer:
I haven't looked at this in detail yet, but just noting that we have a similar PR: https://github.com/denoland/fresh/pull/870. If we go with this one, then we should close 870.

docs/concepts/route-layout.md
```diff
@@ -0,0 +1,28 @@
...
+```tsx
+// routes/sub/_layout.tsx
+
+import { LayoutProps } from "$fresh/server.ts";
+
+export default function Layout({ Component, state }: LayoutProps) {
+  //do something with state here
+  return (
+    <div class="layout">
+      <Component />
```
NuroDev: I guess the aim of this here is to match the `_app.tsx` format of a `Component` being passed as a prop & that be the slot point to hand off to the child component(s).

However is there any argument or reason why something more like [Next.js' style ](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#layouts)of just having a `children` prop instead? The only thought I can think for the former being thinking of prop drilling but I don't think that really applies here does it?

mcgear: The `children` prop feels a little wrong to me in this case. I look at the `children` prop as a tool for use when creating a custom component (server/island). Its a property of the custom component, and signifies where the "inner" contents of the element go `<CustomComponent>{...children}</CustomComponent>`.  

In the case of a Page, template (_app), or layout (_layout) there isn't actually "children" of the component, but rather the rendering of a component based on data that comes from outside of the properties. For this reason, passing in the `Component` feels like the right API as the children are unknown, but the inner Component is known.

marvinhagemeister: +1 on going with `children`. That's the way [nextjs does it too](https://nextjs.org/docs/pages/building-your-application/routing/pages-and-layouts) and to me it seems worth it to be familiar to users coming from other frameworks.

I don't think there is a strong distinction from Preact's point of view between a layout, a page or a component. For Preact all of these are standard components.

mcgear: I'm not as familiar with NextJS, so can't speak much to that. However, i do think that going with `children` is semantically different than the current framework approach to an Application wrapper.  As a user, it would feel awkward to me to use injected `<Component />` in the app wrapper and then to use `children` in a route layout.

So then the question is, would we want to change App wrapper to use children as well? So we could in theory be inline with those migrating from NextJS... or do we keep Route Layouts inline with the existing App wrapper implementation.  My +1 is towards a similar API to what we already have in App wrapper.

marvinhagemeister: That's a good point. I hadn't thought about there being `Component` already in `_app.tsx` and it makes sense to stay consistent. For me that resolves the discussion. Let's go with `Component`

lino-levan: Seems a little off-topic but I'm just thinking. Does it make sense to keep the `_app.tsx` with layouts being introduced? I don't think it does anything differently. I'm +1 for deprecating it (with a nice warning message telling users to move to layout). My understanding was that the `_app.tsx` was always going to be removed but was just a temporary measure until layouts get introduced.

If we do decide to deprecate `_app.tsx`, I'm strongly for using the `children` prop over an arbitrary `Component` since there is no use in being consistent with a deprecated utility.  If Remix has taught us anything over the years, it's to stop using non-standard components when standards exist.

marvinhagemeister: @lino-levan I think remix use an `<Outlet />` component which is kinda similar to our `<Component />` for that.

Yeah with layouts having `_app.tsx` is a kinda the same thing and becomes unnecessary. We should consider nudging people to a top level `_layout.tsx` file instead. I'd still only soft deprecate `_app.tsx` though by printing a warning or something. That way the upgrade process is easier and we don't have to do a breaking change. This imo should be done in a separate PR though.

lino-levan: A separate PR makes sense to me. Printing a message for a few updates of fresh stating that it's deprecated and will be removed "at some point in the future" is probably good enough.

marvinhagemeister: To close this discussion: We'll go with `<Component />` for now and we can easily add support for `props.children` on top of it later.

marvinhagemeister (approved):
Took the liberty of adding a couple of tests. Thank you so much for adding this to Fresh ❤️

bendik:
How should i go about toggling routeConfig wrapper layouts based on a header? 'hx-request' in my case.

marvinhagemeister:
@bendik Can you please open a new issue instead of commenting on merged PRs?

PR #1507, opened 2023-07-22
Accessbility tweaks and fixes for docs/www

timharek:
First let me say thank you so much for Deno Fresh – it's a trill to use and learn!

With this PR I want to give back with some of my knowledge.

This PR focuses on accessibilty fixes that makes a screen reader read redundant stuff which can make it harder to understand what is being read.

To summarize:
- Added `aria-hidden="true"` to a bunch of decorative icons.
- Some text had too low contrast.
- Replaced the container for `<LemonDrop />` to be a `<div>` instead of `<section>`, as a section should have a heading and this element is purely a decorative feature ([from w3c](https://www.w3.org/TR/2014/PR-html5-20140916/sections.html#the-section-element)).

If there are any questions, suggestions or anything, please let me know :)


marvinhagemeister (approved):
This is fantastic! Thank you so much for the PR 💯

PR #1508, opened 2023-07-23
feat(server): plugin type-safety

Jabolol:
## Overview

This pull request aims to enhance type-safety for the `Plugin` type by introducing the option to pass an array of used `keys` as a type parameter, thereby enabling better checks for plugin configuration.

When creating Plugins, developers might unintentionally misspell an entrypoint import specifiers key, resulting in a 404 error. With this update, users will be able to explicitly specify those keys they intend to use when creating a plugin. This addition ensures that any potential mistakes or typos are caught.

## API changes

The new way of creating a plugin would be as follows:

> **Note:** omitting the type parameter fallbacks to the default behavior with no type-safety
```ts
import { Plugin } from "$fresh/server.ts";

export const info: Plugin<["log"]> = {
  name: "info",
  entrypoints: {
    log:
      `data:application/javascript,export default function e(n){ console.log(n + 1) }`,
  },
  render: (ctx) => {
    ctx.render();
    return {
      scripts: [{
        entrypoint: "log",
        state: 1,
      }],
    };
  },
};
```

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -306,7 +306,7 @@ export interface Island {
 
 // --- PLUGINS ---
 
-export interface Plugin {
+export interface Plugin<T extends string[] = string[]> {
   /** The name of the plugin. Must be snake-case. */
   name: string;
 
@@ -314,7 +314,7 @@ export interface Plugin {
    * declared here can later be used in the "scripts" option of
    * `PluginRenderResult` to load the entrypoint's code on the client.
    */
-  entrypoints?: Record<string, string>;
+  entrypoints?: Record<T[number], string>;
 
   /** The render hook is called on the server every time some JSX needs to
    * be turned into HTML. The render hook needs to call the `ctx.render`
@@ -323,7 +323,7 @@ export interface Plugin {
    * The hook can return a `PluginRenderResult` object that can do things like
    * inject CSS into the page, or load additional JS files on the client.
    */
-  render?(ctx: PluginRenderContext): PluginRenderResult;
+  render?(ctx: PluginRenderContext): PluginRenderResult<T[number]>;
 
   /** The asynchronous render hook is called on the server every time some
    * JSX needs to be turned into HTML, wrapped around all synchronous render
@@ -336,7 +336,9 @@ export interface Plugin {
    * unlike the synchronous render hook, you can not use global variables to
    * propagate state between the render hook and the renderer.
    */
-  renderAsync?(ctx: PluginAsyncRenderContext): Promise<PluginRenderResult>;
+  renderAsync?(
+    ctx: PluginAsyncRenderContext,
+  ): Promise<PluginRenderResult<T[number]>>;
 
   routes?: PluginRoute[];
 
@@ -351,11 +353,11 @@ export interface PluginAsyncRenderContext {
   renderAsync: PluginAsyncRenderFunction;
 }
 
-export interface PluginRenderResult {
+export interface PluginRenderResult<T = string> {
   /** CSS styles to be injected into the page. */
   styles?: PluginRenderStyleTag[];
   /** JS scripts to ship to the client. */
-  scripts?: PluginRenderScripts[];
+  scripts?: PluginRenderScripts<T>[];
 }
 
 export interface PluginRenderStyleTag {
@@ -364,14 +366,14 @@ export interface PluginRenderStyleTag {
   id?: string;
 }
 
-export interface PluginRenderScripts {
+export interface PluginRenderScripts<T = string> {
   /** The "key" of the entrypoint (as specified in `Plugin.entrypoints`) for the
    * script that should be loaded. The script must be an ES module that exports
    * a default function.
    *
    * The default function is invoked with the `state` argument specified below.
    */
-  entrypoint: string;
+  entrypoint: T;
   /** The state argument that is passed to the default export invocation of the
    * entrypoint's default export. The state must be JSON-serializable.
    */


deer:
Just a heads up that this will conflict with my suggested fix for a different "plugin typing" issue: https://github.com/denoland/fresh/issues/1531#issuecomment-1652953414. Maybe yours will get merged first, since I haven't even created a PR yet!

Btw, thanks for the nice article about plugins: https://javi-blog.deno.dev/fresh_plugins. FYI @bjesuiter

I will play with this PR later today, in order to more fully understand how this is all working.

Jabolol:
> Btw, thanks for the nice article about plugins: https://javi-blog.deno.dev/fresh_plugins. FYI @bjesuiter

You're welcome! Happy to help.

> I will play with this PR later today, in order to more fully understand how this is all working.

It's a fairly easy addition in order to type the `entrypoint` keys and avoid errors, I'm open to suggestions.

deer:
Ok, the idea seems good, but somehow I can't get it working. I tried applying this to the three plugins that use entrypoints in the codebase:
1) `js-inject-plugin.ts` this was the only one I could get it working for
2) `twind.ts`
3) `twindv1.ts`

Maybe you could update these to make use of the new feature so there's always an example of how it works? And you could consider adding a dummy plugin that doesn't work, and then use `// @ts-expect-error` to assert the error. Or a new plugin that has multiple entrypoints.

Jabolol:
The plugin itself would need some tweaking to match the new implementation due to the way this plugin is done. I'll use the one at [`twind.ts`](https://github.com/denoland/fresh/blob/main/plugins/twind.ts) as an example.

> Ok, the idea seems good, but somehow I can't get it working.

TypeScript complains because the [`scripts`](https://github.com/denoland/fresh/blob/efbff06fcaf30e8a5600d9fd299b75ed590c31b8/plugins/twind.ts#L23) variable is set to `any[]`, and when adding scripts [`here`](https://github.com/denoland/fresh/blob/efbff06fcaf30e8a5600d9fd299b75ed590c31b8/plugins/twind.ts#L40), the type of the `entrypoint` is `string` instead of the one that we want because we are just pushing an object.

The required change would be to annotate the expected `entrypoints`:
```ts
export default function twind(options: Options): Plugin<["main"]> {
```

There are two ways to do this, the `easy` way and the `dynamic` way.

## Easy way

Change the [`scripts`](https://github.com/denoland/fresh/blob/efbff06fcaf30e8a5600d9fd299b75ed590c31b8/plugins/twind.ts#L23) variable to match the expected type

```ts
const scripts: PluginRenderScripts<"main">[] = [];
```

### Caveats

The maintainers need to manually update the return type and the `scripts` variable so they match whenever they want to add or remove an `entrypoint`.


## Dynamic way

### Approach One: extract union from `Plugin`

The main idea behind is to build an `union` with all the types specified in the `Plugin` type parameter.

First, we extract the `entrypoints` array from the generic type
```ts
type ExtractEntrypoint<PluginType> = PluginType extends
      Plugin<infer Entry> ? Entry : never;
```
Second, we make the union type from the `entrypoints` array
```ts
type MakeUnion<Types extends unknown[]> = Types extends
      [infer First, ...infer Rest] ? First | MakeUnion<Rest>
      : Types extends [infer Only] ? Only
      : never;
```
Third, we build the `union` from the array supplied as the type parameter
```ts
type Entrypoints = MakeUnion<ExtractEntrypoint<ReturnType<typeof twind>>>;
```
Lastly, we type that array that will hold our `strongly` typed scripts.
```ts
const scripts: PluginRenderScripts<Entrypoints>[] = [];
```

### Approach Two: extract union from `renderAsync` return type

The main idea behind is to get the return type of the `twind` function, that is the `Plugin` type, and through the return type of `render` or `renderAsync` get the union of `entrypoint` literals that we can pass as a type parameter to `PluginRenderScripts` to maintain type safety.

In order to get this working dynamically we would need to add more types to help us out. I will extract the type from the `renderAsync` function.

First, one utility type to get the type parameter of a `Promise`
```ts
type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
```
Second, a type that gets the type parameter of a `PluginRenderResult`, whose type parameter is the `entrypoint` as a narrowed `string` literal, that is one of the exact entrypoints specified
```ts
type ExtractEntrypoints<T> = T extends
  PluginRenderResult<infer Entrypoints> ? Entrypoints : never;
```
Third, a type that gets the type of the plugin returned to avoid repetitions.
```ts
type TwindReturnType = ReturnType<typeof twind>;
```
> Since the only difference between the return type of `renderAsync` and `render` is that the former is wrapped in a `Promise`, this type could be inferred from the `render` method, not needing the `UnwrapPromise` type and changing the `renderAsync` references for `render`

Fourth, a type that extracts the return type of the `renderAsync` function and unwraps any `Promise`.
```ts
type UnwrappedRenderAsyncReturn<
  T extends TwindReturnType["renderAsync"],
> = T extends (...args: infer _Args) => infer Return
  ? UnwrapPromise<Return>
  : never;
```
Fifth, a type that defines the type for the result of the `renderAsync` function
```ts
type RenderAsyncResult = UnwrappedRenderAsyncReturn<
  TwindReturnType["renderAsync"]
>;
```
Sixth, we get the `union` of all types specified when creating the `Plugin` through the type parameter
```ts
type EntrypointsType = ExtractEntrypoints<RenderAsyncResult>;
```
Lastly, we type that array that will hold our `strongly` typed scripts.
```ts
const scripts: PluginRenderScripts<EntrypointsType>[] = [];
```

### Caveats

The complexity is way higher, making it more difficult to maintain should a change be needed.

Jabolol:
Regarding the refactor and examples you suggested I could update those too if this PR goes forward. We will have to choose the `easy` or `dynamic` approach for it, though.

Jabolol:
Closed since there's an ongoing plugin discussion at #1602

PR #1510, opened 2023-07-24
fix: async route typing

### LINKED ISSUE #1509: bug: `AsyncRoute` doesn't support passing state type variable
See https://github.com/denoland/saaskit/pull/393#issuecomment-1647142200

iuioiua:
The `ctx.state` type can now be passed in an async route component. I omitted tests, as this is just a matter of typing. Happy to add some if they're deemed needed.

Fixes #1509
Related https://github.com/denoland/saaskit/pull/393#issuecomment-1647142200

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -135,13 +135,15 @@ export interface RouteModule {
   config?: RouteConfig;
 }
 
-export type AsyncRoute<T> = (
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Oh great catch! Thanks for adding this 💯

PR #1511, opened 2023-07-24
fix: remove twind from import map

### LINKED ISSUE #1379: New twind plugin (fresh ^1.2.0) requires you to add `twind/` to your import map
## Reason

The new plugin imports `twind/sheets` which can't be resolved, because the import map is only for this project

https://github.com/denoland/fresh/blob/18bf977052a589cba5a70ecf0da35e0b577e3b0a/.vscode/import_map.json#L3

```
error: Uncaught (in promise) TypeError: Relative import path "twind/sheets" not prefixed with / or ./ or ../ and not in import map from "https://deno.land/x/fresh@1.2.0/plugins/twind.ts"
    at https://deno.land/x/fresh@1.2.0/plugins/twind.ts:1:30
  await import(entrypoint);
  ^
    at async dev (https://deno.land/x/fresh@1.2.0/src/dev/mod.ts:140:3)
    at async file:///some_path/dev.ts:5:1
```

## Proposed fix

Update the import to be a URL instead of something from a import map (as it was in the v1 of the plugin).

https://github.com/denoland/fresh/blob/18bf977052a589cba5a70ecf0da35e0b577e3b0a/plugins/twind.ts#L1

https://github.com/denoland/fresh/blob/18bf977052a589cba5a70ecf0da35e0b577e3b0a/plugins/twindv1.ts#L1

## Workaround

Add this to your fresh projects import map

```json
"twind/": "https://esm.sh/twind@0.16.19/",
```

deer:
closes https://github.com/denoland/fresh/issues/1379

There seems to be a bit of confusion about the "new" twind plugin. `twind.ts` is the old plugin, which is what the issue is complaining about. The actual new plugin is `twindv1.ts` which has a nice line like `import { stringify, virtual } from "https://esm.sh/@twind/core@1.1.3";`. The _old_ plugin should probably also do this. I don't see any reason why we should dirty the import map.

This PR changes the old plugin to use absolute imports. I suppose this could cause problems if someone has changed their import map to use a newer version of twind... But they should probably just use `twindv1.ts` at that point instead. Thoughts?


deer:
Hi @marvinhagemeister, I thought I would take a look at Fresh again because it's been too long. My favorite way of getting back to a project is by dusting off some old PRs. Fortunately some of mine are still floating around.

What do you think about this? I know there's the plan to switch to uno, but that has also been planned for a while. This seems like a small simplification in the mean time.

deer:
Closing due to migration to tailwind.

PR #1515, opened 2023-07-24
fix: allow string based event handlers in SSR

### LINKED ISSUE #1309: Error: Component's "onInvalid" property should be a function
After upgrading from 1.1.6 to 1.2.0.

`<input onInvalid="this.classList.add('foo')" />`

_Error: Component's "onInvalid" property should be a function, but got [string]_

Is it possible to use inline event handlers anymore?

marvinhagemeister:
We don't officially recommend it as string based event handlers are only safe when it's a static string to avoid any form of injections. The fresh linting preset will throw an error if it's not a static string already, but `preact/debug` still throws on these. At runtime we can't really differentiate where a string was coming from, but user's must actively bypass TS warnings AND the linting rule, which to me seems like an adequate guardrail.

Fixes #1309

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -586,21 +586,19 @@ options.vnode = (vnode) => {
         );
       };
     }
-  } else if (typeof vnode.type === "string") {
+  } else if (typeof vnode.type === "string" && vnode.props !== null) {
     // Work around `preact/debug` string event handler error which
     // errors when an event handler gets a string. This makes sense
     // on the client where this is a common vector for XSS. On the
     // server when the string was not created through concatenation
     // it is fine. Internally, `preact/debug` only checks for the
     // lowercase variant.
-    if (vnode.props !== null) {
-      const props = vnode.props as Record<string, unknown>;
-      for (const key in props) {
-        const value = props[key];
-        if (key.startsWith("on") && typeof value === "string") {
-          delete props[key];
-          props["ON" + key.slice(2)] = value;
-        }
+    const props = vnode.props as Record<string, unknown>;
+    for (const key in props) {
+      const value = props[key];
+      if (key.startsWith("on") && typeof value === "string") {
+        delete props[key];
+        props["ON" + key.slice(2)] = value;
       }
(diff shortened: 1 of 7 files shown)


tests/test_utils.ts
```diff
@@ -137,3 +137,24 @@ async function spawnServer(
 
   return { serverProcess, lines, address };
 }
+
+/**
+ * Wait until the passed function returns a truthy value or the time runs out
+ */
+export async function waitFor(
```
iuioiua: Perhaps, you could use https://deno.land/std@0.195.0/async/mod.ts?s=retry

It's default values look like they might suffice here.

src/server/render.ts
```diff
@@ -586,7 +586,25 @@ options.vnode = (vnode) => {
...
+    // server when the string was not created through concatenation
+    // it is fine. Internally, `preact/debug` only checks for the
+    // lowercase variant.
+    if (vnode.props !== null) {
+      const props = vnode.props as Record<string, unknown>;
+      for (const key in props) {
+        const value = props[key];
+        if (key.startsWith("on") && typeof value === "string") {
+          delete props[key];
+          props["ON" + key.slice(2)] = value;
```
```suggestion
          props[key.replace("on", "ON")] = value;
```
Nit: tiny bit clearer what's going on

src/server/render.ts
```diff
@@ -586,7 +586,25 @@ options.vnode = (vnode) => {
         );
       };
     }
+  } else if (typeof vnode.type === "string") {
```
```suggestion
  } else if (typeof vnode.type === "string" && vnode.props !== null) {
```
Nit

tests/test_utils.ts
marvinhagemeister: Good point, I really should look more as to what's available in the standard library.

src/server/render.ts
marvinhagemeister: -1 I always get confused if `.replace()` replaces only the first occurrence or more.

iuioiua (approved):

PR #1518, opened 2023-07-24
fix: parse multiple route params in same part

### LINKED ISSUE #1516: Not found route with extension in the file name
The following file:

```
routes/catalog/movie/[id].json.ts
```

Will not match the path "/catalog/movie/foo.json".

I traced it down to [this branch](https://github.com/denoland/fresh/blob/e18a60851503b4b24956fbf889338a4db7d6f3d5/src/server/context.ts#L1014-L1016), and it won't replace `[id].json` with `:id.json`

marvinhagemeister:
Our existing way to parse file paths to patterns for routes was a bit rudimentary. It only allowed parameters to be the full path segments:

Before:

- `foo/[bar]@foo` -> `foo/[bar]@foo`

After:

- `foo/[bar]@foo` -> `foo/:bar@foo`

Fixes #1516

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -1013,6 +1013,14 @@ export function pathToPattern(path: string): string {
         return `:${part.slice(4, part.length - 1)}*`;
       }
 
+      // Disallow neighbouring params like `/[id][bar].tsx` because
+      // it's ambigious where the `id` param ends and `bar` begins.
+      if (part.includes("][")) {
+        throw new SyntaxError(
+          `Invalid route pattern: "${path}". A parameter cannot be followed by another parameter without any charactes in between.`,
+        );
+      }
+
       // Case: /[id].tsx
       // Case: /[id]@[bar].tsx
       // Case: /[id]-asdf.tsx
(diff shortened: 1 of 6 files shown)


src/server/context.ts
```diff
@@ -1001,20 +1001,47 @@ function getPriority(part: string) {
...
+      for (let i = 0; i < part.length; i++) {
+        const char = part[i];
+        if (char === "[") {
+          // Disallow neighbouring params like `/[id][bar].tsx` because
+          // it's ambigious where the `id` param ends and `bar` begins.
+          if (i > 0 && part[i - 1] === "]") {
+            throw new SyntaxError(
+              `Invalid route pattern: "${path}". A parameter cannot be followed by another parameter without any charactes in between.`,
+            );
+          }
```
iuioiua: This case can be caught earlier in the callback function via `if (part.includes("][") ...`

marvinhagemeister: updated the PR.

marvinhagemeister: updated the PR.

bartlomieju (approved):
LGTM

iuioiua (approved):

PR #1520, opened 2023-07-25
add basic Dockerfile support to project generator

wsb1994:
Will add a basic functioning docker file in the base directory of the generated project.


lino-levan:
I'm pretty strongly of the opinion that this should be behind a flag of some sort. This feels like an extra configuration file that a lot of people will never use.

init.ts
```diff
@@ -109,6 +109,28 @@ await Deno.writeTextFile(
   GITIGNORE,
 );
 
+const DOCKERFILE_TEXT = `
+FROM denoland/deno:1.35.0
```
lino-levan: This should match the version of Deno that is currently running.

wsb1994: I'll take care of this and re-ping, already know how to take care of it 👍

lino-levan:
Looks better to me. I still think this should be behind a `--docker` flag. Could you try to implement that?

wsb1994:
> Looks better to me. I still think this should be behind a `--docker` flag. Could you try to implement that?

I don't mind at all.

wsb1994:
> Looks better to me. I still think this should be behind a `--docker` flag. Could you try to implement that?

would something like "would you like to generate a docker file? y/n" be more appropriate for consistency?

marvinhagemeister:
This is a great idea, thanks for making a PR! I agree that putting this behind a `--docker` flag is the way to go. As great as docker is, it's something you usually come across at a later stage in your career. If you're completely new to web development and want to build a website then you don't know what docker is. Because of that it feels like adding a prompt to the init command adds unnecessary friction for newer users.

So lets keep it behind `--docker`. That way advanced users can benefit from Fresh generating the Dockerfile and newer users are not confronted with a choice they might not know about. Happy to merge it once the prompt is removed.

init.ts
```diff
@@ -88,6 +92,10 @@ const useVSCode = flags.vscode === null
   ? confirm(USE_VSCODE_MESSAGE)
   : flags.vscode;
 
+const useDocker = flags.docker === null
+  ? confirm(USE_DOCKER_MESSAGE)
+  : flags.docker;
```
lino-levan: I don't think we should prompt users here. Completely agree with Marvin here.

lino-levan (changes requested):

init.ts
wsb1994: sure thing 🌞

lino-levan (approved):
LGTM. Nice work!

marvinhagemeister (approved):
LGTM, thanks for adding this 💯

osddeitf:
@wsb1994 Something wrong here:
```ts
const flags = parse(Deno.args, {
  boolean: ["force", "twind", "vscode, docker"],
  default: { "force": null, "twind": null, "vscode": null, "docker": null },
});
```
It would be `"vscode", "docker"` instead.

PR #1522, opened 2023-07-25
www: move latest docs into `latest` folder

marvinhagemeister:
This PR moves the markdown files for the latest docs into a `latest` folder, similar to how we already do it for `canary`. This should make it more obvious which files to change.


bartlomieju (approved):
LGTM

JTRNS:
You are probably already aware and working on a fix, but just in case, all the docs seem to return an HTTP 500 error response since this commit

marvinhagemeister:
@JTRNS Thanks for pointing it out. Resolved with #1523

y12studio:
It looks like this PR change broke the remote case of rendering-markdown example.

https://github.com/denoland/fresh/blob/9f97e3423cc57f141a0b5819a1c028e967d49ef3/docs/latest/examples/rendering-markdown.md?plain=1#L31-L38

PR #1525, opened 2023-07-26
fix(init): typo in flags parsing

lino-levan:
I missed this while reviewing #1520. Thanks to @osddeitf for pointing that out!


marvinhagemeister (approved):
Good catch! Missed this myself 👀

PR #1526, opened 2023-07-26
fix: lots of typos

lino-levan:


marvinhagemeister (approved):
Thanks for fixing these. Half of those mistakes are done by me 😬 😅

PR #1532, opened 2023-07-26
feat: add support for view transitions

marvinhagemeister:
This is a very rough draft for natively supporting view transitions in Fresh. It's inspired by the recent [astro announcement](https://astro.build/blog/astro-290/). I quite like the way they handle the CSS generation via attributes.

Keep in mind that this is a very early prototype.

Rough demo:

https://github.com/denoland/fresh/assets/1062408/996eb844-e648-4758-8f23-d2123daca65b

TODO:
- [x] generate CSS bindings on the fly
- [x] Ensure correctness of history entries
- [ ] Only load view transition code when used.
- [ ] tests


marvinhagemeister:
Updated video:

https://github.com/denoland/fresh/assets/1062408/abb659d1-6ad7-4f3a-870a-7c94eab59547

(I know it's not the prettiest animation, it's just to show that it works)

lino-levan:
Excited for this. Definitely a more "fresh-y" way of doing this.

welschmoor:
Dear Santa...

really looking forward to it!

mgenereu:
Do you think this will eventually work with partials?

5hanth:
still on cards?

bartlomieju:
Closing in favor of #3708 which integrates View Transitions into Fresh 2.x's existing partials system. Instead of a parallel navigation system, the new approach wraps partial updates in `document.startViewTransition()` — opt-in via `f-view-transition` attribute. Thanks for the original prototype\!

PR #1534, opened 2023-07-27
fix(server): add type to Plugin so that MiddlewareHandler is typed correctly

### LINKED ISSUE #1531: [Plugins] PluginMiddleware Type isn't  conforms to existing Type ?
I am in need of advice. Please let me know if you have a better implementation and I would be happy to help.

I have been creating middleware for some time and responded to the middleware as it became available to be applied via plugins.

However, I have one question.
Without error, it seems to conform to existing types but not to new types.

Therefore, I had to make a special assertion.

Is this the recommended way to write this?

For example.

```typescript
export function getLoggerHandler(logSender?: RemoteLogSenderFunction): Plugin {
  const handler: MiddlewareHandler<Logger> = async function (
    _req: Request,
    ctx: MiddlewareHandlerContext<Logger>  // <= Logger can be applied.
  ) {
    ctx.state.logger = new LogModule(logSender);
    return await ctx.next();
  };

  return {
    name: "loggerPlugin",
    middlewares: [
      {
        middleware: {
          handler: handler as MiddlewareHandler<Record<"logger", unknown>>, // <= Logger can't be applied. use Record<"logger", unknown>
        },
        path: "/",
      },
    ],
  };
}
```
(https://github.com/Octo8080X/fresh-logger/blob/main/src/loggerPlugin.ts)

`MiddlewareHandler<Record<"logger", unknown>>`  is not applicable. However, `MiddlewareHandlerContext<Logger>` is applicable.

MiddlewareHandler MiddlewareHandlerContext

The generics for both `MiddlewareHandler` and `MiddlewareHandlerContext`, though, is `Record<string, unknown>`,　and `Logger` is compatible.

Is there any other recommended way to write this?

deer:
closes https://github.com/denoland/fresh/issues/1531

So the user's code goes from:
```tsx
export function getLoggerHandler(logSender?: RemoteLogSenderFunction): Plugin {
  const handler: MiddlewareHandler<Logger> = async function (
    _req: Request,
    ctx: MiddlewareHandlerContext<Logger>
  ) {
    ctx.state.logger = new LogModule(logSender);
    return await ctx.next();
  };

  return {
    name: "loggerPlugin",
    middlewares: [
      {
        middleware: {
          handler: handler as MiddlewareHandler<Record<"logger", unknown>>,
        },
        path: "/",
      },
    ],
  };
}
```
to
```tsx
export function getLoggerHandler(logSender?: RemoteLogSenderFunction): Plugin<Logger> {
  const handler: MiddlewareHandler<Logger> = async function (
    _req: Request,
    ctx: MiddlewareHandlerContext<Logger>
  ) {
    ctx.state.logger = new LogModule(logSender);
    return await ctx.next();
  };

  return {
    name: "loggerPlugin",
    middlewares: [
      {
        middleware: {
          handler: handler,
        },
        path: "/",
      },
    ],
  };
}
```

Thoughts?


deer:
Hmmm, the test plugin should probably also use this...

iuioiua (approved):

marvinhagemeister (approved):
Nice 👍

PR #1536, opened 2023-07-28
fix(server): change RouteContext data type parameter default to any

### LINKED ISSUE #1528: async route incompatible with index
Creating `async.tsx` file in `routes` according to https://fresh.deno.dev/docs/concepts/routes#async-route-components results in a cryptic typescript error/warning:

image

Contents of `/routes/async.tsx`:

```javascript
import { RouteContext } from "$fresh/server.ts";

// deno-lint-ignore require-await
export default async function (
    req: Request,
    ctx: RouteContext,
  ) {
    return <div>async</div>;
  }
```

The problem seems to be related to `RouteContext`. If I remove it, the error disappears. Running `deno check` reports no error however. So it might be related to the language server or the vscode extension. I am not sure how to debug it further.

deer:
closes https://github.com/denoland/fresh/issues/1528

I have a large comment in 1528 explaining what's going on here.

During the course of researching this, I found `deno check **/*.ts` to type check all the files in the project. It turns out we have this issue in our code base as well! `tests/fixture_server_components/main.ts` shows this exact error, because `routes/context/[id].tsx` and `routes/fail.tsx` don't qualify `RouteContext` with any type parameters.

In a soon-to-be-opened PR, I will suggest that we type check our entire project as part of CI. If we had always been doing this, then we would have caught this problem before users had to report an issue.

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -59,11 +59,9 @@ export interface PageProps<T = any, S = Record<string, unknown>> {
 /**
  * Context passed to async route components.
  */
-export type RouteContext<T = unknown, S = Record<string, unknown>> =
-  & Omit<
-    HandlerContext<T, S>,
-    "render"
-  >
+// deno-lint-ignore no-explicit-any
+export type RouteContext<T = any, S = Record<string, unknown>> =
+  & Omit<HandlerContext<T, S>, "render">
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
LGTM, thanks for the PR 👍

PR #1537, opened 2023-07-28
chore: clean up remaining type errors in project

deer:
Resulting from the investigation in https://github.com/denoland/fresh/pull/1536, this resolves three type errors in the project. The import_map one is for `www/routes/docs/[...slug].tsx`.


marvinhagemeister (approved):
LGTM, thanks for the PR 💯

PR #1538, opened 2023-07-28
chore(ci): type check entire project in ci

deer:
The logical continuation of https://github.com/denoland/fresh/pull/1536 and https://github.com/denoland/fresh/pull/1537. Now that the project is clean (after those are merged), we should prevent any such errors from sneaking in again in the future.

A question here: why are individual things type checked in the CI? Specifically:
```yaml
      - name: Type check init script
        run: deno check --remote init.ts

      - name: Type check website
        run: deno check main.ts dev.ts
        working-directory: www/

      - name: Type check demo
        run: deno check --remote main.ts dev.ts
        working-directory: demo

      - name: Type check tests/fixture
        run: deno check main.ts dev.ts
        working-directory: tests/fixture/

      - name: Type check tests/fixture_error
        run: deno check main.ts dev.ts
        working-directory: tests/fixture_error/
```

Opening as a draft because:
1) this will fail due to the other PRs not being merged yet
2) open question regarding duplicate type checking
3) should I pull the check into the matrix like the other type checks?

I do agree that `deno check --remote init.ts` should remain separate, but the other four seem superfluous if we decide to check the whole project.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Type check project
         if: startsWith(matrix.os, 'ubuntu') && matrix.deno == 'canary'
-        run: deno check **/*.ts
+        run: deno task check:types
 
       - name: Cache dependencies
         if: startsWith(matrix.os, 'ubuntu') && matrix.deno == 'v1.x'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,6 @@ jobs:
         uses: crate-ci/typos@master
 
       - name: Type check project
-        if: startsWith(matrix.os, 'ubuntu') && matrix.deno == 'canary'
         run: deno task check:types
 
       - name: Cache dependencies
@@ -84,10 +83,6 @@ jobs:
       - name: Type check init script
         run: deno check --remote init.ts
 
-      - name: Type check website
-        run: deno check main.ts dev.ts
-        working-directory: www/
-
       - name: Test website
         run: deno test -A
         working-directory: www/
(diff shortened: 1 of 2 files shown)


deer:
Thoughts on removing the extra type checks? Thoughts on why we type check individual items in every matrix run? I picked `startsWith(matrix.os, 'ubuntu') && matrix.deno == 'canary'` for type checking since that's what Lino did for typos, but maybe we should do the full type check in every run.

This is fortunately passing now.

iuioiua (commented):
Suggestions:
- Create a new `check:types` task in `deno.json` and use `demo task check:types` in CI
- Also check `.tsx` files

WDYT?

deer:
Thanks, I think that's very similar to what I've done in https://github.com/denoland/fresh/pull/1454. So maybe I'll pull the changes to `deno.json` out of 1454 and leave that focused on `init.ts`.

So here in this PR I'd have... 
* `check:types` which does `deno check **/*.ts && deno check **/*.tsx`
* change the existing `run: deno check **/*.ts` to `run: deno task check:types`

Any thoughts on my confusion around the existing type checks in CI? And how it runs for all matrix variations?

iuioiua (approved):
LGTM! It's OK to run the type check on all platforms + versions.

deer:
> LGTM! It's OK to run the type check on all platforms + versions.

Ok, I can remove the condition then. Should I also remove the other type checks, now that we're checking everything at once?

iuioiua:
Sure 🙂

deer:
>error: Import 'https://deno.land/std@0.173.0/path/posix.ts' failed: 500 Internal Server Error
    at https://deno.land/std@0.173.0/path/mod.ts:26:25

Please rerun CI 🤞

marvinhagemeister (approved):
LGTM, thanks for adding this 👍

PR #1539, opened 2023-07-28
chore(init): remove blank line from _404 template

deer:
Found in https://github.com/denoland/fresh/pull/1454#issuecomment-1655170384. The project produced by `init` doesn't pass the formatting check. Perhaps we should expand the scope of this PR to have a test run the `safe` command (from 1454) on the newly initialized project? I can happily do this in this PR or a separate one. Or if we decide against the `safe` command, we could manually run type checks, linting, and formatting on the initialized project in a test.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -229,8 +229,7 @@ await Deno.writeTextFile(
 );
 
 // 404 page
-const ROUTES_404_PAGE = `
-import { Head } from "$fresh/runtime.ts";
+const ROUTES_404_PAGE = `import { Head } from "$fresh/runtime.ts";
 
 export default function Error404() {
   return (


deer:
ci failed due to `error: Import 'https://deno.land/std@0.193.0/streams/text_line_stream.ts' failed: 500 Internal Server Error`
please rerun

marvinhagemeister (approved):

marvinhagemeister:
We could add a step to the CI that creates a new Fresh project and type checks it. To me that's preferable because that way the creating a new project stays fast.

deer:
I'm rereading my comment and it definitely could have been phrased better. I'm suggesting an addition to `cli_test.ts` which runs `deno task check` on the newly initialized project. This wouldn't impact the creation of a new project. And then this would get caught when a developer changes the init script locally and runs tests. (Hopefully we're all running tests locally before pushing to remote, right!?)

I can add the new test, if you think this is a good idea.

Issue #1514, opened 2023-07-24
[Plugins] PluginRoutes seem not to be registered correctly in fresh.gen.ts


bjesuiter:
I want to update my fresh_openprops integration to use a real Fresh Plugin for the integration:  

https://github.com/codemonument/deno_fresh_openprops/blob/main/src/fresh_openprops_plugin.ts   

It has a factory function, which returns the plugin object like so: 

```ts

export async function FreshOpenProps(rawOptions?: PluginOptions) {
  // Throws when option parsing fails
  const { cssInputPath, postcssModuleDirs, isProd, doPrefillCssCache } =
    PluginOptions.parse(
      rawOptions,
    );

  // Should only happen once, since this plugin is only initialized once
  await initPostcssInstance(postcssModuleDirs);

  if (doPrefillCssCache) {
    await prefillCssCache({ cssInputPath });
  }

  const postcssRoute = {
    path: "postcss/[...path]",
    handler: async (
      req: Request,
      ctx: HandlerContext,
    ): Promise<Response> => {
       // The handler code 
    } 
  } satisfies PluginRoute; 
  
  return {
    name: "openprops",
    routes: [postcssRoute],
  } satisfies Plugin;
```

It integrates like this: 

See my example in the repo for full code: https://github.com/codemonument/deno_fresh_openprops/blob/main/example/main.ts

```ts

// inside main.ts file 

const openpropsPlugin = await FreshOpenProps({
  isProd: false,
  doPrefillCssCache: true,
  cssInputPath: "example/css",
  postcssModuleDirs: ["example/css_deps"],
});

await start(manifest, {
  plugins: [
    openpropsPlugin,
  ],
});

```

First, I had the openpropsPlugin object inlined but extracted it afterwards bc. I suspected the await of the factory function to mess up the plugin initialization. 

## Error Observations 

- The route /postcss/global.css cannot be found (404) 
- After Inspection: The route from the plugin is not even registered in fresh.gen.ts 

```ts 
// Content of fresh.gen.ts
import * as $0 from "./routes/index.tsx";

const manifest = {
  routes: {
    "./routes/index.tsx": $0,
  },
  islands: {},
  baseUrl: import.meta.url,
};

export default manifest;

```

Source: https://github.com/codemonument/deno_fresh_openprops/blob/main/example/fresh.gen.ts

What may be wrong here?

deer:
[...]

Issue #1416, opened 2023-07-06
Introduce development time checks to remove remaining foot guns in the project


deer:
I read our documentation and found the following assumptions in the getting started and concepts pages. But we don't check for any of these to give good error messages.

- Only one _app.tsx is allowed per application and it must have default export
- One 500.tsx and it must have default export
- One 404.tsx and it must have default export
- Pascal case or kebab case islands (island name)
- Route must have handler or component
- If static override, ensure there is no normal static directory
- Static file conflict with routes (because static files have higher priority)
- Plugin doesn't call render
- Plugin Each defined module must be a JavaScript/TypeScript module

As an example, if you remove the default export from `_app.tsx` then the following unhelpful error message is presented (note how it doesn't mention `_app.tsx`:
```
An error occurred during route handling or page rendering. Error: Undefined component passed to createElement()

You likely forgot to export your component or might have mixed up default and named imports<#text params="[object Object]" url="http://localhost:8000/" route="/" data="undefined" Component="function Component() {}" />

  in #text

    at i.__b (https://esm.sh/stable/preact@10.15.1/denonext/debug.js:11:362)
    at k (https://esm.sh/v125/preact-render-to-string@6.1.0/X-ZS8q/denonext/preact-render-to-string.mjs:2:2086)
    at k (https://esm.sh/v125/preact-render-to-string@6.1.0/X-ZS8q/denonext/preact-render-to-string.mjs:2:2591)
    at k (https://esm.sh/v125/preact-render-to-string@6.1.0/X-ZS8q/denonext/preact-render-to-string.mjs:2:2591)
    at T (https://esm.sh/v125/preact-render-to-string@6.1.0/X-ZS8q/denonext/preact-render-to-string.mjs:2:1312)
    at renderInner (https://deno.land/x/fresh@1.2.0/src/server/render.ts:191:16)
    at renderSync (https://deno.land/x/fresh@1.2.0/src/server/render.ts:209:7)
    at https://deno.land/x/fresh@1.2.0/src/server/render.ts:240:38
    at Object.DEFAULT_RENDER_FN [as renderFn] (https://deno.land/x/fresh@1.2.0/src/server/context.ts:792:3)
    at renderAsync (https://deno.land/x/fresh@1.2.0/src/server/render.ts:240:18)
Error: Undefined component passed to createElement()

[...]

PR #1545, opened 2023-07-29
fix(server): merge competing PluginMiddleware handlers together

deer:
blocked by https://github.com/denoland/fresh/pull/1534

This issue arises when multiple plugins try to create middleware at the same path. We need to merge them together, similar to [here](https://github.com/denoland/fresh/blob/75f5761ca460efb9517887433064561fcb4392db/tests/fixture/routes/_middleware.ts#L62).

I've added a test for this, which is currently not passing the type checks, due to 1534 mentioned above. The test throws some extra middleware onto the root route, and also creates a new route that has multiple handlers from a single plugin. Another plugin then adds here as well. The final result is asserted in the test. Prior to this change, the existing `plugin routes and middleware` test was failing if another middleware was added from a separate plugin (which happened to me in a side project). The test is left in place.

Additionally, I've fixed an issue with specifying routes with the introduction of `formatMiddlewarePath`. Now developers of plugins no longer have to specify leading and trailing slashes. This maintains backwards compatibility.


deer:
Once #1534 is merged, then I will update the plugins with types so that the tests run.

deer:
@bjesuiter, I think this solves the issue you mentioned [here](https://github.com/denoland/fresh/issues/1514#issuecomment-1648545715), at least for middleware. If we think this is the right approach, then I can create a separate PR for fixing/improving the way plugin authors specify routes. (This PR just improve how it's done for middleware, in addition to the bug fix.)

bjesuiter:
@deer I looked at this issue here and all the other connected issues we created to try to make sense, what exactly solves the problem I mentioned in https://github.com/denoland/fresh/issues/1514#issuecomment-1648545715. 

## My Thoughts on the middleware handler thematic

### Bloated syntax for defining middleware handlers
Considering the "routePlugin" definition in [the PR](https://github.com/denoland/fresh/pull/1545/files#diff-5d3be94b363b7606991803e441fec019f4a09884d8aff3e3ecbf0998e08ea9ef) (for some reason i cannot permalink to the file div for this): 

```ts
export default function routePlugin(options: Options): Plugin {
  return {
    name: "routePlugin",
    middlewares: [{
      middleware: { handler: testMiddleware },
      path: "/",
    }, {
      middleware: {
        handler: twoPointlessMiddlewares,
      },
      path: "lots-of-middleware",
    }],
    routes: [{
      path: "/_app",
      component: AppBuilder(options),
    }],
  };
}
```

1. Why do we accept an array of handlers for the same path here: 
```ts
{
      middleware: {
        handler: twoPointlessMiddlewares,
      },
      path: "lots-of-middleware",
    }
```

2. Couldn't this structure be simplified to: 

```ts
return {
    name: "routePlugin",
    middlewares: [
    {
      handler: testMiddleware
      path: "/",
    }, 
    {
       handler: twoPointlessMiddlewares,
      path: "lots-of-middleware",
    }]
}
```

3. Does it need to be called "handler"? 
```ts
  handler: twoPointlessMiddlewares,
```
If it supports an array of handlers, I'd like to call it "handlers". 
But this might be confusing either way depending on the coding level of the user. 

### Merging of middlewares at the same path seems nice 

If I see this correctly, this PR adds the option to merge 2 or more middleware handlers for the same route by simply making a concatenation of them. Do I view this correctly? 

This might explain, why you thought of allowing multiple middleware handlers for one route in a plugin definition, since the merging code is in the framework at the end. 

#### Suggestion: Disallow for multi-handlers per route in one plugin
But I think, that we should limit the possibilities for plugin authors a little bit here. 
My rule would be: A plugin author can only add one middleware handler per route. 
Then the code in src/server/context.ts of the fresh framework would still merge the multiple handlers from different Plugins, but each plugin can only have one handler per route. 

Questions/Thoughts for that: 
1. Would that limit some functionality I didn't thought of? 
2. This would also solve the naming problem: each path in a plugin middleware definition would've only one "handler" again ^^
3. Do you want to allow multiple handlers per path from a plugin because of plugin combinations/chaining of plugins??

bjesuiter:
@deer 

# Now my thoughts on how this could apply to routes from plugins 

_put this in a second comment for you to be able to answer differently to both of them ^^_

## Doing route merging the same as middleware is harder 

I think it's not possible to simply use the same logic for merging routes with the same path from different plugins as it is for middlewares. 
Reason: routes can return various different things, like components, raw Results, multiple handlers depending on HTTP verbs. 

=> Therefore I'd forbid two plugins registering the same routes. 
=> **Suggestion**:  This would be a nice use-case for relative URLs in Plugins! 
When plugins define their routes as relative, we can make the basePath for these relative urls configureable from the outside world of the plugin! Depending on the plugin, it doesn't matter for the functionality at which basePath it's attached. 
If the plugin really needs an absolute path, we should throw an error when there's more than one route at the same absolute path. 

## Which issue is solved with this PR compared to my original problem with the absolute path [here](https://github.com/denoland/fresh/issues/1514#issuecomment-1648545715)? 

When I look at the original thread, my main problem was with the  missing '/' at the beginning of the route path. 
I'm not able to see how this is improved by the ideas in this PR, other than the "setting the base path outside the plugin" idea from above. 

What did you mean exactly when you said: 
> I think this solves the issue you mentioned https://github.com/denoland/fresh/issues/1514#issuecomment-1648545715, 
> at least for middleware. If we think this is the right approach`
- which issue of the ones I mention do you mean here? 
- what approach do you mean might be the right one in this PR? 

## I like the "Improving dev time dev errors" PR 

I think, [this PR with improving the dev time dev errors you mentioned](https://github.com/denoland/fresh/issues/1416) below my original problem with the absolute/relative paths, is really nice. 
I haven't looked into it in detail, but more error when developing seems very good.

deer:
Thanks for the very detailed posts. Let's see if I can correctly respond to all the points. I clearly didn't express myself clearly. Let me try to correct that: all I meant was that "this approach" can be used to solve the problem of specifying paths, if we apply it to routes. Prior to this fix, if a plugin author wanted to specify middleware that runs at a particular location, they would have to write `path: /my-location/` which is perhaps a bit silly. Now they can just write `path: my-location`. Now, on to your great feedback...

> 1. Why do we accept an array of handlers for the same path here:

One middleware route can already specify multiple handlers, e.g. [this](https://github.com/denoland/fresh/blob/75f5761ca460efb9517887433064561fcb4392db/tests/fixture/routes/_middleware.ts#L62). This functionality was introduced by https://github.com/denoland/fresh/pull/553 and I think we should offer plugin authors the same functionality.

>2. Couldn't this structure be simplified to:

Yes, I see your point that it is currently a bit bloated. From `types.ts` we have...
```tsx
export interface PluginMiddleware {
  /** A path in the format of a filename path without filetype */
  path: string;

  middleware: Middleware;
}

export interface Middleware<State = Record<string, unknown>> {
  handler: MiddlewareHandler<State> | MiddlewareHandler<State>[];
}
```
This makes it slightly easier to inject the middleware into the server context, at the cost of developer experience to plugin authors. I agree it would be simpler if we did something like:
```tsx
export interface PluginMiddleware {
  /** A path in the format of a filename path without filetype */
  path: string;
  middleware?: Middleware;
  handler?: MiddlewareHandler<State> | MiddlewareHandler<State>[];
}
```
This would enable your suggested:
```tsx
return {
    name: "routePlugin",
    middlewares: [
    {
      handler: testMiddleware
      path: "",
    }, 
    {
      handler: twoPointlessMiddlewares,
      path: "lots-of-middleware",
    }]
}
```
>3. Does it need to be called "handler"?

I'm just copying what's already in `types.ts`. Of course we could consider changing it, but this seems relatively minor, especially when hovering over `handler:` in IDEs shows that it supports an array as well. And once we have a great "Writing a plugin" example, there should be no problem here. If only someone were to write that... 🙃 

Ok, so now your second post.

>Doing route merging the same as middleware is harder

Agree, I don't think there should be any route merging for plugins. Middleware merging is already a thing that happens, so I'm just making it happen for plugins. But there's no route merging as is, so we certainly shouldn't introduce it for plugins. 

> Which issue is solved with this PR compared to my original problem with the absolute path https://github.com/denoland/fresh/issues/1514#issuecomment-1648545715?

I believe I answered that at the beginning of this post.
> `path: /my-location/` which is perhaps a bit silly. Now they can just write `path: my-location`

So extending this idea to routes, the existing:
```tsx
routes: [{
      path: "/_app",
      component: AppBuilder(options),
    }
```
would change its path declaration to "_app" if it should run at the root route. The core issue is that the plugin loader does this:
```tsx
`./routes${route.path}.ts`
```
It shouldn't force plugin developers to start their path with a slash. We should instead be able to do things like this:
```ts
path: "index"
path: "folder/[asdf]"
path: "folder/subfolder/index"
```
We should instead do a conditional check on the path and prepend a slash if necessary. I believe this would solve your problem of `/postcss/[...path]` vs `postcss/[...path]`.

> I like the "Improving dev time dev errors" PR
We should extend this concept to do a few things for plugins. Given that a plugin author might not be authoring their plugin in the context of a fresh application, we can't rely on any sort of dynamic analysis there. (Perhaps some improved linting at some point?)

More likely would be two things:
1) a single plugin has configured conflicting routes. There should be a message to the user encouraging them to contact the plugin author to let them know that they've done something really silly.
2) Multiple plugins conflict with each other. If two plugins are trying to push handlers to the same route, then we should let the user know that the plugins they're trying to use are conflicting with each other.

Please let me know if I've missed something, or if anything is still unclear.

bjesuiter:
@deer Thanks for the detailed response! Going over it now. 

1. Thanks for pointing out #553, this was exactly what I missed as explanation for why we should have multiple middlewares per route. 

2. Happy that you like the new style of 

```ts
return {
    name: "routePlugin",
    middlewares: [
    {
      handler: testMiddleware
      path: "",
    }, 
    {
      handler: twoPointlessMiddlewares,
      path: "lots-of-middleware",
    }]
}
```

=> Do we go on implementing this change? (Probably in another PR, to not bloat this one :D) 

3. Yes, the `handler` naming is not that critical, i only find it annoying from time to time and wanted to know if it was intentional or some kind of leftover :D  
Imo we can leave it like this for now to not introduce more construction sites in the code than necessary. 

4. I support this: 
```ts
We should instead do a conditional check on the path and prepend a slash if necessary. I believe this would solve your problem of /postcss/[...path] vs postcss/[...path].
```

Thanks for the clarification about `postcss/[...path]` and `/postcss/[...path]`, i really couldn't figure it out while reading the PR first time. 

5. I also see these two cases for dev errors: 

> More likely would be two things:
> 
> a single plugin has configured conflicting routes. There should be a message to the user encouraging them to contact the > plugin author to let them know that they've done something really silly.
> Multiple plugins conflict with each other. If two plugins are trying to push handlers to the same route, then we should let the user know that the plugins they're trying to use are conflicting with each other.

I think I don't have more questions for this PR, will comment on some other ones in the future again :) 
Working with you feels great, thanks for that! :)

deer:
Now that 1534 was merged, I will clean this up later today.

deer:
ci / test (v1.x, macOS-latest) (pull_request) is failing due to
>error: Import 'https://deno.land/x/imagescript@1.2.15/utils/wasm/jpeg.js' failed: 500 Internal Server Error
    at https://deno.land/x/imagescript@1.2.15/ImageScript.js:7:26

I'm see this more and more as of late.

deer:
To summarize the work produced by the discussions:
1. simplify plugin middleware by doing the following:
```tsx
export interface PluginMiddleware {
  /** A path in the format of a filename path without filetype */
  path: string;
  middleware?: Middleware;
  handler?: MiddlewareHandler<State> | MiddlewareHandler<State>[];
}
```
This will make the developer experience easier for plugin creators, at the slight cost of more complicated code to load plugins.

2. Simplify the creation of routes from plugins, by not requiring slashes at the beginning. This is confusing as is.
3. Create a "Writing a plugin" example.
4. Add checks to https://github.com/denoland/fresh/blob/main/src/server/dev_checks.ts for two things:
4a. routes within a plugin conflicting
4b. routes across plugins conflicting

@bjesuiter, did I miss anything? I will create issues for these tonight.

marvinhagemeister (approved):
LGTM 👍

bjesuiter:
@deer please tag me in the "how to write a plug-in issue" I already started a branch for that locally :)

PR #1489, opened 2023-07-19
feat(server): dev safety checks

mbhrznr:
closes #1416  
closes #1524

as discussed in the aforementioned issue, @deer was so kind to let me finish this one off.
as expected it gave me quite some nice understanding of `fresh`'s internals.

this pr introduces the following check when being in `dev` mode:  
- [x] only one _app.tsx is allowed per application and it must have default export
- [x] one 500.tsx and it must have default export
- [x] one 404.tsx and it must have default export
- [x] ~~pascal case or kebab case islands (island name)~~ (this one became obsolete w/ 1.3)
- [x] route must have handler or component
- [x] route patterns must be unique 
- [x] if static override, ensure there is no normal static directory
- [x] static file conflict with routes (because static files have higher priority)
- [x] plugin doesn't call render
- [x] plugin each defined module must be a js/ts module

the functionality can be tested by navigating into `tests/fixture_dev_checks/` and running `deno task start`

PR #1550, opened 2023-07-31
fix(ci): type error in dev_checks fixture

deer:
I think https://github.com/denoland/fresh/pull/1538 was merged without merging in https://github.com/denoland/fresh/pull/1489 and rerunning CI.

@marvinhagemeister, I try to avoid pinging you specifically to merge my PRs, but in this case I'll make an exception. Without this fix I think all future CI runs will fail.

### DIFF
--- a/tests/fixture_dev_checks/main.ts
+++ b/tests/fixture_dev_checks/main.ts
@@ -9,6 +9,8 @@ import manifest from "./fresh.gen.ts";
 import notCallingCtxRender from "./plugins/not_calling_ctx_render.ts";
 import notHavingModules from "./plugins/not_having_modules.ts";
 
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):

marvinhagemeister:
Whoops, I thought somebody had rerun the CI already because it was green.

deer:
Hmmm, not sure how it works for you as maintainer, but when I look at a PR (e.g. #1454 -- look, I did it!) I see:
> This branch is out-of-date with the base branch

Perhaps this doesn't happen if you don't refresh the page or something when merging multiple PRs in a short time frame.

marvinhagemeister:
Ohh right that makes sense. I read over that part and only looked at the green checkmark 😅

PR #1551, opened 2023-07-31
When specifying the path when importing, it is now easier to specify …

EdamAme-x:
The method of specifying from `"~/"` is very straightforward. For example, using `import Counter from "./islands/Counter.tsx"` instead of `import Counter from "~/islands/Counter.tsx"`. We believe that the former is less likely to lose the directory structure.

Oops, I feel like I already know this.

Anyway, this should be accepted as a memorable first directory notation...

### DIFF
--- a/demo/deno.json
+++ b/demo/deno.json
@@ -9,7 +9,8 @@
     "preact/": "https://esm.sh/preact@10.15.1/",
     "preact-render-to-string": "https://esm.sh/*preact-render-to-string@6.2.0",
     "@preact/signals": "https://esm.sh/*@preact/signals@1.1.3",
-    "@preact/signals-core": "https://esm.sh/@preact/signals-core@1.2.3"
+    "@preact/signals-core": "https://esm.sh/@preact/signals-core@1.2.3",
+    "~/": "./"
   },
   "compilerOptions": {
     "jsx": "react-jsx",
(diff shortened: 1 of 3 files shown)


demo/deno.json
```diff
@@ -9,7 +9,8 @@
     "preact/": "https://esm.sh/preact@10.15.1/",
     "preact-render-to-string": "https://esm.sh/*preact-render-to-string@6.2.0",
     "@preact/signals": "https://esm.sh/*@preact/signals@1.1.3",
-    "@preact/signals-core": "https://esm.sh/@preact/signals-core@1.2.3"
+    "@preact/signals-core": "https://esm.sh/@preact/signals-core@1.2.3",
+    "~/": "./"
```
iuioiua:
```suggestion
    "@/": "./"
```
Fresh projects typically use `@/` instead of `~/` (see [Deno SaaSKit](https://github.com/denoland/saaskit/blob/1061b81d8fed9e23f7039bf2446629623fc4ac5f/deno.json#L20)). Let's stick with that convention. I also think it'd be a good idea to move this to the top of the `imports` object.

demo/deno.json
iuioiua, line 1: Can you please modify components, routes, islands, etc., to use this new import prefix?

iuioiua (changes requested):

EdamAme-x:
I see.
I am ashamed to say that I have never heard of `@/` before.
I'm going to go change the import prefix for each component.

EdamAme-x:
Oops, warn of types

EdamAme-x:
How should we deal with...

deer:
This reminds me of https://github.com/denoland/fresh/pull/471.

EdamAme-x:
I can't do anything about this warn...

I wish for solve of #471 issue...

bartlomieju:
Closing as stale — the review feedback to update all imports across the demo was never completed. This is a minor cosmetic change to the demo app and can be revisited in a fresh PR if desired. Thanks for the contribution!

PR #1554, opened 2023-08-01
showcase: add over-ip

sirlan-ff00ff:
Add [over-ip](https://over-ip.dragon-vi.be) to the showcase, a simple _chat_ program about shouting noises into the void. Made as a test/showcase of a 'full stack' program using deno kv and websockets. The source code is public on GitLab (link on the 🦕🍋 emojis)


marvinhagemeister (approved):
LGTM

PR #1557, opened 2023-08-01
chore(CI): combine install Chrome jobs

iuioiua:
Reference: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsenv


marvinhagemeister (approved):
LGTM 👍

PR #1559, opened 2023-08-01
chore(CI): cache deps before type checks

iuioiua:
This makes more sense to do, as checks require modules to be downloaded first. Hopefully, this helps with CI flakiness.


deer:
Thank you 🙌

iuioiua:
Closing in favour of #1571.

PR #1561, opened 2023-08-01
chore: simplify route creation logic

marvinhagemeister:
No behaviour changes, just noticed that it could be written with less loops. Saves about `1ms` (not much really) but is a little shorter to read.


bartlomieju (approved):
LGTM, CI failures seems to be intermittent error.

deer:
This is much better, thanks.

PR #1562, opened 2023-08-01
refactor: pre-sort all routes

marvinhagemeister:
Instead of sorting `routes`, `layouts` and `middlewares` individually we can sort them all at once. Didn't notice any perf differences but to me it makes the code a bit easier to read. As a plus we only have one sort function to deal with instead of 3 separate ones.


bartlomieju (approved):
LGTM

PR #1563, opened 2023-08-01
feat: add support for route groups

### LINKED ISSUE #1150: Route grouping support
## What

Coming from trying out the new Next.js app directory (Not an apples to apples comparison I know) I've found one feature that has made structuring my file-based routes 100x easier to manage is [route groups](https://beta.nextjs.org/docs/routing/defining-routes#route-groups) which allows you to group pages together in a directory & wrapping that directory name in brackets, like `(foo)` for example, to keep the route pathname the same isolate stuff to be scoped specifically.

Wondering what others thoughts on this feature are & if this is something that either Fresh could add or someone else could implement via a 3rd party module?

## Why

One of the best examples I have found for this is isolating the "app" part of your application & the "marketing" page(s). So you have `(marketing)` that groups together say `index.tsx`, `pricing.tsx`, `about.tsx`, etc & then you have `(dashboard)` for the actual application which is all passed through a `_middleware.ts` to add, for example, some authentication check.

marvinhagemeister:
This PR adds support for route groups which allows you to use different layouts for routes in the same URL segment.

```txt
/routes
  /(marketing)
    /_layout.tsx   <-- only applies to about.tsx and career.tsx
    /about.tsx
    /career.tsx
  /(info)
    /_layout.tsx   <-- only applies to index.tsx and contact.tsx
    /index.tsx
    /contact.tsx
```

To be able to do this we need to keep track of the original route paths. Previously, matching `_layout` and `_middleware` was done via their `URLPattern`, but this doesn't work with groups anymore. With groups multiple `_layout` files can have the same `URLPattern` and it becomes ambiguous as to which file to pick.

To solve this we need to match on the original file path semantics instead of their resulting `URLPattern`. To do that the PR adds an additional property `baseRoute` to each route which is the normalised file path. When a route matches we can retrieve that (like `/foo/(bar)/(bob)/about`) and easily match the needed layout files.

The PR is a bit longer than I'd hoped. Most of the changes relate to changing the internal route signature from:

```ts
interface Route {
  [K in "GET" | "PATCH" ...]: ...
}
```

to

```ts
interface Route {
  baseRoute: string;
  methods: {
    [K in "GET" | "PATCH" ...]: ...
  }
}
```

Instead of making `baseRoute` just a string, I made it a nominal type. Noticed that I ran into a couple of errors during development where some values started with a `/`, others didn't etc. With the nominal type we can force it to have been parsed first and ensure it always has the expected format.

Fixes #1150 .

### DIFF
--- a/docs/canary/concepts/route-groups.md
+++ b/docs/canary/concepts/route-groups.md
@@ -1,0 +1,58 @@
+---
+description: |
+  Route groups allow you to organize your route segments into groups and share layout or middleware code without affecting the URL structure.
+---
+
+When working with [route layouts](/docs/canary/concepts/route-layout) or
+[route middlewares](/docs/canary/concepts/middleware) you'll sometimes come
+across a situation where you want your routes to inherit from a different layout
+than another route that's in the same URL segment.
+
+Let's illustrate that with an example:
+
+```txt
+/about -> layout A
+/career -> layout A
+/archive -> layout B
+/contact -> layout B
+```
+
+Without any way to group routes this is a problem because every route segment
+can only have one `_layout` file.
+
+```txt
+/routes
+  /_layout.tsx   <-- applies to all routes here :(
+  /about.tsx
+  /career.tsx
+  /archive.tsx
+  /contact.tsx
+```
+
+We can solve this problem with route groups. A route group is a folder which has
+a name that is wrapped in braces. For example `(pages)` would be considered a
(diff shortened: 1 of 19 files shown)


marvinhagemeister:
oh no, looks like I broke some tests 🙈

marvinhagemeister:
Alright, ready to review now 🎉

bartlomieju (approved):
LGTM, no comments. I love the approach of using a nominal type 👍

PR #1565, opened 2023-08-01
chore: speed up layout tests

marvinhagemeister:
The layout tests were a bit slow. Noticed that they only check if the server sent HTML is correct and don't need any JS to be executed client side. This means we can skip launching a full `puppeteer` process and simply assert against the fetched HTML. This brings down the layout test time from a couple of seconds down to `<100ms`.

Wrote a tiny pretty printer that parsers the linkedom document into something for humans to debug when for example a selector isn't found:

Screenshot 2023-08-01 at 15 33 18

_Note the charset casing issue is resolved in `preact-render-to-string`, just need to cut a new version_

### DIFF
--- a/tests/deps.ts
+++ b/tests/deps.ts
@@ -19,4 +19,9 @@ export {
   default as puppeteer,
   Page,
 } from "https://deno.land/x/puppeteer@16.2.0/mod.ts";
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
LGTM, very nice!

PR #1567, opened 2023-08-01
Revert: feat(server): add safety checks during `dev`

marvinhagemeister:
Revert: feat(server): add safety checks during `dev` (#1489)

I merged #1489 too quickly. Reverting this PR because it has a couple of false positives that block me on feature work and I don't have time to look into these at the moment.


deer:
Does the entire thing need to be reverted or can you just comment them out to disable the checks?

marvinhagemeister:
> Does the entire thing need to be reverted or can you just comment them out to disable the checks?

I'm not a fan of having commented out code in the repo. We can always re-land PRs once their fixed.

PR #1568, opened 2023-08-01
feat(server): add safety checks during dev 

### LINKED ISSUE #475: Behavior of `/[dynamic].ext` routes together with `/static.ext`

### LINKED ISSUE #1558: add dev checks for plugin route conflicts
We should add two checks here:
1. routes within a plugin conflicting are conflicting with each other. The plugin author has done something really silly! (Maybe this isn't even necessary?)
2. routes across plugins conflicting with each other, or with the user's routes. This is very possible to happen.

from: https://github.com/denoland/fresh/pull/1545#issuecomment-1658412159
@mbhrznr, any interest in picking this up? 🤞

### LINKED ISSUE #1678: 404 on loading plugin-twind-main.js
I've just upgraded to Fresh 1.4.2.  When I build my project with `deno task build` and launch the site I now get a 404 from attempting to load plugin-twind-main.js.

> http://localhost:8000/_frsh/js/62ccd15cffd2ab97bfc5746f8d22ba03af18915b/plugin-twind-main.js

This can be seen in Deploy as well by visiting, e.g., https://dev-tools.dev/base64-decode.

The HTML loaded contains the following:
```html
<!DOCTYPE html>
<html>
    <head>
        <title>dev-tools</title>
        <meta charset="utf-8"/>
        <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
        <link rel="modulepreload" href="/_frsh/js/acf84d6ccb255abb4480e6a3b8270863025b62b6/plugin-twind-main.js"/>
        <link rel="modulepreload" href="/_frsh/js/acf84d6ccb255abb4480e6a3b8270863025b62b6/main.js"/>
        <link rel="modulepreload" href="/_frsh/js/acf84d6ccb255abb4480e6a3b8270863025b62b6/chunk-XZNBMHWY.js"/>
```

Other js files in _frsh load as expected.  I've tried clearing and disabling the cache to no avail.

Code can be found in [GitHub](https://github.com/cknight/dev-tools).

Locally there are no build errors or other messages.  Unfortunately this 404 breaks my site as other javascript fails to load.

mbhrznr:
this is the second attempt to land #1489, as the initial version has been reverted due to unwanted side effects.  

closes #475.
closes #1558.
closes #1678.

supersedes #539.
supersedes #1414.

this pr introduces the following check when being in `dev` mode:

- [x] only one _app.tsx is allowed per application and it must have a default export
- [x] only one 500.tsx and it must have a default export
- [x] only one 404.tsx and it must have a default export
- [x] ~~pascal case or kebab case islands (island name)~~ (this one became obsolete w/ 1.3)
- [x] routes must have exported handler or component
- [x] route patterns must be unique
- [x] route patterns w/ dynamic params must not have conflicts w/ other routes
- [x] if static override is used, ensure that there is no normal static directory
- [x] static file conflicts with routes (because static files have higher priority)
- [x] plugin doesn't call render/renderAsync
- [x] each defined module in a plugin must be a js/ts module
- [x] plugins should be present in the build snapshot 

the functionality can be tested by changing the `fixture` task in `deno.json` to: 
```
"fixture": "deno run -A --watch=static/,routes/ tests/fixture_dev_checks/dev.ts",
```

this time, the pr also contains unit tests for each individual check to (hopefully 🤞 ) prevent further mishaps.


deer:
You should go into each fixture and start up the dev server to confirm this isn't causing issues. Perhaps we could even add a test to assert that all the fixtures adhere to these checks.

mbhrznr:
> You should go into each fixture and start up the dev server to confirm this isn't causing issues. Perhaps we could even add a test to assert that all the fixtures adhere to these checks.

will do!
currently i'm writing unit tests for each of the checks to hopefully cover all of the cases 🤞 
will keep the pr in draft for the time being, yet wanted to bring it up asap again for transparency reasons.

i'd also incorporate the feedback and remarks from the first pr.

deer:
I see some changes of `as` -> `satisfies`. I guess this makes sense, but I wonder why just these three places? It seems like there should be a larger audit of the codebase to make this switch where appropriate. Therefore I think you should revert these from this PR, and then log an issue or open a separate PR if you're willing to take on the work yourself.

Some of the names in the error messages don't seem quite right:
```
See: _app
See: test
See: /custom.txt
```
I would expect:
```
See: routes/_app.tsx
See: routes/test.tsx
See: static/custom.txt
```

mbhrznr:
as always, thank you for the feedback!

> I see some changes of `as` -> `satisfies`. I guess this makes sense, but I wonder why just these three places? It seems like there should be a larger audit of the codebase to make this switch where appropriate. Therefore I think you should revert these from this PR, and then log an issue or open a separate PR if you're willing to take on the work yourself.

sorry about that!
there wasn't an immediate need to do so, i just stumbled upon this while i was testing each of the available fixtures.
will revert these once i push the final changes.

> Some of the names in the error messages don't seem quite right:
> 
> ```
> See: _app
> See: test
> See: /custom.txt
> ```
> 
> I would expect:
> 
> ```
> See: routes/_app.tsx
> See: routes/test.tsx
> See: static/custom.txt
> ```

you're absolutely right!  
this has also been brought up in the initial pr and is something i'm currently working on. 
this would be the final task before i'd comfortably consider this pr as "ready for review".

mbhrznr:
improved the warnings output as suggested.
also checked all of the fixtures to have some certainty that the dev checks don't introduce any unwanted behavior.

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/** Asserts that only one module exists with the given name. */
+export function assertSingleModule(
+  routes: Route[],
+  moduleName: string,
+): CheckResult[] {
+  const moduleRoutes = routes.filter((route) =>
+    route.name.includes(moduleName)
+  );
```
marvinhagemeister: Does this work for routes like `/foo/_app/bar` or would that classify this as an invalid route?

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+    }];
+  }
+  return [];
+}
+
+/** Asserts that each route has a unique pattern. */
+export function assertSingleRoutePattern(routes: Route[]) {
+  const routeNames = new Set(routes.map((route) => route.pattern));
+
+  return routes.flatMap((route) => {
```
marvinhagemeister: Should this be a `.map`?  The `Route[]` types hints at that being a non-nested array.

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+    }
+    return [];
+  });
+}
+
+/** Asserts that each route has at least one handler or component. */
+export function assertRoutesHaveHandlerOrComponent(
+  routes: Route[],
+): CheckResult[] {
+  return routes.flatMap((route) => {
```
marvinhagemeister: Should this be `.map`?

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+    const hasComponent = !!route.component;
+    let hasHandlerMethod = false;
+
+    if (typeof route.handler === "function") {
+      hasHandlerMethod = true;
+    }
```
marvinhagemeister:
```suggestion
    let hasHandlerMethod = typeof route.handler === "function"
```

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+}
+
+/** Asserts that here are no route conflicts with static files. */
+export function assertNoStaticRouteConflicts(
+  routes: Route[],
+  staticFiles: StaticFile[],
+): CheckResult[] {
+  const routePatterns = new Set(routes.map((route) => route.pattern));
+
+  return staticFiles.flatMap((staticFile) => {
```
marvinhagemeister: Should this be `.map`?

src/server/dev_checks.ts
```diff
@@ -0,0 +1,255 @@
...
+    if (typeof plugin.renderAsync === "function") {
+      try {
+        const renderAsyncSpy = spy(() => {
+          return Promise.resolve({
+            htmlText: "",
+            requiresHydration: false,
+          });
+        });
+
+        plugin.renderAsync({ renderAsync: renderAsyncSpy }).then(() => {});
```
marvinhagemeister: Nit: the `.then(() => {});` doesn't seem to do anything

src/server/dev_checks_test.ts
```diff
@@ -0,0 +1,368 @@
...
+    localUrl: new URL("http://localhost:3000/foo.bar"),
+    path: "/foo.bar",
+    size: 0,
+    contentType: "text/plain",
+    etag: "foo",
+    ...file,
+  };
+}
+
+Deno.test("assertModuleExportsDefault", async (t) => {
```
marvinhagemeister, line 66: Love all the tests here 👍

src/server/dev_checks.ts
mbhrznr: whoops, good catch!
i changed the logic to `.endsWith` and added a test case.

src/server/dev_checks.ts
mbhrznr: yes, yet each route initially returned `CheckResult[]`, which was the reason to have `.flatMap`.
however i homogenized each of the checks to have a `results` array to push to.

src/server/dev_checks.ts
mbhrznr: cleaned that up during the homogenization.

src/server/dev_checks.ts
mbhrznr: added a comment for the "why", as i wanted to keep each of the checks as sync operations.
however whenever i tried to convert it to something like this, i ran into timing issues, where the check resolved before the plugin was done:  
```ts
plugin.renderAsync({ renderAsync: renderAsyncSpy }).then(() => {
  assertSpyCalls(renderAsyncSpy, 1)
}).catch(() => {
  results.push({
    category: CheckCategory.PluginRenderAsync,
    message:
      `Plugin ${plugin.name} must call ctx.renderAsync() exactly once.`,
  });
})
```

deer:
Seems like this should eventually cover https://github.com/denoland/fresh/issues/1678#issuecomment-1683034081 as well.

mbhrznr:
> Seems like this should eventually cover https://github.com/denoland/fresh/issues/1678#issuecomment-1683034081 as well.

found a way to naively check the existence plugins in the current `BuildSnapshotJson`.  
thanks for the hint!

deer:
https://github.com/denoland/fresh/issues/1950#issuecomment-1768959095 is another example of a good check to add.

deer:
An addendum to https://github.com/denoland/fresh/issues/1558: we also need to think about conflicting islands as well. And if https://github.com/denoland/fresh/pull/2253 goes through, then also static files.

bartlomieju:
Closing this — Fresh 2.0 was a major rewrite and this PR targets the 1.x architecture, so the code no longer applies. The idea of dev-mode safety checks is still valuable and could be revisited for Fresh 2.x. Thanks for the thorough work on this!

PR #1570, opened 2023-08-01
chore(CI): rework lighthouse action

iuioiua:
This action is slightly simpler and more performant.

Reference: https://github.com/marketplace/actions/lighthouse-ci-action

### DIFF
--- a/.github/workflows/lighthouse.yml
+++ b/.github/workflows/lighthouse.yml
@@ -13,15 +13,15 @@ jobs:
       URL: ${{ github.event.deployment_status.environment_url }}
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Setup repo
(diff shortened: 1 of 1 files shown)


iuioiua:
Is having a Lighthouse action a requirement for this repo? Genuine question.

marvinhagemeister:
I don't think it's a requirement per se. When working on the website I tend to run it locally myself. But then again it's kinda nice to have maybe.

marvinhagemeister (approved):

iuioiua:
Perhaps, it'd be of greater justifiable value if we were to require performance budgets to be met. References:
* https://github.com/marketplace/actions/lighthouse-ci-action#budgetpath
* https://web.dev/use-lighthouse-for-performance-budgets

PR #1571, opened 2023-08-01
chore(CI): fix and simplify caching

iuioiua:
This change also removes the `deno cache ...` step, which isn't needed, and moves type checks to after caching is complete. This should reduce the amount of import errors we're seeing in CI.

Reference: https://deno.land/manual@v1.35.3/advanced/continuous_integration#speeding-up-deno-pipelines


marvinhagemeister (approved):
Nice 💯

PR #1573, opened 2023-08-02
chore: `deno task ok`

iuioiua:
This task checks formatting, linting, types and performs tests in one. Now, before committing, a contributor only needs to run `deno task ok` without worrying about each aspect of the repo is ok. This has worked great in Deno SaaSKit's repo.

If reasonable, this task should be used in CI to (almost) guarantee parity in checks between the local machine and CI.

Note to self: once this is merged, update the manual.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -1,18 +1,16 @@
 {
   "lock": false,
   "tasks": {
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Nice, I like that the test task only tests things now 👍

PR #1574, opened 2023-08-02
fix: typo `boder-box` to `border-box`

vrunt:
minor typographical error in init.ts setup, affects one line in style.css resulting in an invalid property


marvinhagemeister (approved):
Good catch! Not sure how this slipped through. Thank you for fixing this 💯

PR #1579, opened 2023-08-02
fix: add missing state property to interface

### LINKED ISSUE #1578: 404 page props are missing state
#1264 added the `state` property automatically to all pages (`PageProps`, `AppProps`, `UnknownPageProps`), since the code code was updated to automatically pass this in. But `state` was not added to the `UnknownPageProps` interface in the same PR, which seems like an oversight given that [`ctx.state` passed to and used in `renderNotFound`](https://github.com/denoland/fresh/blob/main/src/server/context.ts#L699).

It would make it a lot easier to use shared state data for missing pages if this was corrected.

csvn:
Corrects the props interface used in 404 pages to reflect that `state` is available as well.

Closes #1578

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -186,7 +186,7 @@ export interface LayoutRoute {
 // --- UNKNOWN PAGE ---
 
 // deno-lint-ignore no-explicit-any
-export interface UnknownPageProps<T = any> {
+export interface UnknownPageProps<T = any, S = Record<string, unknown>> {
   /** The URL of the request that resulted in this page being rendered. */
   url: URL;
 
@@ -199,6 +199,7 @@ export interface UnknownPageProps<T = any> {
    * `undefined`.
    */
   data: T;
+  state: S;
 }
 
 export interface UnknownHandlerContext<State = Record<string, unknown>>
--- a/tests/fixture/routes/_404.tsx
+++ b/tests/fixture/routes/_404.tsx
@@ -1,10 +1,19 @@
 import { UnknownPageProps } from "$fresh/server.ts";
 
-export default function NotFoundPage({ data, url }: UnknownPageProps) {
+type Data = { hello: string };
+type State = { root: string };
+
+export default function NotFoundPage(
+  { data, state, url }: UnknownPageProps<Data | undefined, State>,
+) {
+  // Checks that we have the correct type for state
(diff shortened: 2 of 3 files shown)


deer:
Thanks for taking the initiative to open a PR. What about adding a test to make sure this doesn't get broken in the future?

csvn:
@deer I could not find any references on how to test pure Typescript changes/types. There's no existing `types_test.ts` to add to. Would this mean creating a new `fixture_???` purely to check types? Should we create a `types_test.ts` that is just type checked and does something like below?

```ts
export default function NotFound(props: UnknownPageProps<{ foo: number }, { bar: boolean }>) {
  props.data.foo satisfies number;
  props.state.bar satisfies boolean;
}
```

The above would be enough to test the `UnknownPageProps` type, but not sure where to put it to be type checked.

deer:
I meant a functional test that confirms the state is actually passed into the test _404 page. Currently there's a `tests/fixture/routes/_404.tsx` file that you could possibly modify. Look at how the tests are done in the `main_test.ts` file. Perhaps you can piggyback on an existing `_middleware.ts` which sets the state, but try going to an invalid url. That should get the state set, and then route to the 404.

deer:
I meant a functional test that confirms the state is actually passed into the test _404 page. Currently there's a `tests/fixture/routes/_404.tsx` file that you could possibly modify. Look at how the tests are done in the `main_test.ts` file. Perhaps you can piggyback on an existing `_middleware.ts` which sets the state, but try going to an invalid url. That should get the state set, and then route to the 404.

csvn:
I gave it a try. I added some types for the generics, and updated the `_404.tsx` file.
Not sure if I should add a new test case, or updating existing one was enough

Let me know what you think 😄

deer (approved):
I think this testing approach is fine.

marvinhagemeister (approved):
LGTM, thanks for making a PR 👍

marvinhagemeister:
Looks like the CI is failing due to formatting issues. You can resolve these by running `deno fmt` locally and committing the changes.

csvn:
Sorry, thought I had format on automatically on save, and workflow did not when I made the PR. Hope fully it should be OK now 😀

marvinhagemeister (approved):

PR #1583, opened 2023-08-03
chore(CI): parallel tests

iuioiua:
Let's give this a few tries and see if we can weed out the root causes.


iuioiua:
5x CI runs without a single failure. It seems parallel testing and caching are working well! 🚀

marvinhagemeister (approved):
I've disabled parallel testing die to instability of the windows runner, but maybe the caching helps. Let's give this another go.

PR #1584, opened 2023-08-03
chore: format on save

iuioiua:
I feel as though this is a reasonable default to have. This eliminates one more thing that a contributor must remember to do before submitting a PR.


marvinhagemeister (approved):
Good idea 👍

PR #1591, opened 2023-08-04
fix(update): update script actually supports comments in jsonc

### LINKED ISSUE #1576: The `update.ts` script fails when the `deno.jsonc` files uses comments
Files with the `.jsonc` extension allow comments, but the update script can't cope with that because it tries to parse `deno.jsonc` with a normal JSON parser:

```
error: Uncaught (in promise) SyntaxError: Expected double-quoted property name in JSON at position 182
let denoJson = JSON.parse(denoJsonText);
                    ^
    at JSON.parse (<anonymous>)
    at https://deno.land/x/fresh@1.3.1/update.ts:52:21
    at eventLoopTick (ext:core/01_core.js:183:11)
```

Reading, modifying and writing back out the file while preserving comments is probably tricky. But I guess not preserving comments would be preferable to failing.

deer:
closes https://github.com/denoland/fresh/issues/1576


marvinhagemeister (approved):
This is great!

PR #1592, opened 2023-08-04
fix(server): make LayoutProps generic

### LINKED ISSUE #1580: Cannot override type for state in `AppProps` and `LayoutProps`
It's not possible to override the `state` property generic type in `AppProps`/`LayoutProps`. This makes us stuck with `Record<string, unknown>` as the type for `props.state`, even if we have a correct interface for use for state based on what global middleware adds.

```ts
export default function App(props: AppProps) {
  props.state.foo;
  // type: unknown
}
```

I suggest adding a generic to make it possible to configure the `state` type:

```ts
export default function App(props: AppProps<{ foo: string }>) {
  props.state.foo;
  // type: string
}
```

`PageProps` has generic for both `data` and `state`, but it does not make sense to be able to set a type for `data`, as neither App and Layout can load data, and it's the page that determines what data is available. Either we can keep the `any` type for `data`, or set it to something safer than `any`, like `unknown` or `undefined`.

```ts
export interface AppProps<S = Record<string, unknown>> extends PageProps<undefined, S> { ... }
export interface LayoutProps<S = Record<string, unknown>> extends PageProps<undefined, S> { ... }
```

This change should be backwards compatible, and allow typing `state` properly.

deer:
closes #1580

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -172,7 +172,9 @@ export interface AppModule {
   default: ComponentType<AppProps>;
 }
 
-export interface LayoutProps extends PageProps {
+// deno-lint-ignore no-explicit-any
+export interface LayoutProps<T = any, S = Record<string, unknown>>
+  extends PageProps<T, S> {
   Component: ComponentType<Record<never, never>>;
 }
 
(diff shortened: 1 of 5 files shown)


marvinhagemeister (approved):
Great work!

PR #1594, opened 2023-08-05
chore(server): simplify plugin route creation

### LINKED ISSUE #1560: simplify plugin route creation
#1545 recently simplified the way that plugin authors specify middleware. We should do a similar simplification for routes: don't require slashes at the beginning, as this is confusing. The plugin loader does this:
```tsx
`./routes${route.path}.ts`
```
but it should instead check to see if `route.path` has a leading slash, and add one if not present. We'll need tests for this new functionality, and to make sure the old style doesn't break.

@bjesuiter, I think this is the last of the work resulting from our discussion.

from: https://github.com/denoland/fresh/pull/1545#issuecomment-1658412159

deer:
closes #1560

As specified in the issue, we should have a simpler way of specifying routes. Forcing the leading slash is confusing and pointless.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -1289,10 +1289,14 @@ function getMiddlewareRoutesFromPlugins(
   return Object.values(mws);
 }
 
+function formatRoutePath(path: string) {
+  return path.startsWith("/") ? path : "/" + path;
+}
+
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):
Great idea!

PR #1597, opened 2023-08-05
feat: support async layout functions

marvinhagemeister:
This PR adds support for making Layouts behave similar to async route components. At some point I'll probably need to add proper async components in Preact, but for now this works and aligns well with the existing capabilities we have for routes.

```tsx
export default async function AsyncLayout(
  req: Request,
  ctx: RouteContext,
  { Component }: LayoutProps,
) {
  return (
    <div class="async-layout">
      <p>Async layout</p>
      <Component />
    </div>
  );
}
```

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -329,9 +329,8 @@ export async function render<Data>(
           component,
         ];
 
-        const componentStack: Array<VNode | null> = new Array(
-          renderStack.length,
-        ).map(() => null);
+        const componentStack: (VNode | null)[] = new Array(renderStack.length)
+          .fill(null);
 
         // deno-lint-ignore no-explicit-any
         const preparedLayouts: ComponentType<any>[] = [];
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -11,10 +11,12 @@ import {
 } from "preact";
 import {
   AppModule,
+  AsyncLayout,
   AsyncRoute,
   ErrorPage,
   Island,
   LayoutModule,
+  LayoutProps,
   Plugin,
   PluginRenderFunctionResult,
   PluginRenderResult,
@@ -137,6 +139,19 @@ function defaultCsp() {
   };
 }
 
+function getSyncPluginsMessage(plugins: Plugin[]) {
+  return `Async server components cannot be rendered synchronously. The following plugins use a synchronous render method: "${
+    plugins.map((plugin) => plugin.name).join('", "')
+  }"`;
+}
+
+function checkAsyncComponent(
+  component: unknown,
+): component is AsyncRoute | AsyncLayout {
+  return typeof component === "function" &&
+    component.constructor.name === "AsyncFunction";
+}
+
 /**
  * This function renders out a page. Rendering is synchronous and non streaming.
  * Suspense boundaries are not supported.
@@ -145,8 +160,8 @@ export async function render<Data>(
   opts: RenderOptions<Data>,
 ): Promise<[string, ContentSecurityPolicy | undefined] | Response> {
   const component = opts.route.component;
-  const isAsyncComponent = typeof component === "function" &&
-    component.constructor.name === "AsyncFunction";
+  const layouts = opts.layouts;
+  const isAsyncComponent = checkAsyncComponent(component);
 
   const props: Record<string, unknown> = {
     params: opts.params,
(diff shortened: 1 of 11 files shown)


tests/layouts_test.ts
```diff
@@ -64,3 +64,44 @@ Deno.test("check file types", async (t) => {
...
+        const doc = await fetchHtml(`${address}/async/redirect`);
+        assertSelector(
+          doc,
+          ".app .root-layout .async-layout .async-sub-layout .async-sub-page",
+        );
+      },
+    );
+  },
+  sanitizeOps: false,
+  sanitizeResources: false,
```
bartlomieju: Is there a reason why you need to disable sanitizers here?

bartlomieju (commented):
Can you add a test that exercises `getSyncPluginsMessage`?

tests/fixture_layouts/routes/async/redirect/_layout.tsx
```diff
@@ -0,0 +1,13 @@
+import { LayoutProps, RouteContext } from "$fresh/server.ts";
+
+export default async function AsyncSubLayout(
+  req: Request,
+  ctx: RouteContext,
+  { Component }: LayoutProps,
+) {
+  await new Promise((r) => setTimeout(r, 10));
+  return new Response("", {
```
iuioiua:
```suggestion
  return new Response(null, {
```
Nit

tests/fixture_layouts/routes/async/redirect/_layout.tsx
```diff
@@ -0,0 +1,13 @@
+import { LayoutProps, RouteContext } from "$fresh/server.ts";
+
+export default async function AsyncSubLayout(
+  req: Request,
+  ctx: RouteContext,
+  { Component }: LayoutProps,
+) {
+  await new Promise((r) => setTimeout(r, 10));
```
iuioiua:
```suggestion
  await delay(10);
```
Nit. And other instances. Reference: https://deno.land/std@0.197.0/async/mod.ts?s=delay

src/server/render.ts
```diff
@@ -137,6 +139,19 @@ function defaultCsp() {
...
+  return `Async server components cannot be rendered synchronously. The following plugins use a synchronous render method: "${
+    plugins.map((plugin) => plugin.name).join('", "')
+  }"`;
+}
+
+function checkAsyncComponent(
+  component: unknown,
+): component is AsyncRoute | AsyncLayout {
+  return typeof component === "function" &&
+    component.constructor.name === "AsyncFunction";
```
iuioiua, line 146:
```suggestion
  return component.constructor.name === "AsyncFunction";
```

src/server/render.ts
```diff
@@ -190,16 +205,41 @@ export async function render<Data>(
...
+    });
+  }
+
+  function renderInnerSync(vnode: ComponentChildren): string {
     // deno-lint-ignore no-explicit-any
     let finalAppComp: VNode<any> = vnode as any;
 
+    if (layouts.some((layout) => checkAsyncComponent(layout.default))) {
+      throw new Error(getSyncPluginsMessage(syncPlugins));
+    }
```
iuioiua: Is this assertion able to be moved to earlier in the order of execution? Genuine question.

src/server/render.ts
```diff
@@ -284,28 +304,93 @@ export async function render<Data>(
...
+        };
+
+        const renderStack = [
+          ...layouts.map((layout) => layout.default),
+          component,
+        ];
+
+        const componentStack: Array<VNode | null> = new Array(
+          renderStack.length,
+        ).map(() => null);
```
```suggestion
        const componentStack: (VNode | null)[] = new Array(renderStack.length)
          .fill(null);
```
Nit

iuioiua (approved):
LGTM. Just a couple of nits.

tests/layouts_test.ts
iuioiua: Tests pass when I removed these options on my local machine.

src/server/render.ts
marvinhagemeister: `route.component` can be `undefined` in the case a route only has handlers.

tests/layouts_test.ts
marvinhagemeister: Good point. Probably copy & pasted it from somewhere else.

src/server/render.ts
marvinhagemeister: Good call out. Just noticed that this check doesn't make sense here it all. We always start rendering by calling into `renderAsync()` and fall back to sync rendering from there. This means we already run into the other check for async routes + layouts and would never reach this here with an async layout anyway.

PR #1599, opened 2023-08-06
feat(server): support https

### LINKED ISSUE #901: Support for TLS, SSL and HTTPS missing from Deno.Serve

deer:
closes https://github.com/denoland/fresh/issues/901, closes https://github.com/denoland/fresh/issues/809

We currently have (all referenced and relevant types included for clarity):
```tsx
export type StartOptions = ServeInit & FreshOptions;

export interface ServeInit extends Partial<Deno.ListenOptions> {
  signal?: AbortSignal;
  onError?: (error: unknown) => Response | Promise<Response>;
  onListen?: (params: { hostname: string; port: number }) => void;
}

export interface ListenOptions {
  port: number;
  hostname?: string;
}

```

And after this change we have (all referenced and relevant types included for clarity):
```tsx
export type StartOptions = Partial<Deno.ServeTlsOptions> & FreshOptions;

export interface ServeTlsOptions extends ServeOptions {
  cert: string;
  key: string;
}

export interface ServeOptions {
  port?: number;
  hostname?: string;
  signal?: AbortSignal;
  reusePort?: boolean;
  onError?: (error: unknown) => Response | Promise<Response>;
  onListen?: (params: { hostname: string; port: number }) => void;
}
```

So this change adds the necessary `cert` and `key` to the options bag, and the unnecessary (at least for this change, but overall possibly beneficial) `reusePort`.

Now that we're type checking the project, I created a "test" for this by introducing a `main_tls.ts` file which tries to start a fresh server with `cert` and `key` specified. We'll rely on the deno runtime tests themselves to confirm this works, but this at least asserts that the properties are passable. Note that we get a type error if the actual change here is reverted, proving that https isn't currently supported.

### DIFF
--- a/src/server/deps.ts
+++ b/src/server/deps.ts
@@ -12,7 +12,6 @@ export {
   type Handler as ServeHandler,
   serve,
 } from "https://deno.land/std@0.193.0/http/server.ts";
-export type { ServeInit } from "https://deno.land/std@0.193.0/http/server.ts";
 export { Status } from "https://deno.land/std@0.193.0/http/http_status.ts";
 export {
   typeByExtension,
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -1,11 +1,10 @@
 import { ComponentChildren, ComponentType, VNode } from "preact";
-import { ServeInit } from "./deps.ts";
 import * as router from "./router.ts";
 import { InnerRenderFunction, RenderContext } from "./render.ts";
 
 // --- APPLICATION CONFIGURATION ---
 
-export type StartOptions = ServeInit & FreshOptions;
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):
LGTM, neat that this turned out to be only a typing issue 👍

PR #1600, opened 2023-08-06
feat: co-locate islands + non-route files

### LINKED ISSUE #1486: Allow putting islands + routes + components in the same folder
Currently, the way Fresh works is that it scans for an `islands/` and a `routes/` folder in which Islands and Routes have to live respectively. This grouping by purpose is fine for smaller projects, but grouping by feature makes a lot more sense. Putting everything related to a certain route in the same folder makes a lot of sense.

Another motivation for this feature is that I notice folks putting everything in one single route file as a workaround to having to put the code outside of `routes/`.

We have the distinct notion of the `islands/` and `routes/` folder, because it signals to Fresh what to bundle for the browser and what routes to resolve.

The details on this still need to be worked out.

marvinhagemeister:
This PR allows non-route files to be co-located in the `routes/` directory. Whenever we encounter a folder named `(_foo)` where `foo` can be any word, we'll skip that when crawling the file system for routes. When such a folder is named `(_islands)`, we'll treat the files inside it as islands.

```sh
/routes
  /(_islands)     # files treated as islands
    ...
  /(_components)  # ignored, no special meaning
    ...
  /index.tsx      # your typical route
``` 

Fixes #1486

### DIFF
--- a/src/dev/deps.ts
+++ b/src/dev/deps.ts
@@ -13,7 +13,11 @@ export {
 } from "https://deno.land/std@0.193.0/path/mod.ts";
 export { DAY, WEEK } from "https://deno.land/std@0.193.0/datetime/constants.ts";
 export * as colors from "https://deno.land/std@0.193.0/fmt/colors.ts";
-export { walk, WalkError } from "https://deno.land/std@0.193.0/fs/walk.ts";
+export {
+  walk,
+  type WalkEntry,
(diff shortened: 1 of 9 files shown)


bartlomieju (approved):
LGTM, this seems super useful 👍

lino-levan:
Really like this. Nice work!

PR #1601, opened 2023-08-06
Showcase Github Land

roberto-morado:


marvinhagemeister:
I'm hesitant to merge this as the site uses the GitHub branding very liberally. I'm worried that we'd open ourselves up for potential trademark issues by having it on our site.

PR #1450, opened 2023-07-11
Allow prebuilding an ESBuild snapshot

zaynetro:
These changes allow users to define a build step where they can prebuild the assets prior to running the server. An attempt to solve https://github.com/denoland/fresh/issues/1062

I thought that a good first step would be to expose the necessary functionality to support prebuilding the assets. Users are free to implement their own snapshot persistence strategies (at least for now).

<details> 
<summary>Generate a snapshot example</summary>

```ts
// build.ts
import manifest from "@/fresh.gen.ts";
import * as base64 from "$std/encoding/base64.ts";
import { ServerContext } from "$fresh/server.ts";

const ctx = await ServerContext.fromManifest(manifest, {});
// Bundle assets
const snapshot = await ctx.buildSnapshot();

// Prepare a JSON
const files: Record<string, string> = {};
for (const p of snapshot.paths) {
  const data = snapshot.read(p) as Uint8Array;
  const v = base64.encode(data);
  files[p] = v;
}

const deps: Record<string, string[]> = {};
for (const p of snapshot.paths) {
  const v = snapshot.dependencies(p);
  deps[p] = v;
}

const content = JSON.stringify(
  { files, deps },
  null,
  2,
);

// Write to disc
await Deno.writeTextFile("snapshot.json", content);
```

```ts
// main.ts
import { start } from "$fresh/server.ts";
import manifest from "@/fresh.gen.ts";

import { BuildSnapshot } from "$fresh/src/build/mod.ts";
import * as base64 from "$std/encoding/base64.ts";

// Load the snapshot
const snapshotText = await Deno.readTextFile("snapshot.json");
const snapshotJson = JSON.parse(snapshotText);

// Implement BuildSnapshot interface
export const snapshot: BuildSnapshot = {
  get paths() {
    return Object.keys(snapshotJson.files);
  },

  read(path: string): Uint8Array | null {
    const v = snapshotJson.files[path];
    if (!v) {
      return null;
    }

    return base64.decode(v);
  },

  dependencies(path: string): string[] {
    return snapshotJson.deps[path] ?? [];
  },
};

// Pass snapshot to the server
await start(manifest, {
  snapshot,
});
```

Now before deployment you can run `deno run -A build.ts` which will generate a `snapshot.json` file that could be used in `main.ts` to load a build snapshot and skip ESBuild phase at runtime.

Use this import URL:

```diff
-    "$fresh/": "https://deno.land/x/fresh@1.2.0/",
[...]

PR #1604, opened 2023-08-07
feat: prebundle code + restore from snapshot

### LINKED ISSUE #1062: Pre-build islands

marvinhagemeister:
This PR allows you to prebundle all islands and other code needed for the browser ahead of time. The workflow with this is that you have a GitHub action where you build the assets and then deploy from there.

Pros:
- Files are served as quick as deploy + fresh can serve them
- No need to fetch dependencies when an isolate spins up
- Makes deployments reproducible since they cannot be affected by outages from third parties like esm.sh anymore
- The bigger the island code the more noticeable the speed improvement

Cons:
- Makes deployments 10-15s longer
- Requires GH Actions to be set up

Assets can be generated by running `deno task build` and the `main.ts` script will pick up the snapshots automatically if one is there. When it does it prints a message to the terminal.

Screenshot 2023-08-07 at 17 47 24

The speed difference is best shown with a video. In the ahead of time (AOT) bundled version the agolia search loads nearly instantaneously. Go to https://fresh-www-test.deno.dev/ to see that for yourself.

https://github.com/denoland/fresh/assets/1062408/a7191855-bd5b-423e-a36b-2bdde763660f

The main difference to #1450 is, that this accounts for plugins and includes a way to generate the bundles + snapshots directly in Fresh.

Fixes #1062


lino-levan:
This is a very similar approach to what I did in #1084 for a PoC. I think this is the right way to do it. +1 for the concept.

marvinhagemeister:
Gonna do the documentation update in another PR since the deploy page needs an overall in general.

src/dev/build.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    return Deno.writeFile(join(outDir, fileName), data);
+  }));
+
+  // Write dependency snapshot file to disk
+  const deps: Record<string, string[]> = {};
+  for (const filePath of snapshot.paths) {
+    const dependencies = snapshot.dependencies(filePath);
+    deps[filePath] = dependencies;
+  }
```
bartlomieju, line 39: Is the ordering here deterministic? Ie. will no change cause a potential snapshot file change?

src/server/context.ts
```diff
@@ -157,6 +160,40 @@ export class ServerContext {
       );
     }
 
+    // Restore snapshot if available
+    let snapshot: BuildSnapshot | null = null;
+    // Load from snapshot if not explicitly requested not to
+    const loadFromSnapshot = opts.loadSnapshot ?? true;
+    if (loadFromSnapshot) {
```
bartlomieju: Is this actually backwards compatible?

src/dev/build.ts
lino-levan: We could always sort the paths to ensure determinism. They may be deterministic in one build of esbuild but not in another.

src/server/context.ts
marvinhagemeister: Not sure what you mean. This is the first time we're exposing the snapshot.

fdemir:
We should add this to the documentation.

bartlomieju (approved):
LGTM 🚀

marvinhagemeister:
Added a page to our documentation and I think the PR is ready to be merged.

lino-levan:
Let me take a look!

.gitignore
```diff
@@ -0,0 +1,4 @@
+_fresh/
+deno_modules/
```
lino-levan: Not sure why you include this, given it was only named that for one unstable version? I would remove this to avoid any confusion.

docs/canary/concepts/ahead-of-time-builds.md
```diff
@@ -0,0 +1,135 @@
...
+To have Fresh optimize all the assets, run one of the following commands:
+
+```sh
+# As a task in newer Fresh projects
+deno task build
+# or invoke it manually
+deno run -A dev.ts build
+```
+
+This will create a `_fresh/` folder in the project directory. That folder
```
lino-levan: Should this be `_fresh` and not the static prefix?

init.ts
```diff
@@ -105,6 +105,9 @@ const GITIGNORE = `# dotenv environment variable files
 .env.test.local
 .env.production.local
 .env.local
+
+# Fresh build directory
+fresh/
```
lino-levan: There isn't an underscore here.

src/build/esbuild.ts
```diff
@@ -163,7 +163,7 @@ export class EsbuildSnapshot implements BuildSnapshot {
   }
 
   get paths(): string[] {
-    return Object.keys(this.#files);
+    return Array.from(this.#files.keys());
```
lino-levan, line 166: What is this change about?

src/dev/build.ts
```diff
@@ -0,0 +1,42 @@
...
+import { dirname, fromFileUrl, join } from "../server/deps.ts";
+import { fs } from "./deps.ts";
+
+export async function build(
+  manifestPath: string,
+  opts: FreshOptions,
+) {
+  const manifest = (await import("file://" + manifestPath)).default as Manifest;
+
+  const outDir = join(dirname(fromFileUrl(manifest.baseUrl)), "_fresh");
```
lino-levan, line 13: Again, should this be the prefix?

src/dev/build.ts
```diff
@@ -0,0 +1,42 @@
+import { ServerContext } from "../server/context.ts";
+import { FreshOptions, Manifest } from "../server/mod.ts";
+import { dirname, fromFileUrl, join } from "../server/deps.ts";
+import { fs } from "./deps.ts";
+
+export async function build(
+  manifestPath: string,
+  opts: FreshOptions,
+) {
+  const manifest = (await import("file://" + manifestPath)).default as Manifest;
```
lino-levan: Is this safe to do? This feels a bit janky.

tests/build_test.ts
```diff
@@ -0,0 +1,89 @@
+import * as path from "$std/path/mod.ts";
```
lino-levan, line 1: praise: nice test

tests/build_test.ts
```diff
@@ -0,0 +1,89 @@
...
+      );
+    });
+
+    await t.step("restore from snapshot", async () => {
+      const { lines, serverProcess, address, output } = await startFreshServer({
+        args: [
+          "run",
+          "--allow-env",
+          "--allow-read",
+          "--allow-net=0.0.0.0",
```
lino-levan: Is this a safe `--allow-net`? I have bad memories around `0.0.0.0`, not sure if they apply here.

src/build/esbuild.ts
marvinhagemeister: Calling `Object.keys()` on a `Map` gives you the keys of the `Map` object itself and not the keys of the data it holds. I'm not sure how this bug has survived this long

lino-levan: huh, nice catch!

tests/build_test.ts
marvinhagemeister: I'm admittedly not sure why deno needs this. Normally `127.0.0.1` or `localhost` should suffice, but when I use these I get the following error:

```txt
error: Uncaught (in promise) PermissionDenied: Requires net access to "0.0.0.0:8000", run again with the --allow-net flag
    await Deno.serve(opts, handler).finished;
```

lino-levan: Interesting. Seems this works fine then, I'll keep an eye on it for the future.

src/dev/build.ts
marvinhagemeister: Good callout, yeah that should've been `toFileUrl`.

lino-levan:
Nice! Excited to have this land.

PR #1606, opened 2023-08-08
chore: add test for checking initialized project

deer:
This is, I think, the natural followup to https://github.com/denoland/fresh/pull/1539 and https://github.com/denoland/fresh/pull/1454. Now that we have an easy way to type check the initialized project, we should assert that there are no type errors in that project.

This adds three seconds on my slow computer, but I think it's worth it; releasing code with red squiggles or linting or formatting errors is a bit silly.


marvinhagemeister (approved):
LGTM 👍

PR #1611, opened 2023-08-09
feat: allow layout and _app overrides

marvinhagemeister:
This PR adds two new options to route configuration to override previously inherited layouts or skip the `_app` template.

```tsx
export const config: RouteConfig = {
  rootLayout: true, // Mark this route as the root layout
  appTemplate: false // Disable `_app` template
};

export default function MyPage() {
  return <p>Hello world</p>;
}

### DIFF
--- a/docs/canary/concepts/route-layout.md
+++ b/docs/canary/concepts/route-layout.md
@@ -15,8 +15,8 @@ routes/
     index.tsx
     page.tsx
   /other
-    _layout.tsx
-    page.tsx # will be applied on top of `routes/_layout.tsx`
+    _layout.tsx # will be applied on top of `routes/_layout.tsx`
+    page.tsx
 ```
 
 The component to be wrapped is received via props, in addition to a few other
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by marvinhagemeister** [comment]
Ah looks like I need to take another look at the tests.

**Comment by marvinhagemeister** [comment]
Alright, addressed the test failures.

File: docs/canary/concepts/route-layout.md [hunk 1]
```diff
@@ -7,6 +7,18 @@ An route layout is defined in a `_layout.tsx` file in any sub directory (at any
...
+```sh
+routes/
+  _app.tsx
+  _layout.tsx # will be applied to all routes
+  /sub
+    index.tsx
+    page.tsx
+  /other
+    _layout.tsx
+    page.tsx # will be applied on top of `routes/_layout.tsx`
```
bartlomieju: Was this comment supposed to be here or next to `_layout.tsx`?

marvinhagemeister: oops, yeah that's supposed to be one line above that.

bartlomieju (approved):
LGTM!

Issue #1582, opened 2023-08-03
Islands.md needs a bit of love.


rawlingsr:
I was reading through the fresh documentation, more or less in order, and all was well up til I got to https://fresh.deno.dev/docs/concepts/islands

I found 3 problems:
1. MyIsland.tsx is the file name of the first island component but later the file is imported as my-island.tsx
This commit should resolve the issue: https://github.com/setibeings/fresh/commit/8ca01f64746627b96cd942f48dd15d1e3ba19a11

2. The "Passing JSX to islands" example listed appears to require a bit of code to be present on the island in order to receive the children as props
A bit of tinkering led me to this solution:
https://github.com/setibeings/fresh/commit/778d5852d3a36d83020cdea4653ae2a02309a1c1

3. I'm wondering if the second half of the section expresses an idea separate from the idea of client side components existing to add interactivity. That (p)react components which are islands can have children makes them similar to other components, I think, so perhaps this part belongs in a new section in the concepts chapter. I'm not sure how to cover the main points of that new section, if that's what's needed, or how to shorten this one while still covering the essentials.

I just wanted to call out that this stack overflow post was really helpful:
https://stackoverflow.com/questions/67144030/type-children-element-has-no-properties-in-common-with-type-intrinsicat
I found this question because of the error "Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'."

marvinhagemeister:
@setibeings It looks like you've already made some commits for this in your fork. What do you think about PR'ing those fixes here?

PR #1614, opened 2023-08-10
docs(concepts) make islands.md examples run without errors

rawlingsr:
Addresses https://github.com/denoland/fresh/issues/1582

I've tried to break up my changes so that only the first one or two commits can be taken if they are the only ones that are helpful.

### DIFF
--- a/docs/latest/concepts/islands.md
+++ b/docs/latest/concepts/islands.md
@@ -108,6 +108,18 @@ enough information to put the islands where they are supposed to go without
 requiring hydration for the static children of interactive islands. No
 Javascript is sent to the client when no interactivity is needed.
 
+```HTML
+<!--frsh-myisland_default:default:0-->
+<div>
+  Counter is at 0. 
+  <button>+</button>
+  <!--frsh-slot-myisland_default:children-->
+  <p>This text is rendered on the server</p>
+  <!--/frsh-slot-myisland_default:children-->
+</div>
+<!--/frsh-myisland_default:default:0-->
+```
+
 ### Nesting islands
 
 Islands can be nested within other islands as well. In that scenario they act
--- a/docs/latest/concepts/islands.md
+++ b/docs/latest/concepts/islands.md
@@ -15,20 +15,14 @@ island.
 // islands/my-island.tsx
 
 import { useSignal } from "@preact/signals";
-import { VNode } from "preact";
 
-interface Data {
-  children: VNode<Element>;
-}
-
-export default function MyIsland({ children }: Data) {
+export default function MyIsland() {
   const count = useSignal(0);
 
   return (
     <div>
       Counter is at {count}.{" "}
       <button onClick={() => (count.value += 1)}>+</button>
-      {children}
     </div>
   );
 }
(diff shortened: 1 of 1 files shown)


docs/latest/concepts/islands.md
```diff
@@ -31,6 +31,60 @@ export default function MyIsland() {
...
+// islands/my-island.tsx
+
+import { useSignal } from "@preact/signals";
+import { VNode } from "preact";
+
+interface Data {
+  children: VNode<Element>;
+}
+
+export default function MyIsland({ children }: Data) {
```
marvinhagemeister:
```suggestion
import { ComponentChildren } from "preact";

interface Props {
  children: ComponentChildren;
}

export default function MyIsland({ children }: Props) {
```

docs/latest/concepts/islands.md
```diff
@@ -77,18 +113,32 @@ Islands can be nested within other islands as well. In that scenario they act
...
+interface Data {
+  children: VNode<Element>;
+  foo: string;
+}
+
+function randomNumber() {
+  return Math.floor(Math.random() * 100);
+}
+
+export default function MyIsland({ children, foo }: Data) {
```
marvinhagemeister:
```suggestion
import { ComponentChildren } from "preact";

interface Props {
  children: ComponentChildren;
  foo: string;
}

function randomNumber() {
  return Math.floor(Math.random() * 100);
}

export default function MyIsland({ children, foo }: Props) {
```

marvinhagemeister (commented):
Thanks for making a PR! This is great!!

Left two minor nit comments because the first argument to components is usually called `Props` not  `Data`. It's more of a convention not a hard rule, but that makes it easier for folks coming from other frameworks.

The correct type for `children` is `ComponentChildren` in Preact instead of `VNode<Element>`

rawlingsr:
@marvinhagemeister Thank you for the changes, and for the feedback. 

I wanted to make one other change, but I am not sure what the text should say instead of what it says now. There is a paragraph that talks about not needing to serialize children: https://github.com/setibeings/fresh/blob/eafe8a901a3c9496561cc6b0059882a44e649747/docs/latest/concepts/islands.md?plain=1#L105C1-L109C1 

It uses the pronoun "we" which I think could trip somebody up if they aren't sure if they're reading about something that THEY should be able to tell by looking at the HTML returned by the server.

marvinhagemeister:
@setibeings Good catch. Looks like that is a copy & paste error from the blog post on my part. That section is meant to refer to Fresh the framework itself.

rawlingsr:
I'm not sure if this is the blog post, but I'll just plan on reading it today and take a crack at rephrasing the paragraph I mentioned. 

https://deno.com/blog/intro-to-islands

marvinhagemeister:
@setibeings it references a section from this blog post https://deno.com/blog/fresh-1.2#passing-jsx-to-islands-and-nesting-islands-within-each-other

marvinhagemeister:
Looks like there is a formatting error. Running `deno fmt` locally will resolve these and make the CI green 👍

rawlingsr:
Converting to draft because I now want to add an HTML Example.

rawlingsr:
Take a look at what I wrote and let me know if it matches what you think when you get a moment. Thanks for bearing with me.

marvinhagemeister (approved):
LGTM, thanks for making the PR 💯

PR #1617, opened 2023-08-10
feat: Support setting `<html>` in `_app` template

### LINKED ISSUE #1529: Allow head + body creation in `_app.tsx`
This week I've come across a couple of fresh projects which add the `<html>`, `<body>` and `<head>` tag respectively inside `_app.tsx`. Most other frameworks like Remix, astro or nextjs work that way, so it's a natural assumption that this would work in Fresh too.

Example:

```tsx
// _app.tsx
  return (
    <html class="h-full bg-[#010524] overflow-hidden">
      <head>
        <title>Just Lists</title>
        <meta name="title" content="" />
        <meta name="description" content="" />
      </head>
      <body class="h-full text-slate-600">
        <Component />
      </body>
    </html>
  )
}
```

### LINKED ISSUE #1595: Setting attributes on `<html>` element
Hi,

I noticed that it is not possible to use async server components with a custom render function. But why? I'm providing my own render function to set the HTML language to 'de':

```typescript
await start(manifest, {
  plugins: [kvInsightsPlugin()],
  render: (ctx, render) => {
    ctx.lang = 'de';
    render();
  },
  router: {
    trailingSlash: true,
  },
});
```

Is there maybe another way to set the language?

marvinhagemeister:
This PR allows us to create the `<html>`-tag ourselves in `_app.tsx`. In case we didn't render any `<html>` tag we fall back to the current way of adding an internal outer template. This allows you to set the `lang` attribute on `<html>` without having to go through various abstractions, etc.

```tsx

  return (
    <html lang="de">
      <head>
        ...
      </head>
      <body>
        <Component />
      </body>
    </html>
  );
}
```

I tried various approaches to achieve this like adding a marker and replacing that in the HTML string. That proved to be bad for security as the marker could be injected by users. So I landed on the approach in this PR where we essentially swap out user supplied `<html>`, `<head>` and `<body>` tags with an invisible Fragment. We store the props of the old vnode and after the body has rendered we'll render our internal outer template with the user supplied values. Basically this works close to how it worked before.

Fixes #1529, fixes #1595


bartlomieju (commented):
From an initial review this looks good to me, but I must admit I'm not so knowledgable about this part. I would appreciate another pair of eyes for an approval.

src/server/render.ts
```diff
@@ -209,6 +233,17 @@ export async function render<Data>(
   // Clear previous slots
   SLOTS_TRACKER.clear();
 
+  // Clear rendering state
+  RENDERING_USER_TEMPLATE = false;
```
fdemir, line 237: RENDERING_USER_TEMPLATE should not be defined as screaming snake case. it confuses the reader, sounds constant.

marvinhagemeister: It's in uppercase to signal that it's a global variable in the context of the module and not a local one. I can rename it though. I'm not attached to the name

marvinhagemeister:
Merging, as I got verbal approval internally.

lino-levan:
Looks great. Excited to be moving in the right direction!

PR #1619, opened 2023-08-11
fix: rename layout config options

marvinhagemeister:
This addresses some feedback on the naming of the layout config options. Whilst changing that I noticed that we don't have a test for when a layout disables the `_app` template.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -253,8 +253,8 @@ export class ServerContext {
           component,
           handler,
           csp: Boolean(config?.csp ?? false),
-          appLayout: Boolean(config?.appTemplate ?? true),
-          rootLayout: Boolean(config?.rootLayout ?? false),
+          appTemplate: Boolean(config?.appTemplate ?? true),
+          inheritLayouts: Boolean(config?.inheritLayouts ?? true),
         };
         routes.push(route);
       } else if (isMiddleware) {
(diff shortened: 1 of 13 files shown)


src/server/context.ts
```diff
@@ -253,8 +253,8 @@ export class ServerContext {
           component,
           handler,
           csp: Boolean(config?.csp ?? false),
-          appLayout: Boolean(config?.appTemplate ?? true),
-          rootLayout: Boolean(config?.rootLayout ?? false),
+          appTemplate: Boolean(config?.appTemplate ?? true),
+          inheritLayouts: Boolean(config?.inheritLayouts ?? true),
```
lucacasonato: True by default is very confusing. Can we invert the naming so it is `false` by default? Same for `appLayout`.

lucacasonato: https://www.w3.org/TR/design-principles/#optional-parameters

tests/layouts_test.ts
```diff
@@ -120,6 +120,17 @@ Deno.test("disable _app layout", async () => {
   );
 });
 
+Deno.test("disable _app in _layout", async () => {
```
lino-levan, line 123: praise: thank you for the test!

src/server/context.ts
marvinhagemeister: @lucacasonato Good point. Renamed it to `skipAppTemplate` and `skipInheritedLayouts` respectively.

bartlomieju (approved):
LGTM

PR #1623, opened 2023-08-12
www: add language switcher to code blocks

marvinhagemeister:
This PR spices up the code blocks quite a bit. It adds a language selector so that you can switch between `TypeScript/JavaScript` code. The setting is persisted in local store and is applied globally to all code blocks.

On top of that the PR adds the ability to add a title to a code block so that it tells the user what they're looking at. In most cases that will be the file name.

The marked smartypants plugin doesn't work with the latest marked version. The previous plugin would parse the final generated HTML again anyway instead of just tapping into marked's parser. So I've hand rolled our own smartypants.

https://github.com/denoland/fresh/assets/1062408/518b6d3c-0697-4a63-b397-92395925ca31

### DIFF
--- a/docs/canary/concepts/middleware.md
+++ b/docs/canary/concepts/middleware.md
@@ -17,9 +17,8 @@ special [\_app](/docs/concepts/app-wrapper.md) wrapper and normal
 properties, e.g. `ctx.state.loggedIn = true`, but you can also replace the
 entire object like `ctx.state = { loggedIn = true }`.
 
-```ts
-// routes/_middleware.ts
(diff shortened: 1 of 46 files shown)


lino-levan:
Would be cool if there was a copy button, giving that we're rolling our own codeblocks anyways.

bartlomieju:
The docs, www site structure, and markdown rendering have all been rewritten for Fresh 2.x. The code examples use old 1.x patterns (`$fresh/server.ts`, `MiddlewareHandlerContext`, etc.) and the www routes (`www/routes/_app.tsx`, `www/fresh.gen.ts`) no longer match the current structure.

Closing as stale.

PR #1624, opened 2023-08-12
www: remove counters in sidebar

marvinhagemeister:
Small design change that removes the decimal numbered list in the sidebar. Always found they include visual noise and don't really have a purpose.

Before:

| Before | After |
|---|---|
| Screenshot 2023-08-12 at 16 00 32 | Screenshot 2023-08-12 at 16 00 36 |


fdemir (approved):

iuioiua:
Glad to see this has been done!

PR #1626, opened 2023-08-12
feat: support async app template

marvinhagemeister:
This PR adds support for async `_app` templates. It was the last thing which didn't support that. Now all routes, layouts or the app template all behave the same internally. In doing so I noticed that we can make the rendering function a lot simpler and make a bunch of error checks redundant.


marvinhagemeister:
FYI: I've simplified a bit the async route component TS types too, since they all have the same args, except for a route not having `Component`

tests/fixture_layouts/routes/async/_layout.tsx
```diff
@@ -1,15 +1,14 @@
-import { LayoutProps, RouteContext } from "$fresh/server.ts";
+import { LayoutContext } from "$fresh/server.ts";
 
 export default async function AsyncLayout(
   req: Request,
-  ctx: RouteContext,
-  { Component }: LayoutProps,
+  ctx: LayoutContext,
 ) {
   await new Promise((r) => setTimeout(r, 10));
```
iuioiua:
```suggestion
  await delay(10);
```
Nit. Just for consistency.

src/server/types.ts
```diff
@@ -162,7 +170,7 @@ export interface Route<Data = any> {
   pattern: string;
   url: string;
   name: string;
-  component?: PageComponent<Data>;
+  component?: PageComponent<Data> | AsyncLayout | AsyncRoute;
```
iuioiua:
```suggestion
  component?: PageComponent<Data> | AsyncLayout<Data> | AsyncRoute<Data>;
```

iuioiua (commented):
I don't entirely understand the implementation but spotted a couple nits 🙂

tests/fixture_layouts/routes/async/_layout.tsx
marvinhagemeister: At some point I'll remember that we have this in our standard library 👀

bartlomieju (approved):

PR #1627, opened 2023-08-12
feat: add helpers to make Fresh less TypeScripty

### LINKED ISSUE #1608: bug: type checks fail when type variables are passed to `RouteContext` in async route component
See https://github.com/denoland/saaskit/pull/411

marvinhagemeister:
Had an idea to make async route component less TypeScripty by not requiring you to add all the type arguments yourself. Similar to vite's `defineConfig` function, the `defineRoute` function is mostly for automatic type inference.

Before:

```tsx
import { RouteContext } from "$fresh/server.ts";

export default async function Page(req: Request, ctx: RouteContext) {
 const name = await fetchSomeData()

  return (
    <div class="home-page">
      <p>Hello {name}</p>
    </div>
  );
});
```

After:

```tsx
import { defineRoute } from "$fresh/server.ts";

export default defineRoute(async (req, ctx) => {
 const name = await fetchSomeData()

  return (
    <div class="home-page">
      <p>Hello {name}</p>
    </div>
  );
});
```

It doesn't seem like much, but it makes quickly hacking on Fresh a lot more enjoyable because you don't have to look up the correct types all the time anymore.

Fixes #1608


marvinhagemeister:
Needed to widen the manifest types because TS cannot match based on arity. See https://twitter.com/drosenwasser/status/1691130933704036352 and related issue https://github.com/microsoft/TypeScript/issues/55370.

I think it's fine since we generate the manifest ourselves anyway. As a bonus this fixes #1608

bartlomieju (approved):
LGTM, very useful 👍

PR #1629, opened 2023-08-13
www: new fenced code block design

marvinhagemeister:
This extracts the design bits out of #1623 so that this can be merged in advance. The language switcher in #1623 itself needs a bit more work.

Screenshot 2023-08-13 at 10 43 57


marvinhagemeister:
Merging myself, since this PR is only a minor UX change for the docs site.

PR #1630, opened 2023-08-13
chore: speed up some tests

marvinhagemeister:


marvinhagemeister:
Merging, because this just makes tests faster.

PR #1635, opened 2023-08-15
www: Various documentation improvements

marvinhagemeister:
- We call `_app.tsx` app wrapper in our docs but in the code they were called `appTemplate`. Switched that over to `appWrapper` everywhere
- Updated the `Layouts` docs and call it `Layouts` instead of `Route Layouts`
- Delete the `setting the lang` page in the canary docs as that was only necessary in older Fresh version's, because you couldn't render the `<html>` tag yourself. With Fresh 1.4 you can.
- Merge `Route Groups` doc into `Routing`.
- Change title of middleware docs to `Middlewares` instead of `Route middlewares`.
- Add documentation for `defineApp`, `defineLayout` and `defineRoute`


docs/toc.ts
```diff
@@ -72,7 +71,6 @@ const toc: RawTableOfContents = {
         link: "latest",
         pages: [
           ["modifying-the-head", "Modifying the <head>", "link:latest"],
-          ["setting-the-language", "Setting the language", "link:latest"],
```
deer, line 75: Is setting the language now such an obvious thing that the documentation should be deleted? Or could it be beneficial to completely redo the example to account for the new functionality?

docs/latest/concepts/routing.md
```diff
@@ -51,4 +51,60 @@ export const config: RouteConfig = {
 // ...
 ```
 
+## Route Groups
+
+When working with [route layouts](/docs/canary/concepts/route-layout) or
+[route middlewares](/docs/canary/concepts/middleware) you'll sometimes come
+across a situation where you want your routes to inherit from a different layout
+than another route that's in the same URL segment.
```
deer:
```suggestion
When working with [route layouts](/docs/canary/concepts/route-layout) or
[route middlewares](/docs/canary/concepts/middleware), you'll sometimes come
across a situation where you want your routes to inherit from a layout
other than what's suggested by the URL segment.
```

deer (approved):
Great cleanup!

docs/toc.ts
marvinhagemeister: Setting the language is now mostly:
1. Open the `_app.tsx` file
2. Add the `lang` attribute to the `<html>`-tag

Not sure if we need a full page for that. I'd rather have a note for that on a page that deals with internationalization overall. Like it shows how to parse the `Accept` header (or uses a lib for that), how to user translation strings and as part of that a short section on setting the `lang` attribute.

docs/latest/concepts/routing.md
marvinhagemeister: Thanks!

PR #1637, opened 2023-08-15
Fix example routes

nathanwaldman:
I have never used Fresh, but I was glancing through the documentation and this looked incorrect.


marvinhagemeister (approved):
Good catch! Thanks for filing a PR ❤️

nathanwaldman:
Hm, this test failure should be unrelated to my change.

PR #1640, opened 2023-08-16
fix: load snapshot message shown for build

marvinhagemeister:


marvinhagemeister:
Merging, because that's super harmless and the test verifies the change.

PR #1651, opened 2023-08-16
fix: don't load snapshot in dev mode

### LINKED ISSUE #1650: bug: running dev.ts uses the snapshot
Using `deno run -A dev.ts` after building the assets I get `Using snapshot found at ..`

I expected the dev builds to not use the snapshots, and thats what the docs here https://fresh.deno.dev/docs/concepts/ahead-of-time-builds seems to say 
>When Fresh is started in **non-development** mode (usually via main.ts), Fresh will automatically pick up optimized assets

marvinhagemeister:
TODO: add tests

Fixes #1650

### DIFF
--- a/src/dev/build.ts
+++ b/src/dev/build.ts
@@ -1,25 +1,10 @@
 import { ServerContext } from "../server/context.ts";
-import { FreshOptions, Manifest } from "../server/mod.ts";
-import { dirname, fromFileUrl, join, toFileUrl } from "../server/deps.ts";
-import { fs } from "./deps.ts";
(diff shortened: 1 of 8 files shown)


marvinhagemeister:
I think this PR is definitely the right way to go long term as it properly differentiates between `dev` and non-dev mode. Doing so would be a bit of a disruptive change for folks not using `fresh.config.ts`. So I'm leaving this PR open here until the time is ready.

marvinhagemeister:
Replaced by https://github.com/denoland/fresh/pull/1735

PR #1653, opened 2023-08-17
docs(www): update hello bar for Fresh 1.4

iuioiua:


marvinhagemeister (approved):
Ohh good catch! I totally forgot we had this 🙈

iuioiua:
Perhaps, we can keep a small release checklist somewhere.

PR #1657, opened 2023-08-17
fix: remove debug code

marvinhagemeister:
Accidentally committed these with https://github.com/denoland/fresh/pull/1617/files .

Opened a feature request that we should have a `no-console` lint rule https://github.com/denoland/deno/issues/20187


marvinhagemeister:
Merging myself since it's just the removal of an accidental `console.log`

PR #1658, opened 2023-08-17
fix: twind duplicating css classes when `className` is used instead of `class`

### LINKED ISSUE #1655: `Duplicate class in <html></html> when use className`
# Duplicate class when use `className` inside `<html></html>` 

```tsx
<html className='bg-slate-800'>
</html>
```

**In Devtools**
```html
<html class='bg-slate-800 bg-slate-800'>
</html>
```

# `className` or `class` not work in `<body></body>`

```tsx
<body className='bg-slate-800'>
</body>
```

**In Devtools**
```html
<html>
...
   <body>
  </body>
</html>
```

marvinhagemeister:
Both the `twind` and `twindV1` plugin didn't properly handle the case when `className` is used to set classes instead of `class`. This would lead to CSS classes being applied twice.

Fixes #1655


bartlomieju (approved):
LGTM, no comments

PR #1684, opened 2023-08-19
chore: explicitly version rts in deps.ts

deer:
I got an annoying error last night when trying to run a custom script:
```
error: Relative import path "preact" not prefixed with / or ./ or ../ at my-file-here.ts
```
This got me thinking: why are preact and preact-render-to-string pushed into the import maps of users? Why doesn't Fresh explicitly version them in a `deps` file so that we could remove them from the import maps of users? I see this was added in https://github.com/denoland/fresh/pull/258 with no explanation. (I'm also unsure about the name of #258. It seems like it should be 'version preact implicitly via import map' since the project is no longer using an explicit version, but is instead implicitly relying on the version from the user's project.)

All the tests pass locally -- let's see what CI says. Additionally, I'm happy for this to be closed if there's a solid reason why it's like this. Sadly I couldn't find any explanation. As a user of Fresh, I've never needed to call `renderToString` myself, and I would guess we don't want users to do this...

Also, I know this wouldn't have helped my issue last night. If this approach is desired, I would apply the same treatment to preact in a separate PR.


marvinhagemeister (approved):
Good call! We can definitely drop v5 👍

PR #1696, opened 2023-08-23
fix: Add `_fresh` to .gitignore on update

### LINKED ISSUE #1680: Add `_fresh` to gitignore on update
Seems like we don't add that on update.

smitssjors:
Fixes #1680

### DIFF
--- a/tests/cli_test.ts
+++ b/tests/cli_test.ts
@@ -548,3 +548,189 @@ Deno.test("fresh-update add _app.tsx if not present", async function fn(t) {
     assertEquals(code, 0);
   }
 });
+
+Deno.test(
+  "fresh-update add _fresh to .gitignore if not present",
+  async function fn(t) {
+    // Preparation
+    const tmpDirName = await Deno.makeTempDir();
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister (commented):
Awesome, this is great! Let's add a quick test to the existing update test that verifies that this works 👍

smitssjors:
I added some test but there are some things I would like to add.

- I found it difficult to find the tests
- There is some room for refactoring the update tests

I don't mind picking up some of this work. I would suggest splitting the tests into `init_test.ts` and `update_test.ts`.

Also, why import the asserts from `deps.ts` and not directly from `$std`?

marvinhagemeister:
> I don't mind picking up some of this work. I would suggest splitting the tests into `init_test.ts` and `update_test.ts`.

Agree, splitting these up makes sense.

> Also, why import the asserts from `deps.ts` and not directly from $std?

That was before my time, but I'd guess that this was done so that there are less places to update when moving to a newer version of std.

marvinhagemeister (approved):
This is great! Thank you so much for fixing this 👍

PR #1702, opened 2023-08-24
docs: fix small typos

smitssjors:


docs/toc.ts
```diff
@@ -53,7 +53,7 @@ const toc: RawTableOfContents = {
           ["forms", "Forms", "link:latest"],
           ["islands", "Interactive islands", "link:latest"],
           ["static-files", "Static files", "link:latest"],
-          ["middleware", "Middlewares", "link:latest"],
+          ["middleware", "Middleware", "link:latest"],
```
marvinhagemeister: Let's keep the visible page title as `Middlewares`. All other things in the docs are in plurals too. Rather we should redirect the url to the plural version at some point, but that can happen in another PR.

docs/latest/concepts/middleware.md
```diff
@@ -15,7 +15,7 @@ be used to pass arbitrary data to downstream (or upstream) handlers. This
 special [\_app](/docs/concepts/app-wrapper.md) wrapper and normal
 [routes](/docs/concepts/routes.md). `ctx.state` is normally set by modifying its
 properties, e.g. `ctx.state.loggedIn = true`, but you can also replace the
-entire object like `ctx.state = { loggedIn = true }`.
+entire object like `ctx.state = { loggedIn: true }`.
```
marvinhagemeister, line 18: Great catch!

docs/toc.ts
smitssjors: I did not consider that they were all plural, I've reverted the change.

marvinhagemeister (approved):

Issue #1690, opened 2023-08-21
Optional Dynamic Routes


cymkd:
**Fresh DX would benefit with supporting optional dynamic routes via filename convention**

Common scenario:
- multi-language routing where default language is not shown in the url.

Using an example for [lang] parameter.
/home 
/en/home 
/de/home

Whilst the folder structure might be
- [[lang]]
- - home.tsx

[[lang]]
[+lang]
[?lang]

mbhrznr:
fwiw we could already apply optional dynamic params via `RouteConfig`.
successfully tested the following example:
```tsx
import { RouteConfig, RouteContext } from "$fresh/server.ts";

export const config: RouteConfig = {
  routeOverride: "{/(de|en)}?/home",
};

export default function HomePage(req: Request, ctx: RouteContext) {
  console.log(ctx.params);
  return <h1>i'm home!</h1>;
}
```

this would successfully catch `/home`, `/de/home` and `/en/home`.  
`ctx.params` would be logged as `{}`, `{ "0": "de" }` and `{ "0": "en" }` respectively.

might be worth some exploration as well?

cymkd:
That is a great solution.

If I am not mistaken, when having 3000/4000 pages, we would need to maintain this Config, with routes and with allowed parameters. (or maybe just use .* and alike).
It's another thing to take care of.

In terms of DX it feels an improvement to have the optional dynamic routes filename convention, as we already have routing conventions.

marvinhagemeister:
Yeah, the route config is more of an escape hatch when you need more control. It can do everything you need to, but nothing wrong with making common scenarios easier. It's the reason we have `[id]` or `[...slug]` after all.

mbhrznr:
with #1706 being merged we could probably close this, wdyt?

marvinhagemeister:
Ah good point. I missed this 👍

PR #1706, opened 2023-08-25
feat: add support for optional dynamic route params

mbhrznr:
tackles #1690.

this pr is a first attempt to support optional dynamic routes.  
the file name convention is inspired by [sveltekit](https://kit.svelte.dev/docs/advanced-routing)'s established convention.

the feature can be tested in the primary fixture by running:
```
deno task fixture
```

once started, navigate to `/i18n/lang` and `/i18n/:lang/lang`.


mbhrznr:
would add some documentation once we've settled on the general approach.

src/server/context_test.ts
```diff
@@ -21,11 +21,20 @@ Deno.test("pathToPattern", async (t) => {
...
 
+  await t.step("parses optional params", () => {
+    assertEquals(pathToPattern("foo/[[name]]"), "/foo{/:name}?");
+    assertEquals(pathToPattern("foo/[name]/[[bob]]"), "/foo/:name{/:bob}?");
+    assertEquals(pathToPattern("foo/[[name]]/bar"), "/foo{/:name}?/bar");
+    assertEquals(
+      pathToPattern("foo/[[name]]/bar/[[bob]]"),
+      "/foo{/:name}?/bar{/:bob}?",
+    );
+  });
```
marvinhagemeister, line 36: Let's add one test case for invalid usage like:

```sh
foo/foo-[[name]]-bar/baz
foo/[[name]]-bar/baz
foo/foo-[[name]]/baz
foo/foo-[[name]]
foo/[[name]]-bar
```

Currently, that returns this path `/foo{/:name}?-bar/baz` which would match `foo-bar/baz`. To me that shouldn't be allowed. When using the double bracket notation `[[asd]]` it should be only possible if it's a full segment of the path.

I think adding a check a double `[[` is not at the start of a segment or a `]]` is not at the end of a segment should suffice.

marvinhagemeister (commented):
This is fantastic! Going with the way Svelte does it makes a lot of sense. Like this much more then the original proposed `[...id]`. Found one additional case to account for, but other than that it looks really good and I'm excited to have this in Fresh 👍

src/server/context_test.ts
mbhrznr: thanks for writing down the tests cases, this made it incredible easy to add!

mbhrznr:
would it make sense to refactor `www`'s `docs/[...slug].tsx` to make use of the new concept?
or would you prefer to have this tackled in another pr?

marvinhagemeister:
> would it make sense to refactor www's docs/[...slug].tsx to make use of the new concept?
or would you prefer to have this tackled in another pr?

Let's tackle that in another PR

marvinhagemeister (approved):
Awesome, this is great!

marvinhagemeister (approved):
Awesome, this is great!

PR #1713, opened 2023-08-29
docs: fix app wrapper docs

jiawei397:
This is just a document mistake, but it would be better to make some modifications:

image


marvinhagemeister (approved):
Good catch! Thanks for making a PR 👍

Issue #1728, opened 2023-09-03
Updating the twind template or providing a section for Twind v1 plugin


ghost:
Hello 👋🏼.

Today Twind is providing a template for using tailwind CSS transparently with twind. However it is providing twind v0 despite v1 being available as an official plugin. I was wondering if we could update it or provide a documentation page to explain how to setup Twind V1 ? Currently from looking at the docs, I can only see a page for describing how to write [plugins](https://fresh.deno.dev/docs/concepts/plugins).

WDYT ?

---

# How I did it

If anyone coming across the same issue wants to do it here's a tl;dr regarding how to do it, it can also give a hint as-to how could it look like if we end up providing a tutorial or an updated template regarding this issue:

1. Create a `twind.config.ts` to bootstrap twind and the tailwind prefix lib since `@twind/core` does not include tailwind base rules. Here, autoprefix is included as well but it's not necessary.

```ts
import { Options } from "$fresh/plugins/twindv1.ts";

import { Preset } from "https://esm.sh/v131/@twind/core@1.1.3/core.d.ts";

import presetAutoprefix from "https://esm.sh/@twind/preset-autoprefix@1.0.7";
import presetTailwind from "https://esm.sh/@twind/preset-tailwind@1.1.4";

interface OptionsWithPresets extends Options {
  presets: Preset[];
}

export default {
  presets: [presetAutoprefix(), presetTailwind()],
  selfURL: import.meta.url,
} as OptionsWithPresets;
```

2. Then, edit the `fresh.config.ts` to include twind:

```ts
import { defineConfig } from "$fresh/server.ts";
import twindPlugin from "$fresh/plugins/twindv1.ts";
import twindConfig from "./twind.config.ts";

export default defineConfig({
  plugins: [twindPlugin(twindConfig)],
});
```

rojvv:
@marvinhagemeister This can be closed.

marvinhagemeister:
Closing as we're officially recommending to use the tailwindcss plugin instead.

PR #1729, opened 2023-09-03
Use Twind v1 instead of v0 on init

ghost:
Following up on #1728, I created this PR in order for twind v1 to be used by default on newly generated projects that require tailwind.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -287,11 +287,20 @@ await Deno.writeTextFile(
   ROUTES_API_JOKE_TS,
 );
 
-const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twind.ts";
+const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twindv1.ts";
+
+import { Preset } from "https://esm.sh/v131/@twind/core@1.1.3/core.d.ts";
+
+import presetTailwind from "https://esm.sh/@twind/preset-tailwind@1.1.4";
+
+interface OptionsWithPresets extends Options {
+  presets: Preset[];
+}
 
 export default {
+  presets: [presetTailwind()],
   selfURL: import.meta.url,
-} as Options;
+} as OptionsWithPresets;
 `;
 if (useTwind) {
   await Deno.writeTextFile(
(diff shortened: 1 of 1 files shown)


init.ts
```diff
@@ -287,11 +287,20 @@ await Deno.writeTextFile(
   ROUTES_API_JOKE_TS,
 );
 
-const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twind.ts";
+const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twindv1.ts";
+
+import { Preset } from "https://esm.sh/v131/@twind/core@1.1.3/core.d.ts";
```
king8fisher, line 292: Would this `import { Preset } from "https://esm.sh/@twind/core@1.1.3";` look more the rest of imports?

init.ts
```diff
@@ -503,7 +512,7 @@ try {
 
 let FRESH_CONFIG_TS = `import { defineConfig } from "$fresh/server.ts";\n`;
 if (useTwind) {
-  FRESH_CONFIG_TS += `import twindPlugin from "$fresh/plugins/twind.ts"
+  FRESH_CONFIG_TS += `import twindPlugin from "$fresh/plugins/twindv1.ts"
```
king8fisher, line 515: Thanks for this. This patch is missing in https://fresh.deno.dev/docs/examples/using-twind-v1 docs.

init.ts
```diff
@@ -287,11 +287,20 @@ await Deno.writeTextFile(
...
 );
 
-const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twind.ts";
+const TWIND_CONFIG_TS = `import { Options } from "$fresh/plugins/twindv1.ts";
+
+import { Preset } from "https://esm.sh/v131/@twind/core@1.1.3/core.d.ts";
+
+import presetTailwind from "https://esm.sh/@twind/preset-tailwind@1.1.4";
+
+interface OptionsWithPresets extends Options {
```
NuroDev, line 296: Should this type be included as part of the v1 plugin exports? That way every user does not need to define this wrapper type & instead Fresh can offer everything needed.

ghost: I think so! I provided this in the init to avoid messing with too much files but it would be better if it was provided directly from the plugin exports 👍🏼

deer:
@marvinhagemeister, another tailwind victim here.

PR #1731, opened 2023-09-04
fix: wrong type for html boolean attrs

### LINKED ISSUE #1725: checked attributed ignored on JSX passed to an island
I have an island form that renders regular components as children (as described in https://fresh.deno.dev/docs/concepts/islands#passing-jsx-to-islands)

```tsx
import { ComponentChildren } from "preact";

export default function IslandForm(props: {
  children?: ComponentChildren;
}) {
  return (
    <form action="javascript:">
      <fieldset>
        {props.children}
      </fieldset>
    </form>
  );
}

```

When the child is either a checkbox or a radio button, the `checked` attribute is ignored:

```tsx
import IslandForm from "../islands/IslandForm.tsx";

export default function () {
  return (
    <IslandForm>
      <label>
        <input type="radio" checked />first option
      </label>
      <label>
        <input type="checkbox" checked />checkbox
      </label>
    </IslandForm>
  );
}
```

The result:
image

If the checkbox or radio button is directly in the island, it works as expected:

```tsx
import { ComponentChildren } from "preact";

export default function IslandForm(props: {
  children?: ComponentChildren;
}) {
  return (
    <form action="javascript:">
      <fieldset>
        <label>
          <input type="radio" checked />first option
        </label>
        <label>
          <input type="checkbox" checked />checkbox
        </label>
      </fieldset>
    </form>
  );
}

```

Results in
image

marvinhagemeister:
HTML boolean attributes must be treated as being `true` when present. See: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

Fixes https://github.com/denoland/fresh/issues/1725

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -84,9 +84,6 @@ function addPropsChild(parent: VNode, vnode: ComponentChildren) {
   }
 }
 
-const BOOLEAN_ATTR =
-  /(async|autofocus|autoplay|checked|controls|default|defer|disabled|formnovalidate|inert|ismap|itemscope|loop|multiple|muted|nomodule|novalidate|open|playsinline|readonly|required|reversed|selected)/;
-
 const enum MarkerKind {
   Island,
   Slot,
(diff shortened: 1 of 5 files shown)


src/runtime/entrypoints/main.ts
```diff
@@ -323,7 +326,12 @@ function _walkInner(
         };
         for (let i = 0; i < sib.attributes.length; i++) {
           const attr = sib.attributes[i];
-          props[attr.nodeName] = attr.nodeValue;
+
+          // Boolean attributes are always `true` when present.
+          // See: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML
+          props[attr.nodeName] = BOOLEAN_ATTR.test(attr.nodeName)
+            ? true
+            : attr.nodeValue;
```
bartlomieju: Have you benched if regex is faster than searching in an array?

bartlomieju (approved):

src/runtime/entrypoints/main.ts
marvinhagemeister: Good call, did a quick benchmark and given that we're dealing with short arrays it's faster. Thought about this some more if there is a way not to ship those strings at all to the browser and found an even better approach. Since we already _have_ all attribute names we care about and they all exist as properties on DOM nodes we can do a simple `typeof value === "boolean"` check.

PR #1735, opened 2023-09-05
fix: dev command loading snapshot

### LINKED ISSUE #1650: bug: running dev.ts uses the snapshot

marvinhagemeister:
This PR changes the `dev.ts` script so that it doesn't import `main.ts` anymore if the new `config` parameter was passed. This changes the architecture in that the `dev` mode doesn't sit on top of `main` anymore which allows us to properly detect if we're in dev mode without other means. By calling into `ServerContext` directly instead of somewhere abstracted away behind `start` we can set all options ourselves, like if we want to skip loading the snapshot or not.

Note that this still preserves the previous behavior when no `fresh.config.ts` file is used. So this isn't a breaking change.

Fixes #1650

### DIFF
--- a/src/dev/dev_command.ts
+++ b/src/dev/dev_command.ts
@@ -1,8 +1,13 @@
 import { updateCheck } from "./update_check.ts";
-import { DAY, dirname, fromFileUrl, join } from "./deps.ts";
-import { FreshOptions } from "../server/mod.ts";
+import { DAY, dirname, fromFileUrl, fs, join, toFileUrl } from "./deps.ts";
+import {
+  FreshOptions,
+  Manifest as ServerManifest,
+  ServerContext,
+} from "../server/mod.ts";
 import { build } from "./build.ts";
 import { collect, ensureMinDenoVersion, generate, Manifest } from "./mod.ts";
+import { startFromContext } from "../server/boot.ts";
 
 export async function dev(
   base: string,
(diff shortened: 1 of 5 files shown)


src/dev/dev_command.ts
```diff
@@ -34,9 +39,28 @@ export async function dev(
 
   if (manifestChanged) await generate(dir, newManifest);
 
-  if (Deno.args.includes("build")) {
+  const manifest = (await import(toFileUrl(join(dir, "fresh.gen.ts")).href))
+    .default as ServerManifest;
```
bartlomieju, line 43: If `fresh.gen.ts` is not present on disk, this will throw. Is it handled anywhere?

src/server/boot.ts
```diff
@@ -0,0 +1,68 @@
...
+        await bootServer(handler, { ...opts, port });
+        firstError = undefined;
+        break;
+      } catch (err) {
+        if (err instanceof Deno.errors.AddrInUse) {
+          // Throw first EADDRINUSE error
+          // if no port is free
+          if (!firstError) {
+            firstError = err;
+          }
```
bartlomieju, line 45: Probably not related to this PR but if I read it right this will at most check port `8000` and `8001`. Is that expected?

bartlomieju (approved):

src/dev/dev_command.ts
marvinhagemeister: In this dev script we generate the `fresh.gen.ts` manifest a few lines before this one ourselves. I think it's ok to assume that it will be present here.

Issue #1674, opened 2023-08-17
Issue with children as props in interactive island components


mcgear:
I am trying to figure out how to support this, and will try to setup the scenario.

We have a shared component library with some of our companies reusable deno building blocks.  Specifically in there, we have a control to help show items or hide them behind a popup based on screen size.  We call this our ResponsiveSet:
https://github.com/fathym-deno/atomic/blob/integration/src/molecules/ResponsiveSet.tsx

Its heavy lifting is done by the MenuButton control which hides its children behind a button click:
https://github.com/fathym-deno/atomic/blob/integration/src/molecules/MenuButton.tsx

You'll see the `{props.toggleChildren}` is rendered inside an Action to wire up the onclick to show the menu. 

In order to make this control usable we have to wrap the ResponsiveSet in an island component to make the MenuButton interactivity work. We do this in our Fresh project with the InteractiveResponsiveSet:
 https://github.com/o-biotech/biotech-manager-web/blob/integration/islands/molecules/InteractiveResponsiveSet.tsx
 
 This works and the whole control works, the only issue is that the toggleChildren won't render the icon we are trying to get inside the action:
https://github.com/o-biotech/biotech-manager-web/blob/7d075f0b62c001d219fad0d8233fe3e71b0e19e6/components/organisms/BiotechHeader.tsx

There seems to be something going on when registering it with the InteractiveResponsiveSet, if you take the BiotechHeader and instead use the ResponsiveSet itself, the MenuIcon renders (but interactivity is lost).

In our latest integration branch, you can see that if we set the triggerChildren to a simple `toggleChildren="☰"` value, it is passed through just fine:
https://github.com/o-biotech/biotech-manager-web/blob/integration/components/organisms/BiotechHeader.tsx

Does this have to do with issues in passing props between server and island components?  Is there a better way i should be registering Islands in the component library (i know i'd like a better way than the wrapper).

Thanks for the help, let me know how else i can help.

kitsonk:
When I upgraded to Fresh 1.4, I am finding that children get set to undefined when rendered. I moved the same value to a property and everything worked as before.

marvinhagemeister:
@kitsonk do you have a link to a GitHub repo where the problem can be reproduced?

mcgear:
This commit of the site is what should be able to reproduce it:
https://github.com/o-biotech/biotech-manager-web/blob/7d075f0b62c001d219fad0d8233fe3e71b0e19e6/

[...]

PR #1736, opened 2023-09-05
fix: Missing error on invalid island children

marvinhagemeister:
In #1674 a reproduction was posted where a json object was passed to an island instead of an expected JSX element.

```jsx
<Island>{{ foo: 123 }}</Island>
```

Whilst this is technically possible in Preact, _if_ the receiving component knows how to deal with that, it's not something that's encouraged. It breaks the compositional nature of components and leads to difficult to understand errors.

Instead users should pass data via any other prop where the already existing serialization mechanism we have in Fresh will kick in. The `children` prop is reserved for JSX nodes only.

All this PR does, is throw an error when it detects that the user is trying to pass something invalid via `children` to an island

### DIFF
--- a/src/server/rendering/preact_hooks.ts
+++ b/src/server/rendering/preact_hooks.ts
@@ -3,6 +3,7 @@ import {
   type ComponentChildren,
   Fragment,
   h,
+  isValidElement,
   type Options as PreactOptions,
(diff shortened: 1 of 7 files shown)


bartlomieju (approved):
LGTM!

PR #1738, opened 2023-09-05
chore: fix update check tests downloading deps

marvinhagemeister:
Deno looks up the cache dir via `HOME`. We replaced that with a tmp dir in our tests, which lead to all dependencies being downloaded from scratch. This PR changes that so that we use a different environment variable for our tests. That way we can reuse the existing cache which is much faster.


marvinhagemeister:
Merging myself since this just speeds up tests.

PR #1740, opened 2023-09-05
fix: incorrect island child validation

marvinhagemeister:
I messed up the check in #1736 a little and threw on valid children. This PR addresses that.


marvinhagemeister:
Merging, since that's blocking another PR of mine.

PR #1742, opened 2023-09-06
fix: only show update message once per day

### LINKED ISSUE #1673: Show update message less frequently
[From discord](https://discord.com/channels/684898665143206084/991511118524715139/1141791394084692050):

> absolutely not a fan of my terminal being hammered by the update warning
> it's probably better to just tell to update once, when cold-starting a project, than on every save/reload/refresh

marvinhagemeister:
Changes the update check behavior so that the update message is only shown once per day.

Fixes: https://github.com/denoland/fresh/issues/1673


lucacasonato (approved):
LGTM

PR #1757, opened 2023-09-09
docs: remove unneeded `Counter` import from Head page

mattjbones:
The `Counter` import isn't used in the example and likely copypasta?


marvinhagemeister (approved):
Good catch, that's indeed a copy & paste error.

PR #1758, opened 2023-09-09
www: remove extra whitespace in class

ryuapp:
Header of Fresh Components(https://fresh.deno.dev/components#Header) have extra whitespace in class.
This PR removes that whitespace and improves the usability of the component.


marvinhagemeister (approved):
Thanks!

PR #1764, opened 2023-09-11
fix dev_command.ts

chibat:
The `if` shown in the following URL will always be true.
https://github.com/chibat/fresh/blob/02a3cb4dc3f7e71ae32877e46f281e6c82100a14/src/dev/dev_command.ts#L58

Therefore, the line shown by the following URL will never be executed.
https://github.com/chibat/fresh/blob/02a3cb4dc3f7e71ae32877e46f281e6c82100a14/src/dev/dev_command.ts#L64

I have fixed the code to what I believe is expected.


chibat:
The following URL is an alternate fix.
https://github.com/denoland/fresh/compare/main...chibat:fresh:dev_command2

src/dev/dev_command.ts
```diff
@@ -55,7 +55,7 @@ export async function dev(
     // Ensure that build dir is empty
     await fs.emptyDir(outDir);
     await build(join(dir, "fresh.gen.ts"), options);
-  } else if (options) {
+  } else if (Object.keys(options).length) {
```
marvinhagemeister: I think what this check was supposed to do, was check if a third argument was passed to the function. Passing an empty object is valid and should fall into this branch too.

chibat: In the v1.4.3, the third argument defaults to `{}`, so the following code is unreachable code.
```ts
await import(entrypoint);
```
`Boolean({})` is `true`.

marvinhagemeister: Exactly, that's the issue.

chibat: I have updated the fix.

marvinhagemeister (commented):
This looks great! Only thing missing now is a test case 👍

chibat:
I have added a test.
Could you please check it.

marvinhagemeister:
I took the liberty to update the tests. They were passing even without the changes in this PR.

marvinhagemeister (approved):
Awesome, thanks for taking the time to make a PR 👍

PR #1768, opened 2023-09-13
Fix build path

jiawei397:
The `snapshotDirPath` path in the current version is different from the one used in `deno task build`.

Under normal circumstances, there is no problem, but if you move `dev.ts` to the `src/` directory, a bug will occur.


marvinhagemeister (commented):
Good catch, let's add a test case for that to ensure that it works.

jiawei397:
> Good catch, let's add a test case for that to ensure that it works.

I added a test case based on `build_test`. Do you think it's okay?

marvinhagemeister (approved):
Awesome, this is great! Love the additions of tests 👍

PR #1772, opened 2023-09-13
feat: add nonce for inline scripts

marvinhagemeister:
This PR changes our content security handling so that we'll automatically add a `nonce` attribute when we render an inline script tag. This will make it easier to integrate Google Tag Manager and things like that.


bartlomieju (approved):
LGTM

PR #1776, opened 2023-09-14
twindv1: Enable setting tailwind theme variables in Options parameter

MadTyres:
currently, the v1 plugin locks the Options to BaseTheme from @twind/core

this change aims to allow other options to be passed, for example TailwindTheme from @twind/preset-tailwind
dunno if this is desirable, but when i migrated my project from 0.16 to v1 i could not set `config.theme.fontFamily`
This should fix that issue.


marvinhagemeister (approved):
Nice catch! Thanks for making a PR 👍

Issue #1716, opened 2023-08-30
fix flaky tests


deer:
This can be a list of flaky tests that need to get fixed at some point.

https://github.com/denoland/fresh/pull/1715 failed with the following problems:
* https://github.com/denoland/fresh/actions/runs/6021026772/job/16333211788?pr=1715
```
Resolves routes with non-latin characters => ./tests/main_test.ts:890:6
error: Leaking resources:
  - A fetch response body (rid 109) was created before the test started, but was consumed during the test. Do not close resources in a test that were not created during that test.
```
* https://github.com/denoland/fresh/actions/runs/6021026772/job/16333211975?pr=1715
```
revive island in lazy server rendered children conditionally => ./tests/islands_test.ts:498:6
error: Leaking async ops:
  - 1 async operation to op_read was started before this test, but was completed during the test. Async operations should not complete in a test if they were not started in that test.
```

marvinhagemeister:
@lucacasonato was able to trace the warnings down to a bug in the test ops sanitization code. The fix is not yet ready to land, but hopefully soon. https://github.com/denoland/deno/pull/20501

Apart from that it seems like `puppeteer` is leaking timers. One idea there is to extract puppeteer tests into separate files and disable ops sanitization there. Disabling this in a file which has other tests trips up the leak detection.

marvinhagemeister:
Resolved by https://github.com/denoland/fresh/pull/1786

PR #1777, opened 2023-09-14
chore: fix wait for child process shutdown

marvinhagemeister:
@lucacasonato discovered this in a debugging session earlier today.

related #1716


marvinhagemeister:
Merging myself, because it's harmless.

PR #1781, opened 2023-09-15
fix: server initialized twice in old dev mode

marvinhagemeister:
Noticed this will working on #1780 . Initializing the server twice didn't lead to any errors, but makes dev bootup time a hair split slower.


marvinhagemeister:
Merging, because CI is flaky because of faulty ops sanitization. The core cli team is currently investigating this.

PR #1787, opened 2023-09-18
feat: add nonce attribute to style tags

marvinhagemeister:


bartlomieju:
Superseded by a new implementation against the current Fresh 2.x codebase. The new PR adds nonce support for both inline `<style>` and `<script>` tags, plus a `useNonce` option on the CSP middleware.

PR #1788, opened 2023-09-18
feat: exclude _fresh dir from everything

marvinhagemeister:
Turns out that we can set a top level `exclude` property in `deno.json` to exclude it from everything. This simplifies the configuration generated by Fresh to exclude the `_fresh` directory quite a bit.


marvinhagemeister:
Merging myself, since it's harmless.

PR #1790, opened 2023-09-18
feat: reduce manifest merge conflict potential

marvinhagemeister:
This PR rewrites the manifest (= `fresh.gen.ts`) generation to reduce the likelihood of git merge conflicts. Previously, we'd crawl all files and generate a sequential identifier based on this list like `$1`. When a new route was added the existing order would change and lots of unrelated routes needed to be renumbered.

This PR changes that by turning the import specifier into a JS identifier and thereby reducing the reliance on ordering. When it sees that an identifier is already used we'll append a numeric suffix like `foo -> foo_1`.


marvinhagemeister:
Closing in favour of https://github.com/denoland/fresh/pull/2020

PR #1799, opened 2023-09-19
fix: Fix typo in hover rule for default non-twind CSS

mmuddasani:
In a new Fresh project, I noticed a typo in the generated file `static/styles.css`. The generated CSS file has the line `.hover:bg-gray-200:hover {`, and the hover style did not work by default. When I locally replaced that line with `.hover\:bg-gray-200:hover {`, the hover style started to work as expected.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -423,7 +423,7 @@ html {
 .rounded {
   border-radius: 0.25rem;
 }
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Ohh that's a great catch! Thanks for fixing this 👍

PR #1808, opened 2023-09-21
refactor: simplified Deno KV OAuth plugin

iuioiua:
This implementation has far fewer moving parts but still allows for the flexibility of multiple providers.


plugins/kv_oauth.ts
```diff
@@ -1,206 +1,107 @@
...
-            req,
-            providersOrOAuth2Client,
-          );
-          return response;
-        },
+export function createRoutes(
+  oauthConfig: OAuth2ClientConfig,
+  /** Parent path for sign-in, callback and sign-out pages. */
+  oauthPath = "/oauth",
+): PluginRoute[] {
```
mbhrznr: **suggestion:** would it make sense to convert this into a tuple rather than having an array?
this way, we could simply pass `oauthPath` as a generic to infer the paths, if needed:   

```tsx
export function createRoutes<T extends string = "/oauth">( // <-- `/oauth` would be needed so that we can infer the default
  oauthConfig: OAuth2ClientConfig,
  /** Parent path for sign-in, callback and sign-out pages. */
  oauthPath: T = "/oauth" as T,
): [
  { path: `${T}/signin`; handler: Handler }, // <-- each of these would still satisfy `PluginRoute`
  { path: `${T}/callback`; handler: Handler },
  { path: `${T}/signout`; handler: Handler },
] {
  return [
    {
      path: `${oauthPath}/signin`,
      handler: async (req) => await signIn(req, oauthConfig),
    },
    {
      path: `${oauthPath}/callback`,
      handler: async (req) => {
        // Return object also includes `accessToken` and `sessionId` properties.
        const { response } = await handleCallback(
          req,
          oauthConfig,
        );
        return response;
      },
    },
    {
      path: `${oauthPath}/signout`,
      handler: signOut,
    },
  ];
}

const routes = createRoutes(createGitHubOAuthConfig()); // <-- hovering over `routes` would already give us some nice intellisense

type Routes = typeof routes; // <-- this would let us assign it as a type
type Route = typeof routes[number]["path"]; // <-- this would give us a union type consisting of `"/oauth/signin" | "/oauth/callback" | "/oauth/signout"`
```

wdyt?

plugins/kv_oauth.ts
```diff
@@ -1,206 +1,107 @@
-import { handleCallback, signIn, signOut } from "./kv_oauth/plugin_deps.ts";
-import { OAuth2Client } from "./kv_oauth/plugin_deps.ts";
-import type { Plugin } from "../server.ts";
+import {
+  createAuth0OAuthConfig,
```
mbhrznr, line 2: **praise:** great idea to re-export all of the providers!

plugins/kv_oauth.ts
```diff
@@ -1,206 +1,107 @@
...
-        path: options?.callbackPath ?? "/oauth/callback",
-        handler: async (req) => {
-          // Return object also includes `accessToken` and `sessionId` properties.
-          const { response } = await handleCallback(
-            req,
-            providersOrOAuth2Client,
-          );
-          return response;
-        },
+export function createRoutes(
```
mbhrznr: **praise:** nice to have such a simple helper fn!

plugins/kv_oauth.ts
NuroDev: I like this idea. I was originally going to try diving into the idea of making each array element a constant but I think using tuples is much smarter here.

Then if we want to update the `InferOAuthProviders` type we can update it to something like this to make it as easy as possible to get a typed route ([Playground](https://tsplay.dev/NDAEVm)):
```ts
type InferOAuthProviders<T extends PluginRoute[]> = T[number]["path"];
```

Though I think it would be better to rename that type to something more accurate now like `InferOAuthRoutes`

NuroDev: The one thing I will add:
> each of these would still satisfy `PluginRoute`

Do we want to rely on the explicit return for this type satisfaction or add a `satisfies PluginRoute[]` to the actual array being returned?
```ts
return [
	...
] satisfies PluginRoute[];
```

iuioiua: Yeah, I was thinking that too.

mbhrznr: > Do we want to rely on the explicit return for this type satisfaction or add a satisfies PluginRoute[] to the actual array being returned?

thought about that as well, yet i wasn't able to appease ts.  
would absolutely prefer the usage of `satisfies PluginRoute[]`.

iuioiua: I went with the approach that explicitly defines route paths. It provides greater detail.

marvinhagemeister (approved):

iuioiua:
Note to self: update documentation where needed.

PR #1816, opened 2023-09-25
feat(dev): Allow setting host and port

bastilian:
This adds `--host` and `--port` flags to allow running fresh with a custom IP and port to listen on for requests.

### DIFF
--- a/docs/latest/getting-started/running-locally.md
+++ b/docs/latest/getting-started/running-locally.md
@@ -35,11 +35,16 @@ passing the extra folder as an argument: `--watch=static/`. You should also add
 `routes/` to the watch list, so that the server restarts automatically whenever
 you add a new route.
 
-If you want to change the port or host, modify the config bag of the `start()`
-call in `main.ts` to include an explicit port number:
+If you want to change the port or host, modify the config bag of the
+`defineConfig()` call in `fresh.config.ts` to include an explicit port number:
 
-```js main.ts
-await start(manifest, { port: 3000, hostname: "0.0.0.0" });
+```js fresh.config.ts
+export default defineConfig({
+  server: {
+    port: 3000,
+    hostname: "0.0.0.0",
+  },
+});
 ```
 
 You can also change the port and hostname by setting the `PORT` and `HOST`
--- a/docs/latest/getting-started/running-locally.md
+++ b/docs/latest/getting-started/running-locally.md
@@ -39,13 +39,20 @@ If you want to change the port or host, modify the config bag of the `start()`
 call in `main.ts` to include an explicit port number:
 
 ```js main.ts
-await start(manifest, { port: 3000 });
+await start(manifest, { port: 3000, hostname: "0.0.0.0" });
 ```
 
-You can also change the port by setting the `PORT` environment variable:
+You can also change the port and hostname by setting the `PORT` and `HOST`
+environment variables:
 
 ```sh Terminal
-$ PORT=3000 deno task start
+$ HOST=0.0.0.0 PORT=3000 deno task start
+```
+
+There is also a `--port` and a `--host` flag that can be passed to set them.
+
+```sh Terminal
+$ deno task start --port 3000 --host 0.0.0.0
 ```
 
 Combining all of this we get the following `deno run` command:
--- a/src/server/boot.ts
+++ b/src/server/boot.ts
@@ -1,26 +1,40 @@
-import { colors } from "./deps.ts";
+import { colors, parse } from "./deps.ts";
 import { ServeHandler } from "./types.ts";
+
+const onListen =
+  (opts: Partial<Deno.ServeTlsOptions>) =>
+  (params: { hostname: string; port: number }) => {
+    const address = colors.cyan(
+      `http://${opts.hostname || params.hostname}:${opts.port || params.port}/`,
+    );
+    console.log(
+      "\n",
+      colors.bgRgb8(colors.rgb8(" 🍋 Fresh ready ", 0), 121),
+      `\n    ${colors.bold("Local:")} ${address}\n\n`,
+    );
+  };
 
 export async function startServer(
   handler: Deno.ServeHandler,
   opts: Partial<Deno.ServeTlsOptions>,
 ) {
-  if (!opts.onListen) {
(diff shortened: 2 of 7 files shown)


tests/test_utils.ts
```diff
@@ -426,7 +426,8 @@ async function spawnServer(
   // @ts-ignore yes it does
   for await (const line of lines.values({ preventCancel: true })) {
     output.push(line);
-    const match = line.match(/https?:\/\/localhost:\d+/g);
+    const match = line.match("Fresh ready");
```
bastilian: Since the hostname/IP can now be variable this needed to be changed to assert the fresh is running. "Fresh ready" seemed to be better as it'll always be there and not dynamically change.

tests/test_utils.ts
```diff
@@ -427,7 +427,9 @@ async function spawnServer(
   for await (const line of lines.values({ preventCancel: true })) {
     output.push(line);
     const match = line.match(/https?:\/\/localhost:\d+/g);
-    if (match) {
+    const running = line.match("Fresh ready");
+
+    if (running) {
       address = match[0];
```
marvinhagemeister, line 464: This line is likely causing the tests to fail. Previously the regex was used to grab the actual server address out of the child process stdout and that's missing now.

bastilian: I reverted the change to "Fresh ready" and just made the regex less strict, which should do the trick.

bastilian:
The test added for the `HOST` env var currently covers just the env var, is there a place (already) where the the flags and/or the `dev()` options should be tested? I couldn't find an existing test for the `port`, or have I missed it?

bastilian:
Just checking up on here. If this should have more tests, it might be a better path to first add general tests for the `dev` command, which might be better in a separate PR (that I'd be happy to submit).

bastilian:
@marvinhagemeister After playing with this more, I felt it would be better if flags or environment variables would override what is passed in the options.

bastilian:
ping. Let me know if this PR needs more work. :)

docs/latest/getting-started/running-locally.md
```diff
@@ -39,13 +39,20 @@ If you want to change the port or host, modify the config bag of the `start()`
 call in `main.ts` to include an explicit port number:
 
 ```js main.ts
```
deer: I know this isn't from you, but we no longer modify `main.ts` to do this. The correct spot is now in `fresh.config.ts`.  (Otherwise these changes wouldn't take place when running in dev mode.) Additionally, setting `port` and `hostname` directly are deprecated. Please see `FreshConfig` for details.

I guess the correct thing would be something like:
```ts
export default defineConfig({
  server: {
    port: 3000,
    hostname: "0.0.0.0",
  },
});
```

src/server/boot.ts
```diff
@@ -1,26 +1,40 @@
...
+  });
+  const portEnv = Deno.env.get("PORT");
+  const hostEnv = Deno.env.get("HOST");
 
-      const address = colors.cyan(`http://localhost:${params.port}/`);
-      const localLabel = colors.bold("Local:");
-      console.log(`    ${localLabel} ${address}\n`);
-    };
+  if (portEnv !== undefined || portFlag !== undefined) {
+    opts.port = parseInt((portFlag || portEnv) as string, 10);
```
deer, line 29: Isn't 10 the default? Do you really need to declare it?

tests/main_test.ts
```diff
@@ -841,6 +841,27 @@ Deno.test("PORT environment variable", async () => {
   for await (const _ of lines) { /* noop */ }
 });
 
+Deno.test("HOST environment variable", async () => {
```
deer: I see your comment about testing port, but since Fresh relies on `Deno.serve`, I guess there aren't any tests for testing hostname/port from the options bag. We can assume it works from upstream.

Perhaps you could modify this to also set the port via an environment variable?

deer (commented):
Hi @bastilian, cool feature. A few comments:
* looks like a conflict has crept in (I think due to one of my PRs that got merged) -- please merge main and clean that up
* the rest are inline

src/server/boot.ts
marvinhagemeister: It's unfortunately a little more complex. Quoting from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#description):

> - If the input string, with leading whitespace and possible +/- signs removed, begins with 0x or 0X (a zero, followed by lowercase or uppercase X), radix is assumed to be 16 and the rest of the string is parsed as a hexadecimal number.
> - If the input string begins with any other value, the radix is 10 (decimal).

So better to always pass the radix argument when using `parseInt`

deer: Thanks for the link and the clarification!

docs/latest/getting-started/running-locally.md
bastilian: I wasn't sure on how to properly update the docs properly. This helped! Thank you for the hint!

tests/main_test.ts
bastilian: Add the PORT to this test. Thank you!

docs/latest/getting-started/running-locally.md
deer, line 1: I unfortunately think my [comment](https://github.com/denoland/fresh/pull/1963#discussion_r1381213369) from here applies as well:
> Now that I think about it more, this should probably be branched to a canary version.

If someone reads this prior to the next release, they could end up confused. https://github.com/denoland/fresh/pull/1984/files#diff-40ec5d53f77de7b64b64ce31562e1717e678c4c55640531b6a55d2cbc9f53936L57 provides an example of how to modify the `toc.ts` file to have a different canary page.

In the future I'll make sure to make this suggestion straight away when people are adding new features.

deer (changes requested):
New test looks good. Just a minor comment about canary that I should have made yesterday!

bastilian:
@deer Thank you for taking a look again! The updated docs should now be under canary.

bastilian:
I'm closing this. I'm too busy at the moment with work stuff and I'm unsure when I'd be getting back to finalise this.

PR #1829, opened 2023-09-28
chore: append newline to produced json

king8fisher:
Every time ` deno run -Ar https://fresh.deno.dev/update .` is run, the `deno.json[c]` file gets a newline removed which contradicts formatting rules.

### DIFF
--- a/update.ts
+++ b/update.ts
@@ -331,8 +331,7 @@ async function findSrcDirectory(
   return resolvedDirectory;
 }
 
-// deno-lint-ignore no-explicit-any
-export function stringifyFormattedJson(jsonStr: any): string {
-  // Ensure formatting rules are met
-  return JSON.stringify(jsonStr, null, 2) + "\n";
+export function stringifyFormattedJson(data: unknown): string {
+  // Append newline to ensure formatting rules are met
+  return JSON.stringify(data, null, 2) + "\n";
 }
--- a/update.ts
+++ b/update.ts
@@ -139,7 +139,7 @@ freshImports(denoJson.imports);
 if (denoJson.imports["twind"]) {
   twindImports(denoJson.imports);
 }
-denoJsonText = JSON.stringify(denoJson, null, 2);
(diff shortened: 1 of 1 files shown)


update.ts
```diff
@@ -330,3 +330,9 @@ async function findSrcDirectory(
   }
   return resolvedDirectory;
 }
+
+// deno-lint-ignore no-explicit-any
+export function stringifyFormattedJson(jsonStr: any): string {
+  // Ensure formatting rules are met
+  return JSON.stringify(jsonStr, null, 2) + "\n";
+}
```
marvinhagemeister: The argument name is a bit misleading because we're not passing in a string.

```suggestion
export function stringifyFormattedJson(data: unknown): string {
  // Append newline to ensure formatting rules are met
  return JSON.stringify(data, null, 2) + "\n";
}
```

marvinhagemeister (commented):
LGTM, left a minor comment. Let me know what you think

marvinhagemeister (approved):

king8fisher:
Thank you very much!

mgenereu:
This was driving me nuts. Thank you for fixing it. I started a PR to call out to Deno to do the formatting to make sure this file is always compliant with the Deno fmt rules as configured for the project.  But a squirrel walked by and then I saw this.  Do you think there's value in the Deno fmt call out instead?

marvinhagemeister:
@mgenereu Yeah, I think ultimately we should call `deno fmt` on the whole project after an update or init. It's fast enough that we can do that.

king8fisher:
So much better than my simplistic approach!

mgenereu:
> @mgenereu Yeah, I think ultimately we should call `deno fmt` on the whole project after an update or init.

I didn't do it because I didn't know if the user is using `deno fmt` on their project.  I just figured `deno.json` was getting completely reserialized anyways that additional formatting from `deno fmt` was a "best default" until the user can decide to run their formatter or not afterward.

PR #1830, opened 2023-09-29
showcase: Add Deno Artwork

jasonjgardner:
- Added Deno Artwork to showcase. This Fresh project uses Deno KV to allow GitHub users to react to Deno-inspired artwork. (Initial artwork comes from the archived `denoland/dotland` repository.)


marvinhagemeister (approved):
nice

PR #1839, opened 2023-10-02
fix: `remoteAddr` hardcoded to `localhost:1234`

### LINKED ISSUE #1585: [Bug Report] Can't get client's Real IP / remoteAddr after upgrading to v1.3.0
Hi, after I updated fresh to v1.3.0 and after. I cannot get the client's real IP by `ctx.remoteAddr` method.
This is the test code of `_middleware.ts`

```typescript
interface State {
  ip: string;
}

  req: Request,
) {
  const ip = (ctx.remoteAddr as Deno.NetAddr).hostname;
  ctx.state.ip = ip;
  const resp = await ctx.next();
  console.debug(`Addr: ${(ctx.remoteAddr as Deno.NetAddr).hostname}, ${(ctx.remoteAddr as Deno.NetAddr).port}`);
  return resp;
}
```

The server can only get **Addr: localhost, 1234** after being upgraded to v1.3.0, but the real IP from the client is expected. I read the code and I found that the issue might come from this PR: https://github.com/denoland/fresh/pull/1278

I can see there is a static code to set the `DEFAULT_CONN_INFO` in the file `src/server/context.ts` , which overwritten the `remoteAddr`

I am not good at coding can you please help me check and fix the issue? Thank you!

martinrempel:
Fixes #1585.

This fix is based on the implementation of [Jabolol](https://github.com/Jabolol) which he [linked](https://github.com/Jabolol/fresh/blob/ee0d00c717e9acbad44080599696e9ea27144ead/src/server/mod.ts#L168-L181) in the issue.

### DIFF
--- a/src/server/boot.ts
+++ b/src/server/boot.ts
@@ -67,7 +67,7 @@ async function bootServer(handler: ServeHandler, opts: StartOptions) {
           remoteAddr,
           localAddr: {
             transport: "tcp",
-            hostname: opts.hostname,
+            hostname: opts.hostname ?? "localhost",
             port: opts.port,
           } as Deno.NetAddr,
         }),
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):

PR #1844, opened 2023-10-04
Switched to use Deno formatting rules

mgenereu:
Switched `update.ts` to use Deno fmt command on `deno.json` so it respects the `deno.json` `fmt` configuration.  Tested with 4 space indentation.


mgenereu:
Builds on #1829's hard work.

marvinhagemeister (approved):
Thanks 👍

PR #1868, opened 2023-10-05
Expose metafile.json

tlgimenes:
This PR exposes ESbuild's metafile at `/_frsh/js/{build_id}/metafile.json` so we can use ESbuild's bundle analyzer https://esbuild.github.io/analyze/ to analyze our bundle sizes. 

A future work would be to merge and not use `snapshot.json` in favor `metafile.json` since the later contains the snapshot info.


marvinhagemeister (approved):
Nice! Agree with dropping `snapshot.json` for esbuild's meta file in the future.

PR #1870, opened 2023-10-05
feat: Use href instead of origin on Partial 

tlgimenes:
Use `location.href` on partials so we can create refresh buttons without relying on islands or on the request

The idea is to be able to write code like this:
```tsx
const Refresh () => {
   return <button f-partial="?refresh">Refresh Page</button>
}
```

instead of this
```tsx
const Refresh ({href}: {href: string}) => {
   const url = new URL(href)
   
   url.searchParams.set('refresh', "true")
   
   return <button f-partial=`${url.href}`>Refresh Page</button>
}
```

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -948,7 +948,7 @@ document.addEventListener("click", async (e) => {
 
         const partialUrl = new URL(
           partial ? partial : nextUrl.href,
-          location.origin,
+          location.href,
         );
         await fetchPartials(partialUrl);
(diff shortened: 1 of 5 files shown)


marvinhagemeister (commented):
Great idea! Let's add a test case for this.

tlgimenes:
Just added some tests. Please let me know if you need anything else :)

marvinhagemeister (approved):
Awesome, looks great! Thanks for adding this 👍

tlgimenes:
Just saw the tests were failing, now they are all passing :)

PR #1873, opened 2023-10-05
fix: Partial state deserialization on simple islands

tlgimenes:
This PR fixes an issue with deserialization of Fresh Partial.

When no special type is used on __FRESH_STATE, the requiresSerialization variable is false, which causes the`deserialize` function to be null, thus making a void state, thus not hydrating correctly the islands inside the fresh state. 

This PR addresses this issue by using the good&old JSON.parse in these cases.

### DIFF
--- a/tests/fixture_partials/routes/island_props/index.tsx
+++ b/tests/fixture_partials/routes/island_props/index.tsx
@@ -1,21 +1,19 @@
 import { Partial } from "$fresh/runtime.ts";
-import { Fader } from "../../islands/Fader.tsx";
+// import { Fader } from "../../islands/Fader.tsx";
 import PropIsland from "../../islands/PropIsland.tsx";
 
 export default function PropsDemo() {
   return (
     <div>
       <Partial name="slot-1">
-        <Fader>
-          <p class="status-initial">initial</p>
-          <PropIsland
-            boolean={true}
-            number={1}
-            obj={{ foo: 123 }}
-            strArr={["foo"]}
-            string="foo"
-          />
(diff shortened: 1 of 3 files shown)


marvinhagemeister (commented):
Nice one! Let's add a test case for this and then it's ready to be merged 👍

tlgimenes:
Hi @marvinhagemeister. 

Looking at the code, I've noticed a test for `serializes props`. Taking a deeper look on why this test was not failing with mainstream code, I've figured out that the `<Fader/>` component was making the deserialize function needed (since it's an island). Removing the `<Fader/>` made the tests fail with mainstream code and pass with the proposed changes. 

Since there's already a test for nested islands, removing <Fader/> from `routes/island_props` test this case. 

Tell me what you think, thanks!

tests/fixture_partials/routes/island_props/index.tsx
```diff
@@ -1,21 +1,19 @@
 import { Partial } from "$fresh/runtime.ts";
-import { Fader } from "../../islands/Fader.tsx";
+// import { Fader } from "../../islands/Fader.tsx";
```
marvinhagemeister:
```suggestion
```

marvinhagemeister (approved):
@tlgimenes that makes perfect sense! Thanks for getting to the bottom of it. Let's remove the island from that test fixture entirely rather then commenting it out to avoid confusion 👍

tlgimenes:
Done, thanks @marvinhagemeister !

PR #1892, opened 2023-10-10
Adding the ability to prefix assets with a CDN url

MoaathAlattas:
### Summary
Adding the ability to prefix assets urls with a string for static files `(./static)` and internal files `(./_frsh/js/)`. This adds the ability to set a CDN url and load those files outside of fresh server.

### Changes
- [x] Add a new config key `cdnUrl`
- [x] Add `ASSET_PATH_PREFIX` in a similar fashion to `BUILD_ID`
- [x] Add a runtime utility to determine static file path `staticFile`
- [x] Skip building static and islands js routes if `cdnUrl` is defined.
- [x] Add tests

### Why
I am not using Deno deploy as I am hosting my app on one region closer to the database. However, I would like to host all static files on a CDN closer to the end user. There might be other benefits but this was it for me.

### How It works
1. Set the cdn url as part of the config passed to the start function (`./main.ts`) 
  ```ts
  await start(manifest,{
    ...config,
    cdnUrl: "https://my-cdn-url.com/my-bucket-name"
  })
  ```
2. Build your islands [Ahead-of-time Builds](https://fresh.deno.dev/docs/concepts/ahead-of-time-builds)
3. Create a script to upload `./fresh` and `./static` folders to your CDN, _**see an example below**_.
4. Add the script as a pre-deploy step, similar to the one here: [Deploying an optimized Fresh project](https://fresh.deno.dev/docs/concepts/ahead-of-time-builds#deploying-an-optimized-fresh-project) 
5. Your app will prefix static and island js files with a `cdnUrl/[build_id]` value:
  ```ts
  /_frsh/js/d763ab9889e57eba8c364daf5de8367e4746e61d/main.js
  =>
  https://my-cdn-url.com/my-bucket-name/d763ab9889e57eba8c364daf5de8367e4746e61d/_frsh/js/main.js
  
  /favicon.ico =>
  https://my-cdn-url.com/my-bucket-name/d763ab9889e57eba8c364daf5de8367e4746e61d/favicon.ico
  ```

### Upload to CDN script (example)
This is an example to upload `./fresh` and `./static` folders to AWS S3 (or any compatible service such as cloudflare R2)
```ts 
// deno run -A ./upload-to-cdn.ts
import { INTERNAL_PREFIX } from "$fresh/runtime.ts";
import { JS_PREFIX } from "$fresh/src/server/constants.ts";
import {
  PutObjectCommand,
  S3Client,
} from "https://esm.sh/@aws-sdk/client-s3@3.427.0";
import { walk } from "https://deno.land/std@0.201.0/fs/walk.ts";
import { contentType } from "https://deno.land/std@0.201.0/media_types/content_type.ts";

const appAbsolutePath = Deno.cwd()
const cdnFolderPath = "/" + await getBuildID();
const BUCKET_NAME = "simple-pages";

const client = new S3Client({
    region: "auto",
    endpoint: '',
    credentials: {
      accessKeyId: '',
      secretAccessKey: '',
    },
  });

await uploadFolder(
  appAbsolutePath + "/_fresh",
  cdnFolderPath + INTERNAL_PREFIX + JS_PREFIX,
);
await uploadFolder(appAbsolutePath + "/static", cdnFolderPath);

// functions
async function getBuildID() {
  try{
    const { default: json } = await import(
      `${appAbsolutePath}/_fresh/snapshot.json`,
      { with: { type: "json" }}
    );
    return json.build_id;
  } catch (error) {
    console.log("did you build your app islands? ")
    console.log(error)
    Deno.exit(1)
  }

}

async function uploadFile(filePath: string, objectKey: string) {
  const fileContent = await Deno.readFile(filePath);
  const fileContentLength = fileContent.byteLength;
  const fileContentType = contentType(filePath.split(".").pop()!)?.split(
    ";",
  )[0];

  const command = new PutObjectCommand({
    Bucket: BUCKET_NAME,
    Key: objectKey,
    Body: fileContent.byteLength === 0 ? new Blob([""]) : fileContent,
    ContentType: fileContentType,
    ContentLength: fileContent.byteLength === 0
      ? new Blob([""]).size
      : fileContentLength,
    ACL: "public-read",
  });

  try {
    await client.send(command);
  } catch (error) {
    console.error(`Failed to upload: ${objectKey}`);
    console.error(error);
    Deno.exit(1);
  }
}

async function uploadFolder(fromPath: string, toPath: string) {
  console.log(`Uploading folder ${fromPath} to the cdn (${toPath}/)...`);
  for await (const file of walk(fromPath)) {
    if (file.isFile) {
      await uploadFile(file.path, toPath + "/" + file.name);
    }
  }
  console.log(`"Done uploading folder ${fromPath} to the cdn (${toPath}/)`);
}

```


src/server/boot.ts
```diff
@@ -10,7 +10,8 @@ export async function startFromContext(ctx: ServerContext, opts: StartOptions) {
         colors.bgRgb8(colors.black(colors.bold(" 🍋 Fresh ready ")), 121),
       );
 
-      const address = colors.cyan(`http://localhost:${params.port}/`);
+      const host = opts.hostname ?? "localhost";
+      const address = colors.cyan(`http://${host}:${params.port}/`);
```
MoaathAlattas, line 16: bonus change 👀

src/server/asset_path.ts
```diff
@@ -0,0 +1,4 @@
+export let ASSET_PATH_PREFIX = "";
+export function setAssetPathPrefix(prefix: string) {
+  ASSET_PATH_PREFIX = prefix + ASSET_PATH_PREFIX;
+}
```
MoaathAlattas: It is important to have this to be able to create a `staticFile` utility where `ASSET_PATH_PREFIX` is updated on app start.

src/server/context.ts
```diff
@@ -672,7 +691,7 @@ export class ServerContext {
         baseRoute: toBaseRoute(ALIVE_URL),
         methods: {
           default: () => {
-            let timerId: number | undefined = undefined;
+            let timerId: Timer | undefined = undefined;
```
MoaathAlattas: bonus change 👀

src/server/constants.ts
```diff
@@ -7,5 +8,8 @@ export const JS_PREFIX = `/js`;
 export const DEBUG = !Deno.env.get("DENO_DEPLOYMENT_ID");
 
 export function bundleAssetUrl(path: string) {
+  if (ASSET_PATH_PREFIX !== "") {
+    return `${ASSET_PATH_PREFIX}${INTERNAL_PREFIX}${JS_PREFIX}${path}`;
+  }
   return `${INTERNAL_PREFIX}${JS_PREFIX}/${BUILD_ID}${path}`;
```
marvinhagemeister: Nit: Since by default `ASSET_PATH_PREFIX` is an empty string, so we can just do:

```suggestion
  return `${ASSET_PATH_PREFIX}${INTERNAL_PREFIX}${JS_PREFIX}${path}`;
```

It's one less `if`-branch to worry about.

src/server/context.ts
marvinhagemeister: Can you change it back to `number`? There is no `Timer` type in Deno as their `setTimeout` return type is based on the browser API which returns a `number`. The `Timer` type only exists in node.

marvinhagemeister (commented):
This looks great! Excited to have this land in Fresh. Left some minor comments, let me know what you think. Let's add some tests too and then it's ready to be merged.

src/server/context.ts
MoaathAlattas: looks like i need to do something with my VSCode settings as it was showing an error. I will revert this.

src/server/constants.ts
MoaathAlattas: I have tried this in the beginning but I found that it is a tricky one! This will remove the build id for anyone not using a CDN which might cause issues with cache busting on new versions. 

Another option is to keep the build id and set the prefix as the cdn url only, i.e `setAssetPathPrefix(${opts.cdnUrl}/)` but that would mean static files won't be prefixed with a build id when using a CDN which is the same issue just in a different place.

We could also go with what i did in the first iteration where I created two variables: `STATIC_ASSET_PATH_PREFIX` and `INTERNAL_ASSET_PATH_PREFIX`. This will allow us to do `setStaticAssetPathPrefix(${opts.cdnUrl}/${BUILD_ID})` and `setInternalAssetPathPrefix(${opts.cdnUrl}/)`. It felt unnecessary but I think that aligns with the current setup. I also find this option makes storing files on the cdn confusing:
```
/static/[build_id]/...
/_frsh/js/[build_id]/...
```

rather than
```
[build_id]/static/...
[build_id]/_frsh/js/...
```

curious to hear your thoughts.

src/server/context.ts
marvinhagemeister: Could be that the Deno vscode extension wasn't active or something.

src/server/constants.ts
marvinhagemeister: Ohh i totally missed that. You're right, ignore my suggestion.

src/runtime/entrypoints/main.ts
```diff
@@ -856,7 +857,9 @@ export async function applyPartials(res: Response): Promise<void> {
 
 const originalHook = options.vnode;
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing assets when asset path prefixed,
+  // the CDN should handle caching
+  if (!hasAssetPrefix()) assetHashingHook(vnode);
```
MoaathAlattas, line 862: came across this when writing tests ([caching](https://fresh.deno.dev/docs/concepts/static-files#caching)). 

> Fresh also does this automatically for src and srcset attributes in  and <source> HTML tags. These will automatically use “locked” paths if Fresh deems it safe to do so. You can always opt out of this behaviour per tag, by adding the data-fresh-disable-lock attribute.

When using a CDN usually this will be handled there as every build will upload the files again with a different url.

src/server/rendering/preact_hooks.ts
```diff
@@ -174,7 +175,9 @@ const oldRender = options.__r;
 const oldHook = options.__h;
 
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing assets when asset path prefixed,
+  // the CDN should handle caching
+  if (!hasAssetPrefix()) assetHashingHook(vnode);
```
MoaathAlattas, line 180: same as https://github.com/denoland/fresh/pull/1892/files#r1355976552

src/runtime/entrypoints/main.ts
MoaathAlattas: a future improvement would be to combine both and make caching only opt-in, less magic. However, this is a breaking change for sites that expects caching is already there so I prefer to leave that as a future improvement.

MoaathAlattas:
@marvinhagemeister added tests and left comments inline for new updates since last time. Let me know if there is anything else needed.

src/runtime/entrypoints/main.ts
```diff
@@ -856,7 +857,9 @@ export async function applyPartials(res: Response): Promise<void> {
 
 const originalHook = options.vnode;
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing prefixed assets  because
+  // cache busting is expected to ba handled externally
+  if (!hasAssetPrefix()) assetHashingHook(vnode);
```
MoaathAlattas: came across this when writing tests ([caching](https://fresh.deno.dev/docs/concepts/static-files#caching)). 

> Fresh also does this automatically for src and srcset attributes in  and <source> HTML tags. These will automatically use “locked” paths if Fresh deems it safe to do so. You can always opt out of this behaviour per tag, by adding the data-fresh-disable-lock attribute.

When using a CDN usually this will be handled there as every build will upload the files again with a different url.

MoaathAlattas: a future improvement would be to combine both and make caching only opt-in, less magic IMO. However, this is a breaking change for sites that expect caching works by default so I prefer to leave that as a future improvement.

src/server/rendering/preact_hooks.ts
```diff
@@ -174,7 +175,9 @@ const oldRender = options.__r;
 const oldHook = options.__h;
 
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing prefixed assets  because
+  // cache busting is expected to ba handled externally
+  if (!hasAssetPrefix()) assetHashingHook(vnode);
```
MoaathAlattas: same as https://github.com/denoland/fresh/pull/1892/files#r1356019025

src/runtime/entrypoints/main.ts
```diff
@@ -856,7 +857,9 @@ export async function applyPartials(res: Response): Promise<void> {
 
 const originalHook = options.vnode;
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing prefixed assets  because
+  // cache busting is expected to ba handled externally
```
marvinhagemeister: Super tiny nit:

```suggestion
  // avoid hashing prefixed assets because
  // cache busting is expected to be handled externally
```

src/server/asset_path.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Checks if the asset path has a prefix.
+ */
+export function hasAssetPrefix() {
+  return ASSET_PATH_PREFIX !== "";
+}
+
+/**
+ * Helper to use in templates. It returns the full path of an asset by concatenating the ASSET_PATH_PREFIX with the provided path.
+ * This is useful to get the right path durning development and production.
```
marvinhagemeister: Nit: typo

```suggestion
 * This is useful to get the right path during development and production.
```

src/server/rendering/preact_hooks.ts
```diff
@@ -174,7 +175,9 @@ const oldRender = options.__r;
 const oldHook = options.__h;
 
 options.vnode = (vnode) => {
-  assetHashingHook(vnode);
+  // avoid hashing prefixed assets  because
+  // cache busting is expected to ba handled externally
```
marvinhagemeister: Super tiny nit:

```suggestion
  // avoid hashing prefixed assets because
  // cache busting is expected to be handled externally
```

marvinhagemeister (commented):
@MoaathAlattas  This looks really great! Love the direction and I think it's a very useful feature to have in Fresh. Left some super minor comments (typos mostly).

One thing I was wondering while going through the changes if there is a way to combine `staticFile` with `asset`. It seems like the intention of the CDN url prefix is to be sort of a global flag for assets and static files alike. That way you could easily switch between using a CDN or not, via a config flag without having to go through the actual code. What do you think?

EDIT: Forgot to add. Love that you added tests 👍

MoaathAlattas:
> One thing I was wondering while going through the changes if there is a way to combine `staticFile` with `asset`. It seems like the intention of the CDN url prefix is to be sort of a global flag for assets and static files alike. That way you could easily switch between using a CDN or not, via a config flag without having to go through the actual code. What do you think?

That's a good point, separating `asset` and `staticFile` doesn't feel right and could be confusing. I did combine them initially then I found out `asset` is applied automatically on src and srcset attributes **in `` and `<source>` tags only** which doesn't cover other static files. In case of using a cdn url you have to use `asset` explicitly for other static files to prefix it with the cdn url.

Also, I found a case where some applications, instead of uploading files manually, they just use a proxy on another url. Then use the proxied url to serve static files (files will be cached on the first hit). This wouldn't work with the current changes because we changed assets paths and disabled static and internal routes.

I will think about both for a bit and see what would be the best way without changing the current behavior.

PR #1893, opened 2023-10-10
feat: dynamic file imports

### LINKED ISSUE #1405: proposal: use dynamic imports for routes
[Deno Deploy now supports statically analyzable dynamic imports.](https://deno.com/deploy/changelog#statically-analyzable-dynamic-imports) This may remove, if not reduce, reliance on the `fresh.gen.ts` file.

iuioiua:
Closes #1405


iuioiua:
How can the failing tests be fixed, @marvinhagemeister?

src/dev/mod.ts
```diff
@@ -107,41 +107,28 @@ function toImportSpecifier(file: string) {
...
-      .join("\n")
-  }
-
 const manifest = {
   routes: {
     ${
-    routes.map((file, i) =>
-      `${JSON.stringify(`${toImportSpecifier(file)}`)}: $${i},`
-    )
+    routes.map((file) => `${toPath(file)}: await import(${toPath(file)}),`)
```
marvinhagemeister, line 125: This will cause the route to be eagerly imported the same way as before, just with an additional promise tick. Did you meant to go with `() => await import(...)` instead?

src/dev/mod.ts
```diff
@@ -107,41 +107,28 @@ function toImportSpecifier(file: string) {
...
       .join("\n    ")
   }
   },
   islands: {
     ${
-    islands.map((file, i) =>
-      `${JSON.stringify(`${toImportSpecifier(file)}`)}: $$${i},`
-    )
+    islands.map((file) => `${toPath(file)}: await import(${toPath(file)}),`)
```
marvinhagemeister, line 131: same here.

src/dev/mod.ts
iuioiua: Yeah, that was intentional. I don't see the benefit in them being functions that return the awaited imports. Just resolving up front straight away seems simpler.

iuioiua:
Any idea how to fix the failing tests, @marvinhagemeister?

PR #1894, opened 2023-10-10
chore: synchronize CI and local tests

iuioiua:
These should be kept the same in the future so tests are run identically locally and in CI.


marvinhagemeister (approved):
Thanks

PR #1903, opened 2023-10-11
refactor: replace references of options with config

iuioiua:
Previously, there were repeated inconsistencies in the codebase regarding whether Fresh's settings object was named `options` or `config`. @marvinhagemeister and I have agreed to fix these inconsistencies and name everything `config`. This way, everything is clear. Along with public API deprecations, where appropriate, this change replaces the name and use of:
* `options` and `opts` variables with `config`
* `*Options` types/interfaces with `*Config`
* `options.ts` files with `fresh.config.ts`


marvinhagemeister (approved):
Nice, thanks for taking the time to make it more unified 👍

PR #1910, opened 2023-10-11
fix: Remove deno_kv_auth plugin

marvinhagemeister:
After a bit of internal review, we deemed that it is better if the the KV auth plugin is developed independently from Fresh. It's too early for it to be treated as an official plugin just yet.


sant123:
Should the [blog](https://deno.com/blog/fresh-1.5) be modified to not include the `Deno KV OAuth` highlight?

PR #1930, opened 2023-10-13
feat: add support for boosting forms

### LINKED ISSUE #1928: Partials feedback
Just playing around with Partials and forms.  Some feedback:

In the [optimising partial requests](https://fresh.deno.dev/docs/concepts/partials#optimizing-partial-requests) documentation it shows:

image

From my experiments, I believe this file location should be `routes/partials/docs/[id].tsx`.

With regards to form usage I ran into several problems:
1. In order to support deep linking for forms and load a specific snapshot of my search, I rely on the `onSubmit` event to update the URL search parms with the values of all the form fields.  However, the `f-partial` form submission isn't triggering this event.
2. The partial fetch is a GET request, meaning no form data is sent.  I could in theory add all form values (for my form anyway) as search parameters onto the `f-partial` value but this needs to be done at submission and I'm not able to hook into this event (but really, I think the fetch needs to be a POST with all form data).

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/1928

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -1049,12 +1049,37 @@ addEventListener("popstate", async (e) => {
 document.addEventListener("submit", async (e) => {
   const el = e.target;
   if (el !== null && el instanceof HTMLFormElement && !e.defaultPrevented) {
-    const partial = el.getAttribute(PARTIAL_ATTR);
-    if (partial !== null) {
+    if (!checkClientNavEnabled()) {
+      return;
+    }
+
+    const lowerMethod = el.method.toLowerCase();
+    if (
+      lowerMethod !== "get" && lowerMethod !== "post" &&
+      lowerMethod !== "dialog"
+    ) {
+      return;
+    }
+
+    const action = el.getAttribute(PARTIAL_ATTR) ?? el.action;
+    if (action !== "") {
       e.preventDefault();
 
-      const url = new URL(partial, location.href);
-      await fetchPartials(url);
+      const url = new URL(action, location.href);
+
+      let init: RequestInit | undefined;
+
+      // GET method appends form data via url search params
+      if (lowerMethod === "get") {
+        // TODO: Looks like constructor type for URLSearchParam is wrong
+        // deno-lint-ignore no-explicit-any
+        const qs = new URLSearchParams(new FormData(el) as any);
+        qs.forEach((value, key) => url.searchParams.set(key, value));
+      } else {
+        init = { body: new FormData(el), method: el.method };
+      }
+
+      await fetchPartials(url, init);
     }
   }
 });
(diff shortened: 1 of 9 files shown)


src/runtime/entrypoints/main.ts
```diff
@@ -1045,12 +1045,37 @@ addEventListener("popstate", async (e) => {
 document.addEventListener("submit", async (e) => {
   const el = e.target;
   if (el !== null && el instanceof HTMLFormElement && !e.defaultPrevented) {
-    const partial = el.getAttribute(PARTIAL_ATTR);
-    if (partial !== null) {
+    if (!checkClientNavEnabled()) {
+      return;
+    }
+
+    const lowerMethod = el.method.toLowerCase();
```
csvn: It might be out of scope for this MR, or not intended to be supported right now, but it could be a bonus to check `e.submitter`. The submitter element that submits the form can override the forms fields. I think the only relevant ones are:

- `<button type="submit" formmethod="..">` ([MDN][mdn1])
- `<button type="submit" formaction="..">` ([MDN][mdn2])

```ts
const method = e.submitter?.formMethod ?? el.method;
const lowerMethod = method.toLowerCase();
```

The same might be needed for `action` below, but I don't think it's needed for the others; `enctype`, `target` & `novalidate` (as they are not used currently when fetching partial).

[mdn1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#formmethod
[mdn2]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#formaction

src/runtime/entrypoints/main.ts
```diff
@@ -1045,12 +1045,37 @@ addEventListener("popstate", async (e) => {
 document.addEventListener("submit", async (e) => {
   const el = e.target;
   if (el !== null && el instanceof HTMLFormElement && !e.defaultPrevented) {
-    const partial = el.getAttribute(PARTIAL_ATTR);
-    if (partial !== null) {
+    if (!checkClientNavEnabled()) {
```
adamgreg: Very minor point, but could you make this check right at the start of the event handler instead?

adamgreg:
LGTM (not that anyone asked! :smile:). What remains before this can be merged? I'm going to merge this branch in my own repo and give it a try.

EDIT: I'm using it now - works great, thanks!

src/runtime/entrypoints/main.ts
adamgreg: Ah no sorry, looks like it actually needs `el` as an argument

marvinhagemeister:
> What remains before this can be merged?

I'll likely cut another patch release this week and after that it's ready to land. I tend to delay feature PRs for about a week after the monthly Fresh release, because then I don't have to deal with backporting fixes between various branches.

src/runtime/entrypoints/main.ts
marvinhagemeister: Thanks for pointing this out. Added it to this PR 👍

PR #1936, opened 2023-10-14
feat: adds easier Headers configuration

RoeHH:
This change would make it easier to add static http headers that are better separate from the logic inside Middlewares


RoeHH:
Take a look at docs/latest/concepts/middleware.md on how to use :)

marvinhagemeister:
Thanks for opening a PR. I'm a bit worried by the increase in special files that trigger specific behavior in Fresh. When it comes to HTTP headers in particular, I'm not sure if the overhead and additional teachings we have to do is worth it. Whenever you encounter a `Response` instance, you can always add headers on the fly to it. Doesn't matter if it's inside a middleware, a route handler or somewhere else.

```tsx
export async function handler(req, ctx) {
  const res = await ctx.next();
  res.headers.set("X-Foo", "My HTTP header value");
  return res;
}
```

The nice thing about this is that we don't have to teach users new Fresh specific concepts about how to add HTTP headers. They can just add them like they normally would. What do you think?

RoeHH:
Sure, that makes a lot of sense. I would suggest we delete the PR then 👌
It was fun implementing this though :)
And I really appreciate the polite answer <3

marvinhagemeister:
Yeah, these explorations are necessary to test how it feels like in practice. A lot of code I write isn't merged either, but it was necessary to write it down to know if that was a good idea or not. 👍

Issue #1920, opened 2023-10-12
Client-side navigation doesn't work with 404 routes


lionel-rowe:
Adding client-side navigation with the new `f-client-nav` and `Partial` doesn't work with 404 routes.

Repro ([CodeSandbox](https://codesandbox.io/p/sandbox/fresh-1-5-1-client-nav-repro-486mzt?file=%2Froutes%2F_layout.tsx%3A7%2C27) — I've re-used the same one from https://github.com/denoland/fresh/issues/1919):

```tsx
// routes/_layout.tsx

import { Partial } from '$fresh/runtime.ts'

export default function Layout({ url, Component, state }: LayoutProps<unknown, unknown>) {
  return <div f-client-nav>
    <nav>
      <ul class='text-blue-500 underline m-8 flex gap-12'>
        <li><a href='/'>red bg</a></li>
        <li><a href='/blue'>blue bg</a></li>
        <li><a href='/no-route-here'>404 page</a></li>
      </ul>
    </nav>
    <div class='mx-auto max-w-screen-md'>
      <Partial name='main-content'>
        <Component />
      </Partial>
    </div>
  </div>
}

// routes/_404.tsx

import { PageProps } from "$fresh/server.ts";

export default function _404() {
  return <div>404 not found!</div>;
}
```

When starting from `/`, navigating to `/no-route-here` fails to load the 404 page. Upon full-page reloading from `/no-route-here`, the 404 page is displayed.

PR #1937, opened 2023-10-14
fix if route not found for partial

fdemir:
Aims to fix the #1920.


deer:
@marvinhagemeister, I think this is done by #2072.

PR #1944, opened 2023-10-16
fix: dev client script being loaded twice

marvinhagemeister:
Follow up to https://github.com/denoland/fresh/pull/1943


src/runtime/client.ts
```diff
@@ -1,7 +1,5 @@
 let ws: WebSocket;
 
-let closed = false;
-
 let reconnectTimer: number;
 const backoff = [
```
iuioiua, line 4: FYI, there's https://deno.land/std@0.204.0/async/mod.ts?s=retry

Issue #1649, opened 2023-08-16
Feat: Expose routes to be able to generate swagger docs etc


marvinhagemeister:
Feature request [from twitter](https://twitter.com/JohnRodkey/status/1691911722834002108):

> Is it possible to also generate openapi/swagger docs?

I think a good way to explore this would be a way to expose the final routes to plugins.

fdemir:
is it possible to get handlers from a plugin?

marvinhagemeister:
No, neither routes, not available handlers are exposed to plugins.

joeky888:
Something like https://github.com/tazo90/next-openapi-gen would be nice.

PR #1946, opened 2023-10-16
feat: Expose routes #1649

RoeHH:
I saw this issue (https://github.com/denoland/fresh/issues/1649) here and thought, 'Hmm, why not just give it a go?' I am not sure if my approach to it was good, but I kinda like the idea.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -18,6 +18,7 @@ import {
   BaseRoute,
   ErrorPage,
   ErrorPageModule,
+  FreshInfo,
   Handler,
   InternalFreshConfig,
   Island,
@@ -413,6 +414,7 @@ export class ServerContext {
   #plugins: Plugin[];
   #builder: Builder | Promise<BuildSnapshot> | BuildSnapshot;
   #routerOptions: RouterOptions;
+  #freshInfo: FreshInfo;
 
   constructor(
     routes: Route[],
@@ -452,6 +454,7 @@ export class ServerContext {
       target,
     });
     this.#routerOptions = routerOptions;
+    this.#freshInfo = { routes };
   }
 
   /**
@@ -623,6 +626,7 @@ export class ServerContext {
         },
         destination: "route",
         params: paramsAndRouteResult.params,
+        info: this.#freshInfo,
       };
 
       for (const { module } of mws) {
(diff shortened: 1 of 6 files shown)


RoeHH:
While playing around with this i saw that it would be possible too add all Middlewares that apply to this Route or all handlers that exist in the fresh project

RoeHH:
@marvinhagemeister review :) ?

deer:
Hey @RoeHH, can you please merge main and deal with the conflicts? A lot has changed in `context.ts` since you created this a few weeks ago. It's probably best to make sure the PR is ready to go before trying to convince Marvin to take a look.

Two logistical points: 
* what about some documentation? As a user, how would I even discover this exists?
* I could also imagine the test being a bit more "functional" in the sense that your handler could create a "site map" or something similar. Then you could verify that your favorite route (perhaps your new "info-ctx" route) is included in the map. Of course the only way such a test could pass is if your code is working correctly.

For the actual code: 
* could you please clarify the name `FreshInfo`? I'm not sure we need a new type here, unless you have some ideas on other stuff that belongs inside this "info". It seems like you could just add `routes: Route<any>[]` to the context types and be done with it. 
* what about `UnknownHandlerContext` and `ErrorHandlerContext`? Right now they're feeling a bit left out. I guess there should at least be an explanation for why we don't want to pass the routes to these.

deer (changes requested):
Hey @RoeHH, can you please merge main and deal with the conflicts? A lot has changed in `context.ts` since you created this a few weeks ago. It's probably best to make sure the PR is ready to go before trying to convince Marvin to take a look.

Two logistical points: 
* what about some documentation? As a user, how would I even discover this exists?
* I could also imagine the test being a bit more "functional" in the sense that your handler could create a "site map" or something similar. Then you could verify that your favorite route (perhaps your new "info-ctx" route) is included in the map. Of course the only way such a test could pass is if your code is working correctly.

For the actual code: 
* could you please clarify the name `FreshInfo`? I'm not sure we need a new type here, unless you have some ideas on other stuff that belongs inside this "info". It seems like you could just add `routes: Route<any>[]` to the context types and be done with it. 
* what about `UnknownHandlerContext` and `ErrorHandlerContext`? Right now they're feeling a bit left out. I guess there should at least be an explanation for why we don't want to pass the routes to these.

PR #1948, opened 2023-10-17
fix: missing generics for plugin entrypoint

MadTyres:
Follow-up for https://github.com/denoland/fresh/pull/1776

I forgot to add generics to the plugin entrypoint 🤦 
It still work as is, but you'd get red squiggly lines


marvinhagemeister (approved):
Good catch 👍

PR #1953, opened 2023-10-17
feat: PostCSS & Tailwind Plugins

jasonjgardner:
I've added two plugins: a PostCSS plugin and a Tailwind plugin (which uses the PostCSS plugin).

- Has dev mode and build mode
- Inlines source map
- Works with Tailwind IntelliSense
- Should work with Tailwind and PostCSS plugins if they are imported and instantiated successfully

### Example usage
```ts
import { defineConfig } from "$fresh/server.ts";
import postcssPlugin from "$fresh/plugins/postcss.ts";
import tailwindPlugin from "$fresh/plugins/tailwind.ts";
import pfm from "https://esm.sh/postcss-font-magician";

export default defineConfig({
  plugins: [
    postcssPlugin({
      css: ["./src/font_magician.css"],
      dest: "./static/styles",
      mode: "render",
      sourceMap: true,
      setup: () => {
        return [
          pfm({
            variants: {
              "Inter": {
                // ...
              },
            },
            foundries: ["google"],
          }),
        ];
      },
    }),
    // Development mode
    tailwindPlugin(
      {
        css: [
          "body { @apply bg-yellow-500 }",
          "./src/style.css",
        ],
      },
      (await import("./tailwind.config.ts")).default,
    ),
    // Production mode
    // Outputs to ./static/styles/myDestinationFileName.css
    tailwindPlugin(
      {
        css: { myDestinationFileName: "./src/style.css" },
        mode: "build",
        dest: "./static/styles",
      },
      (await import("./tailwind.config.ts")).default,
    ),
  ],
});
```

### Result
Fresh test


plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  if (mode === "render") {
+    plugin.renderAsync = async (ctx) => {
+      const res = await ctx.renderAsync();
+
+      return {
+        styles: await getStyles(res.htmlText),
+      };
+    };
+  }
```
marvinhagemeister, line 114: Let's drop the `"render"` mode. To make this work with conditional code inside islands, this would require shipping tailwind to the browser which we don't want. We're actively trying to move people away from JIT building.

plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+        await ensureDir(options.dest);
+      }
+
+      const styles = await getStyles();
+
+      await Promise.all(styles.map(async (style, idx) => {
+        const fileName = style.id ?? `style_${idx}`;
+        await Deno.writeTextFile(
+          join(
+            options.dest ?? "./static",
```
marvinhagemeister, line 128: I feel like we should write into Fresh's build directory which lies at `_fresh`. Once we expose those options to the plugin API we could retrieve the value from there. Main reason being that the `_fresh` folder is usually ignored by git etc already.

plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+): Promise<PluginRenderStyleTag[]> {
+  const postCssPlugins = options.setup
+    ? options.setup(content)
+    : [autoprefixer()];
+  const cssSrc = Array.isArray(src) ? src : [src];
+  const dest = options.dest ?? "./static";
+  return await Promise.all(
+    cssSrc.map(async (src, idx): Promise<PluginRenderStyleTag> => {
+      const isFile = src.startsWith("./") || src.startsWith("/") ||
+        src.startsWith("file://");
```
marvinhagemeister, line 38: Does the tailwind plugin pass inline styles to the `postcss` plugin? If yes, then I'd vote to keep the API surface small for now. That way we can also skip having to test this on all 3 OS platforms.

marvinhagemeister (commented):
This looks very promising! Let's add some tests to it to ensure that future changes don't break things

plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+      : await Promise.all(
+        Object.entries(options.css).map(async ([key, value]) => {
+          return {
+            id: `${STYLE_ELEMENT_ID}_${key}`,
+            cssText: (await processPostCss(value, options, content)).map((
+              style,
+            ) => style.cssText).join("\n"),
+          };
+        }),
+      );
```
iuioiua, line 104: This becomes simpler if we remove the `Record<string, string>` union member from `css`'s type, which seems to be a little extraneous.

plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export interface PostCssPluginOptions {
+  css: string | string[] | Record<string, string>;
+  sourceMap?: boolean;
+  dest?: string;
+  mode?: "render" | "build";
+  setup?: (
+    content?: string,
+  ) => AcceptedPlugin[];
+}
```
iuioiua, line 21: Are you able to add some JSDocs explaining these properties?

plugins/postcss.ts
```diff
@@ -0,0 +1,139 @@
...
+ *  plugins: [
+ *      postcssPlugin({
+ *          css: "./src/styles.css",
+ *          setup: (content) => [autoprefixer(), tailwind({ content })],
+ *          mode: "render",
+ *          dest: "./static/styles",
+ *      })
+ *    ],
+ *  });
+ * ```
```
iuioiua, line 83:
```suggestion
 * export default defineConfig({
 *   plugins: [
 *     postcssPlugin({
 *       css: "./src/styles.css",
 *       setup: (content) => [autoprefixer(), tailwind({ content })],
 *       mode: "render",
 *       dest: "./static/styles",
 *     }),
 *   ],
 * });
 * ```
```
To align with Deno's formatter.

plugins/tailwind.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @example
+ * ```ts
+ * import { defineConfig } from "$fresh/server.ts";
+ * import tailwindPlugin from "$fresh/plugins/tailwind.ts";
+ *
+ * export default defineConfig({
+ *   plugins: [
+ *     tailwindPlugin({
+ *       css: ["./src/styles.css"],
+ *     }, (await import("./tailwind.config.ts")).default),
```
iuioiua, line 19: It'd be clearer if we just passed the basic config object here. Otherwise, a newcomer might wonder what the config object is supposed to look like. E.g.:
```ts
{
  content: [
    "./routes/**/*.{ts,tsx}",
    "./islands/**/*.tsx",
    "./components/**/*.tsx",
  ]
}
```

plugins/tailwind.ts
```diff
@@ -0,0 +1,52 @@
...
+ *   ],
+ * });
+ * ```
+ */
+export default function tailwindPlugin(
+  options: PostCssPluginOptions | PostCssPluginOptions["css"],
```
iuioiua, line 25: I think the additional logic to account for union types should be worth the DX gained. In this instance, I think it's fine just to have `PostCssPluginOptions`.

plugins/postcss.ts
iuioiua: +1

iuioiua (changes requested):
Great start!

A few thoughts:
1. Is it worth having a separate PostCSS plugin, or rather, could we use PostCSS in the background powering the Tailwind plugin? I have yet to see any demand for a PostCSS plugin, and having another plugin adds 1 more consideration that a dev might have to make. If demand grows, we could factor out the logic into its own plugin like now.
2. Can this implementation be simpler? I think starting with a small plugin covering the basics and most use cases is best. If demand for further functionality emerges, we can add that in later. If the implementation already does that, cool! But it wouldn't be ideal if we implement functionality that's not commonly used. I lack experience setting up Tailwind to point out examples, but it's something that comes to mind when looking at the implementation.

Regarding tests, you want to use `/tests/twind_test.ts`, `/tests/fixture_twind_app` and `/tests/fixture_twind_hydrate` as a guide, but perhaps in a simpler manner. Also, having the main Fresh site (`/www`) use this plugin might be good. Arguments can be made for why that should be done in this PR or a follow-up PR. Either way, it'd be a comprehensive stress test for this implementation.

iuioiua:
Also, I tested this with Deno SaaSKit, and it's not working. Have I made a mistake with the configuration of the plugin?

https://github.com/denoland/saaskit/tree/tailwind-plugin-test

image

plugins/tailwind.ts
jasonjgardner: Yeah, plus I later realized an `import tailwindConfig from ./tailwind.config.ts` at the top might have made more sense too.

jasonjgardner:
@iuioiua 
> Also, I tested this with Deno SaaSKit, and it's not working. Have I made a mistake with the configuration of the plugin?

Switching the `mode` to `"render"` will let Tailwind process styles during dev. Otherwise, you'd have to run the build step to output the .css file, then add them to your `<head>`. 

But as @marvinhagemeister suggested, we can probably drop that now. I'm going to see what I can do with the features added in #1961. I think I can make the plugin much more succinct. Closing this PR for take-two.

Thank you both for your advice!

marvinhagemeister:
@jasonjgardner I'm planning to work on a fallthrough route for static assets so that generated ones can be stored in `build.outDir`. Hoping to get that done on Monday.

PR #1955, opened 2023-10-18
feat: Nested partials support

tlgimenes:
This PR adds initial support for nested Partials.

# How it works?
Suppose you have a scenario where you don't really have control over where/how your component is rendered, but you still want to be able to partially update/render this component:
```tsx
// Comp1.tsx
function Comp1({ children }) {
  return (
    <Partial name="comp-1">
      {children}
    </Partial>
  );
}

// Comp2.tsx
function Comp2() {
  return (
    <Partial name="comp-2">
      {/* component layout */}
    </Partial>
  );
}

defineRoute(() => {
  return (
    <>
      <Comp1>
        <Comp2 />
      </Comp1>
      <button f-partial="/update-comp1">Update 1</button>
      <button f-partial="/update-comp2">Update 2</button>
    </>
  );
});
```

With the current Fresh implementation, If you hit `Update 1`, everything works fine. However, if you press `Update 2`, an error is logged. This PR makes pressing `Update 2` to work.

# Limitations
Once `Update 2` is clicked, `<Comp2/>` is updated. However, If you then click on `Update 1`, both `<Comp1/>` and `<Comp2/>` will be updated. This behavior can be seen on the added tests.

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -103,16 +103,12 @@ export function revive(
   );
 
   for (let i = 0; i < result.length; i++) {
-    const { vnode, marker, rootFragment } = result[i];
+    const { vnode, rootFragment } = result[i];
     const _render = () => {
       render(
         vnode,
         rootFragment,
       );
-
-      if (marker.kind === MarkerKind.Partial) {
-        partials.set(marker.text, vnode as VNode<PartialComp>);
-      }
     };
 
     "scheduler" in window
@@ -148,6 +144,10 @@ function addPropsChild(parent: VNode, vnode: ComponentChildren) {
 class PartialComp extends Component<
   { children?: ComponentChildren; mode: number; name: string }
 > {
+  componentDidMount() {
+    partials.set(this.props.name, this);
+  }
+
   render() {
     return this.props.children;
   }
@@ -175,7 +175,7 @@ export interface RenderRequest {
 // Useful for debugging
 const SHOW_MARKERS = false;
 
-const partials = new Map<string, VNode<PartialComp>>();
+const partials = new Map<string, PartialComp>();
 
 /**
  * Replace comment markers with empty text nodes to hide them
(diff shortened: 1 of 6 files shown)


marvinhagemeister (commented):
Thanks for making a PR, love the tests!

Storing the vnode itself is a bit risky, because a new vnode will be created with every single render. If a parent does a state update, like an upper island then the partial vnode will be a different reference and the value that was stored in the Map becomes stale. The only thing that remains the same is the component instance (=`__c`).

The fix in this PR is very close and I only adapted it slightly to not rely on storing vnode in the Map. Here is the patch against the original code in `main` that addresses the issue here:

 ```patch
 diff --git a/src/runtime/entrypoints/main.ts b/src/runtime/entrypoints/main.ts
index 339efb8..dd1383d 100644
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -149,6 +149,9 @@ function addPropsChild(parent: VNode, vnode: ComponentChildren) {
 class PartialComp extends Component<
   { children?: ComponentChildren; mode: number; name: string }
 > {
+  componentDidMount() {
+    partials.set(this.props.name, this);
+  }
   render() {
     return this.props.children;
   }
@@ -432,11 +435,6 @@ function _walkInner(
             const parent = vnodeStack[vnodeStack.length - 1]!;
             addPropsChild(parent, vnode);
 
-            if (marker.kind === MarkerKind.Partial) {
-              // deno-lint-ignore no-explicit-any
-              partials.set(marker.text, (vnode as any).__c);
-            }
-
             sib = marker.endNode.nextSibling;
             continue;
           }
```

tlgimenes:
Nice! Thanks for the support @marvinhagemeister! I didn't know this about the vnode. Now the solution seems much more elegant.

marvinhagemeister (approved):
Sweet, thanks for fixing this 🎉

PR #1977, opened 2023-10-26
feat(server): add isPartial to context types

### LINKED ISSUE #1935: Add a way to identify if the current request is a Fresh Partial in routes/middleware
I was thinking of doing a PR for this, but realized it's probable best to check interest first. Similar to the reasoning for having `ctx.destination` in middleware, it would be nice for middleware/handlers/async routes to check if the current request is a partial. It's currently possible to do this by checking the URL `?fresh-partial=true`, but this feels a bit brittle (`PARTIAL_SEARCH_PARAM` is not exported).

My use case is having a form both be usable with the regular action, and when updating via partials. For instance, if a `<form>` posts to `/foo` and we want to return partials:

- One way would be to always render partials on the `/foo` endpoint, but this becomes weird if users disable their JS, where we get a strange result back (partial content)
- Another way is to create both a `/routes/foo` and `/routes/partials/foo` and set `action="/foo" f-partial="/partials/foo"` on the form. but it's annoying to have to duplicate the handling of the form in both places
- The way I intended to do it is to only have one the `/foo` route, but check if it's a partial (via query param currently). We first handle the form, then if it's a partial we return partial content, if not we redirect to a suitable place after the form

To make the last point easier it would be great if there was something built in to do something like the below:

```tsx
export const handler: Handlers<unknown, unknown> = {
  GET(req, ctx) {
    const fd = await req.formData();
    // Handle the form

    if (ctx.isPartial) return ctx.render();

    return new Response(undefined, {
      status: 303,
      headers: { location: '/' },
    });
  },
};
```

Another possible solution would be to export a helper function for checking this:
```tsx
import { isPartial } from '$fresh/server.ts';

    if (isPartial(req)) return ctx.render();
    // or
    if (isPartial(ctx)) return ctx.render();
```

deer:
closes https://github.com/denoland/fresh/issues/1935

`MiddlewareHandlerContext` and `HandlerContext` continue to get more and more similar, this time with `RouteContext` participating as well: all three get a new ` isPartial: boolean;` added.

The key to this change is the realization that we need to capture this when parsing the URL, i.e. inside the function returned by `getParamsAndRoute`. A small change there allows us to now return:
```tsx
       {
          route: route,
          params: groups,
          isPartial: urlObject.searchParams.has(PARTIAL_SEARCH_PARAM),
        }
```
Then this is available when we build the two context objects in `compose.ts`. After that I just had to clean up type errors and add some tests.

### DIFF
--- a/src/server/compose.ts
+++ b/src/server/compose.ts
@@ -1,4 +1,4 @@
-import { ErrorHandler, FinalHandler, InternalRoute } from "./router.ts";
+import { ErrorHandler, FinalHandler, RouteResult } from "./router.ts";
 import {
   BaseRoute,
   MiddlewareHandlerContext,
@@ -55,10 +55,7 @@ export function composeMiddlewares(
   errorHandler: ErrorHandler<RouterState>,
   paramsAndRoute: (
     url: string,
-  ) => {
-    route: InternalRoute<RouterState> | undefined;
-    params: Record<string, string>;
-  },
+  ) => RouteResult<RouterState>,
 ) {
   return (
     req: Request,
@@ -103,6 +100,7 @@ export function composeMiddlewares(
       },
       destination: "route",
       params: paramsAndRouteResult.params,
+      isPartial: paramsAndRouteResult.isPartial,
     };
 
     for (const { module } of mws) {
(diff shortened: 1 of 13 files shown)


marvinhagemeister (approved):
This is great! Thanks for making a PR 👍 

Yeah, agree with you that we have too many different context things. We should merge them all into one context type and make the properties that can only occur in other handlers optional. Would be much cleaner than what we have today.

PR #1979, opened 2023-10-26
Added perpinya.eu to the showcase

evilmonkey19:
Hello! 

I did my site using Fresh and Deno. As something special, I have added translations in 3 languages. I hope it can be added 🙇 . Thank you beforehand 😄


marvinhagemeister (approved):
Thanks 👍

PR #1982, opened 2023-10-27
docs: update using twindv1 page

### LINKED ISSUE #1672: Documentation: examples: 'Using Twind v1' is partially outdated
Create new fresh project with ```deno run -A -r https://fresh.deno.dev``` and both main.ts and dev.ts import config from same place: ```import config from "./fresh.config.ts";``` but example suggests to import plugin and config separately and do the same as  fresh.config.ts already does:
```
import twindPlugin from "$fresh/plugins/twindv1.ts";
import twindConfig from "./twind.config.ts";

await start(manifest, { plugins: [twindPlugin(twindConfig)] });
```

Following instructions as they are now leaves fresh.config.ts as it is and that leads to an type error in ```twindPlugin(twindConfig)```
in fresh.config.ts which still imports twindPlugin from old version.

To fix the example update it with suggested modifications for main.ts to be included for the fresh.config.ts instead

deer:
closes https://github.com/denoland/fresh/issues/1672

As correctly pointed out in the issue, this page is no longer up-to-date. Corrected that here.

### DIFF
--- a/docs/latest/examples/using-twind-v1.md
+++ b/docs/latest/examples/using-twind-v1.md
@@ -13,37 +13,46 @@ you'll end up with a `main.ts` like the following:
 /// <reference lib="dom.asynciterable" />
 /// <reference lib="deno.ns" />
 
+import "$std/dotenv/load.ts";
+
 import { start } from "$fresh/server.ts";
 import manifest from "./fresh.gen.ts";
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks ❤️

PR #1983, opened 2023-10-27
Update documentation for 'Internal Server Error' page behavior

gustavomaltez:
As described on #1978, the custom '500 Internal Server Error' page is not displayed when the app is running in dev mode. This can be confusing for developers, and it took me some time to realize this behavior.

This PR aims to address the issue by updating the documentation to clarify the default behavior of 'Internal Server Error' pages. It will help other developers understand how these pages work.

closes #1978

PR #1984, opened 2023-10-28
feat(server): throwing Deno.errors.NotFound triggers 404

### LINKED ISSUE #1759: suggestion: automatically render not found page
If [`Deno.errors.NotFound`](https://deno.land/api@v1.36.4?s=Deno.errors.NotFound) or [`errors.NotFound`](https://deno.land/std@0.201.0/http/http_errors.ts?s=errors) is thrown within a handler, the "HTTP 404 Not Found" page should be automatically rendered, by default. In other words, `throw new Deno.errors.NotFound()` would be the same as `return ctx.renderNotFound()`.

deer:
closes https://github.com/denoland/fresh/issues/1759

I've branched the documentation to canary because I've added the following snippet:
````
This can also be achieved by throwing an error, if you're uninterested in
passing specific data to your 404 page:

```tsx
import { Handlers } from "$fresh/server.ts";

  GET(_req, _ctx) {
    throw new Deno.errors.NotFound();
  },
};
```
````

Note: this soft-conflicts with https://github.com/denoland/fresh/pull/1983, since I don't have that addition to the page.

### DIFF
--- a/docs/canary/concepts/error-pages.md
+++ b/docs/canary/concepts/error-pages.md
@@ -1,0 +1,82 @@
+---
+description: |
+  Error pages can be used to customize the page that is shown when an error occurs in the application.
+---
+
+Fresh supports customizing the `404 Not Found`, and the
+`500 Internal Server Error` pages. These are shown when a request is made but no
+matching route exists, and when a middleware, route handler, or page component
+throws an error respectively.
+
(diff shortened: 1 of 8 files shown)


marvinhagemeister (approved):
This is a great DX feature 👍

PR #1987, opened 2023-10-29
docs: Add Tailwind plugin example

jasonjgardner:
- Added article to examples which outlines Tailwind plugin usage
- Updated TOC to include new article
- Added Fresh Images and Fresh Tailwind plugins to integrations list


docs/latest/examples/using-tailwindcss.md
```diff
@@ -0,0 +1,186 @@
...
+  plugins: [
+    tailwindPlugin({
+      css: "./src/style.css",
+    }),
+  ],
+});
+```
+
+The `dest` option can be specified to save output to a CSS file during
+[ahead-of-time builds](../concepts/ahead-of-time-builds.md).
```
deer, line 39: I can recommend running `deno task www` and clicking through things. This doesn't actually work! I think you need to remove the .md from the end.

jasonjgardner: cool

PR #1991, opened 2023-10-31
docs: add jsdoc for IS_BROWSER

### LINKED ISSUE #1719: Missing JSDocs for `IS_BROWSER` constant
An explanation and example would be great to have in the [API reference](https://deno.land/x/fresh/runtime.ts?doc=&s=IS_BROWSER). I'd also suggest linking any instances of `IS_BROWSER` in the documentation to the API reference page.

deer:
closes https://github.com/denoland/fresh/issues/1719

@iuioiua, what do you think of this? It's only referenced once in the documentation, hence I only modified `islands.md`.

I'm not sure if this is what you had in mind for the example. Let me know if anything can be improved.

### DIFF
--- a/docs/latest/concepts/islands.md
+++ b/docs/latest/concepts/islands.md
@@ -183,7 +183,8 @@ An error occurred during route handling or page rendering. ReferenceError: Event
     ....
 ```
 
-Use the [`IS_BROWSER`](https://deno.land/x/fresh/runtime.ts?doc=&s=IS_BROWSER) flag as a guard to fix the issue:
+Use the [`IS_BROWSER`](https://deno.land/x/fresh/runtime.ts?doc=&s=IS_BROWSER)
+flag as a guard to fix the issue:
 
 ```tsx islands/my-island.tsx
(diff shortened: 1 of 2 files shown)


docs/latest/concepts/islands.md
iuioiua, line 1: What about just linking the "`IS_BROWSER`" from the line above the snippet to the docs? Seems tidier.

src/runtime/utils.ts
iuioiua, line 1: I'd maybe re-word to say something like "Returns true when the current runtime is the browser and false otherwise. This is used for guard runtime-dependent code." and provider an `@example` snippet. WDYT?

docs/latest/concepts/islands.md
deer: 🤦 yes of course

src/runtime/utils.ts
deer: sounds great. didn't know about `@example`. I even learned to demarcate it with ```

marvinhagemeister (approved):
Sweet, thank you for adding more docs to this 👍

PR #2000, opened 2023-11-03
chore: bump deps

deer:
This primarily updates `std` to 0.205.0. We were previously using 0.193.0, which was just about four months old. (Although some places were using other versions as well).

I also pulled lots of import statements out of individual files and pushed them into various `deps.ts` files, in order to make this sort of thing easier in the future.

This also updates ts-morph to 20.0.0.

### DIFF
--- a/update.ts
+++ b/update.ts
@@ -231,7 +231,7 @@ await start(manifest, { plugins: [twindPlugin(twindConfig)] });\n`;
     for (const n of nodes) {
       if (!n.wasForgotten() && Node.isJsxAttribute(n)) {
         const init = n.getInitializer();
-        const name = n.getName();
+        const name = n.getStructure().name;
         if (
           Node.isJsxExpression(init) &&
           (name === "class" || name === "className")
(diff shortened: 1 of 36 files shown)


update.ts
```diff
@@ -231,7 +231,7 @@ await start(manifest, { plugins: [twindPlugin(twindConfig)] });\n`;
     for (const n of nodes) {
       if (!n.wasForgotten() && Node.isJsxAttribute(n)) {
         const init = n.getInitializer();
-        const name = n.getName();
+        const name = n.getStructure().name;
```
deer, line 234: I don't think there's a test for this, because just commenting the line out also seems to work. I wonder how important all of this is, because I guess not many people are still using the old style:
```
This project is using an old version of the twind integration. Would you like to
update to the new twind plugin? This will remove the 'class={tw\`border\`}'
boilerplate from your source code replace it with the simpler 'class="border"'.
```

But I did do some due diligence, and logged out the result of `n.getName()` with 17.0.1, and that seems to perfectly match what `n.getStructure().name` is doing with 20.0.0, so I guess these two are the same.

marvinhagemeister: Sounds good to me. No idea why it was renamed.

marvinhagemeister (approved):
Awesome! Thanks for updating these 👍

PR #2002, opened 2023-11-03
chore: refactor dev client

### LINKED ISSUE #1992: Can't navigate when layout has more than one island
See https://github.com/adamzerner/nav-issue-demo for reproduction.

First I created a new Fresh project. Then I set up a `_layout.tsx` file like so:

```
import { LayoutProps } from "$fresh/server.ts";
import { Foo } from "../islands/foo.tsx";
import { Bar } from "../islands/bar.tsx";

export default ({ Component }: LayoutProps) => {
  return (
    <div>
      <Foo />
      <Bar />
      <Component />
    </div>
  );
};
```
Where `Foo` and `Bar` are the following (no JavaScript):
```
export const Foo = () => {
  return <div>Foo</div>;
};
```
and
```
export const Bar = () => {
  return <div>Bar</div>;
};
```
Then, in `routes/_index.tsx`, I added `<a href="/greet/alice">Greet Alice</a>`. If I remove `<Bar />`, the link works. But with both `<Foo />` and `<Bar />` in the layout, it does not.

marvinhagemeister:
Most of it was already fixed by #1956 but #1992 gives us a much better test case.

Fixes https://github.com/denoland/fresh/issues/1992

### DIFF
--- a/src/runtime/entrypoints/client.ts
+++ b/src/runtime/entrypoints/client.ts
@@ -1,4 +1,5 @@
 let ws: WebSocket;
+let revision = 0;
 
 let reconnectTimer: number;
 const backoff = [
@@ -37,7 +38,7 @@ function reconnect() {
     backoffIdx++;
 
     try {
-      connect(true);
+      connect();
       clearTimeout(reconnectTimer);
     } catch (_err) {
       reconnect();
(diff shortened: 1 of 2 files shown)


deer:
When I comment out the fix, the tests still pass ten times locally. What happens on your computer?

marvinhagemeister:
Right, looks like we've already got this covered. Changed the PR to be a refactor of the dev client.

PR #2004, opened 2023-11-04
docs: update active link page to clarify twindv1 plugin usage

### LINKED ISSUE #1925: Docs: Update "Styling active links" example for twindv1
In the example for T(ail)wind it says that

> you can style `data-*`-attributes by prepending the classes with `[data-current]:`

When using the twindv1 plugin however, you need to use it like this: `data-[current]:`

```jsx
function Menu() {
  return (
    <a href="/foo" class="data-[current]:text-green-600">
      link to some page
    </a>
  );
}
```

deer:
closes https://github.com/denoland/fresh/issues/1925

I did test this out and what Felix reports is true: twindv1 requires a slightly different syntax. I explicitly called out the plugins as subheadings.

### DIFF
--- a/docs/latest/examples/active-links.md
+++ b/docs/latest/examples/active-links.md
@@ -29,7 +29,7 @@ a[data-ancestor] {
 
 ## Tailwind / Twind
 
-In tailwind or tailwind like styling solutions you can style `data-*`-attributes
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks!

PR #2007, opened 2023-11-05
feat(server): automatically extract forwarded-for header

### LINKED ISSUE #1827: add connInfo in custom handlers
It would be really cool to be able to access connInfo from custom handlers
```
   async GET(_req, ctx) {
  // access to connInfo somewhere here
     return await ctx.render();
   },
};
```

deer:
closes https://github.com/denoland/fresh/issues/1827
Or rather, the discussion that happens later in the issue.

This change modifies the call to `Deno.serve` to automatically extract the `X-Forwarded-For` header, if it exists. This is then available in `HandlerContext`, `MiddlewareHandlerContext`, and `RouteContext`.

### DIFF
--- a/src/server/boot.ts
+++ b/src/server/boot.ts
@@ -60,21 +60,27 @@ async function bootServer(
   handler: ServeHandler,
   opts: Partial<Deno.ServeTlsOptions>,
 ) {
-  // @ts-ignore Ignore type error when type checking with Deno versions
   if (typeof Deno.serve === "function") {
-    // @ts-ignore Ignore type error when type checking with Deno versions
-    await Deno.serve(
-      opts,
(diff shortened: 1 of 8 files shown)


deer:
@marvinhagemeister, based on https://github.com/denoland/fresh/issues/1827#issuecomment-1794435015 I guess I should close this. Before I do though: do you see any value in having the `handler` passed to `Deno.serve` configurable/overridable? If not, I'll close it. If yes, I can rework this.

Right now I don't see any way of controlling middleware execution order, so if someone absolutely needs something to run first, they're out of luck. Being able to override the `Deno.serve` handler and having the ability to order middleware (especially from plugins) seems like it would increase flexibility. I can open a separate issue for middleware ordering if desired.

PR #2017, opened 2023-11-06
fix: twind styles not updated on partial nav

### LINKED ISSUE #1919: Client-side navigation breaks tailwind styles
Adding client-side navigation with the new `f-client-nav` and `Partial` breaks tailwind styles. The breakage seems flaky from my testing — occasionally it seems to work, but the logic for when that happens isn't immediately obvious.

Repro ([CodeSandbox](https://codesandbox.io/p/sandbox/fresh-1-5-1-client-nav-repro-486mzt?file=%2Froutes%2F_layout.tsx%3A7%2C27)):

```tsx
// routes/_layout.tsx

import { Partial } from '$fresh/runtime.ts'

export default function Layout({ url, Component, state }: LayoutProps<unknown, unknown>) {
  return <div f-client-nav>
    <nav>
      <ul class='text-blue-500 underline m-8 flex gap-12'>
        <li><a href='/'>red bg</a></li>
        <li><a href='/blue'>blue bg</a></li>
      </ul>
    </nav>
    <div class='mx-auto max-w-screen-md'>
      <Partial name='main-content'>
        <Component />
      </Partial>
    </div>
  </div>
}

// routes/index.tsx

  return <div class='bg-red-500'>
    Should have red bg
  </div>
}

// routes/blue.tsx

export default function Blue() {
  return <div class='bg-blue-500'>
    Should have blue bg
  </div>
}
```

When starting from `/`, navigating to `/blue` fails to load the `bg-blue-500` styles (though the class is applied correctly). Upon full-page reloading from `/blue`, the `bg-blue-500` styles are correctly applied; however, upon navigating back to `/`, the `bg-red-500` styles are no longer loaded.

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/1919

### DIFF
--- a/plugins/twind.ts
+++ b/plugins/twind.ts
@@ -17,32 +17,28 @@ export default function(state) { hydrate(options, state); }`;
     entrypoints: { "main": main },
     async renderAsync(ctx) {
       sheet.reset(undefined);
-      const res = await ctx.renderAsync();
+      await ctx.renderAsync();
       const cssTexts = [...sheet.target];
       const snapshot = sheet.reset();
-      const scripts = [];
-      let cssText: string;
-      if (res.requiresHydration) {
-        const precedences = snapshot[1] as number[];
-        cssText = cssTexts.map((cssText, i) =>
-          `${cssText}/*${precedences[i].toString(36)}*/`
-        ).join("\n");
-        const mappings: (string | [string, string])[] = [];
-        for (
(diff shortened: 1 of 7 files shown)


lucacasonato:
This is very sad :(

lucacasonato:
We should only do this if there is a `fresh-client-nav` attribute in the page

PR #2019, opened 2023-11-06
docs: switch to aria-current in active links page

### LINKED ISSUE #2018: docs: Point to `aria-current` instead of `data-*` attributes for active links
We need to update our docs to point to `aria-current` instead of `data-*` properties for styling active links. Both are supported, but going with `aria-current` has benefits for accessibility and we should remove the `data-current` and `data-ancestor` at some point.

deer:
closes https://github.com/denoland/fresh/issues/2018

I'm not sure I've handled the removal of `data-ancestor` correctly. Based on reading the code and the tests, it seems like `aria-current='true'` is equivalent to `data-ancestor` and `aria-current='page'` is equivalent to `data-current`.

Although the usage of `aria-current='true'` also doesn't line up with the spec I saw. Is `true` a non-standard usage?


marvinhagemeister (approved):
Yeah I think `true` is a non-standard value. Not sure what the appropriate one would be. For now I think this is fine as it's better than recommending `data-current`

PR #2028, opened 2023-11-07
showcase: streamdal.com

Dosplatos:
The Console UI for Streamdal is a Deno + Fresh project. 

We thought it would be fantastic for the showcase!


marvinhagemeister (approved):
Nice! Was initially a bit confused, because the main site is built with astro, but the actual product is built with Fresh from what I can tell.

PR #2030, opened 2023-11-08
feat(server): disable 404 response

### LINKED ISSUE #1918: way to disable 404 route
Follow-up from #1916

Might it be valuable to provide a way to disable the 404 route?

deer:
closes https://github.com/denoland/fresh/issues/1918

@harrysolovay, what do you think of this? This change allows you to do something like this:
```ts
export default defineConfig({
  router: {
    disabled404response: () => {
      throw new Error("my custom error");
    },
  },
});
```
or
```ts
export default defineConfig({
  router: {
    disabled404response: () => {
      return new Response("asdf");
    },
  },
});
```

I can look to see what it would take to allow just returning `undefined`, but that seems like a larger change.


deer:
No feedback from the requester, so I'll just mark this as ready and see what happens. I'm totally open to reworking this, if desired.

harrysolovay:
Apologies I haven't had time to give an appropriate review. TBH I don't believe I'd be the right reviewer. Not familiar with the codebase.

deer:
@harrysolovay, sorry for the miscommunication. All I was looking for was whether the interface of the config is acceptable to you. I know you were hoping for returning `undefined`, but I thought perhaps a configurable function allowing you to return whatever error you want would possibly be acceptable.

harrysolovay:
I suppose it's an improvement over needing to create a dedicated 404 route file. Still, I believe getting back a `Response | undefined` from the render fn would be preferable.

PR #2034, opened 2023-11-12
fix(server): use relative import instead of relying on an import map

### LINKED ISSUE #2033: Internal import should be relative instead of relying on an import map
https://github.com/denoland/fresh/blob/7ff2781fcd0420b9755c24a634dbb81dd6392c34/src/server/default_error_page.tsx#L4

```
$ deno cache src/server/main.ts 
error: Relative import path "$fresh/src/server/deps.ts" not prefixed with / or ./ or ../ and not in import map from "https://deno.land/x/fresh@1.5.4/src/server/default_error_page.tsx"
    at https://deno.land/x/fresh@1.5.4/src/server/default_error_page.tsx:4:24
```

This issue probably went unnoticed because the standard import map for new fresh projects does define the specifier "$fresh".  For me it is problematic because the import map of my fresh project does use the specifier "fresh" instead of "$fresh" (for consistency reasons).

I've temporarily fixed this by defining both mappings for my fresh project:
```json
	"$fresh/": "https://deno.land/x/fresh@1.5.4/",
	"fresh/": "https://deno.land/x/fresh@1.5.4/",
```

kellnerd:
I haven't setup a local development environment to test this, but I'm pretty sure this fixes #2033

### DIFF
--- a/src/server/default_error_page.tsx
+++ b/src/server/default_error_page.tsx
@@ -1,7 +1,7 @@
 import { ComponentChildren, h } from "preact";
 import { DEBUG } from "./constants.ts";
 import type { ErrorPageProps, RouteConfig } from "./types.ts";
-import { colors } from "$fresh/src/server/deps.ts";
+import { colors } from "./deps.ts";
 
 export const config: RouteConfig = {
   skipAppWrapper: true,


marvinhagemeister (approved):
Thanks 👍

Issue #2039, opened 2023-11-15
Docs: get params from ctx instead of props in defineRoutes.


harshmangalam:
In Partials docs getting params from props instead of ctx

deer:
@marvinhagemeister, this can be closed.

PR #2040, opened 2023-11-15
docs: fix get params from ctx instead of props

harshmangalam:
#2039


marvinhagemeister (approved):
Whoops, good catch! Thanks for fixing this ❤️

PR #2051, opened 2023-11-20
chore: merge language-specific VSCode settings

interrrp:


marvinhagemeister (approved):
Thanks!

marvinhagemeister:
For some reason this confuses my vscode. Not sure if it's a bug in vscode itself or in the deno extension. What happens on my end is that the default formatter is always reset to prettier instead of the one from Deno since landing this PR. Gonna revert it.

PR #2061, opened 2023-11-22
chore: switch init to use tailwind

deer:
A small change to help out with the migration from `twind` to `tailwind`. I suppose this is a breaking change, because it changes the command line args. If someone has scripted `deno run -Ar https://fresh.deno.dev` with `--twind`, then this will no longer work. I can redo the change to avoid this, but scripting this seems like a strange thing to do, so I thought this change might be fine.

In addition, I can think of two more PRs to tackle as well:
* change `update.ts`
* ensure documentation is correct (docs on how to use the new plugin, remove lingering references to `twind` in code snippets)

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -4,7 +4,7 @@ import { collect, ensureMinDenoVersion, generate } from "./src/dev/mod.ts";
 import {
   dotenvImports,
   freshImports,
-  twindImports,
+  tailwindImports,
 } from "./src/dev/imports.ts";
 
 ensureMinDenoVersion();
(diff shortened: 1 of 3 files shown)


deer:
Additionally it looks like "someone" snuck `asdf` into a test name. I guess they were doing this to work around the bug in the test runner to ensure only that particular test would run. Anyway, this has now been removed. 🙈

marvinhagemeister (approved):
Sweet, thank you for adding this 🙌

PR #2079, opened 2023-11-28
chore: remove $fresh from import statements

deer:
There have been a few issues where `$fresh` has crept into import statements in the source code. I thought I would go through and purge all the remaining usages.


marvinhagemeister (approved):
Nice! Yeah I've been bugging folks here to also suggest the relative path in the VSCode extension. It's too easy to add the wrong path.

PR #2081, opened 2023-11-28
chore: bump std to 208 and deal with deprecations

deer:
Note the change to `main_test` where the expected value becomes e.g. `text/plain; charset=UTF-8` instead of `text/plain`. Thoughts? It seems more correct now, but I guess this could break people who depend on the old, less correct version. Additionally, should I add something to `src/server/mod.ts` like the following?
```ts
import { STATUS_CODE } from "./deps.ts";
export const Status = STATUS_CODE;
```
This would allow people to still do `import { Status } from "$fresh/server.ts";`, but they should probably just be importing it from `std` directly, right?


marvinhagemeister (approved):
Sweet, thank you 👍 I wouldn't worry about the encoding being part of the `Content-Type` header now. That's a perfectly valid value.

Issue #2062, opened 2023-11-22
flaky test -- Error: Text file busy 


deer:
I've seen a new issue pop up on some of the recently merged commits:

* https://github.com/denoland/fresh/actions/runs/6955879237/job/18925570743#step:10:937
* https://github.com/denoland/fresh/actions/runs/6942541716/job/18885712146#step:10:915
* https://github.com/denoland/fresh/actions/runs/7020837075/job/19101492223?pr=2083#step:10:1245
* https://github.com/denoland/fresh/actions/runs/7044513115/job/19172340792?pr=2088#step:10:1279

```
esbuild snapshot with cwd=Deno.cwd() => ./src/build/esbuild_test.ts:22:6
error: Error: Text file busy (os error 26)
      const child = Deno.run({
                         ^
    at opRun (ext:runtime/40_process.js:49:14)
    at Object.run (ext:runtime/40_process.js:135:15)
    at https://deno.land/x/esbuild@v0.19.4/mod.js:1880:26
    at eventLoopTick (ext:core/01_core.js:178:11)
    at async Module.initialize (https://deno.land/x/esbuild@v0.19.4/mod.js:1768:3)
    at async EsbuildBuilder.build (file:///home/runner/work/fresh/fresh/src/build/esbuild.ts:53:7)
    at async file:///home/runner/work/fresh/fresh/src/build/esbuild_test.ts:35:20

esbuild snapshot with cwd=/ => ./src/build/esbuild_test.ts:39:6
error: Error: Text file busy (os error 26)
      const child = Deno.run({
                         ^
    at opRun (ext:runtime/40_process.js:49:14)
    at Object.run (ext:runtime/40_process.js:135:15)
    at https://deno.land/x/esbuild@v0.19.4/mod.js:1880:26
    at eventLoopTick (ext:core/01_core.js:178:11)
    at async Module.initialize (https://deno.land/x/esbuild@v0.19.4/mod.js:1768:3)
    at async EsbuildBuilder.build (file:///home/runner/work/fresh/fresh/src/build/esbuild.ts:53:7)
    at async file:///home/runner/work/fresh/fresh/src/build/esbuild_test.ts:35:20
```

marvinhagemeister:
Closing as this was an issue in the Fresh 1.x test suite.

PR #2083, opened 2023-11-28
chore: rename referrer to referer

deer:
As per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer, which links to the interesting https://en.wikipedia.org/wiki/HTTP_referer, the incorrect `referer` is actually correct when using headers.

I noticed this when frantically trying to keep up with your flood of PRs while trying to get the base path stuff working. The header coming in does indeed use `referer`.

@marvinhagemeister


deer:
Ubuntu is failing due to the issue reported in #2062.

src/server/context.ts
```diff
@@ -484,7 +484,7 @@ export class ServerContext {
         methods: {
           default: async (req, ctx) => {
             if (
-              req.headers.get("referrer")?.includes("/_frsh/error_overlay")
+              req.headers.get("referer")?.includes("/_frsh/error_overlay")
             ) {
               throw new Error("fail");
             }
```
marvinhagemeister: Ah I think I accidentally committed that whilst adding the error overlay. It can be removed.
```suggestion
```

marvinhagemeister (approved):

PR #2089, opened 2023-11-28
docs/examples: Client side components and libraries

guilherme-miyake:
> I've struggled one too many times fighting front-end libs that are not exactly compatible with SSR, Deno, Fresh or any combination of those, so this is my current go-to solution for those cases
> Any suggestion/recommendation/alternatives to this? Do you think this would fit in the examples doc page with some explanations?
@guilherme-miyake

> Go for it. That's a great example to show client only components 👍
@martinrempel

Here it is!

### DIFF
--- a/docs/latest/examples/client-side-components-and-libraries.md
+++ b/docs/latest/examples/client-side-components-and-libraries.md
@@ -1,0 +1,176 @@
+---
+description: |
+  Client side components and libraries
+---
+
+## Introduction
+
+Some components depend on client environments, browser-specific features, or dynamic user interactions, making
+them incompatible or non-functional during server-side rendering.
+
+By employing conditional rendering and state management techniques, we can ensure graceful handling of library or data loading, improving workflow and usability of
+such components.
+
+Let's explore an example utilizing this solution with Leaflet, a popular mapping library, in a Fresh application. The
+objective is to ensure the proper rendering of Leaflet components on the client side while gracefully handling them on
+the server side.
+
+The full code is available at the end of the page
+
+---
+
+## Explanation
+
+The first step is creating the context variable to enhance usability across various components within a
+Fresh application. By initializing these variables with a null value and integrating type references, developers can
+streamline the use of client side features while adapting to scenarios where server side rendering might not be
+feasible.
+
+:warning: Proper typing might not be easily available, so we might need to define our own types or not use types at all.
+
+```ts
+export const leafletContext = createContext<typeof Leaflet | null>(null);
+```
+
+Then, we should implement a Provider Component, this will handle loading and passing down values to be used in other
+components, other than that, we also need to handle the server side case as well.
+
+In this example, for the server side we are simply rendering a placeholder in place of our component tree. As for the
+context value, we are using html tags to inject the library on the window and a onLoad callback to set the value of our
+state, and this value will be handled/shared with our other components.
+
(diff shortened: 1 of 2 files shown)


docs/latest/examples/client-side-components-and-libraries.md
```diff
@@ -0,0 +1,176 @@
...
+    </LeafletProvider>
+  );
+}
+```
+
+---
+
+## Full code:
+
+```ts MapIsland.tsx
```
marvinhagemeister: Super tiny nit:

```suggestion
```tsx MapIsland.tsx
```

docs/latest/examples/client-side-components-and-libraries.md
```diff
@@ -0,0 +1,176 @@
...
+  });
+  return <div id="map" class={"relative w-[80vw] h-[50vh]"}></div>;
+}
+```
+
+Here is an example island encapsulating both the provider and component in order to demonstrate a simple usage.
+In real cases, it's usually better to add the Provider directly to our Page and then use Components that depend on that
+provider inside it.
+
+```ts
```
marvinhagemeister: Super tiny nit:

```suggestion
```tsx
```

docs/latest/examples/client-side-components-and-libraries.md
```diff
@@ -0,0 +1,176 @@
...
+    </>
+  );
+}
+```
+
+In order to utilize the context, call the useContext hook with the context variable this will give us access to the
+value set in the Provider. Handling cases where the context has not loaded values yet is a good practice as well,
+in this way we can have a smooth integration and manipulation of client-side data and logic on our server-side code.
+
+```ts
```
marvinhagemeister: Super tiny nit:

```suggestion
```tsx
```

docs/latest/examples/client-side-components-and-libraries.md
```diff
@@ -0,0 +1,176 @@
...
+components, other than that, we also need to handle the server side case as well.
+
+In this example, for the server side we are simply rendering a placeholder in place of our component tree. As for the
+context value, we are using html tags to inject the library on the window and a onLoad callback to set the value of our
+state, and this value will be handled/shared with our other components.
+
+:warning: Be careful with providers, the manner in which they load/inject both script and css may cause issues. Leaflet,
+for instance, will throw errors if we try to load it again.
+
+```ts
```
marvinhagemeister: Super tiny nit:

```suggestion
```tsx
```

docs/latest/examples/client-side-components-and-libraries.md
```diff
@@ -0,0 +1,176 @@
...
+---
+
+## Explanation
+
+The first step is creating the context variable to enhance usability across various components within a
+Fresh application. By initializing these variables with a null value and integrating type references, developers can
+streamline the use of client side features while adapting to scenarios where server side rendering might not be
+feasible.
+
+:warning: Proper typing might not be easily available, so we might need to define our own types or not use types at all.
```
marvinhagemeister: A little nicer styling with our markdown renderer.

```suggestion
> [warn]: Proper typing might not be easily available, so we might need to define our own types or not use types at all.
```

marvinhagemeister (approved):
This is great! Thank you so much for writing a page for client side components 👍  Left some super minor nits. But I can address them myself tomorrow when I'm back at work.

marvinhagemeister:
The CI errors can be resolved by running `deno fmt` locally.

PR #2106, opened 2023-11-30
docs: make use of props.isPartial to increase performance

deer:
I tried showing a friend how cool Fresh is, and I said that the doc site is a great example of partials. Then we looked, and much to my dismay, it seems like the entire page is being re-computed and sent to the browser. Of course the script in the browser is doing its job and only changing what's necessary, but it seems like we could be more aggressive.

docs old
docs new

In this pass I've gone after `nav` elements and similar things by adding `{!props.isPartial &&` around them. This cut the size of the responses to around 1/3 of the original size. I also moved the `<Partial>` tag down, to hopefully just cover the changing content. Perhaps there is room for more optimization in the future?


deer:
@marvinhagemeister, what do you think about this one? It makes the docs site slightly faster by reducing the amount of content sent out of the server.

The test failures are unrelated to the change (they were previously passing).

deer:
@iuioiua, in our conversation you said to add you to PRs that don't require lots of context. I think this one fits the bill.

www/routes/docs/[...slug].tsx
```diff
@@ -161,17 +161,20 @@ export default function DocsPage(props: PageProps<Data>) {
...
-                <Header title="docs" active="/docs" />
-                <main class="mt-4 min-w-0 mx-auto">
-                  <MobileSidebar page={page} />
+            {!props.isPartial && (
+              <nav class="w-[18rem] flex-shrink-0 hidden md:block px-4">
+                <DocsSidebar
+                  versionLinks={page.versionLinks}
+                  selectedVersion={page.version}
+                />
+              </nav>
```
Comment (iuioiua) [inline www/routes/docs/[...slug].tsx:170]: What happens to the rendering of this element if `props.partial` is true? Ditto elsewhere. Excuse me if I'm missing something.

Comment (deer) [reply www/routes/docs/[...slug].tsx:170]: It's not changing if we're staying on the docs page, so there's no point to recompute it. Without this change it's recomputed, but then ignored by the partials code in the client. The whole point of this change is to stop computing stuff the client will ignore.

Comment (iuioiua) [reply www/routes/docs/[...slug].tsx:170]: Oh! Yep. Makes total sense.

www/routes/docs/[...slug].tsx
```diff
@@ -161,17 +161,20 @@ export default function DocsPage(props: PageProps<Data>) {
...
+            )}
+            <div class="w-full min-w-0">
+              {!props.isPartial && <Header title="docs" active="/docs" />}
+
+              <main class="mt-4 min-w-0 mx-auto">
+                {!props.isPartial && <MobileSidebar page={page} />}
+                {!props.isPartial && (
```
Comment (iuioiua) [inline www/routes/docs/[...slug].tsx]: Can these two be combined?

iuioiua (approved):
LGTM! Just a single nit.

www/routes/docs/[...slug].tsx
Comment (deer) [reply www/routes/docs/[...slug].tsx]: yes, thanks

iuioiua (approved):
LGTM

iuioiua:
@marvinhagemeister, can this PR be updated and merged still?

bartlomieju:
Closing as stale — this targets the Fresh 1.x docs site which no longer exists after the 2.0 rewrite. The optimization idea (using isPartial to skip rendering static elements) is solid and worth applying if the pattern comes up in the new docs. Thanks for the contribution!

Issue #1632, opened 2023-08-13
Docs: Improve getting started section


marvinhagemeister:
The "getting started" section in our docs is way too wordy and could be simplified. I want to avoid a first impression that you'd need to read half a book to get started with Fresh.

These are best to be worked on in separate PRs to make it easier to review.

A couple of thoughts looking at the getting started pages:

- The "Introduction" page should be merged with the "Getting Started" overview page. The text in the introduction page uses too many buzzwords and needs to many paragraphs to explain what Fresh is about
- The content on the "Getting Started" page feels like filler content
- Description of files on "Create a project" page is too long. Should be a short bullet point list
  - I like the file + folder explanation though, maybe we can make it even better
- The "Running locally" page should be merged with "Create a project". Feels odd to have a whole page describing how to execute one command
  - Commands should use `-A` there instead of listing each flag individually. We should have an accompanying info box that explains the permissions Fresh need. That makes it clear that this is secondary information. We use `-A` in the project that's bootstrapped via the init script too
- The "Create a route" page is too much text for explaining what a route is
- The "Dynamic routes" page should be merged into "Create a route" and simplified as well
- The "Fetching data" page should be before "custom handlers", The content in "Fetching data" should come first as it's the immediate onboarding path to get something visible on the screen for users
- The "Custom handlers" page should be merged into "Form submissions". Instead of going the URL route, the "Form submission" example should do a proper submit
- The "Deploy to production" page should not just be text. It should show how to deploy Fresh.

In general the getting started guide should be aimed at getting new users up and running and have a deployed site as quickly as possible.

marvinhagemeister:
Done for the Fresh 2 canary docs. It's overall still WIP but the issues mentioned here have been addressed.

PR #2107, opened 2023-11-30
docs. Merge data fetching pages (#1974)

### LINKED ISSUE #1974: Docs: Merge data fetching pages
As per '[Fetching Data](https://fresh.deno.dev/docs/getting-started/fetching-data)' section in the docs, In order to fetch data, the route component must be asynchronous. Unfortunately when doing so the parameters are overwritten and you no longer have access to the props.

`/routes/[user].jsx`
```
const User = (props) => {
    console.log(props);
```

Compared to

`/routes/[user].jsx`
```
const User = async (props) => {
    console.log(props);
```

Is there a correct way for handling this I'm maybe missing?

CAYdenberg:
Attempts to clear up the confusion highlighted in #1974 by introducing async routes as an optional shortcut to async handlers combined with a route.

Note that I also re-wrote the first data fetching example to invoke `ctx.renderNotFound` rather than a `Reponse`. I think this is more realistic since it would be unusual to return a raw response in one case and an actual page in another, and it fits better with the async route example (which returns an actual page).

This also deletes an entire page from Getting Started (see #1632).

Closes #1974


marvinhagemeister (approved):
This is great! Thank you so much for PRing this 👍

PR #2110, opened 2023-12-01
chore: reuse puppeteer instance

deer:
Refreshing #1418, along with some minimal perf numbers by running `deno test -A tests/partials_test.ts --reporter=dot`:

before the change
* ok | 50 passed | 0 failed (1m11s)
* ok | 50 passed | 0 failed (1m16s)
* ok | 50 passed | 0 failed (1m20s)

after the change:
* ok | 51 passed | 0 failed (55s)
* ok | 51 passed | 0 failed (50s)
* ok | 51 passed | 0 failed (53s)

So this seems like a sizable improvement. It seems like the test runner doesn't wait for `globalThis` to finish doing its thing before starting, so I added some dummy tests at the top of the two modified files to make sure everything is warm before continuing. If there's a better way, please let me know.

### DIFF
--- a/tests/partials_test.ts
+++ b/tests/partials_test.ts
@@ -1,4 +1,4 @@
-import { assert, assertEquals, assertMatch, Page } from "./deps.ts";
+import { assert, assertEquals, assertMatch, delay, Page } from "./deps.ts";
 import {
   assertMetaContent,
   assertNoComments,
@@ -6,7 +6,9 @@ import {
   assertNotSelector,
   assertSelector,
   assertTextMany,
+  closeBrowser,
   getErrorOverlay,
+  launchOrGetBrowser,
   parseHtml,
   waitFor,
   waitForText,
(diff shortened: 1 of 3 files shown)


marvinhagemeister:
@lucacasonato is there a way to disable the tracing ops feature of `deno test` globally?

lucacasonato:
@marvinhagemeister No. But usually the solution to the sanitizer failing is not to disable the sanitizer, but to fix the test.

In this PR for example, you leak `puppeteer.Page` instances all over the place. The place where you removed `await browser.close()`, you should replace with `await page.close()`

tests/partials_test.ts
```diff
@@ -19,6 +21,21 @@ async function assertLogs(page: Page, expected: string[]) {
...
+
+globalThis.addEventListener("beforeunload", async () => {
+  await closeBrowser();
+});
+
+Deno.test("dummy test", {
+  sanitizeOps: false,
+  sanitizeResources: false,
+}, async () => {
+  await delay(1000);
```
lucacasonato, line 36: Rather than awaiting on a delay here (flaky), `await launchOrGetBrowser()`

PR #2122, opened 2023-12-02
chore: update types in tests, init, update

deer:
Of particular note is the change to `tests/fixture_plugin/routes/test.tsx`. Note how I do this:
```diff
-  async GET(_req, ctx: HandlerContext<unknown, { test: string }>) {
+  async GET(_req, ctx: FreshContext<{ test: string }, unknown>) {
```

I find it strange and confusing that the order of the type parameters has "changed". This makes it impossible to do a simple search and replace, because people now have to change order as well.

Most of the types have been declared as `<T, S>`, i.e. data before state:
* `PageProps<T = any, S = Record<string, unknown>>`
* `RouteContext<T = any, S = Record<string, unknown>>`
* `Handler<T = any, State = Record<string, unknown>>`
* `Handlers<T = any, State = Record<string, unknown>> `
* `AsyncRoute<T = any, S = Record<string, unknown>>`
* `PageComponent<T = any, S = Record<string, unknown>> `
* `AsyncLayout<T = any, S = Record<string, unknown>>`

And then `FreshContext` comes around like 
```
export interface FreshContext<
  State = Record<string, unknown>,
  Data = any,
  NotFoundData = Data,
>
```
and switches it all up.

Not a huge issue, but just something to be aware of that others might report.

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -273,7 +273,7 @@ await Deno.writeTextFile(
   ROUTES_GREET_TSX,
 );
 
-const ROUTES_API_JOKE_TS = `import { HandlerContext } from "$fresh/server.ts";
+const ROUTES_API_JOKE_TS = `import { FreshContext } from "$fresh/server.ts";
 
 // Jokes courtesy of https://punsandoneliners.com/randomness/programmer-jokes/
 const JOKES = [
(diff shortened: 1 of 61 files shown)


marvinhagemeister:
Main reason for changing the order was that I want to deprecate the `Data` generics long term.

marvinhagemeister (approved):
Nice, thank you for updating those types 👍

PR #2129, opened 2023-12-03
feat: Support Partial navigation on every element

### LINKED ISSUE #1952: feat: Allow partials on every element
There is a strong desire to expand support for partials to more scenarios. Fresh 1.5 shipped with partial support for navigation and https://github.com/denoland/fresh/pull/1930 landed recently in main that enables it for form elements. But there are lots of scenarios where you want to trigger a partial update based on an event. Like when changing the value of a select in a form or something. We need a way to allow that.

Something I was pondering about is whether we should introduce some sort of object syntax for events.

```tsx
<select onChange={{ partial: "/foo/bar" }}>...</select>

// or maybe abstracted via a function
<select onChange={partialTrigger("/foo/bar")}>...</select>
```

I haven't yet decided what it should look like and open to ideas.

osddeitf:
Fix #1952.

The idea is to assign some common functions / state variables to `window`. So both the inline-script embeded to html, and islands may be able to import and use it.


osddeitf:
@marvinhagemeister I need your review on this. I have tested and it worked fine, but still need more testing and clean up.
The exposed function is currently `navigate` from `/src/runtime/entrypoints/impl/global.ts`

marvinhagemeister:
This PR is a bit difficult to review because it moves a bunch of things around. Is the idea to expose the `navigate` function for users?

osddeitf:
Yes.
The file `entrypoints/main.ts` is an entrypoint input to esbuild.
User defined islands are different entrypoints.
It would be hard for them to share the same state variables (different entrypoints are minify differently by esbuild).
So my idea is to assign them to global `window` object.

It would help if you try viewing different commits one by one.
This PR is just a draft anyway, just to show it could work.

bartlomieju:
Thanks for the contribution! This PR is from the Fresh 1.x era. The runtime entrypoints, server context, and partials system have all been completely rewritten for Fresh 2.x. The files modified here (`src/runtime/entrypoints/main.ts`, `src/server/context.ts`, etc.) no longer exist.

Closing as stale.

Issue #2131, opened 2023-12-03
Route error generates `_frsh/error_overlay` 404 on preview/production


csvn:
It does not seem like error pages are rendered correctly either in dev or preview/production mode.

When running `dev.ts`, there is an error frame for the error. When that is closed, behind it's not rendering the real `_500.tsx` page.

When running `main.ts` after doing `dev.ts --build`, the `_500.tsx` page is rendered, but it still loads `_frsh/error_overlay` which then renders a 404 page over the 500 page in the error frame.

So I currently cannot get the correct 500 page rendered correctly in either dev mode or preview mode.

## Reproduction steps

1. Initialize a new project:
    ```
    deno run -A -r https://raw.githubusercontent.com/denoland/fresh/main/init.ts
    ```
2. Add an error page:
    ```tsx /routes/_500.tsx
    export default function Error500() {
      return <p>Internal server error ⚠️</p>;
    }
    ```
3. Update the home page to throw errors:
    ```tsx /routes/_500.tsx
    export default function Home() {
      throw new Error('err');
    }
    ```
4.
   a. Run `deno task start`
   b. Run `deno task build` and then `deno task preview`
5. Open http://localhost:8000

### Expected outcome

The 500 page is rendered for both, in local mode we also get a error frame

### Actual outcome

In local mode we get the wrong error page behind the error frame, in preview mode we get the error page first, but a 404 page is rendered above it in an error frame.

marvinhagemeister:
Resolved by #2133 and #2134

PR #2133, opened 2023-12-03
fix: remove error frame HTML for non-dev mode

csvn:
Addresses part of #2131, by removing the HTML of the code-frame when running in production/preview mode. This avoids a 404 error being rendered over the 500 page due to the `_frsh/error_overlay` returning a 404 error.


marvinhagemeister (approved):
Nice, thanks for making a PR 👍

Issue #1978, opened 2023-10-26
"_500.tsx" Not Rendering in Dev Mode


gustavomaltez:
I had some trouble setting up a custom error page for my app. I found that the _500.tsx page only seems to show up when I run the app in preview mode. Is there a way I can make it appear in dev mode? In dev mode, the error message is quite descriptive, but I'm wondering if it's possible to show the custom _500.tsx page instead of the default dev error page.

I have a handler intentionally throwing an error:

```ts
export const handler: Handlers<Props> = {
  async GET(_req, ctx) {
    throw new Error('Some error inside the handler');
  }
};
```

When I run in dev mode using `deno run -A --watch=static/,routes/ dev.ts`, I see this screen:

![image](https://github.com/denoland/fresh/assets/57905553/2b1c562d-49ca-4d74-a2ac-0c2611951ba7)

When running in preview mode with `deno run -A main.ts`, this is the screen that appears (my _500.tsx component):

![image](https://github.com/denoland/fresh/assets/57905553/4414fcf3-c215-4d21-9807-49e951ceced3)

These are my dev.ts and main.ts files:

`dev.ts`
```ts
#!/usr/bin/env -S deno run -A --watch=static/,routes/

import dev from 'fresh/dev.ts';
import config from './fresh.config.ts';

import '$std/dotenv/load.ts';

await dev(import.meta.url, './main.ts', config);
```

`main.ts`
```ts
/// <reference no-default-lib='true' />
/// <reference lib='dom' />
/// <reference lib='dom.iterable' />
/// <reference lib='dom.asynciterable' />
/// <reference lib='deno.ns' />

import '$std/dotenv/load.ts';

import { start } from 'fresh/server.ts';
import manifest from './fresh.gen.ts';
import config from './fresh.config.ts';

await start(manifest, config);
```

Fresh version: 1.5.2
Deno version: 1.37.2

Thanks in advance :)

deer:
[Here](https://github.com/denoland/fresh/blob/7ad4610e3a42aba42638cbc1041b96ee58a9b29e/src/server/fs_extract.ts#L215) we check if we're in dev mode and shove in the `DefaultErrorHandler` which has the code box you're seeing.

[This](https://github.com/denoland/fresh/blob/7ad4610e3a42aba42638cbc1041b96ee58a9b29e/src/server/mod.ts#L106) command gets invoked from your `main.ts`. So you can see dev being set to false and then we don't use the `DefaultErrorHandler`.

[...]

PR #2134, opened 2023-12-03
fix: use real 500 error page in dev mode

csvn:
In #1978 it was said the "real" error page behind the error code frame should now be shown. But I don't think this  works correctly, as in dev-mode, the `DefaultErrorHandler` is always used over the real `_500.tsx` error page.

This solves part of #2131.

### DIFF
--- a/src/server/fs_extract.ts
+++ b/src/server/fs_extract.ts
@@ -227,16 +227,6 @@ export async function extractRoutes(
         name,
         component,
         handler: (req, ctx) => {
-          if (config.dev) {
-            const prevComp = error.component;
-            error.component = DefaultErrorHandler;
-            try {
-              return ctx.render();
-            } finally {
-              error.component = prevComp;
-            }
-          }
-
           return handler
             ? handler(req, ctx)
             : router.defaultErrorHandler(req, ctx, ctx.error);
--- a/src/server/fs_extract.ts
+++ b/src/server/fs_extract.ts
@@ -226,11 +226,7 @@ export async function extractRoutes(
         url,
         name,
         component,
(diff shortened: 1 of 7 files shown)


tests/dev_command_test.ts
```diff
@@ -115,13 +115,13 @@ Deno.test("shows custom 500 page for rendering errors when not in dev", async (t
...
-      assertSelector(doc, ".frsh-error-page");
-    },
-  );
+  // await withFakeServe(
+  //   "./tests/fixture_dev_codeframe/dev.ts",
+  //   async (server) => {
+  //     const doc = await server.getHtml(`/`);
+  //     assertSelector(doc, ".frsh-error-page");
+  //   },
+  // );
```
csvn: I'm not sure what to do about this test (or the one below). There seems to be a bit of overlap between this one and `tests/error_test.ts`, and I do not feel like this test has the desirable result.

src/server/fs_extract.ts
```diff
@@ -227,16 +227,6 @@ export async function extractRoutes(
         name,
         component,
         handler: (req, ctx) => {
-          if (config.dev) {
```
marvinhagemeister, line 230: Oh no, I thought I had removed that part. I'm a bit mad at myself that I didn't catch that with a test. 🤦

marvinhagemeister (approved):
Sweet, thanks for catching this and opening a PR 👍

src/server/fs_extract.ts
csvn: I know the feeling. I usually need 4 commits instead of one when changing CI stuff because I forget stupid things too 😅

PR #2136, opened 2023-12-04
chore: Use `satisfies` instead of `as`

justjavac:
Use `as`

image

Use `satisfies`

image


deer:
Cool! Please do a full text search for `as Config` and get the rest as well.

marvinhagemeister (approved):
Awesome, thanks for updating this 👍

PR #2139, opened 2023-12-04
fix: PluginRoute supports AsyncRoute and AsyncLayout

### LINKED ISSUE #2112: Asynchronous app as plugin
I'm currently trying to re-use as much of the same work as possible between applications, and the plugin API makes sense as a starting point. 

However, I'm starting with the `_app.tsx` as a skeleton route, which doesn't seem to work for an async function. Is this by design, or am I missing some information on how to use async routes in plugins?

Using synchronous is fine usually, but I'm doing a lot of data and asset fetching, using `fetch` the simplest approach is to use `await`. Is there an alternative recommended approach whereby some async fetch routines can be used, in the synchronous routes / components without a need to re-render?

deer:
closes https://github.com/denoland/fresh/issues/2112

Turned out to just be a typing issue.

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -572,7 +572,11 @@ export interface PluginRoute {
   /** A path in the format of a filename path without filetype */
   path: string;
 
-  component?: ComponentType<PageProps> | ComponentType<AppProps>;
+  component?:
+    | ComponentType<PageProps>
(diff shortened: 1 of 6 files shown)


marvinhagemeister (approved):
Good catch!

PR #2141, opened 2023-12-05
fix: Unknown at rule @apply css(unknownAtRules)

justjavac:
see https://github.com/tailwindlabs/tailwindcss/discussions/5258#discussioncomment-1979394


marvinhagemeister (approved):
Sweet, nice catch! I didn't know that we could make vscode aware of this 👍

PR #2143, opened 2023-12-05
docs: add more description to deprecated data property

### LINKED ISSUE #2127: `data` @deprecated - "Pass values to state instead"
Congrats on the release of `v1.6.0`! 🎉 

After `v1.6.0` `data` in `PageProps` are being deprecated, and says we should use `state` instead. The docs still says we should use `data`, so I guess those need to be updated.

So to ask the question, how am I supposed to use `state` instead? Should I define a new prop in my State type in the middleware that is for page specific data?

(not sure to title the issue, so please rename if it doesn't make sense)

csvn:
Tries to improve the description for the `@deprecation` not on `FreshContext.data`.

- Adds mention that async route component can replace `.data` usage
- Adds links to `FreshContext.state´ and `AsyncRoute`

Adds some JSDoc for `FreshContext.state` as well.

Closes #2127

### DIFF
--- a/src/server/types.ts
+++ b/src/server/types.ts
@@ -189,9 +189,30 @@ export interface FreshContext<
   destination: router.DestinationKind;
   params: Record<string, string>;
   isPartial: boolean;
+  /**
+   * An object where data can be assigned to and that is accessible
+   * across middlewares, handlers and routes.
+   * @example
+   * ```tsx
+   * // routes/_middleware.ts
+   * export type State = { data: string };
+   *
+   * export async function handler(_req: Request, ctx: FreshContext<State>) {
+   *   ctx.state.data = "myData";
+   *   return ctx.next();
+   * }
+   * // routes/data.tsx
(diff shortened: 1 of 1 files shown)


src/server/types.ts
```diff
@@ -189,9 +189,30 @@ export interface FreshContext<
...
+   *
+   * export async function handler(_req: Request, ctx: FreshContext<State>) {
+   *   ctx.state.data = "myData";
+   *   return ctx.next();
+   * }
+   * // routes/data.tsx
+   * export const handler: Handler<void, State> = (_req, ctx) => {
+   *   return new Response(`middleware data: ${ctx.state.data}`);
+   * };
+   * ```
```
csvn, line 208: Not sure if we need this large example here. It was adapted from https://fresh.deno.dev/docs/concepts/middleware. Maybe we should link to that page instead via `@link`? It feels a bit odd though to link there, since the page is named "Middleware" (though it explains about `state` too).

PR #2088, opened 2023-11-28
chore: create manifest generation task

deer:
Hey @marvinhagemeister, your change in #2082 broke the functionality introduced in #1986. You can see the type checking failures here, now that I've generated the new manifests for all the fixtures.

In particular the manifest now looks like:
```ts
"./islands/MultipleCounters.tsx": $MultipleCounters,
```

but it really needs to be the original
```ts
"./islands/MultipleCounters.tsx": {
      CounterOne: $$MultipleCounters.CounterOne,
      CounterTwo: $$MultipleCounters.CounterTwo,
      default: $$MultipleCounters.default,
    },
```
or something similar. I thought relying on type checking would be enough to prevent regressions, but that's obviously not true. The gap to be filled is that we should always regenerate all manifest files when changing `manifest.ts`. (Your change snuck through because you didn't regenerate `fixture/fresh.gen.ts`.)

I'll open this as a draft for now, since it's late. Feel free to collaborate if you're feeling eager, otherwise I hope to get to this again tomorrow. (Right before basePath!)

PR #2147, opened 2023-12-06
feat(tool): add a tool to release docs

deer:
After playing around with ts-morph in #2088 (the ts-morph part eventually got removed), I wanted to continue trying it out. Here's a tool that should make the release process slightly easier. It does three things:
1. copy canary docs to latest
2. deletes canary docs
3. update the toc.ts

There's possibly an easier way to do, so please let me know. But a bit of learning is never a bad thing, and I think the end result is pretty slick. Here's how I tested this:

```
git checkout 0c3dc04 (the primary commit here)
git reset HEAD^
git checkout bd372b9 (right before 1.6.0 was released)
git add . (just to get these changes out of the working directory)
deno task docs:release
git status
```

So then I have a nice diff showing what would have happened if you used this to release the docs last week. Except this shows there's actually a small bug (I think!) in `toc.ts` right now.
https://github.com/denoland/fresh/blob/c72f534369d6ddcc6e35f69c57b2407a879eeb53/docs/toc.ts#L75 should be at `latest` instead of `canary`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -9,7 +9,8 @@
     "ok": "deno fmt --check && deno lint && deno task check:types && deno task test",
     "install-puppeteer": "PUPPETEER_PRODUCT=chrome deno run -A --unstable https://deno.land/x/puppeteer@16.2.0/install.ts && PUPPETEER_PRODUCT=firefox deno run -A --unstable https://deno.land/x/puppeteer@16.2.0/install.ts",
     "test:www": "deno test -A tests/www/",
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks for the contribution! This PR is from the Fresh 1.x era. The docs structure (canary/latest workflow) and `docs/toc.ts` have been rewritten for Fresh 2.x, and the release process has changed.

Closing as stale.

PR #2162, opened 2023-12-08
fix link to data-fetching

Xernisfy:
Just a fixed link

### DIFF
--- a/docs/latest/getting-started/custom-handlers.md
+++ b/docs/latest/getting-started/custom-handlers.md
@@ -66,5 +66,4 @@ export const handler: Handlers = {
 ```
 
 Handlers can do much more, including fetching data from a database or external
-API and passing it to their route. See [Data fetching](/docs/concepts/data-fetching) for more
-details.
+API and passing it to their route.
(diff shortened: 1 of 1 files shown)


docs/latest/getting-started/custom-handlers.md
```diff
@@ -66,5 +66,5 @@ export const handler: Handlers = {
 ```
 
 Handlers can do much more, including fetching data from a database or external
-API and passing it to their route. See [Data fetching](/docs/concepts/) for more
+API and passing it to their route. See [Data fetching](/docs/concepts/data-fetching) for more
 details.
```
marvinhagemeister: Looks like we removed that page completely in https://github.com/denoland/fresh/pull/2107 . Let's get rid of that "See ..." sentence.

```suggestion
API and passing it to their route.
```

Xernisfy: Thanks for pointing that out!

marvinhagemeister (approved):
Awesome, thanks for making a PR to fix this 👍

PR #2180, opened 2023-12-17
feat: Use `@/` by default to allow absolute path specification.

EdamAme-x:
Use `@/` by default to allow absolute path specification.
This is expected to improve the outlook for dependencies between components.

You should be able to count on this EXAMPLE
```tsx
// routes/orgs/[name].tsx
import OrgProfile from "../../islands/orgs/profile.tsx"

export default function OrgsProfile() {
    return ....
}
```

If absolute paths are possible

```tsx
// routes/orgs/[name].tsx
import OrgProfile from "@/islands/orgs/profile.tsx"

export default function OrgsProfile() {
    return ....
}
```

beautiful : )

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -153,7 +153,7 @@ CMD ["run", "-A", "main.ts"]
 }
 
 const ROUTES_INDEX_TSX = `import { useSignal } from "@preact/signals";
-import Counter from "../islands/Counter.tsx";
+import Counter from "@/islands/Counter.tsx";
 
 export default function Home() {
   const count = useSignal(3);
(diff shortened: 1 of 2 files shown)


EdamAme-x:
I saw the results of workflow.
Deploy seems to be successful.
However, an error message `didn't startup server` is thrown on test.
I don't understand why the error is being thrown up when there are no destructive changes made by this change.

marvinhagemeister:
Closing as this is specific to particular projects. Some use `@`, some use `~` and other's don't have a root mapping but many mappings for specific directories. I'm not yet comfortable to prescribe one specific way of doing this for users of all Fresh projects.

PR #2182, opened 2023-12-18
docs: fix typo in upload example

imaitland:
Fix undefined error on file upload example.


marvinhagemeister (approved):
Good catch, thanks for fixing this!

PR #2192, opened 2023-12-22
docs: fix typo in active links introduction

injms:
Fixes an `<a>` in the active link markdown from being interpreted as an actual link, rather than part of the text.

Before:

Before with rogue anchor element in hover state:

After:


marvinhagemeister (approved):
Awesome, thanks for making a PR to fix this 👍

PR #2198, opened 2023-12-27
docs: fix navigating to components from docs with small screen

### LINKED ISSUE #2190: Fresh Docs website fails to navigate from Docs to Showcase or Components when the Table of Contents appear collapsed
The error is thrown in the browser:
```
main.js:1 Fetch finished loading: GET "https://fresh.deno.dev/showcase?fresh-partial=true".
S @ main.js:1
(anonymous) @ main.js:1
main.js:1 Uncaught (in promise) Error: Found no partials in HTML response. Please make sure to render at least one partial. Requested url: https://fresh.deno.dev/showcase?fresh-partial=true
    at ce (main.js:1:6200)
    at async S (main.js:1:4601)
    at async HTMLDocument.<anonymous> (main.js:1:8950)
ce @ main.js:1
```

deer:
closes https://github.com/denoland/fresh/issues/2190

Marvin, I noticed that adding the fix [here](https://github.com/denoland/fresh/blob/a8a43ff0d4ed7ae9958a09f37ce8dadf219f76e1/www/components/Header.tsx#L41) did not resolve the issue. I see the docs say `Partials are enabled by adding a f-client-nav attribute to an HTML element` -- is it expected that adding `f-client-nav={false}` to a component does not automatically apply it to html element children?


deer:
@marvinhagemeister, this is a one-liner that fixes an issue on the docs website for small screens.

marvinhagemeister (approved):
Nice catch!

PR #2204, opened 2023-12-28
refactor: use `--env` CLI arg

iuioiua:
This change replaces the use of the Standard Library's `dotenv.load()` with the native `--env` CLI flag.

The benefits include:
* One less dependency
* More obvious consumption of the `.env` file in the CLI output


marvinhagemeister (approved):
Sweet, thanks 👍

PR #2210, opened 2024-01-04
docs: add jsx and preact introduction

ooker777:


ooker777:
can you explain why this is not suitable? Or is it because you have a totally new docs?

marvinhagemeister:
1. The content references implementation details like `esbuild`. We expose no part of the esbuild API to the user. If we would decide to swap out bundlers internally, there should be no changes required to the documentation.
2. JSX doesn't stand for "JavaScript XML". It's a word that stands on its own. There is no mention of "JavaScript XML" [in the specification](https://facebook.github.io/jsx/)
3. It's describing implementation details of the JSX transform like how it _can_ be transformed to objects. This is not a hard rule and highly depends on the JSX transform chosen. In Fresh 2 for example we're pushing folks to use [Deno's `precompile` transform](https://docs.deno.com/runtime/reference/jsx/#jsx-precompile-transform) as it's much faster on the server. It doesn't create objects, but precompiles to strings.
4. The future Deno Deploy version (currently under early access for users to try out) supports a build step out of the box. We'll automatically pick up a Fresh preset too, so any mentions of doing ahead of time stuff like in the tailwindcss plugin docs are out of date.

ooker777:
I agree with point 2 and 4. 

In point 1, while I agree with the rationale, I still think it's useful for the users to understand what's going on under the hood. Do you think it's useful to have that documented somewhere? If yes, where it should be?

In point 3, I think it's useful for beginners to learn about basic knowledge like JSX, framework and metaframework, and their applications in Fresh and Preact. And while it's great to improve server performance, my guess is that object transformation of JSX is still needed for islands and partials. So in the context of learning what framework is, starting with object transformation is more beneficial? And it's easy to add a sentence on how JSX is transformed in server to present a full picture. What do you think?

PR #2224, opened 2024-01-09
feat: add more useful data to plugin render ctx

guitavano:


src/server/render.ts
```diff
@@ -261,10 +261,19 @@ export async function render<Data>(
 
   const renderResults: [Plugin, PluginRenderResult][] = [];
 
+  const pluginRenderOptions = {
+    request: opts.request,
```
marvinhagemeister: I'm wondering if it makes more sense to go with `req`. That's the term used in `express` and `hono` and might be familiar to users coming from other frameworks.

src/server/render.ts
```diff
@@ -261,10 +261,19 @@ export async function render<Data>(
...
+    lang: opts.lang,
+    context: opts.context,
```
marvinhagemeister: The outer object is already called `context` and having another property named `context` inside of it feels strange.

```ts
const myPlugin = {
  name: "my-plugin",
  render(ctx) {
    // This feels weird typing context twice
    const config = ctx.context.config
  }
}
```

What do you say about merging `opts.context` directly into this one?

guitavano: I am not sure about bring all with ```...opts.context```, so I am following PageProps, that already exclude render, which one would be duplicated with ```...opts.context```.

What do you think?

guitavano: @marvinhagemeister up

marvinhagemeister (commented):
Apologies for the silent on my part. I have this on my list to look at for tomorrow. I'm currently involved in a few non-fresh projects that eat my time.

deer (changes requested):
Hi @guitavano, cool feature -- I'm always in favor of more plugin functionality, but how would I use it? I think you should update the documentation and add a test plugin that uses this stuff.

Branch `plugins.md` to `/docs/canary/concepts/` and then update `toc.ts` appropriately. You should be able to see the different versions when you do `deno task www`. This way people won't find out about the feature before it's released.

bartlomieju:
Thanks for the contribution! This PR is from the Fresh 1.x era. The plugin render system (`src/server/render.ts`, `src/server/types.ts`) has been completely rewritten for Fresh 2.x.

Closing as stale.

PR #2227, opened 2024-01-09
feat: improve tailwind deployment flow

### LINKED ISSUE #2223: TailwindCSS doesn't load on production but in development it works
This problems happens with the following project: [jheysaav/jheysonsaavedra.com](https://github.com/jheysaav/jheysonsaavedra.com/tree/63184af98ad4ec2d106aae2b28fb88482db465be). The projects is alive [here](https://jheysaav.deno.dev).

TailwindCSS is configured as a the docs explain it and when I run `deno task start` it works fine like this:
image

But in the production mode when I run `deno task preview` or when I deploy it to Deno Deploy it seen like this:
image

The Tailwind css file doesn't work.

This is my `tailwind.config.ts`: 
```typescript
import { type Config } from "tailwindcss";
import colors from "tailwindcss/colors.js";

export default {
  content: [
    "{routes,islands,components}/**/*.{ts,tsx}",
  ],
  darkMode: "class",
  theme: {
    fontFamily: {
      sans: ["Inter Variable", "Inter", "sans-serif"],
    },
    colors: {
      gray: colors.zinc,
      primary: colors.blue,
      secondary: colors.emerald,
    },
  },
} satisfies Config;
```

The `routes/_app.tsx` file looks like this:
```tsx
import { type PageProps } from "$fresh/server.ts";
import Header from "../components/Header.tsx";

export default function App({ Component, url }: PageProps) {
  return (
    <html lang="en">
      <head>
        <meta charset="utf-8" />
        <title>Jheyson Saavedra</title>
        <link rel="stylesheet" href="/styles/tailwind.css" />
        <link rel="stylesheet" href="/styles/inter.css" />
      </head>
      <body className="w-scren h-screen flex flex-col bg-gray-50 dark:bg-gray-900 text-gray-900 dark:text-gray-50">
        <Header url={url} />
        <main>
          <Component />
        </main>
      </body>
    </html>
  );
}

```

Finally the `static/styles/tailwind.css` file looks as this:
```css
@tailwind base;
@tailwind components;
@tailwind utilities;
```

marvinhagemeister:
By moving folks from `twind` over to the `tailwind` plugin we have traded support questions about "how to get twind autocompletion working?" with "my styles are not showing up in automatic mode". The tailwind plugin has a hard requirement on having a build step. We have documented this at various places and have a guide on how to set everything up, but it's much better if we float those steps from Fresh itself when we detect this.

With this PR we will prepare a `.github/workflows/deploy.yml` file already where the user only needs to swap out the project name.

When Fresh is started in production mode, the tailwind plugin is active and no snapshot files were loaded, then we print a helpful hint to the console:

```sh
No pre-compiled tailwind styles found.

Did you forget to run "deno task build" prior to starting the production server?
```

When we are running inside Deno Deploy this PR goes one step further. Since it is the first time many users might use Deno Deploy we'll also show a simple page that shows how to finish the tailwind setup. I've intentionally tried to not present this as a warning but rather as a "finishing" step to avoid scaring users off. It's not the prettiest page, but it's good enough for now.

Screenshot 2024-01-09 at 23 41 27

Fixes https://github.com/denoland/fresh/issues/2223

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -308,10 +308,50 @@ export default {
   ],
 } satisfies Config;
 `;
+const AOT_GH_ACTION = `name: Deploy
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: main
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+
+    permissions:
+      id-token: write # Needed for auth with Deno Deploy
+      contents: read # Needed to clone the repository
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Install Deno
+        uses: denoland/setup-deno@v1
+        with:
+          deno-version: v1.x
+
+      - name: Build step
+        run: "deno task build" # 📝 Update the build command(s) if necessary
+
+      - name: Upload to Deno Deploy
+        uses: denoland/deployctl@v1
+        with:
+          project: "example-project" # 📝 Update the deploy project name if necessary
+          entrypoint: "./main.ts" # 📝 Update the entrypoint if necessary
+`;
 if (useTailwind) {
   await Deno.writeTextFile(
     join(resolvedDirectory, "tailwind.config.ts"),
     TAILWIND_CONFIG_TS,
+  );
+  const ghWorkflowDir = join(resolvedDirectory, ".github", "workflows");
+  await Deno.mkdir(ghWorkflowDir, { recursive: true });
+  await Deno.writeTextFile(
+    join(ghWorkflowDir, "deploy.yml"),
+    AOT_GH_ACTION,
   );
 }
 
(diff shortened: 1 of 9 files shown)


liigo:
There're still complex steps. And maybe it violates https://deno.com/blog/you-dont-need-a-build-step which i think is a regress.

marvinhagemeister:
@liigo sounds like you'd enjoy the existing `twind` plugin more than the `tailwind` plugin. As communicated in past Fresh blog posts already a build step overall is necessary to have a fast site. The main goal now is to make that as smooth as possible from a UX perspective. Long term we're hoping to enhance the Deno Deploy platform to have run the build step there, so that user's don't need to configure anything. But there is no timeline on that yet.

For folks who dislike the build step I'd recommend staying with `twind`.

PR #2244, opened 2024-01-12
fix: Prevent update doc title from partial if not present

### LINKED ISSUE #2243: [Partials] Losing title after partial run
### Issue
Following the doc to [optimize partial requests](https://fresh.deno.dev/docs/concepts/partials#optimizing-partial-requests), to render only the necessary HTML of the partial I observed a problem on the title if the request that answer the partial document doesn't return the tag `<title />`.

Example: 
1. open this page: https://deco-sites-eico.deno.dev/ - the title is present
2. scroll the page 
2.1. find for a request with this pattern `https://deco-sites-eico.deno.dev/deco/render?props=...&fresh-partial=true`
3. the title is empty

### Solution Suggestion
At the partial algorithm to "hydrate" the current page with the new one, if the title isn't present, don't apply the title.
 https://github.com/denoland/fresh/blob/6f6f55124879bbf8b85ea11d9d999cf872b53df1/src/runtime/entrypoints/main.ts
 
 I can implement the solution if is possible

igorbrasileiro:
fixes #2243

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -710,7 +710,9 @@ export async function applyPartials(res: Response): Promise<void> {
   }
 
   // Update <head>
-  document.title = doc.title;
+  if (doc.title) {
+    document.title = doc.title;
+  }
 
   // Needs to be converted to an array otherwise somehow <link>-tags
   // are missing.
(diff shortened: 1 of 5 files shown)


marvinhagemeister (commented):
Oh that's a good catch! Let's add a test case for this and then it's ready to land 👍

igorbrasileiro:
> Oh that's a good catch! Let's add a test case for this and then it's ready to land 👍

Do you have a suggestion of test file to add this or i can create a new one for this case?

marvinhagemeister (approved):
Perfect, those tests look great 👍

PR #2245, opened 2024-01-13
docs: getting started - running locally

greg-zometa:
Removes description of changing the port number with the deprecated `port` property and updates  with `server.port`.

Screenshot 2024-01-12 190452


marvinhagemeister (approved):
Good catch, thanks for updating the docs!

Issue #2254, opened 2024-01-15
More secure implementation for Islands Architecture


arkark:
Hello, I'd like to share the following security considerations (which I don't think are vulnerabilities of Fresh) to make Fresh more secure.

- Props Manipulation for Islands Architecture
- Prototype Pollution in deserialization

This issue is long and some information may be missing. If you have any questions, feel free to ask me.

### Props Manipulation for Islands Architecture

Fresh uses Islands Architecture, and the following article introduces how islands work in Fresh:

- [A Gentle Introduction to Islands - Deno](https://deno.com/blog/intro-to-islands)

An example of generated client-side scripts is as follows (formatted):
```html
<script type="module" nonce="...hex...">
  import { deserialize } from "/_frsh/js/...hex.../deserializer.js";
  import { signal } from "/_frsh/js/...hex.../signals.js";

  const ST = document.getElementById("__FRSH_STATE").textContent;
  const STATE = deserialize(ST, signal);

  import { revive } from "/_frsh/js/...hex.../main.js";
  /* ... snip (import statements for islands) ... */

  const propsArr = typeof STATE !== "undefined" ? STATE[0] : [];
  revive({/* ... snip (a object for islands) ... */}, propsArr);
</script>
```

Fresh renders island components with a JSON value of `document.getElementById("__FRSH_STATE").textContent`. So, if users can inject an HTML element with `id="__FRSH_STATE"`, it is possible to manipulate the rendering process and potentially change the behavior of the application.

For some applications, there are cases where such an assumption is valid. For example, consider an application that allows users to write markdown text freely.

The following documentation introduces how to render markdown and this example uses [deno-gfm](https://github.com/denoland/deno-gfm):

- https://fresh.deno.dev/docs/examples/rendering-markdown

deno-gfm prevents XSS attacks with [sanitize-html](https://github.com/apostrophecms/sanitize-html), but allows adding `id` attributes to HTML elements:

- ref. https://github.com/denoland/deno-gfm/blob/0.2.5/mod.ts#L214-L219

If the following island is defined,
```typescript
import type { Signal } from "@preact/signals";
import { render } from "$gfm";

interface PreviewProps {
  text: Signal<string>;
}

export default function Preview(props: PreviewProps) {
  return (
    <div
      class="markdown-body"
      dangerouslySetInnerHTML={{ __html: render(props.text.value) }}
    />
  );
}
```

the following `text` parameter is valid as a markdown text and it overwrites the island state:
```html
[...]

PR #2255, opened 2024-01-15
fix: prototype pollution in deserializer

lucacasonato:
This is not exploitable in practice unless a user maliciously crafts serialized values in __FRSH_STATE, because `serializer()` never outputs serialized representation that would be vulnerable to prototype pollution. But hey, defense in depth.


marvinhagemeister (approved):

marvinhagemeister:
Just commenting to link issues https://github.com/denoland/fresh/issues/2254

PR #2256, opened 2024-01-15
fix: `__FRSH_STATE` potentially being overwritten by user code

marvinhagemeister:
As outlined in https://github.com/denoland/fresh/issues/2254 it was possible to overwrite the script tag that is used by Fresh to pass island props from the server to the client. This could be done by injecting raw HTML via `dangerouslySetInnerHTML` and simply using the same id that Fresh was using.

With this PR we generate a unique id per render that cannot be guessed anymore. It's composed of `__FRSH_STATE_<uuid>`

### DIFF
--- a/src/runtime/entrypoints/main.ts
+++ b/src/runtime/entrypoints/main.ts
@@ -640,6 +640,7 @@ class NoPartialsError extends Error {}
  */
 export async function applyPartials(res: Response): Promise<void> {
   const contentType = res.headers.get("Content-Type");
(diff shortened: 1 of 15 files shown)


lucacasonato (approved):
LGTM

guitavano:
To share the apprenticeship, at deco.cx we use __FRSH_STATE to see how much data is going to client side

image


After that, a small adjustment was necessary:

``document.getElementById("__FRSH_STATE")?.textContent``
to
``document.querySelector("script[id^='__FRSH_STATE']")?.textContent``

PR #2266, opened 2024-01-19
fix: tailwind supports routes and islands provided by plugins

### LINKED ISSUE #2159: Support tailwind for routes and islands from plugins
Currently the tailwind plugin doesn't styles routes and islands provided by plugins.

deer:
closes https://github.com/denoland/fresh/issues/2159

@mcgear, please take a look at this. I got [my comment](https://github.com/denoland/fresh/issues/2159#issuecomment-1900115007) cleaned up a bit and it's working for my three test cases.
1. The first is included in this PR, where I've hacked a plugin into an existing tailwind test. I added a `components` folder with a component, modified the tailwind config to _not_ scan this folder, and then used this component within a route from the plugin. Without my changes, the text is unstyled; with my changes, it works as expected.
2. The second is outside of this code base (and hasn't left my computer yet), but it's a "real" usage. Previously I was working around the issue by providing a safelist, which I've now removed. Everything is still styled correctly! 🎉  This project follows fresh's pattern of having source and then `tests/fixtures`, so everything was local.
3. Finally I wanted to test a remote plugin. What to do when the remote plugin hasn't been updated to use this branch?
```ts
export default defineConfig({
  plugins: [{...blogPlugin(blogOptions), location: "https://deno.land/x/fresh_blog@0.0.6/src/plugin/blog.ts", projectLocation: "https://deno.land/x/fresh_blog@0.0.6"}, ga4Plugin(), tailwind()],
});
```
After hacking the two new properties into the plugin returned by `blogPlugin`, updating fresh to be my branch (i.e. `"$fresh/": "https://raw.githubusercontent.com/deer/fresh/2159_plugin_routes_tailwind/",`), and removing the safelist, this is also working. 

If you [use](https://fresh.deno.dev/docs/examples/using-fresh-canary-version#forked-fresh) this branch in your project, I'm curious to see if it works.

I haven't tested this with builds yet, and I know this is already going to fail in CI for some web assembly reason (at least on my computer). But I wanted to put this up since it's solving a pretty big problem with plugins.


plugins/tailwind/compiler.ts
```diff
@@ -62,6 +66,23 @@ export async function initTailwind(
     return pattern;
   });
 
+  const imports = (readDenoConfig()?.imports ?? []) as Record<string, string>;
```
iuioiua: `readDenoConfig()` is likely not needed. See https://examples.deno.land/importing-json

plugins/tailwind/compiler.ts
```diff
@@ -76,3 +97,44 @@ export async function initTailwind(
...
+      fileTypes.some((type) =>
+        module.specifier.endsWith(type) &&
+        module.specifier.startsWith(path.dirname(projectLocation))
+      )
+    ) {
+      specifiers.add(module.specifier);
+    }
+  }
+  return Array.from(specifiers);
+}
```
iuioiua:
```suggestion
function extractSpecifiers(graph: ModuleGraphJson, projectLocation: string) {
  return graph.modules
    .filter((module) =>
      module.specifier.endsWith(".tsx") &&
      module.specifier.startsWith(path.dirname(projectLocation))
    )
    .map((module) => module.specifier);
}
```
Nit: `fileTypes` as an array doesn't appear needed. I might be missing context, though.

plugins/tailwind/compiler.ts
```diff
@@ -76,3 +97,44 @@ export async function initTailwind(
...
+  return function customResolve(specifier: string, referrer: string): string {
+    for (const [key, value] of Object.entries(imports)) {
+      if (specifier.startsWith(key)) {
+        const mappedPath = specifier.replace(key, value);
+        return new URL(mappedPath, referrer).toString();
+      }
+    }
+    return new URL(specifier, referrer).toString();
+  };
+}
```
iuioiua:
```suggestion
function createCustomResolver(imports: Record<string, string>) {
  return (specifier: string, referrer: string): string => {
    const importMatch = findSingle(
      imports,
      ({ key }) => specifier.startsWith(key),
    );
    if (importMatch !== undefined) {
      specifier = specifier.replace(importMatch.key, importMatch.value);
    }
    return new URL(specifier, referrer).toString();
}
```
Perhaps, you could do something like this. See https://deno.land/std@0.212.0/collections/mod.ts?s=findSingle

iuioiua (commented):
PTAL at my last suggestion regarding your DM query. `deno_std` can be used, but it doesn't make a huge difference. Nor does this code make me think that there's a need for a new API in `deno_std`.

I'm not really aware of the context behind this PR, so take my suggestions with a pinch of salt, but the code looks good to me 👍🏾

deer:
> PTAL at my last suggestion regarding your DM query. `deno_std` can be used, but it doesn't make a huge difference. Nor does this code make me think that there's a need for a new API in `deno_std`.
> 
> I'm not really aware of the context behind this PR, so take my suggestions with a pinch of salt, but the code looks good to me 👍🏾

Thanks for the suggestions -- these look great. I'll incorporate these now and get this cleaned up some more.

plugins/tailwind/compiler.ts
deer: I've been playing around with this. It still seems like there's room for something standard.
1. Some easy way of saying "find the next applicable `deno.json` or `deno.jsonc` file".
2. It should be parsed into an object and typed properly. I see there is https://docs.deno.com/runtime/manual/getting_started/configuration_file#json-schema, but I guess I'm just greedy and want more.
3. And finally a way of saving it again (not relevant in this case here).

deer: I ended up with the following:
```ts
  const imports =
    (await import(config.denoJsonPath, { with: { type: "json" } }))
      .default.imports as Record<string, string>;
```
This is something of an eyesore, but it's better than what I had before. Thanks.

plugins/tailwind/compiler.ts
deer: Great simplification. Thanks.

deer:
todo:
- [x] fix tests on windows
- [x] write proper tests for this change

deer:
@iuioiua, as we discussed yesterday, I'm working the pluginification of saaskit using this PR. It's going well, except for your usage of `"@/": "./"` in the import map. This is breaking my module resolution. I logged https://github.com/denoland/deno_graph/issues/364 since it doesn't seem like something I should be doing.

Given that I want to get this done sooner rather than later, I'm trying to do the module resolution myself. Unfortunately I'm encountering https://github.com/denoland/vscode_deno/issues/708 when trying to debug the project. So using `console.log` to fix the problem is slow going.

iuioiua:
Perhaps, try get this working for local plugins, where all the code for that plugin lies within that plugin's directory. Then, you might be able to just `walk()` the `.tsx` files and call `Deno.readTextFile()`. That seems like an alright v0 of the implementation and I'm sure there'll be a learning opportunity on the way to get you to the next version.

plugins/tailwind/compiler.ts
```diff
@@ -76,3 +114,27 @@ export async function initTailwind(
...
+  return (specifier: string, referrer: string) => {
+    if (
+      specifier.startsWith("./") || specifier.startsWith("../") ||
+      specifier.startsWith("../../")
+    ) {
+      return path.join(path.dirname(referrer), specifier);
+    }
+    return resolveModuleSpecifier(specifier, imports, baseURL);
+  };
+}
```
deer: The hint I got from Luca in https://github.com/denoland/deno_graph/issues/364#issuecomment-1909835945 made this significantly better than it was, but I was still having issues with relative imports. So the resolution here is still hacky and obviously won't work if there is more relative nesting in the specifier.

I'll keep looking at this until it's better behaving.

deer:
@marvinhagemeister @iuioiua, I think this is ready to look at. There's an uptake of this PR in https://github.com/denoland/saaskit/pull/660.

I've added a few different test cases to make sure that the different ways of adding routes via plugins are covered. (Most of the emphasis is on the module resolution code.)

plugins/tailwind/compiler.ts
```diff
@@ -62,6 +70,31 @@ export async function initTailwind(
     return pattern;
   });
 
+  const imports = (await import(path.toFileUrl(config.denoJsonPath).href, {
+    with: { type: "json" },
+  })).default;
+  for (const plugin of config.plugins ?? []) {
+    if (!plugin.location) continue;
```
iuioiua:
```suggestion
    if (plugin.location === undefined) continue;
```

plugins/tailwind/compiler.ts
```diff
@@ -62,6 +70,31 @@ export async function initTailwind(
...
+    const projectLocation = plugin.projectLocation ??
+      path.dirname(plugin.location);
+    const resolvedImports = await parseFromJson(
+      path.toFileUrl(config.denoJsonPath),
+      imports,
+    );
+
+    const moduleGraph = await createGraph(plugin.location, {
+      resolve: createCustomResolver(resolvedImports),
+    });
```
iuioiua:
```suggestion
    const { resolve } = await parseFromJson(
      path.toFileUrl(config.denoJsonPath),
      imports,
    );

    const moduleGraph = await createGraph(plugin.location, { resolve });
```
`createCustomResolver()` may not be needed.

.vscode/import_map.json
```diff
@@ -3,6 +3,7 @@
     "THIS FILE EXISTS ONLY FOR VSCODE! IT IS NOT USED AT RUNTIME": {}
   },
   "imports": {
+    "@vscode_787_hack/": "../tests/fixture_tailwind_remote_classes/",
```
iuioiua, line 6: Is this here deliberately? This import is already defined in `tests/fixture_tailwind_remote_classes/deno.json`.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
```diff
@@ -0,0 +1,19 @@
...
+export const veryNestedPlugin = {
+  name: "very nested plugin",
+  location: import.meta.url,
+  projectLocation: currentUrl.href,
+  routes: [{
+    path: "routeFromVeryNestedPlugin",
+    handler: (_req, ctx) => {
+      return ctx.render();
+    },
+    component: VeryNestedPluginComponent,
```
iuioiua: Am I right in saying the `handler` property isn't needed if `component` is defined? Ditto for other plugins for this fixture.

tests/fixture_tailwind_remote_classes/fresh.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig, Plugin } from "$fresh/server.ts";
+import tailwind from "$fresh/plugins/tailwind.ts";
+import { basicPlugin } from "./basicPlugin.ts";
+import { nestedPlugin } from "./nestedPlugins/nestedPlugin.ts";
+import { hackyRemotePlugin } from "https://raw.githubusercontent.com/denoland/fresh/727d0e729e154323b99b319c04f9805f382949f0/tests/fixture_tailwind_remote_classes/hackyRemotePlugin.tsx";
```
iuioiua, line 5: Nice.

tests/fixture_tailwind_remote_classes/deno.json
```diff
@@ -0,0 +1,15 @@
+{
+  "lock": false,
+  "imports": {
+    "$fresh/": "../../",
+    "preact": "https://esm.sh/preact@10.15.1",
+    "preact/": "https://esm.sh/preact@10.15.1/",
+    "tailwindcss": "npm:tailwindcss@3.3.5",
+    "@/": "./",
+    "@vscode_787_hack/": "./"
```
iuioiua, line 12: Can you elaborate on why this is needed if we already have the above `@/` import? What that the problematic root import you previously mentioned?

iuioiua (changes requested):

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: You're right -- thanks.

.vscode/import_map.json
deer: It is, in order to get vsocde to behave. The issue is https://github.com/denoland/vscode_deno/issues/787. Note how the path here is slightly different.

tests/fixture_tailwind_remote_classes/deno.json
deer: I should probably remove "@/" since it's not used. But if we define "@/" at the top most level, then other things might break. Scoping it with this hacky identifier keeps it from affecting other people.

plugins/tailwind/compiler.ts
deer: Why? This is the key part of the code to help us identify files that are part of the plugin.

deer: To help clarify: I need fully resolved URLs so that I can fetch them later.

deer: Oh, I see your comment on deno_graph now. I'll play around with it more later, but the suggestion here doesn't immediately work.

deer:
```ts
    const resolvedImports = await parseFromJson(
      path.toFileUrl(config.denoJsonPath),
      imports,
    );

    const moduleGraph = await createGraph(plugin.location, {
      resolve: resolvedImports.resolve.bind(resolvedImports),
    });
```
ended up working. thanks

deer:
I'm not sure whether this should be merged until https://github.com/denoland/vscode_deno/issues/708 is resolved. Otherwise I don't think people will be able to debug their apps.

plugins/tailwind/compiler.ts
```diff
@@ -62,6 +68,39 @@ export async function initTailwind(
...
+      path.toFileUrl(config.denoJsonPath),
+      imports,
+    );
+
+    const moduleGraph = await createGraph(plugin.location, {
+      resolve: resolvedImports.resolve.bind(resolvedImports),
+    });
+
+    for (const file of extractSpecifiers(moduleGraph, projectLocation)) {
+      const response = await fetch(file);
```
iuioiua, line 98: Is `fetch()` being used here to cater to both local (i.e. `file://` scheme) and remote (i.e. `https://` scheme) imports? If so, nice.

plugins/tailwind/compiler.ts
```diff
@@ -62,6 +68,39 @@ export async function initTailwind(
     return pattern;
   });
 
+  let imports;
```
iuioiua:
```suggestion
  let importMap;
```
Using `imports` as a variable name suggests that it represents an import map's `imports` property. Using `importMap` instead more accurately represents an import map object which contains the `imports` property, which this object does.

iuioiua (changes requested):
Mostly LGTM

plugins/tailwind/compiler.ts
deer: yes, this was the reasoning

plugins/tailwind/compiler.ts
deer: ok, thanks

src/server/types.ts
```diff
@@ -495,6 +496,10 @@ export interface Plugin<State = Record<string, unknown>> {
   middlewares?: PluginMiddleware<State>[];
 
   islands?: PluginIslands;
+
+  location?: string;
+
+  projectLocation?: string;
```
iuioiua: Sorry, I forgot to mention, can you please document these new properties with JSDoc blocks?

iuioiua, line 511: Also, I wonder whether it's worth having `projectLocation?: string | URL`. Doing so would:
1. Allow someone to pass in `Deno.cwd()`, like we do in the SaaSKit PR. We would then use `path.toFileUrl()` from `deno_std`.
2. Not require passing in `url.ref`, but rather just the `url`. We could use `path.toFileUrl()` here too.

WDYT?

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
```diff
@@ -0,0 +1,16 @@
...
+import VeryNestedPluginComponent from "../../../../components/VeryNestedPluginComponent.tsx";
+
+const currentUrl = new URL(import.meta.url);
+currentUrl.pathname = currentUrl.pathname.split("/").slice(0, -5).join("/") +
+  "/";
+
+export const veryNestedPlugin = {
+  name: "very nested plugin",
+  location: import.meta.url,
+  projectLocation: currentUrl.href,
```
iuioiua:
```suggestion
import { toFileUrl } from "$std/path/to_file_url.ts";

export const veryNestedPlugin = {
  name: "very nested plugin",
  location: import.meta.url,
  projectLocation: toFileUrl(Deno.cwd()).href,
```
Ditto for `tests/fixture_tailwind_remote_classes/nestedPlugins/nestedPlugin.ts`.

iuioiua (changes requested):
Sorry. Few things that I didn't realise earlier. Definitely looking better, though.

src/server/types.ts
deer: I like the part about accepting URLs. Although I'm not sure that `Deno.cwd()` is always the right thing. I'll definitely play around with this later today and see what happens.

deer: why is there no :saluting_face: here? anyway, of course. this should probably be caught with the linter, right?

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: I'm not sure about this...

What if I do `deno run -A tests/fixture_tailwind_remote_classes/dev.ts`? I don't think `Deno.cwd()` would give the right answer then.

iuioiua: This is true. My suggestion would only work for the Fresh test that spawns a sub-process in the `tests/fixture_tailwind_remote_classes` folder. Is there a cleaner alternative that lets the reader know what `currentUrl.href` evaluates to? It's not so clear right now.

src/server/types.ts
iuioiua: It can, if we use `deno doc --lint`.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: It seems like there should be an easy way of going up N levels, based on starting with `import.meta.url`.

plugins/tailwind/compiler.ts
```diff
@@ -62,6 +68,39 @@ export async function initTailwind(
...
+  if (path.extname(config.denoJsonPath) === ".json") {
+    importMap = (await import(path.toFileUrl(config.denoJsonPath).href, {
+      with: { type: "json" },
+    })).default;
+  } else if (path.extname(config.denoJsonPath) === ".jsonc") {
+    const fileContents = Deno.readTextFileSync(config.denoJsonPath);
+    importMap = jsoncParse(fileContents);
+  } else {
+    throw Error("deno config must be either .json or .jsonc");
+  }
```
deer, line 81: @iuioiua, are you sure there's no room in the std library for something like this? I've just realized this won't work if the user is using `importMap` instead of `imports` in their `deno.json(c)` file. Surely I'm not the only person who wants to read the config file and get a certain property off of it.

Ideally this entire block could be condensed to:
```ts
const importMap = await readConfig(config.denoJsonPath).imports();
```

`readConfig` would take care of handling `.json` and `.jsonc` differences, and `importMap()` would handle whether it's in the config itself, or buried in a separate import map.

src/server/types.ts
deer: When I run `deno doc --lint src/server/types.ts` I get `error: Found 185 documentation lint errors.`
:oof:

deer: Using `toFileUrl` seems risky when needing to deal with remote plugins. I think we should merge this conversation with the other one, since they seem too similar.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: Something like this
```ts
export function upNLevels(url: string | URL, levels: number) {
  const inputUrl = (typeof url === "string") ? new URL(url) : url;

  const segments = inputUrl.pathname.split(path.SEP);

  if (levels >= segments.length) {
    throw new Error("Cannot go up more levels than the current depth");
  }

  const upperPathSegments = segments.slice(0, -levels);
  const newPath = path.join(...upperPathSegments);

  inputUrl.pathname = newPath;

  return inputUrl.href;
}
```

Then we can simplify the plugin declaration like this:
```ts
export const nestedPlugin = {
  name: "nested plugin",
  location: import.meta.url,
  projectLocation: upNLevels(import.meta.url, 2),
  routes: [{
    path: "routeFromNestedPlugin",
    component: NestedPluginComponent,
  }, {
    path: "atRouteFromNestedPlugin",
    component: AtPluginComponent,
  }],
} satisfies Plugin;
```

This approach avoids any issue with file vs http urls and relying on `Deno.cwd`.

plugins/tailwind/compiler.ts
deer: I ended up getting something like this working:
```ts
const importMap = getImportMap(readConfig(config.denoJsonPath));
```

Let me know if I should create a PR in `deno_std` to add this.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: Also let me know if this one could live in `deno_std`.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
```diff
@@ -0,0 +1,13 @@
+import { upNLevels } from "$fresh/plugins/tailwind.ts";
+import type { Plugin } from "../../../../../../src/server/types.ts";
+import VeryNestedPluginComponent from "../../../../components/VeryNestedPluginComponent.tsx";
+
+export const veryNestedPlugin = {
+  name: "very nested plugin",
+  location: import.meta.url,
+  projectLocation: upNLevels(import.meta.url, 5),
```
iuioiua: Try using [`path.normazlie()`](https://deno.land/std@0.214.0/path/mod.ts?s=normalize) from `deno_std` instead.

iuioiua (changes requested):

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
deer: How will this help with a URL? e.g. `https://raw.githubusercontent.com/denoland/fresh/727d0e729e154323b99b319c04f9805f382949f0/tests/fixture_tailwind_remote_classes/hackyRemotePlugin.tsx` I guess I'm unclear what `normalize` would help with.

plugins/tailwind/compiler.ts
iuioiua: Hm... That actually might be good for the Deno ecosystem. Can you open an issue with your proposal?

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
iuioiua: Actually, check out [`url.normalize()`](https://deno.land/std/url/normalize.ts?s=normalize). That might help.

deer: I can of course change it to `projectLocation: normalize(import.meta.url + "../../../../../../").href` if you think that's better, but it seems really awkward to use.

plugins/tailwind/compiler.ts
deer: Sure, I'll open one tomorrow.

tests/fixture_tailwind_remote_classes/nestedPlugins/nested/nested/nested/veryNestedPlugin.ts
iuioiua: Yeah, I do think that's better 🙂

deer: Ok, I've removed the offending code. Note: in order to go up N levels, the user must now include N+1 copies of `../`.  This seems confusing, but maybe I'm just inexperienced with URL path modification.

iuioiua:
Another great PR at the time but can now be closed. Thanks again, @deer.

Issue #2240, opened 2024-01-11
`kv.listenQueue` prevents build mode from working


marvinhagemeister:
The `kv.listenQueue()` function prevents the process from exiting because it waits for messages forever. This breaks the build command.

We should probably add something similar to `IS_BROWSER` like `IS_BUILD_MODE` or something so that users can exclude long running stuff from being executed during build.

Another approach would be to avoid importing user code during build. Off the top of my head I think we only import user code during build for the manifest generation. Maybe we should replace that with AST-based logic. That way user's don't need to add conditionals or anything to their code.

mcgear:
I was just trying to figure out a workaround for this, until we get something going in the framework. Any thoughts?

mcgear:
For now, i have settled on managing a second `dev.build.ts` file that just doesn't import my handlers, and that seems to be working.

marvinhagemeister:
A workaround is to check if Fresh is running in build mode and then skipping the logic. Something like:

```ts
if (!Deno.args.includes("build")) {
  somethingThatShouldNotRunDuringBuild();
}
```

phoens:
Fresh provides `fresh.config.ts` file. In side this file you can pass custom configurations. It take onListen callback which triggers when server starts listening. 

Connect to remote systems here

byvide:
+1 (i had this issue with fresh + cron + deno deploy)

- workaround is simple (exit at dev.ts or guard at cron/kv calls)
- but i would like to see minimum a warning, so ppl arent stuck wondering why the build process hangs

PR #2268, opened 2024-01-19
Shorthand for IS_BUILD_MODE

mcgear:
This is a quick shorthand for IS_BUILD_MODE (Deno.args.includes("build")), should help simplify code that needs to not run during ahead of time builds. Will address #2240 .


deer (changes requested):
My standard comment: tests?

Although you could probably get away in this case without writing anything new if you just refactor the one place in the codebase that uses this line to use your new `const`:
https://github.com/denoland/fresh/blob/34c875c71cc5c25a1e355bee6fcdf798099f38a9/src/dev/dev_command.ts#L41

This is executed in the tests, so if it works here, I'd say it's covered.

mcgear:
This change has been put in place. Thanks for the thoughts.

deer (approved):

mcgear:
Just updated the branch

bartlomieju:
Thanks for the contribution! This PR is from the Fresh 1.x era. The files modified here (`src/dev/dev_command.ts`, `src/server/mod.ts`, `src/server/utils.ts`) no longer exist in Fresh 2.x. The build mode distinction is now handled via `config.mode`.

Closing as stale.

PR #2272, opened 2024-01-23
fix: no-window lint error

deer:
Latest PR runs are failing with linting errors, e.g. https://github.com/denoland/fresh/actions/runs/7625625084/job/20770308634?pr=2266

Maybe this shouldn't be happening like this right now, but we'll still need to clean these up eventually.


deer:
@marvinhagemeister, please take a look.

marvinhagemeister (approved):
Thanks 👍

PR #2274, opened 2024-01-25
bugfix: autofocus didn't work in example Input component

skybrian:
Disabled elements can't receive focus. If autofocus is requested, we shouldn't disable the input element on startup, or it will have no effect.

The easiest way to fix it is to leave the input element enabled. However, I'm not sure what the other consequences of that are.

(Note: this is how I fixed it in my app, but I haven't tested this patch.)


bartlomieju:
Closing as stale — this targets the Fresh 1.x docs site gallery component which no longer exists after the 2.0 rewrite. (Also note there's an operator precedence bug in the fix — the `||` needs parentheses.) Thanks for the contribution!

PR #2275, opened 2024-01-25
Multi plugin islands configs

mcgear:
This PR is a simple change that would allow the plugin islands config to contain multiple configurations of islands.  The purpose of this is to support configuring islands from shared libraries. We have an architecture for our atomic design kit that promotes composition of the final atomic library from base libraries. The issue is that the sub libraries that are composed together all have individual import.meta.url values that need to be used, so a single `baseLocation` does not work.

To prevent breaking changes, the plugin.islands value is updated to support a single PluginIslands config or an array of them.


deer:
This sounds cool, although I haven't had time to look in detail. But after a quick glance at the changes, there are some basics that need taking care of:
1. Don't update the current documentation -- that will cause confusion. Instead, branch `plugins.md` to `/docs/canary/concepts/` and then update `toc.ts` appropriately. You should be able to see the different versions when you do `deno task www`. This way people won't find out about the feature before it's released
2. What about some tests? Something in `/tests/fixture_plugin/` is probably a good starting point.

mcgear:
Will do, thanks for the input. I'll try to work this in today.

mcgear:
I got the docs updated (i think correctly). In terms of testing, i was looking at it more in terms of code coverage, and due to how the code is rewritten, the existing test covers the new code. Should i look at a test that actually utilizes an external library? or given that coverage is 100% with my changes, leave it as is?

deer:
Regardless of what the code coverage says, if you're adding a new type and not using it, then in my opinion it's not covered. At minimum from a "usage guide" (i.e. how does one use this) perspective I would expect to see a plugin that uses your nice new type in this change. This also helps from a type checking perspective, to make sure we catch any possible regressions with code that users have defined.

So something like this:
```ts
{name: "multipleIslandsLocationPlugin",
islands: [{baselocation: "firstPlace", paths: ["a", b"]}, {baselocation: "secondPlace", paths: ["c", d"]}]
}
```

But since I can't merge your change, my opinion is just an opinion. Feel free to ignore me and wait for Marvin 😄

mcgear:
Thank you for your thoughts, much appreciated, I'll work to get this test in place so we can look at merging this in.

mcgear:
Merged in latest

deer:
Yes, but what about the tests?

mcgear:
Shoot, had several contributions going in different repos and forgot to do the tests here, i'll work to get those in today.

mcgear:
Alright, i added a second island in a "sub" folder, and a new plugin that uses it with the original island. The test is `plugin supports multiple islands` and seems to be passing. Not sure if this is exactly how the test should be configured, so open to input on where it belongs and how it should be written.

tests/plugin_test.ts
```diff
@@ -182,6 +182,35 @@ Deno.test({
   sanitizeResources: false,
 });
 
+Deno.test({
+  name: "plugin supports multiple islands",
```
deer: This is a bit pedantic, but plugins already support multiple islands, because `PluginIslands.paths` is a `string[]`. What if you rename the test to "plugins support multiple island configs" or something like that. I think it would then be more clear what the test is asserting.

tests/fixture_plugin/utils/route-plugin-multiple-islands.ts
```diff
@@ -0,0 +1,29 @@
...
+      {
+        path: "pluginroutewithislands",
+        component: IslandsPluginComponent,
+      },
+    ],
+    islands: [
+      {
+        baseLocation: import.meta.url,
+        paths: ["./sample_islands/IslandFromPlugin.tsx"],
+      },
```
deer, line 20: This is interesting because it essentially duplicates the island registered from `route-plugin.ts`. As I was stepping through the code, I saw what was about to happen and kept thinking "no problem, it'll get cleared out later". But it doesn't! But it also doesn't cause problems... 🤔 

I guess it's not a terrible thing to have it like this, because sooner or later someone is bound to end up in a similar situation. I've made note of this in https://github.com/denoland/fresh/pull/1568#issuecomment-1925280649 that we need to detect island conflicts like this. Although this is probably a special case in that the "conflict" isn't an issue, because everything is the same.

deer (approved):
This looks great. Thanks for improving it.

tests/plugin_test.ts
mcgear: done

tests/fixture_plugin/utils/route-plugin-multiple-islands.ts
mcgear: Any changes to make for this?

deer: Nothing to do. I just wanted to highlight that this is happening and isn't causing issues, even though it's unexpected.

mcgear:
Alright, i was testing this out and couldn't get the islands to work. I put together a simple recreation of the issue (in an olde repo i used for something else).

https://github.com/mcgear/deno-fresh-icons-in-island

I was trying to trace through the code to see why MenuButton wasn't being mapped as an island.  Looking for some pointers on where i can look to work through remaining issues.

The island is peing mapped on getServerContext "correctly", the best i can tell. But when scoping out delivered island code in chrome dev tools, i don't see MenuButton code, and it doesn't function as expected. 

Any help on what i'm missing would be great! Thanks in advance

mcgear:
Wondering if it could have to do with the fact that i am defining the plugin inline? Instead of in the sub library?

mcgear:
I can confirm that this is working in multiple deployed applications. One easy to use example is mydiviner.ai. Also using it with https://dashboard.openbiotech.co/

bartlomieju:
Closing — Fresh 2.x completely redesigned the plugin system, so this PR no longer applies:

- The `Plugin` type with `islands: PluginIslands` no longer exists
- Fresh 2.x plugins use the `Builder` class instead (e.g. `builder.registerIsland()`, `builder.onTransformStaticFile()`)
- The files this PR modifies (`src/server/fs_extract.ts`, `src/server/types.ts`) don't exist in the current codebase

The underlying use case (composing islands from multiple libraries) is still valid — in Fresh 2.x it could potentially be addressed via multiple `Builder.registerIsland()` calls, but that would require a fresh PR targeting the new architecture. Thanks for the contribution!

Issue #2285, opened 2024-01-28
[Github Actions] actions/checkout@v3 should be upgraded to v4


bjesuiter:
The actions/checkout@v3 action uses node16 under the hood, which is deprecated by github. 
We should switch to actions/checkout@v4, in the docs and for generating the file. 

=> I'll shortly add a PR for this.

marvinhagemeister:
Resolved by https://github.com/denoland/fresh/pull/2286

PR #2286, opened 2024-01-28
fix: GH actions warning by upgrading actions/checkout to v4 #2285

bjesuiter:
I tested the change in my own repo, the warning on github disapears for the "actions/checkout" action. 
It is still there for the denoland/setup-deno action, but an upgrade PR is already ongoing for this action, see: 
https://github.com/denoland/setup-deno/pull/56


marvinhagemeister (approved):
Good catch. Thanks for updating this 👍

PR #2293, opened 2024-01-31
chore: upgrade to esbuild 0.20.0

deer:
Sadly this isn't working due to failures like this:
```
deno test -A src/build/esbuild_test.ts
running 1 test from ./src/build/esbuild_test.ts
esbuild ...
  esbuild snapshot with cwd=Deno.cwd() ...
ok | 0 passed | 0 failed (88ms)

error: Promise resolution is still pending but the event loop has already resolved.
```
I'm not sure what I'm doing wrong, but I thought I would open this as a draft PR so others can see what I've started.
Note the esbuild_deno_loader import is coming from https://github.com/lucacasonato/esbuild_deno_loader/pull/108.

### DIFF
--- a/src/build/deps.ts
+++ b/src/build/deps.ts
@@ -6,5 +6,5 @@ export {
   toFileUrl,
 } from "https://deno.land/std@0.211.0/path/mod.ts";
 export { escape as regexpEscape } from "https://deno.land/std@0.211.0/regexp/escape.ts";
-export { denoPlugins } from "https://raw.githubusercontent.com/lucacasonato/esbuild_deno_loader/5bd8975c1abcc222024b2b10d902b60377a3b908/mod.ts";
+export { denoPlugins } from "https://deno.land/x/esbuild_deno_loader@0.9.0/mod.ts";
 export { assertEquals } from "https://deno.land/std@0.211.0/assert/mod.ts";
--- a/src/build/deps.ts
+++ b/src/build/deps.ts
@@ -6,5 +6,5 @@ export {
   toFileUrl,
 } from "https://deno.land/std@0.211.0/path/mod.ts";
 export { escape as regexpEscape } from "https://deno.land/std@0.211.0/regexp/escape.ts";
-export { denoPlugins } from "https://deno.land/x/esbuild_deno_loader@0.8.5/mod.ts";
+export { denoPlugins } from "https://raw.githubusercontent.com/lucacasonato/esbuild_deno_loader/5bd8975c1abcc222024b2b10d902b60377a3b908/mod.ts";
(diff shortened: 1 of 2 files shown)


deer:
The error is reproducible with the following minimal case:
```ts
import * as esbuild from "https://deno.land/x/esbuild@v0.20.0/mod.js";

Deno.test("esbuild", async (t) => {
  await t.step("esbuild snapshot with cwd=Deno.cwd()", async () => {
    const builder = new EsbuildBuilder();
    await builder.build();
  });
});

export class EsbuildBuilder {
  async build(): Promise<boolean> {
    await esbuild.initialize({});

    try {
      await esbuild.build({ entryPoints: ["./someFile.ts"] });
      return true;
    } finally {
      await esbuild.stop();
    }
  }
}
```

I'll work on continuing to debug later.

deer:
After taking a break I decided to put this into the esbuild source and modify some things so I could log more. The test is basically the same as before, except now I import like this: `import * as esbuild from "../deno/mod.js";`

I modified the existing:
```js
    close: async () => {
      await writer.close();
      await reader.cancel();
      await child.status;
    },
```
to have some extra logging and to invoke `child.kill`, just for good measure:
```js
    close: async () => {
      try {
        console.log(`resources before writer.close(): ${JSON.stringify(Deno.resources())}`);
        await writer.close();
        console.log(`resources before reader.cancel(): ${JSON.stringify(Deno.resources())}`);
        await reader.cancel();
        console.log(`resources before child.kill(): ${JSON.stringify(Deno.resources())}`);
        child.kill();
        console.log(`resources before child.status: ${JSON.stringify(Deno.resources())}`);
        await child.status;
        console.log("does this log");
      } catch (e) {
        console.log("are there any errors");
        console.log(JSON.stringify(Deno.resources()));
        console.log(e);
      }
    },
```
and then I got the following:
```
esbuild ...
  esbuild snapshot with cwd=Deno.cwd() ...
------- output -------
let a = 2 + 2;
resources before writer.close(): {"0":"stdin","1":"stdout","2":"stderr","3":"childStdin","4":"childStdout","5":"child"}
resources before reader.cancel(): {"0":"stdin","1":"stdout","2":"stderr","4":"childStdout","5":"child"}
resources before child.kill(): {"0":"stdin","1":"stdout","2":"stderr","5":"child"}
resources before child.status: {"0":"stdin","1":"stdout","2":"stderr","5":"child"}

ok | 0 passed | 0 failed (94ms)

error: Promise resolution is still pending but the event loop has already resolved.
```

So it seems like something in the call to `await child.status` is causing the event loop to exit and not return properly, sinice we don't see `does this log` in the logs.

@lucacasonato @marvinhagemeister, I'll pass this over to you guys at this point, since I'm not very good at debugging the deno runtime itself.

PR #2313, opened 2024-02-14
fix: open redirect issue with trailing slash detection

marvinhagemeister:
Passing a path like `/foo/bar` as the location header is interpreted as a relative path. But when it contains two leading slashes like `//evil.com/` it is interpreted as a port relative url. This has the consequence that the origin of a URL can essentially be replaced:

```ts
new URL("//evil.com", "https://example.com/");
// -> 'https://evil.com/'
```

Our trailing slash detection logic didn't guard against this case. By normalizing the pathname of the URL and stripping sibling forward slashes we sidestep this problem.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -234,11 +234,10 @@ export class ServerContext {
         !trailingSlashEnabled
       ) {
         // Remove trailing slashes
-        const path = url.pathname.replace(/\/+$/, "");
-        const location = `${path}${url.search}`;
+        url.pathname = url.pathname.replace(/\/+$/, "");
         return new Response(null, {
           status: STATUS_CODE.TemporaryRedirect,
-          headers: { location },
+          headers: { location: url.toString() },
         });
       } else if (trailingSlashEnabled && !url.pathname.endsWith("/")) {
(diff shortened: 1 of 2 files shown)


crowlKats (approved):
LGTM

src/server/context.ts
```diff
@@ -187,6 +187,11 @@ export class ServerContext {
       connInfo: ServeHandlerInfo = DEFAULT_CONN_INFO,
     ) {
       const url = new URL(req.url);
+      // Syntactically having double slashes in the pathname is valid per
+      // spec, but there is no behavior defined for that. Practically all
+      // servers normalize the pathname of a URL to not include double
+      // forward slashes.
+      url.pathname = url.pathname.replaceAll(/\/+/g, "/");
```
lucacasonato, line 194: Can you also check further down whether the original pathname matches the changed pathname? Because if it doesnt (there were double slashes), we should redirect to the normalized variant. Otherwise there will be a weird edge case where this logic matches as if there are no double slashes, but when a user does `new URL(req.url)`, they still get the "old" non-normalized URL.

PR #2318, opened 2024-02-15
fix: kill zombie builds if they complete

deer:
I was working on a dependency visualizer in order to learn more about `deno_graph` and various network diagram visualization libraries. I've started out with `vis-network`, but I quickly encountered a problem when deploying: despite the `build()` function in `build.ts` finishing, the process doesn't end!

This is different than #2240 because... I dunno, I just think it is. I've gated the code in the fixture with the standard advice: `!Deno.args.includes("build")`, but this does not help. But if you comment out `Deno.exit()` then it starts hanging again. And going a step further, commenting out the code in the effect like this:
```ts
      // network.current = new Network(container.current, {
      //   nodes: nodes,
      //   edges: edges,
      // }, {});
```
causes the build to finish, even though that code isn't even running on the server.

Perhaps I'm going about this wrong and I should instead be importing `vis-network` dynamically, if we're not in build mode? I'm really unsure what's going on here, but using the code in the forked branch unblocked my project, so I thought I would clean it up enough and write a test -- perhaps it's useful for someone else.


deer:
Well, this fixes my problem and passes on my computer. I'm not sure what's going on here. I'll look at the failures in more detail tomorrow.

PR #2321, opened 2024-02-16
startServer onListen logs https when key & cert specified

will-weiss:
Some team members of mine were confused by the default output 

```
 🍋 Fresh ready
    Local: http://localhost:8000/
```

when the server is actually running HTTPS. This should now print

```
 🍋 Fresh ready
    Local: https://localhost:8000/
```

when a key and cert are specified.


marvinhagemeister (approved):
Sweet, thanks 👍

PR #2334, opened 2024-02-23
fix: properly prefix all scripts with base path

### LINKED ISSUE #2326: imported revive function and islands from script into browser is not taking basepath into account
Trying to add a basepath `/neoplayer/` to a deno fresh app, with another app (node project) at the root.
Basepath is setup in deno fresh config and import in main.ts. Everything works fine except the islands.

In the html file loaded in the browser, basepath is applied to src and href (like said in the docs)

image

At the end there is a script that imports the revive function from main.js and islands from their own js files. Those imports are not prefixed with the basepath. 

image

So the request is on the root and the js is not load which leads me to this error in the browser because my root application returns a 404 error page :

image

deno version : 1.36.1
deno fresh version : 1.6.5

deer:
closes https://github.com/denoland/fresh/issues/2326

There was a big misunderstanding on my part -- in particular that a different server might be listening at the root, so the redirect built into `context.ts` will never happen. The existing test (as mentioned in the issue) passes, but only because no one else is listening, and fresh is able to perform the redirect.

I briefly considered writing a docker container to orchestrate this properly, but then I didn't want to have docker running in order to run these tests. (My laptop is old and bad.) So instead the two new tests very thoroughly assert that scripts included in the html are properly prefixed with the base path. And yes, they fail if the fix is reverted.

As for the fix itself: I just needed to prepend the asset URL with the base path.

### DIFF
--- a/src/server/render.ts
+++ b/src/server/render.ts
@@ -358,6 +358,7 @@ export async function render<Data>(
     dependenciesFn: opts.dependenciesFn,
     styles: ctx.styles,
     pluginRenderResults: renderResults,
+    basePath,
   });
 
   // Append error overlay in dev mode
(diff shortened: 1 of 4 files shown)


marvinhagemeister (approved):
This is great, thanks for fixing this 👍

PR #2340, opened 2024-02-26
fix: $fresh ref in import sources

mcgear:
There was an issue with the tailwind error page being pulled in by ref to $fresh, fixed so now it just uses the file


deer (changes requested):
Looks like you didn't base this off main, but rather your `build mode` branch.

mcgear:
Ahh, right. Yea, i forgot to branch for my build mode changes. I removed those from this branch, so it is just the ref fixes.

deer (approved):

marvinhagemeister (approved):
Good catch, thanks for making a PR 👍

PR #2348, opened 2024-03-03
fix: specify array for `on.pull_request.branches`

morinokami:
Currently, the `on.pull_request.branches` in deploy.yml, which is created during project initialization, specifies `main` as a string value. However, according to the [GitHub Actions documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore), it should be an array. I have corrected this inconsistency.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches: [main]
   pull_request:
-    branches: main
+    branches: [main]
 
 jobs:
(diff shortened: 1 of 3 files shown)


bartlomieju:
Closing as stale — this targets Fresh 1.x files (deploy.yml, docs, project init template) which have been replaced in the 2.0 rewrite. Thanks for the contribution!

PR #2349, opened 2024-03-04
doc: Rename _req used to req

narumincho:
Where other documents use req, they use the name req instead of _req

https://github.com/denoland/fresh/blob/b7281792a0a5b2674f659514b1e58116fabd6460/docs/latest/concepts/middleware.md?plain=1#L153

### DIFF
--- a/docs/latest/examples/dealing-with-cors.md
+++ b/docs/latest/examples/dealing-with-cors.md
@@ -47,18 +47,18 @@ deal with "preflight requests". Let's imagine you're trying to support a
 ```ts routes/_middleware.ts
 import { FreshContext } from "$fresh/server.ts";
 
-export async function handler(_req: Request, ctx: FreshContext) {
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Thanks!

kuboon:
I got 
```
TypeError: Cannot change headers: headers are immutable
    at Headers.set (ext:deno_fetch/20_headers.js:425:13)
```
Is this sample code work?

marvinhagemeister:
@kuboon Please don't comment on random PRs and open a new issue instead.

The error message you shared makes it sound like you're trying to update the `Request` headers instead of the `Response` headers.

kuboon:
@marvinhagemeister 
Thank you.

I found that I use `Deno.upgradeWebSocket(req);` in my route and the returned response's headers is locked.
I solved the problem by avoid adding CORS headers when request has `Upgrade: websocket` header.

PR #2356, opened 2024-03-10
feat: allow trailing slash redirect to be disabled

### LINKED ISSUE #2343: Add a way to disable trailing slash logic
A global middleware that is not affected by fresh is required.
The handler defined in the routes/_middleware.ts file is not global middleware and will be affected by the trailingSlashEnabled redirect from fresh.

https://deno.land/x/fresh@1.6.5/src/server/context.ts?source#L175

https://deno.land/x/fresh@1.6.5/src/server/context.ts?source#L237

deer:
closes #2343


marvinhagemeister (commented):
The docs and everything are really great! It's a very thorough PR 👍 

One thing I'm wondering is if we should turn the `trailingSlash` option into a tri-state vs adding a second option. I haven't spent enough thinking on this, but I'm trying to come up with use cases where you'd for example want to have trailing slashes in your URL, but don't want to have the redirect. I might be missing something.

PR #2357, opened 2024-03-10
fix: allow multiple set-cookie headers to be set before render

### LINKED ISSUE #2346: Setting multiple cookies before render will only send the last Set-Cookie header
When creating a `Headers` object, setting multiple cookies on it and passing that to `ctx.render` via the `RenderOptions`, only the last cookie will be set. There is a workaround one can render the response and then modifying the headers on the response but it should work via the other way, too, I suppose.

```ts
// this will not work
const headers = new Headers();
cookie.setCookie(headers, /*...*/);
cookie.setCookie(headers, /*...*/); // <- only this one will be set
return ctx.render({}, { headers });

// this will work
const res = await ctx.render({});
cookie.setCookie(res.headers, /*...*/);
cookie.setCookie(res.headers, /*...*/);
return res
```

deer:
closes #2346

The primary issue was the snippet `headers[key] = value`. This caused only one `set-cookie` to make it through. Switching to a `Headers` object allowed me to use `append` when appropriate, allowing for multiple `set-cookie` entries.

I also cleaned up one of the tests to make use of steps, because I originally broke it and couldn't immediately tell which part was failing.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -785,10 +785,10 @@ function sendResponse(
   },
 ) {
   const [body, uuid, csp] = resp;
-  const headers: Record<string, string> = {
+  const headers: Headers = new Headers({
     "content-type": "text/html; charset=utf-8",
     "x-fresh-uuid": uuid,
-  };
+  });
 
   if (csp) {
     if (options.isDev) {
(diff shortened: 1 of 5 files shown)


marvinhagemeister (approved):
Nice catch!

PR #2358, opened 2024-03-12
feat: add `ctx.redirect()` helper

marvinhagemeister:
This PR adds a `ctx.redirect(path, status)` helper that aims to reduce the verbosity of having to manually create a `Response` obejct for a mere redirect.

```ts
// Before
new Response(null, {
  status: 307,
  headers: {
    Location: "/foo/bar"
  },
})

// After
ctx.redirect("/foo/bar", 307)
```

If the second parameter is not passed, it will default to `302`.

### DIFF
--- a/src/server/context.ts
+++ b/src/server/context.ts
@@ -114,6 +114,30 @@ export async function getServerContext(state: InternalFreshState) {
   );
 }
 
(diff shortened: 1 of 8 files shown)


deer:
Hi @marvinhagemeister, cool feature, but what about docs? 🙃 

I can open a PR if you'd like.

PR #2362, opened 2024-03-14
docs: remove outdated information

rojvv:


deer (approved):
Wow, that's sure out of date. Good catch!

marvinhagemeister (approved):

PR #2365, opened 2024-03-15
fix: move import_map and change import_map scopes

### LINKED ISSUE #2364: Location and description of import_map.json
`.vscode/import_map.json` is used in `deno.json` as well as `.vscode/settings.json`.
If `deno.json` uses it, I thought it was fine to place it in root.

Even if not, I felt that the following notations should be eliminated.
```json
  "scopes": {
    "THIS FILE EXISTS ONLY FOR VSCODE! IT IS NOT USED AT RUNTIME": {}
  },
```
https://github.com/denoland/fresh/blob/main/.vscode/import_map.json#L2-L4

k35o:
closes https://github.com/denoland/fresh/issues/2364


import_map.json
```diff
@@ -1,7 +1,4 @@
 {
-  "scopes": {
-    "THIS FILE EXISTS ONLY FOR VSCODE! IT IS NOT USED AT RUNTIME": {}
-  },
```
lucacasonato, line 4: This is still correct - the import map is not used to remap anything at runtime, as every Fresh project has it's own import map already.

k35o: @lucacasonato 
> THIS FILE EXISTS ONLY FOR VSCODE!

It is written as above, but I think it is used in the lower part and `deno.json`, but is it necessary?
https://github.com/denoland/fresh/blob/4f671ec961c9497c2e2f7c055c0b8e2422b91b49/update.ts#L73

> IT IS NOT USED AT RUNTIME

Does this part only need?

PR #2368, opened 2024-03-18
docs: fix wording related to route groups

dizlexik:


marvinhagemeister (approved):
Good catch. Thanks 👍

marvinhagemeister:
Ah looks like the file is not formatted correctly. A quick `deno fmt` should resolve this.

marvinhagemeister (approved):

PR #2374, opened 2024-03-21
showcase: Remove non-Fresh sites + sites returning errors

rojvv:
These items are no longer Fresh.


marvinhagemeister (approved):
Some seem to be still Fresh, but are not working so it's probably best to remove them. Thanks for making the PR 👍

rojvv:
Wow, that was fast, thank you. You’re welcome!

PR #2389, opened 2024-03-27
chore: satisfy use-import-type lint (biomejs)

sigmaSd:
https://biomejs.dev/linter/rules/use-import-type/

From that page the argument is 

>The rule ensures that all imports used only as a type use a type-only import. It also groups inline type imports into a grouped import type.


marvinhagemeister (approved):
Thanks!

PR #2393, opened 2024-04-08
chore: minor cleanups

iuioiua:
These changes occur by just running `deno task ok` on `main`.


marvinhagemeister (approved):
Nice, thanks!

PR #2407, opened 2024-04-18
www: remov no build step from marketing

marvinhagemeister:
We've been moving away from that for a couple of months now, so featuring the "no build step" prominently in our marketing feels wrong.


rojvv:
@marvinhagemeister Forgot to edit the introduction

image

Issue #2409, opened 2024-04-19
[bug] partial on form submit


emilbillberg:
When using partial with forms the url is updated in the browser to the url of the partial value (`f-partial`).

Example:
```
<div f-client-nav>
  <form method="get" action="/search" f-partial="/partials/test">
    <input type="text" name="q" />
    <button type="submit">Search</button>
  </form>
</div>
```
On `submit` event the url in the browser is updated to `/partials/test?q=<query>`. I would expect the url in the browser to be updated to `/search?q=<query>`.

When digging through the code I can see the function `maybeUpdateHistory` with the url of the partial location is called. 
https://github.com/denoland/fresh/blob/main/src/runtime/entrypoints/main.ts#L1096

A solution would be to create two different url variables, one from `f-partial` and one from `action`. Both of them need to contain the new search params from the form. That approach seems to be done for the `click` event https://github.com/denoland/fresh/blob/main/src/runtime/entrypoints/main.ts#L958-L960

adamgreg:
This has been affecting my app too. Thanks @emilbillberg!

PR #2410, opened 2024-04-19
fix(partials): on form submit update url to correct value

emilbillberg:
Fixes issue: https://github.com/denoland/fresh/issues/2409

Changes:
- uses a url variable for `f-partial` called `partialUrl`
- uses a variable for current url called


emilbillberg:
I'm a bit unsure how to add tests. If it is needed please guide me.

emilbillberg:
fixed in: https://github.com/denoland/fresh/pull/2495

PR #2413, opened 2024-04-26
www: document colocation of route groups

### LINKED ISSUE #2305: Add documentation for colocation
The blog post for Fresh 1.4 [discusses](https://deno.com/blog/fresh-1.4#colocated-islands-components-and-more) how you could use route group directories prefixed with an underscore to colocate files that are related to a given route:

> When the name of a route group folder starts with an underscore, like (_components), Fresh will ignore that folder and it’s effectively treated as private. This means you can use these private route folders to store components related to a particular route.

However, this doesn't seem to be documented anywhere in the docs. It probably should be.

kerimhudson:
This adds in the information regarding co-location of islands and components from [this blog post](https://deno.com/blog/fresh-1.4#colocated-islands-components-and-more) into the main documentation.

A lot of the text is pulled directly from the blog post, but I'm happy to edit it if need be.

Closes #2305

### DIFF
--- a/docs/latest/concepts/routing.md
+++ b/docs/latest/concepts/routing.md
@@ -110,3 +110,41 @@ routes in a folder and use a different `_layout` file for each group.
 > ```
 
 [urlpattern]: https://developer.mozilla.org/en-US/docs/Web/API/URL_Pattern_API
+
+## Co-location
+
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Sweet, this is fantastic! Thanks for making a PR 👍

PR #2414, opened 2024-04-26
fix: add `.js` + `.jsx` extension to tailwind content glob

### LINKED ISSUE #2291: Bug: Tailwindcss not working in js/jsx files
Was reported on discord earlier today. Haven't looked into it yet.

kerimhudson:
Whilst Deno supports Typescript by default, not everyone necessarily will want to write every file in Typescript. If using TailwindCSS, at the moment the files Tailwind searches for classes does not include `.js` and `.jsx` endings. Whilst you can manually add them via the `tailwind.config.ts`, it might be nice to have these added in by default for a better onboarding experience.

Closes #2291

### DIFF
--- a/init.ts
+++ b/init.ts
@@ -330,7 +330,7 @@ const TAILWIND_CONFIG_TS = `import { type Config } from "tailwindcss";
 
 export default {
   content: [
-    "{routes,islands,components}/**/*.{ts,tsx}",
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good idea, thanks for making a PR 👍

PR #2416, opened 2024-04-27
docs: using server components in islands

### LINKED ISSUE #2369: docs: explain better the distinction between islands and components
Quoting @CAYdenberg in: https://github.com/denoland/fresh/issues/2363#issuecomment-2004817651

> Well, if it helps, components that are children of islands can receive functions as props. So, your ComplexOrBigComponent could live in components/, receive onClick when used as a descendent of an island, and not receive it when used outside of an island. Just mark onClick as being optional in your props interface declaration (and handle the cases where it is undefined).

I also found this confusing at first. When should a file go under components/ versus islands/?

 I had thought that islands were client-side and components were server-side, but that's not right. Instead, everything should be a component, except when it needs to be the root of a client-side hierarchy, and then it's an island. (Why? Because there are restrictions on what you can pass in to an island.)

In particular, the [component page](https://fresh.deno.dev/components) has some components tagged as islands that are really just components that can run on both client and the server. (A button could be an island, but why?)

kerimhudson:
After reading the discussion at #2369, this PR introduces a small addition to the docs which highlights how you can use server components within islands.

This, coupled with the existing docs, should hopefully give people enough of an example to understand that islands provide the interactivity layer, and that server components within them can be hydrated and made interactive too.

Closes #2369


docs/latest/concepts/islands.md
```diff
@@ -78,6 +78,32 @@ export default function Home() {
...
+  children: ComponentChildren;
+}
+
+export default function MyIsland({ children }: Props) {
+  const count = useSignal(0);
+
+  return (
+    <Card>
+      Counter is at {count}.{" "}
+      <Button onClick={() => (count.value += 1)}>+</Button>
```
marvinhagemeister, line 102: I file like we should add a sentence that this click handler will only fire when this component is inside `./islands` or referenced inside an island. I'm worried that the component name here suggests that this component would be treated as an island.

kerimhudson: Thanks @marvinhagemeister 

I hoped the positioning in the docs would have helped clarify that as I didn't want to be overly verbose, but I've added a sentence as you mention to make sure it's explicitly known. Let me know what you think.

marvinhagemeister (approved):
Sweet, thanks for adding this to our docs 👍

PR #2449, opened 2024-05-22
major: import Fresh 2.0 code

marvinhagemeister:
🎉


marvinhagemeister:
I think it's time to merge to `main`. There is a little bit of development work left to do, but having it in `main` makes it easier to iterate and work on the docs at the same time.

PR #2451, opened 2024-05-23
Chore: styling and verbiage cleanup

josh-collinsworth:
- Better responsive styling of header, logo, nav menu and docs sidebar
- Specify UTC on homepage (since we can't localize server-side)
- Minor copy updates


marvinhagemeister (approved):
Sweet 🎉

PR #2458, opened 2024-05-24
fix: init script is broken

lucacasonato:


init/src/init.ts
```diff
@@ -350,17 +350,16 @@ html {
...
 
-  const MAIN_TSX = `import { App, staticFiles, fsRoutes } from "@fresh/core";
+  const MAIN_TS = `import { App, staticFiles, fsRoutes } from "@fresh/core";
 import { State } from "./utils.ts";
 
-export const app = new App<State>()
-  .use(staticFiles())
-  .get("/api/:joke", () => new Response("Hello World"))
-  .get("/greet/:name", (ctx) => {
-    return ctx.render(<h1>Hello {ctx.params.name}</h1>);
```
marvinhagemeister, line 360: No more JSX rendering?

lucacasonato: It made the file a tsx file, which was annoying. Most users will not be doing jsx in their main file

marvinhagemeister (approved):

init/src/init.ts
marvinhagemeister: ok shots fired.

PR #2468, opened 2024-05-25
docs: Fix name of JSR update tool

frou:


marvinhagemeister (approved):
Whoops, good catch. Thanks for fixing this 👍

PR #2472, opened 2024-05-27
tests: don't disable all the sanitizers

lucacasonato:
The tests also pass with them enabled for me.


src/build_cache.ts
```diff
@@ -40,6 +40,7 @@ export class ProdBuildCache implements BuildCache {
       hasSnapshot = true;
       setBuildId(snapshot.buildId);
 
+      console.log(snapshot);
```
marvinhagemeister:
```suggestion
```

PR #2475, opened 2024-05-27
fix: show code frame on error

marvinhagemeister:


src/plugins/fs_routes/mod.ts
```diff
@@ -325,6 +334,27 @@ function errorMiddleware<State>(
       return await ctx.next();
     } catch (err) {
       ctx.error = err;
+
+      if (err instanceof HttpError) {
+        if (err.status >= 500) {
+          if (DENO_DEPLOYMENT_ID === undefined) {
```
lucacasonato, line 340: I feel like this check should happen based on `ctx.config.mode`?

bartlomieju:
The codebase has been significantly restructured since this PR. The `fs_routes` plugin, `render_middleware.ts`, and the routing/error handling systems have all been redesigned. The code frame file was never moved as proposed here — it remains in `dev/middlewares/error_overlay/`.

Closing as stale.

PR #2477, opened 2024-05-29
fix: remove @fresh/core/client export

lucacasonato:
These files are now referenced with https URLs.


marvinhagemeister (approved):
Good idea!

PR #2494, opened 2024-06-04
www: Add aria-hidden to icons and illustrations

timharek:
Added `aria-hidden="true"` to all decorative icons and illustrations.

If there are any questions, suggestions or anything, please let me know :)


marvinhagemeister (approved):
Sweet, thanks for addressing this 👍

PR #2417, opened 2024-04-28
docs: Update documentation for active link with Tailwind

bastianplsfix:
**Proposed change:**
- Document styling for ﻿aria-current attribute separately for Tailwind and Twind.

**Disclaimer:**
I am new to open-source contributions and are unfamiliar with the process for submitting a pull request. I have found some more areas of improvement in the Fresh documentation. Should this PR not be up-to-standards on how to contribute, please guide me in the right direction. I have a few more suggestions on issues I found while developing a new Fresh app.

Any guidance would be helpful, either reach out to me on GitHub, Discord (bastianplsfix) or X (bastianplsfix).

Best,
Sebastian 🦕

PR #2505, opened 2024-06-06
Change active link tailwind syntax

### LINKED ISSUE #2123: `[aria-current]` not working with Tailwind CSS
I've just updated to Fresh 1.6 and migrated to Tailwind. I followed the docs after #2019 to style active links with `[aria-current]` but it doesn't work while `aria-[current='page']` works well.

lishaduck:
I haven't independently verified that this works, but it seemed right.

Closes #2123

### DIFF
--- a/docs/latest/examples/active-links.md
+++ b/docs/latest/examples/active-links.md
@@ -41,7 +41,7 @@ Twind. For Tailwind, use the syntax:
 ```tsx
 function Menu() {
   return (
-    <a href="/foo" class="aria-[current]:text-green-600">
+    <a href="/foo" class="aria-[current='page']:text-green-600">
       Link to some page
     </a>
   );


docs/latest/examples/active-links.md
```diff
@@ -41,7 +41,7 @@ Twind. For Tailwind, use the syntax:
 ```tsx
 function Menu() {
   return (
-    <a href="/foo" class="aria-[current]:text-green-600">
+    <a href="/foo" class="aria-[current='page']:text-green-600">
```
marvinhagemeister, line 44: The selector here is already correct. Looks like it was already resolved by https://github.com/denoland/fresh/pull/2417 . Previously, we didn't have the distinction between twind and tailwindcss but that PR added that.

lishaduck: Ah, ok then!

I was thinking I'd run into this before, but now that I think about it, I think it was the aria vs data distinction, not the value of the aria.

PR #2516, opened 2024-06-16
docs: Fix incorrect enctype info on Forms page

### LINKED ISSUE #2515: Forms page in docs is incorrect about multipart/form-data
There are a few issues with the accuracy of the [Forms](https://fresh.deno.dev/docs/concepts/forms) page, especially regarding `multipart/form-data`.

> Forms can either submit as a `GET` request with URL search parameter encoding, or as a `POST` request with `multipart/form-data`.

This is not true. A form will submit a `POST` request (when its `method` attribute is set to `post`) with an `application/x-www-form-urlencoded` body by default. To submit a `multipart/form-data` request, its `enctype` attribute must be set to `multipart/form-data`.

The example given in the section [POST request with `multipart/form-data`](https://fresh.deno.dev/docs/concepts/forms#post-request-with-multipartform-data) does not submit a `multipart/form-data` request. It submits a `application/x-www-form-urlencoded` request.

I can understand the confusion considering that MDN describes the `FormData` interface as using "the same format a form would use if the encoding type were set to `"multipart/form-data"`, but that just refers to what happens if you provide a `FormData` as the body of a request made with `fetch`. When Deno provides a request to a route handler, its `formData()` function will resolve to a `FormData` that is [built from the request's body](https://fetch.spec.whatwg.org/#dom-body-formdata), regardless if it's an `application/x-www-form-urlencoded` or a `multipart/form-data` body.

The second example, [Handling file uploads](https://fresh.deno.dev/docs/concepts/forms#handling-file-uploads), correctly sets the form's `enctype` to `multipart/form-data` since that is required for file uploads.

jordanbtucker:
Fixes #2515

This PR updates the Forms page in the docs with accurate information about a form's `enctype` attribute.

### DIFF
--- a/docs/latest/concepts/forms.md
+++ b/docs/latest/concepts/forms.md
@@ -12,11 +12,12 @@ In the browser, a `<form>` submit will send an HTML action (usually `GET` or
 
 ## POST request with `multipart/form-data`
 
-Forms can either submit as a `GET` request with URL search parameter encoding,
-or as a `POST` request with `multipart/form-data`.
+Forms typically submit as a `GET` request with data encoded in the URL's search
+parameters, or as a `POST` request with either a
+`application/x-www-form-urlencoded` or `multipart/form-data` body.
 
-This example demonstrates how to handle `multipart/form-data` `<form>`
-submissions:
+This example demonstrates how to handle `application/x-www-form-urlencoded`
+`<form>` submissions:
 
 ```tsx routes/subscribe.tsx
 import { Handlers } from "$fresh/server.ts";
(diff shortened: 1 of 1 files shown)


docs/latest/concepts/forms.md
```diff
@@ -12,11 +12,12 @@ In the browser, a `<form>` submit will send an HTML action (usually `GET` or
 
 ## POST request with `multipart/form-data`
```
marvinhagemeister: Nit: We should probably update the heading too

jordanbtucker: Fixed!

marvinhagemeister (approved):
Thanks!

PR #2523, opened 2024-06-18
fix: enhanced config.ts->{parseRootPath}

Iuriiiii:
* Added tests for Windows.
* Enhanced `parseRootPath` function to work with any extension and better windows compatibility.

Tested on Windows 11 and WSL.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -40,18 +40,22 @@ export interface ResolvedFreshConfig {
 }
 
 export function parseRootPath(root: string, cwd: string): string {
-  if (root.startsWith("file://")) {
+  if (root === "\\" || root === "/") {
+    return cwd;
+  } else if (root.startsWith("file://")) {
     root = path.fromFileUrl(root);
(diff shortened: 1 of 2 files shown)


Iuriiiii:
Something is blocking this pr or just is no time to review this?

bartlomieju:
Closing as stale. There are also a few issues with the approach:

1. **Overly broad extension stripping** — replacing the known-extension allowlist (`.ts`, `.tsx`, `.js`, `.jsx`, `.mjs`) with a generic `if (ext)` check means directory names containing dots (e.g. `/app/v2.0`, `/foo/bar.d`) would incorrectly be treated as files and stripped to the parent directory.

2. **Imports private module** — `$std/path/_os.ts` is an internal/private module (underscore prefix) that could break on deno_std updates.

3. **Removed test coverage** — the `file:///foo/bar.ts` test case from the original was dropped.

A safer approach would be to keep the extension allowlist and add Windows path normalization on top. Thanks for the contribution!

PR #2526, opened 2024-06-20
doc: fix the link to sever configuration

mugi0096:
I found a dead link on the "Writing tests" page.


marvinhagemeister (approved):
Good catch!

PR #2529, opened 2024-06-22
fix: ensure deno task check passes in initialized projects

deer:
There were two issues with initialized projects:
1. not every file was formatted properly
2. linting wasn't passing when using `verbatim-module-syntax`. This isn't enabled right now, due to it not having the `recommended` tag, but it seems like a good idea. Actually initializing a project with this can be done in a separate PR.


marvinhagemeister (approved):
Sweet, thanks!

PR #2534, opened 2024-06-23
feat: add example of define.middleware to init

deer:
The documentation for `MiddlewareFn` is out of date, because it references a nonexistent `defineMiddleware` function. It's been updated to reference the correct `define.middleware` function.

Additionally, I included an example of using this in the init project. I also complicated the usage of `app.get` so that the function actually uses the context. To make it clear to users of fresh 1 that this can be equivalent to a file-based-route, I've included such a route as well.

The two new examples in `maini.ts` should make it much more clear to fresh 1 users how the new system can work.


marvinhagemeister (approved):
Sweet thanks!

PR #2535, opened 2024-06-23
chore: cleanup unused exported interfaces

deer:


src/types.ts
```diff
@@ -28,12 +28,6 @@ export interface RouteConfig {
   skipAppWrapper?: boolean;
 }
 
-export interface SignalLike<T> {
-  value: T;
-  peek(): T;
-  subscribe(fn: (value: T) => void): () => void;
-}
```
deer, line 35: This is used later in the file, in some commented code. But I guess it can be re-added once the todo is resolved.

marvinhagemeister: Fair point, opened https://github.com/denoland/fresh/issues/2537 to keep track of that.

marvinhagemeister (approved):
Sweet thanks!

PR #2538, opened 2024-06-23
feat: re-enable lockfile in init

### LINKED ISSUE #2533: feat: re-enable lockfile by default
From https://github.com/denoland/fresh/discussions/2532

We should re-enable the lockfile now that we don't depend as much on `esm.sh` anymore and are moving to JSR.

deer:
closes https://github.com/denoland/fresh/issues/2533

Or does this need a test to assert the file exists?

### DIFF
--- a/init/src/init.ts
+++ b/init/src/init.ts
@@ -531,6 +531,7 @@ if (Deno.args.includes("build")) {
         tags: ["fresh", "recommended"],
       },
     },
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks! Nah, no need for a test for this. It's deno's job to ensure that the lockfile feature works.

PR #2542, opened 2024-06-24
feat: add default GET handler

### LINKED ISSUE #2530: Always add default GET handler
From discord: https://discord.com/channels/684898665143206084/991511118524715139/1254411831041196133

> I've just been reading through some of the decisions made for Fresh 2.0. There's a lot of progress, so great work! I have one question regarding the use of the new define helpers. If a route only needs a POST handler, will we still need to define a GET handler? I did like just exporting a default page component for basic pages and then defining a POST handler. It seems (maybe I'm wrong) in the early alpha I need to define a default GET handler too (one that just returns the page) if I define a POST handler.

We should definitely add a default handler. Was an oversight on my part, there is no reason for it to not be there.

deer:
closes https://github.com/denoland/fresh/issues/2530

### DIFF
--- a/src/plugins/fs_routes/mod.ts
+++ b/src/plugins/fs_routes/mod.ts
@@ -284,13 +284,19 @@ export async function fsRoutes<State>(
       }
     }
 
-    if (routeMod.component !== null) {
-      components.push(routeMod.component);
-    }
-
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thank you!

PR #2565, opened 2024-07-02
docs: update markdown tutorial to use JSR

Mrashes:
Small cleanup of how we recommend importing the gfm package (moved to jsr, old post still works but could be confusing to new users)


marvinhagemeister (approved):
Good catch! Thanks for fixing. this.

Issue #2546, opened 2024-06-26
Dead link


finalclass:
https://fresh.deno.dev/docs/concepts/plugins
![image](https://github.com/denoland/fresh/assets/1343930/73afa03a-0e7b-40a2-b902-210816f669ac)
![image](https://github.com/denoland/fresh/assets/1343930/4616a37d-c974-4ad2-9059-cd0447c69d61)

marvinhagemeister:
We should inline a basic example for Fresh 1.x plugins in the docs instead of referring to a file in the repository.

Mrashes:
Taking a crack at this, PR should be coming in the hour

marvinhagemeister:
Resolved by https://github.com/denoland/fresh/pull/2575

PR #2575, opened 2024-07-03
docs: inline plugin example

Mrashes:
Tackling - https://github.com/denoland/fresh/issues/2546

Basically just inlined the old reference file and removed the islands and app builder to make it a bit more streamlined.

Could make the example more dry but I'm not working from my normal dev environment.


marvinhagemeister (approved):
Sweet, thanks!

PR #2581, opened 2024-07-06
docs: fix typo

behnam-oneschema:
Looks like the intention was to refer to one of the folders in the given project structure listing.


marvinhagemeister (approved):
Good catch, this is better 👍

PR #2584, opened 2024-07-07
fix: support numeric keys when rendering

deer:
I'm trying to migrate one of my projects which publishes a plugin, and this was blocking me.

Looks like key is typed as `export type Key = string | number | any;`, so we might need to make more changes here in the future.


marvinhagemeister (approved):
Good catch!

Issue #2587, opened 2024-07-09
bug: fresh islands are broken with preact 1.22.1


JOTSR:
# Fresh island rendering is broken in preact 1.22.1

## Bug

```
children.js:123 Uncaught (in promise) TypeError: n2.contains is not a function
    at S (children.js:123:16)
    at O (index.js:246:4)
    at S (children.js:85:3)
    at O (index.js:246:4)
    at B (render.js:37:2)
    at _render (main.ts:109:7)
```
![image](https://github.com/denoland/fresh/assets/58234376/feab1653-9b3b-481f-810d-53801dd7702c)

It seems that fresh root fragments of islands are missing `contains` method.

## Bug reason

The PR [replace isConnected with parentDom.contains #4421](https://github.com/preactjs/preact/pull/4421/files) breaks fresh island rendering with new `contains` method.

marvinhagemeister:
Resolved in #2596

PR #2596, opened 2024-07-11
fix: preact 10.22.1 contains error

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/2587


src/runtime/entrypoints/main.ts
```diff
@@ -70,7 +70,20 @@ function createRootFragment(
     removeChild(child: Node) {
       parent.removeChild(child);
     },
+    contains(node: Node | null): boolean {
```
JoviDeCroock, line 73: Sorry about that, will try to think more about this in the future 😅

JoviDeCroock (approved):

PR #2628, opened 2024-08-27
ci: run with DENO_FUTURE=1

bartlomieju:
Add another CI step that runs with `DENO_FUTURE=1` env var
to future-proof for Deno 2 changes.


iuioiua:
I think this can be closed.

PR #2642, opened 2024-09-05
[fresh 2.0] fix: decode static file paths

### LINKED ISSUE #2613: In fresh 2.0 static file path components aren't decoded
> This is only a problem in Fresh 2.0. In version 1.x everything works as expected

## Summary

The static file middleware is not functioning correctly because it fails to decode path components. This results in problems when accessing files and directories with names containing special characters or encoded elements.

Examples of problematic file and directory names include:

- `C#.svg`
- `西安.avif`
- `제주도`

## Steps to Reproduce

1. Place files with these names in the static file directory.
2. Configure the static file middleware to serve files from this directory.
3. Attempt to access these files via their URLs.
  - `/C%23.svg`
  - `/%E8%A5%BF%E5%AE%89.avif`
  - `/%EC%A0%9C%EC%A3%BC%EB%8F%84/sth.txt`

## Expected behaviour

The middleware should decode the URL-encoded path components and correctly serve the requested files:

- `/static/C#.svg`
- `/static/西安.avif`
- `/static/제주도/sth.txt`

## Actual Behavior

The middleware does not decode the path components, resulting in 404 errors.

## Environment

- @fresh/core@2.0.0-alpha.19
- deno 1.45.5
- macos 14.6 (arm64)

felix-schindler:
Fixes #2613

### DIFF
--- a/src/middlewares/static_files.ts
+++ b/src/middlewares/static_files.ts
@@ -17,7 +17,7 @@ export function staticFiles<T>(): MiddlewareFn<T> {
     const { req, url, config } = ctx;
     const buildCache = getBuildCache(ctx);
 
-    let pathname = url.pathname;
+    let pathname = decodeURIComponent(url.pathname);
     if (config.basePath) {
       pathname = pathname !== config.basePath
         ? pathname.slice(config.basePath.length)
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #2646, opened 2024-09-08
feat(docs): dynamic css use tailwind

### LINKED ISSUE #2626: Tailwind classes in Components
Hello, 
changed everything from twind to tailwind.

Maybe it could be documented that currently all classes in components need to be named explicitly.

Let's say I have an encoded link component for mailto: or tel: links and it has
```TS
interface {
  type?: 'mail'|'phone'
}
```
This will _not_ work:
```TS
  <a class={`redaktor-link-${type}`}></a>
```
but this would
```html
  <a class={type === 'phone' ? `redaktor-link-phone` : `redaktor-link-mail`}></a>
```

Not sure if this could be done with the reflection API or if there are better ideas :)

DevJoaoLopes:
close #2626


marvinhagemeister (approved):
Good idea to add it to the docs. Thanks for the PR!

PR #2651, opened 2024-09-16
Adding showcase for wtfiscongressdoingnow.us

gibbygano:
Just adding in my Deno Fresh project. "WTF Is Congress Doing Now?" is a non-partisan up-to-date source of bills introduced congress. Information includes bill metadata, sponsors, links to affected codes and links to PDF copies of bills.


marvinhagemeister (approved):
Thanks!

PR #2656, opened 2024-09-20
Add `dom.iterable` to template `compilerOptions.lib`

timreichen:
Ref: https://github.com/denoland/deno/issues/25741


marvinhagemeister (approved):
Sweet, thanks!

Issue #2660, opened 2024-09-27
Option to let Fresh respect `X-Forwarded-*` headers?


dahlia:
I'd like to run Fresh behind a reverse proxy for both development and production. However, currently Fresh does not respect `X-Forwarded-*` headers, so when I try to get `ctx.url` it returns an `http://` URL instead of `https://` URL even though an `X-Forwarded-Proto: https` header is provided.

So, it would be nice if `App.listen()` has an option to let Fresh respect `X-Forwarded-Proto` and `X-Forwarded-Host` headers. Would it be a good idea? If so, I'd like to send a pull request.

bartlomieju:
Still not implemented — `ctx.url` is constructed directly from `req.url` (line 402 in `app.ts`) with no `X-Forwarded-*` header processing.

**Workaround**: Use a middleware to override `ctx.url` based on the forwarded headers:

```ts
app.use((ctx) => {
  const proto = ctx.req.headers.get("X-Forwarded-Proto");
  const host = ctx.req.headers.get("X-Forwarded-Host");
  if (proto) ctx.url.protocol = proto + ":";
  if (host) ctx.url.host = host;
  return ctx.next();
});
```

This should be placed before any other middleware that reads `ctx.url`.

Leaving open as a feature request — a built-in `trustProxy` option on `App` or `app.listen()` would be the proper solution.

PR #2663, opened 2024-09-30
Add `behindProxy` option to `App.listen()`

dahlia:
Addresses <https://github.com/denoland/fresh/issues/2660>.

The `behindProxy` option lets Fresh respect `X-Forwarded-Proto`, `X-Forwarded-Host`, `X-Forwarded-Port`, and `X-Forwarded-For` headers (if any).  This is useful when Fresh is behind a reverse proxy like nginx or Caddy.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -35,15 +35,24 @@ const DEFAULT_NOT_ALLOWED_METHOD = () => {
   throw new HttpError(405);
 };
 
+export interface HandlerOptions {
+  /**
+   * Whether to respect `X-Forwarded-Proto`, `X-Forwarded-Host`,
+   * `X-Forwarded-Port`, and `X-Forwarded-For` headers.  This is useful when
+   * Fresh is behind a reverse proxy like Nginx or Caddy.
+   * @default {false}
+   */
+  behindProxy?: boolean;
+}
+
 export type ListenOptions =
   & Partial<
     Deno.ServeTcpOptions & Deno.TlsCertifiedKeyPem
   >
+  & HandlerOptions
   & {
     remoteAddress?: string;
   };
-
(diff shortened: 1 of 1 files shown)


marvinhagemeister (commented):
Thanks for making a PR. I feel like this is much better implemented as a middleware instead of a configuration option as it's not something everyone needs. With Fresh 2 the aim is to keep the config as minimal as possible.

dahlia:
@marvinhagemeister Does Fresh's middleware support replacing request objects? At first glance, it doesn't seem to support that.

marvinhagemeister:
Yes, to replace the `Request` object do this:

```ts
function proxyMiddleware(ctx) {
  const req =  new Request()
  
  // Replace current request
  ctx.req = req;
  ctx.url = new URL(req.url)
  
  return ctx.next()
}
```

dahlia:
Apparently [`FreshContext.req` is readonly](https://github.com/denoland/fresh/blob/e0f65319cd48753f5b9b4784f996a0eaf7e6ebff/src/context.ts#L36).

dahlia:
@marvinhagemeister Apparently [`FreshContext.req` is readonly](https://github.com/denoland/fresh/blob/e0f65319cd48753f5b9b4784f996a0eaf7e6ebff/src/context.ts#L36).  Should I suppress the TypeScript error and set `ctx.req` by force?

marvinhagemeister:
Let's remove the `readonly` in `FreshContext`.

dahlia:
> Let's remove the `readonly` in `FreshContext`.

@marvinhagemeister: Just opened a [separate pull request](https://github.com/denoland/fresh/pull/2751).

Issue #2658, opened 2024-09-24
`f-client-nav={false}` fails in islands


adamgreg:
In the Partials doc it says "Passing f-client-nav={false} disables client side navigation for all elements below the current node.". That gives the attribute value `"false"` in SSR (I guess due to special handling in `server/rendering/preact_hooks.ts` ), but if it is done in an island, the attribute is lost after hydration.

Currently (with Fresh 1.x) within an island the `f-client-nav` prop has to be set to `"false"` instead, which angers Typescript.

Could a boolean `false` be handled in a client-side Preact hook, or failing that could the type be extended to accept `"false"`?

marvinhagemeister:
Resolved by #2678

PR #2678, opened 2024-10-04
fix: `f-client-nav={false}` not working inside island

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/2658


adamgreg:
Thanks a lot Marvin! It can't be easy finding time for these at the moment

Issue #2699, opened 2024-10-10
Docs: blockquote text not parsed as markdown


mackieburgess:
👋

The text inside blockquotes isn't being parsed as markdown: this means that – in mild instances – text which should be rendered as a codespan is instead displaying with backticks, and – in severe instances – codespans containing HTML are rendering as actual HTML, instead of being escaped (examples of both below).

Will put in a PR imminently :)

<img width="731" alt="Screenshot 2024-10-10 at 13 41 24" src="https://github.com/user-attachments/assets/1b4cf993-c68b-4fac-96aa-45d2a149d1f8">

<img width="725" alt="Screenshot 2024-10-10 at 13 41 39" src="https://github.com/user-attachments/assets/92ce6736-7810-45c0-9ba4-0c68803feb01">

monooleate:
After your recent commit, the info|warn|tip tags are still not functioning as expected. It seems that with your changes in markdown.ts, these tags will work if you modify the ADMISSION_REG constant by removing the "<" p ">" tag, like so: /^\[(info|warn|tip)\]:\s/. This adjustment should resolve the issue.

mackieburgess:
@monooleate ah yes, I can put this in as a separate PR.

marvinhagemeister:
Resolved with #2702

PR #2700, opened 2024-10-10
fix docs: blockquote text not parsed as markdown

mackieburgess:
See issue #2699.

Blockquote text now renders correctly as markdown.

Screenshot 2024-10-10 at 13 42 26

Screenshot 2024-10-10 at 13 42 47


marvinhagemeister (approved):
Sweet, thanks for fixing this 👍

PR #2704, opened 2024-10-11
fix: ensure file paths in snapshot.json uses /

### LINKED ISSUE #2703: [Fresh-v2] Building assets with subfolders fails on windows
When building Fresh v2 on Windows, using subfolders for `assets` seems to break when running the app in production after building. The reason seems to be inconsistent paths with `\\`:

```json
  "staticFiles": {
    "/images\\logo.svg": {
      "hash": "1e0b7aa79648883158742ec4f064980898c4c8feff0c3300b8c7ea4485ef1354",
      "generated": false
    },
```

Trying to fetch `/images/logo.svg` just gives a HTTP 404 error.

## Reproduction

1. Initialize a Fresh project
    ```bash
    deno run -Ar jsr:@fresh/init@2.0.0-alpha.22
    ```
2. Move the `assets/logo.svg` to `assets/images/svg.logo`
3. Update `routes/index.tsx` with updated image path
4. Build Fresh with `deno task build`
5. Start prod server with `deno task start`
6. Link to image is broken

csvn:
Avoids `\\\\` paths ending up in `snapshot.json` when using `deno task build` on Windows. Backslashes do not which pathnames in URLs, so these assets are not returned.

Closes #2703

### DIFF
--- a/src/dev/dev_build_cache.ts
+++ b/src/dev/dev_build_cache.ts
@@ -1,5 +1,6 @@
 import type { BuildCache, StaticFile } from "../build_cache.ts";
 import * as path from "@std/path";
+import { SEPARATOR as WINDOWS_SEPARATOR } from "@std/path/windows/constants";
 import type { ResolvedFreshConfig } from "../config.ts";
 import type { BuildSnapshot } from "../build_cache.ts";
 import { encodeHex } from "@std/encoding/hex";
@@ -174,7 +175,7 @@ export class DiskBuildCache implements DevBuildCache {
 
   addUnprocessedFile(pathname: string): void {
     this.#unprocessedFiles.set(
-      pathname,
+      pathname.replaceAll(WINDOWS_SEPARATOR, "/"),
       path.join(this.config.staticDir, pathname),
     );
   }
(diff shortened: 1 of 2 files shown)


src/dev/dev_build_cache.ts
```diff
@@ -174,7 +175,7 @@ export class DiskBuildCache implements DevBuildCache {
 
   addUnprocessedFile(pathname: string): void {
     this.#unprocessedFiles.set(
-      pathname,
+      pathname.replaceAll(WINDOWS_SEPARATOR, "/"),
```
csvn, line 178: I don't know if there is a "better" way to normalize file system paths to URL-like paths. This seems to work at least, since manually replacing `\\` with `/` in my `snapshot.json` locally fixes the issue.

csvn:
I think perhaps the `.addProcessedFile()` will produce the same error, but I only fixed this for the "Un"-processed files.

marvinhagemeister (approved):
Sweet, great catch! Love that you added a test as well!

src/dev/dev_build_cache.ts
marvinhagemeister: Seems fine to me.

PR #2712, opened 2024-10-15
www: Fix block quotes not rendering links

efekrskl:
- Fixes a bug where block quotes were not able to properly render URLs in docs (e.g. Info section at https://fresh.deno.dev/docs/getting-started/create-a-route)
- Replaces existing JSX docs url with the suggested, up-to-date docs url

image


marvinhagemeister (approved):
Good catch, thanks for fixing this!

PR #2717, opened 2024-10-17
chore: update 1.x versions

felix-schindler:
Documentation and update script are both still at 1.7.2 even tho [1.7.3](https://github.com/denoland/fresh/releases/tag/1.7.3) has been released to make fresh compatible with deno v2


marvinhagemeister (approved):
Whoops, good catch! Thanks for fixing this.

PR #2725, opened 2024-10-20
docs: fix broken links

marsbird:
This PR:
1. fixes two broken links in the Getting Started documentation
2. updates the link to React's JSX docs to the lastest documentation; (the existing link leads to documentation that is no longer updated)

Screenshot of the broken links - even though they work correctly when viewing the markdown files, they break when rendered in the browser because they're both inside an info block, which breaks the internal link.
Screenshot 2024-10-19 at 10 37 56 PM

Screenshot of the warning seen at the existing JSX link https://react.dev/learn/writing-markup-with-jsx
[
Screenshot 2024-10-19 at 10 41 28 PM
](url)

### DIFF
--- a/docs/latest/getting-started/create-a-route.md
+++ b/docs/latest/getting-started/create-a-route.md
@@ -29,10 +29,11 @@ this, one needs to create a new `routes/about.tsx` file. In this file, we can
 declare a component that should be rendered every time a user visits the page.
 This is done with JSX.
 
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Thanks for your PR. Looks like this was caused by a markdown parsing error. That error was addressed in https://github.com/denoland/fresh/pull/2712and made the links work. Seeing that it's already fixed I'm going to close this one.

PR #2738, opened 2024-10-27
fix(jsonify): parse returning undefined instead of the number

### LINKED ISSUE #2736: jsonify parse -1 into undefined
both in **main** and **jsr:@fresh/core@2.0.0-alpha.23**

here a testing case for jsonify stringify_test.ts
```
Deno.test("stringify - Map with", () => {
  expect(stringify({ foo: -1 }))
    .toEqual(
      '[{"foo":1},-1]',
    );
});
```

while testing in parse_test.ts, it failed expecting `{ foo: undefined }`
```
Deno.test("parse - -1 as undefined", () => {
  expect(parse(`[{"foo":1},-1]`)).toEqual({ foo: -1 });
});
```
//

DevJoaoLopes:
close #2736


src/jsonify/parse.ts
```diff
@@ -149,8 +153,8 @@ function unpack(
     hydrated[idx] = actual;
 
     const keys = Object.keys(current);
-    for (let i = 0; i < keys.length; i++) {
-      const key = keys[i];
```
marvinhagemeister: This change isn't necessary to fix the issue. As a side not in most browsers using a standard for loop is still much faster on arrays than a `for-of` loop.

src/jsonify/parse.ts
```diff
@@ -30,9 +30,13 @@ function unpack(
   custom: CustomParser | undefined,
 ): void {
   if (idx in hydrated) return;
-
+  
   const current = arr[idx];
   if (typeof current === "number") {
+    if(idx !== 0) {
```
marvinhagemeister: Running `deno fmt` should resolve the formatting 👍

marvinhagemeister (commented):
Great catch, thanks for making a PR.

src/jsonify/parse.ts
DevJoaoLopes: @marvinhagemeister Perfectly, I will remove this part from the implementation :smile:

src/jsonify/parse.ts
DevJoaoLopes: Thanks for the comment, I'll do it 👍

DevJoaoLopes:
Taking advantage, I would like to know if it would be possible to join the list of contributors to the project. I've been using deno and fresh in my daily life and I've been trying more and more to contribute here. What do you think @marvinhagemeister ?

DevJoaoLopes:
@marvinhagemeister Can we merge this PR?

DevJoaoLopes:
@marvinhagemeister @lucacasonato  How can we proceed with this PR :smile: ?

DevJoaoLopes:
@marvinhagemeister @lucacasonato Can you take a look at this MR if it still makes sense please?

marvinhagemeister:
Apologies, looks like this was addressed in https://github.com/denoland/fresh/pull/2790

PR #2763, opened 2024-11-14
fix: default get route middleware

lucacasonato:


thegaryroberts:
@lucacasonato you were usurped by https://github.com/denoland/fresh/pull/2807

PR #2767, opened 2024-11-18
Update running-locally.md

prjctimg:
Fixed an anomaly with the tense. It read '...change' instead of 'changing'


marvinhagemeister (approved):
Thanks!

iuioiua (approved):

PR #2786, opened 2024-12-09
feat: add open telemetry instrumentation

lucacasonato:
This commit adds some useful automatic instrumentation for when users are using `--unstable-otel`:

- Fresh renames the span for incoming HTTP requests to `$METHOD $ROUTE` and the `http.route` attribute is added.
- Fresh creates a span around FS route handlers, showing whether they returned a response or data.
- Fresh creates a span around async components, showing how long they took to execute.
- Fresh adds a span around preact rendering, showing how long rendering takes.
- Fresh adds a span around static file serving, showing how long it took and whether there were requests without a cache bust key.

We can improve this in the future with more telemetry, such as metrics.


deno.lock
```diff
@@ -1,75 +1,88 @@
 {
```
lucacasonato, line 1: My previous PR accidentally removed all test dependencies from the lock file :sweat:

marvinhagemeister (approved):
LGTM

PR #2792, opened 2024-12-16
fix(router): exclude route groups in scoring for sorting

marvinhagemeister:
This PR fixes an issue in the route sorting logic when comparing a route segment with a group vs a segment without one.

- `/(app)/[org]/[app]/index.ts`
- `/auth/login.ts`

When determing the order of these two routes we need to skip over the `(app)` group to the next non-group segment. This ensures that we are only scoring actual segments with each other and non-existing ones.


lucacasonato (approved):
LGTM, thanks

PR #2798, opened 2024-12-21
fix: prevent 404 for routes consisting solely of an optional parameter

ghalle:
## Summary

- Fixes an issue where a route consisting solely of an optional parameter (e.g. `/[[name]].tsx`) would 404 on the root path `/`
- The generated pattern `{/:name}?` didn't match `/` because the leading `/` was inside the optional group. This PR changes it to `/{:name}?` which correctly matches both `/` and `/foo`
- Routes with non-optional segments (e.g. `docs/[[version]]/index.ts`) are unaffected
- Adds docs table entry for the standalone optional param pattern

## Details

Prior to this PR, creating a route like `/[[name]].tsx` would result in a URL pattern like `{/:name}?` which would work for `/foo` but would fail for `/`.

After this PR, the pattern is now `/{:name}?` which matches correctly for both `/` and `/foo`. Routes containing more segments like `/[[name]]/bar.tsx` were unaffected and retain the same pattern as before.

## Test plan

- [x] Unit test added for `pathToPattern("(group)/[[name]]/(group2)/index")` → `/{:name}?`
- [x] Existing tests pass
- [x] Docs updated with new routing table entry


bartlomieju (commented):
Nice fix! The approach of post-processing rather than changing the `{/` generation logic makes sense since you don't know at parse time whether non-optional segments will follow.

One suggestion: could you add a direct test for the simplest case (`[[name]]` without groups)? The new test covers the groups+optional combo, but a test for the basic reported bug would make the coverage clearer:

```ts
expect(pathToPattern("[[name]]")).toEqual("/{:name}?");
```

bartlomieju (approved):
Thanks

PR #2804, opened 2025-01-04
Update adding-interactivity.md

2p4b:
Declare interval variable timer before use.


bartlomieju:
Closing — the original `const timer = setInterval(...)` is correct. The `timer` variable is accessible inside the callback via closure (the callback captures the variable binding, not the value at creation time). No change needed here. Thanks for the contribution!

PR #2807, opened 2025-01-08
fix: middlewares not being used with default GET handlers

pcardune:
I noticed a small bug with the default GET handlers where middlewares were not being run. This pull requests fixes the bug and adds some unit tests for this behavior.


marvinhagemeister (approved):
Good catch! Thanks for fixing this 👍

PR #2811, opened 2025-01-16
fix: route middleware order with groups

marvinhagemeister:
This is a follow-up fix to https://github.com/denoland/fresh/pull/2792

The original fix skipped over route groups which lead to them being mixed with non-grouped routes. This in turn confused the middleware processing part when building the final routes.


lucacasonato (approved):
Thanks!

PR #2833, opened 2025-03-04
fix lint errors

timreichen:
This PR fixes current lint errors thrown by `deno lint`.

PR #2836, opened 2025-03-05
fix wrong button type in submit test cases

timreichen:
https://github.com/denoland/fresh/pull/2833 breaks some tests where the button must be of type `submit` rather than `button`. This PR fixes this.


marvinhagemeister (approved):
Thanks!

PR #2842, opened 2025-03-12
chore: fix init tests

marvinhagemeister:


iuioiua:
This is no longer needed

PR #2849, opened 2025-04-06
Fix: Ensure the tasks field for check/fmt/lint can correctly find ts/tsx files(showcase case) & Update showcase.json

SisyphusZheng:
update freshpress and my site, all made by deno and fresh framwork


SisyphusZheng (commented):
forget to add ,

SisyphusZheng:
@marvinhagemeister If u have time plz check this change thx again 🙇‍

SisyphusZheng:
THE FIX IS NOT SUITABLE, SEEM LIKE REPLACE THE ORIGINAL DENO TASKS, BUT I WILL FIND OTHER WAY TO SOLVE THE CONFFLICT OF THE CHECK PROBLEM WITH SHOWCASE

PR #2852, opened 2025-04-14
fix(init): define working directory in `check` task for `init` script

iuioiua:
~~This fixes CI for current PRs until https://github.com/denoland/deno/pull/28655 is added to Deno 2.3.0.~~

This fixes CI until https://github.com/denoland/deno/issues/28923 is fixed. Then, we can use `Deno.makeTempDir()` and not have to use `tmp_*/`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vendor/
 node_modules/
 .docs/
 .DS_Store
+tmp_*
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,5 @@
 _fresh/
-tmp/
 vendor/
 node_modules/
 .docs/
 .DS_Store
-tmp-*
-tmp_*
--- a/deno.json
+++ b/deno.json
@@ -26,7 +26,7 @@
     "test:www": "deno test -A www/main_test.*",
     "release": "deno run -A tools/release.ts"
   },
-  "exclude": ["**/_fresh/*", "**/tmp/*", "*/tests_OLD/**"],
+  "exclude": ["**/_fresh/*", "*/tests_OLD/**"],
   "publish": {
     "include": [
       "src/**",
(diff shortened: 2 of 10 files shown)


kt3k:
CI failures seem caused by https://github.com/denoland/fresh/pull/2837

The test case creates temp directory at the path `path/to/fresh/tmp_<random_hex>`, but that path is now included in `.gitignore`. `deno fmt` and `deno lint` excludes the paths that matches with `.gitignore`, causing `error: No target files found.`

iuioiua:
> `deno fmt` and `deno lint` excludes the paths that matches with `.gitignore`, causing `error: No target files found.`

Sorry, I'm not understanding. Perhaps something obvious I'm missing, but there are plenty of other paths that don't match `path/to/fresh/tmp_<random_hex>`. Why are these other paths not found?

kt3k:
The failing test case `init with tailwind - fmt, lint, and type check project` initializes the project at the path `fresh/tmp_<random_hex>` (with fresh's init template populated in there), and run `deno task check` in that folder, something equivalent of:

```shell
deno run -A -r https://fresh.deno.dev tmp_12345
cd tmp_12345
deno task check
```

and this entire `tmp_12345` directory is ignored by `.gitignore` settings at the root, that causes `deno lint` to fail finding target files

kt3k:
Try remove this line https://github.com/denoland/fresh/blob/d87e47d15fc7d801f5bd4935503a11e0bd94388b/init/src/init_test.ts#L16 and run the test cases. You'll see `tmp_<hash>` folder remaining at the root of the repo

iuioiua:
@kt3k and @marvinhagemeister, I think this is fine to merge as is for now to unblock CI. We can revert the `check` task by before the next release.

marvinhagemeister (approved):
LGTM

PR #2855, opened 2025-04-23
fix(init): fix ignore path, restore check task in init script

kt3k:
This PR restores the original `check` task by reverting the change in https://github.com/denoland/fresh/commit/d87e47d15fc7d801f5bd4935503a11e0bd94388b

The 'init' test cases fail when `tmp_*` is included in `.gitignore` because the test case populates the template in `tmp_<random hex number>`, and that causes `deno lint` and `deno fmt --check` fail.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ vendor/
 node_modules/
 .docs/
 .DS_Store
-tmp_*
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Nice! Should've done that 🤦‍♂️

iuioiua:
This is no longer needed

PR #2858, opened 2025-04-25
refactor(core): standardize `HttpError`

iuioiua:
- Now extends `Error`, so that it satisfies `instanceof Error` operations.
- Narrows the `status` property to be an actual HTTP status code.
- Outsources default evaluation of the `message` property to `@std/http/status`, which already implements this functionality.
- Brings the `stack` property to the class, regardless of its status, which seems like a needless constraint.


marvinhagemeister (approved):
LGTM

PR #2869, opened 2025-04-28
refactor(init): remove `css()`

iuioiua:
It seems this function might not be needed.


marvinhagemeister (changes requested):
Whilst this function does nothing on its own, it allows editors like vscode to syntax highlight the strings. For that reason we should keep it.

iuioiua:
I was wondering if there if it did anything else lol I guess not

PR #2870, opened 2025-04-28
tests: minor cleanups of `tests/precompile_test.ts`

iuioiua:
A tiny lesson that `URL` and `import.meta` can be used instead of `@std/path`. Also, might fix CI thanks to `deno.lock` being updated.


marvinhagemeister (approved):
LGTM

Issue #2850, opened 2025-04-07
[2.0 bug] different/incorrect behaviour due to bump fresh@1.7.3 up to @fresh/core@^2.0.0-alpha.29


calmdowngirl:
i have a partial route [login.tsx](https://github.com/calmdowngirl/sloth/blob/main/routes/login.tsx), it renders a different form depending on what data the handler returns. theres an input field in each form, and theyre pretty much in identical shape except for having different id, name, placeholder etc. what the partial route does is first it asks for email, after user enters email, it then email a code to the user and displays the another form that asks for the code which user would get from their email, user then submit the code, then they get a session in the cookie.

the form looks like this
```html
<form method="POST" autocomplete="off" id="email-form" class="mt-2">
    <input
      class="focus:outline-1 focus:outline-yellow-300"
      type="email"
      id="email-address"
      name="email"
      placeholder=" enter email"
      autocomplete="off"
      autoFocus
      defaultValue=""
      required
    />
    <input
      type="submit"
      value="send code"
      class="px-3 bg-yellow-300 cursor-pointer rounded-sm hover:bg-lime-200"
    />
</form>
```

w/ v1.7.3, when the second form(same as above just having different id, name etc) is displayed, the input field is empty which is as expected (see vid below), as u can see ive explicitly added `autocomplete="off"` and `defaultValue=""` attributes.

w/ fresh v2, when the second form is displayed, the input field is pre-filled w the email address user entered in the email form (see vid below), which is an unexpected and unwanted behaviour change, and i think its incorrect given the form and input has different id and name, and `autocomplete="off"`, `defaultValue=""` attributes.  And the behaviour completely confuses the user as they cant see the placeholder at all, its hard to notice this is another form, and they have to delete the email address for them to be able to enter the code and submit it. its impossible really. 

im tryna start this little side project, do really want to try fresh and v2 now instead of having to upgrade it much later. 

steps to reproduce:
code here: [repo](https://github.com/calmdowngirl/sloth), main branch is fresh v1, fresh v2 is on deno2 branch
ive deployed each branch to a deno deploy project, u can find the link in the gh actions on each branch if u wanna check the behaviour in the vids (i dont wanna post it here even tho the links already public, gcp will charge me $$ for sending email 😂)

https://github.com/user-attachments/assets/be1939ae-d162-4acb-bc03-1c06c4b836f7

https://github.com/user-attachments/assets/416ba967-b856-4221-aa5e-feb352ce4601

[...]

PR #2871, opened 2025-04-29
Fix: Prevent input value leakage by generating unique keys #2850

SisyphusZheng:
Fix: Prevent input value leakage by generating unique keys for Partal different/incorrect behaviour due to bump fresh@1.7.3 up to @fresh/core@^2.0.0-alpha.29 https://github.com/denoland/fresh/issues/2850)


SisyphusZheng:
the fail will slove when emerge the pr #2879

PR #2874, opened 2025-04-30
BREAKING(core): move `FreshConfig.build.outDir` to `FreshConfig.buildOutDir`

iuioiua:
Having `outDir` be a child of `build` would make sense if there were other adjacent properties under `build`. However, this is not the case. Flattening `FreshConfig` opens up the symbols within `src/config.ts` for further simplifications. Also, now `ResolvedFreshConfig` extends `Required<FreshConfig>`.

Depends on #2873
Pre-requisite for #2882


iuioiua:
@marvinhagemeister are you able to take a look or would you prefer we close this?

iuioiua:
@marvinhagemeister, just following up again. Are you able to take a look at this?

PR #2882, opened 2025-05-03
BREAKING(core): merge `FreshConfig` and `ResolvedFreshConfig`

iuioiua:
Depends on #2874


iuioiua:
Will revisit if #2874 is merged

PR #2884, opened 2025-05-03
BREAKING(core): remove `Mode` type

iuioiua:
It doesn't seem like it deserves to be it's own type. It can just be an interface property.


marvinhagemeister (approved):
LGTM

PR #2887, opened 2025-05-05
refactor(update): remove use of `@std/fmt/colors`

iuioiua:
This module simply isn't needed as `console` methods support CSS styles natively.

### DIFF
--- a/update/src/mod.ts
+++ b/update/src/mod.ts
@@ -2,7 +2,6 @@ import { parseArgs } from "@std/cli/parse-args";
 import * as path from "@std/path";
 import { ensureMinDenoVersion, error } from "./utils.ts";
 import { updateProject } from "./update.ts";
-import * as colors from "@std/fmt/colors";
 
 const MIN_DENO_VERSION = "1.43.1";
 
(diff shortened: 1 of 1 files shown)


marvinhagemeister (changes requested):
This breaks color output in macOS' built-in Terminal app. That one famously doesn't support 16bit truecolor and only up to ansi256.

```ts
import * as colors from "jsr:@std/fmt/colors";

console.log("Before:");
console.log(colors.bgRgb8(
  colors.rgb8(" 🍋 Fresh Updater ", 0),
  121,
));
console.log();
console.log("With this PR:");
console.log("%c 🍋 Fresh Updater ", "background-color: #87ffaf; color: black;");
```

Screenshot 2025-05-05 at 08 36 40

iuioiua:
Ah, yes, that's right! In that case, I'd almost go as far as to say that any background styling in `console` methods should be recommended against. Does https://github.com/denoland/deno/pull/18549 or https://github.com/denoland/deno/pull/28122 fix this issue?

marvinhagemeister:
Yeah the first one would fix it. We don't have any code in the console at the moment to convert from 16bit colors to ansi256.

iuioiua:
I've done no prior reading on how it all works, so I have essentially zero context. It'd be great if we can re-open and complete that PR. From what I remember, tests were all that remain. Is that correct? Fixing that would also help fix https://github.com/denoland/std/blob/6a00d6cd34ca70b3a38932e313151a80531c5d2b/internal/build_message.ts#L29.

PR #2888, opened 2025-05-05
chore(core): remove unusued `tailwind_aot_error_page.tsx`

iuioiua:
This appears to not be used anywhere.


marvinhagemeister:
Looks like this page is not wired up anymore. In Fresh 1.x we used this page when people forgot to run the build command and were confused why their css styles weren't being applied. This helped in reducing support requests.

PR #2889, opened 2025-05-05
chore: downgrade lockfile to v4

marvinhagemeister:
This should fix the deploy errors.

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -1,37 +1,55 @@
 {
-  "version": "5",
+  "version": "4",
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
hrmpf looks like this doesn't work either.

iuioiua:
Maybe disable lockfile for now? 🤷🏾‍♂️

iuioiua:
Seems like this is no longer needed.

PR #2890, opened 2025-05-06
fix(core): refine `HttpError.status` property type

iuioiua:
As this is property only ever has an error HTTP status code (>400).


marvinhagemeister (changes requested):
The assumption that only an instance of `HttpError` is thrown is not correct. There is no guarantee for that, see:

```ts
app.use(() => {
  throw "foo"
});
```

iuioiua:
Ah, yes 🤦🏾‍♂️ I did at least assume that an `Error` is thrown.

iuioiua:
Gentle ping @marvinhagemeister

marvinhagemeister (approved):
LGTM

PR #2893, opened 2025-05-06
refactor(core): cleanup jsonify code

iuioiua:
1. Moves `stringify()` and `parse()` tests to `*_test.ts` files. This is usually how tests are structured.
2. Slightly cleans up base64 encoding and decoding functions
3. Removes the use of snapshot assertions, which might not be needed.


src/jsonify/round_trip_test.ts
```diff
@@ -42,9 +40,8 @@ const TESTS = [
 ];
 
 for (const value of TESTS) {
-  Deno.test(`round trip - ${inspect(value)}`, async (t) => {
+  Deno.test(`round trip - ${Deno.inspect(value)}`, () => {
     const str = stringify(value);
-    await assertSnapshot(t, str);
```
iuioiua, line 47: I'm unsure why snapshot assertions were used here originally.

src/runtime/server/preact_hooks.tsx
```diff
@@ -474,9 +474,7 @@ function FreshRuntimeScript() {
       .join(",") +
       "}";
 
-    const serializedProps = JSON.stringify(
-      stringify(islandProps, stringifiers),
-    );
+    const serializedProps = stringify(islandProps, stringifiers);
```
iuioiua, line 477: `stringify()` returns a string. `JSON.stringify(string)` returns that string. So returns the same value.

PR #2898, opened 2025-05-07
fix(www): temporarily unblock Deno Deploy deployments

iuioiua:
Disables `deno.lock`. This should be reverted once Deploy supports lockfile v5.


marvinhagemeister:
Closing as deployments are fixed in `main`.

PR #2900, opened 2025-05-07
refactor: centralise imports and cleanup exports

### LINKED ISSUE #2788: Update `jsr:@fresh/init`'s generated dependencies
By default, the versions of Preact the Fresh v2 alpha and the scaffolded app different, so Deno creates two versions in node_modules, leading to TS errors at JSX expressions about how values of the type `` `ComponentChildren` is not assignable to type `ComponentChildren` ``.  Also, on a fresh :grin: `deno install` (but not `@fresh/init`'s install), it adds a lockfile. I think it's safe to add a lockfile by default now that the starter doesn't depend on `esm.sh`.

iuioiua:
The benefits of these changes are:
1. Managing and updating dependencies is easier. One only needs to run `deno outdated` at the root, instead of in each workspace.
2. The versions of dependencies across workspaces is kept consistent.

Closes #2788


iuioiua:
PTAL @marvinhagemeister

marvinhagemeister:
I fee like this PR is a good argument why Deno should support `workspace:` outside of `package.json`. It's perfectly reasonable to centralise versioning of shared dependencies in the root `deno.json` file like done here, but you still want to be able to see what the dependencies of each package are. I'll bring it up with the CLI team.

iuioiua:
We could just use `deno.jsonc`.

marvinhagemeister:
> We could just use `deno.jsonc`.

I'm not sure I'm following. What would change and what it would look like?

iuioiua:
Imports within `deno.jsonc` sectioned off by comments indicating the workspace the group of imports is for. Happy to implement.

iuioiua:
@marvinhagemeister, is not being able to know which dependencies each workspace uses a blocker for this PR? If so, I can rename the `deno.json` to `deno.jsonc` and section off each group of imports by workspace using a comment. But that'd need other small changes to the codebase that read the root config file. This PR would cure quite a few ills, including potential ones.

marvinhagemeister (approved):
LGTM

> is not being able to know which dependencies each workspace uses a blocker for this PR?

No, I've changed my mind on this one since the recent issue where due to missing `peerDependencies` I published an alpha with duplicate Preact versions. Let's merge this.

PR #2901, opened 2025-05-07
tests: use the same browser instance across tests and better use explicit resource management

iuioiua:
On my machine, reduces time taken to perform `deno task test` from ~1m20s to ~30s.

Inspired by and supersedes #2110.


tests/test_utils.tsx
```diff
@@ -28,6 +28,20 @@ import { NodeProcess } from "./fixtures_islands/NodeProcess.tsx";
 import { FreshAttrs } from "./fixtures_islands/FreshAttrs.tsx";
 import { OptOutPartialLink } from "./fixtures_islands/OptOutPartialLink.tsx";
 
+const browser = await launch({
+  args: [
+    "--window-size=1280,7201",
+    ...((Deno.env.get("CI") && Deno.build.os === "linux")
+      ? ["--no-sandbox"]
```
lino-levan, line 35: https://github.com/lino-levan/astral?tab=readme-ov-file#no-usable-sandbox-with-user-namespace-cloning-enabled

I would recommend removing app armor instead of enabling --no-sandbox

iuioiua: What is the advantage? Genuine question.

iuioiua: Or do you think that's the issue here?

lino-levan: The biggest reason is it cleans up the codebase. I've heard of really obscure bugs not being caught in CI because of `--no-sandbox`, but I've never seen one myself so take that with a grain of salt.

tests/test_utils.tsx
```diff
@@ -28,6 +28,23 @@ import { NodeProcess } from "./fixtures_islands/NodeProcess.tsx";
...
+ * @see {@link https://github.com/lino-levan/astral#no-usable-sandbox-with-user-namespace-cloning-enabled}
+ */
+const browser = await launch({
+  args: ["--window-size=1280,720"],
+  headless: true,
+});
+
+addEventListener("unload", async () => {
+  await browser.close();
+});
```
iuioiua: @lino-levan, when I use `await using`, tests hang.

.github/workflows/ci.yml
```diff
@@ -48,6 +48,10 @@ jobs:
       - name: Type check project
         run: deno task check:types
 
+      - name: Disable AppArmor
```
iuioiua: I think this is worth trying.

iuioiua:
PTAL @marvinhagemeister

tests/test_utils.tsx
```diff
@@ -28,6 +28,23 @@ import { NodeProcess } from "./fixtures_islands/NodeProcess.tsx";
 import { FreshAttrs } from "./fixtures_islands/FreshAttrs.tsx";
 import { OptOutPartialLink } from "./fixtures_islands/OptOutPartialLink.tsx";
 
+/**
+ * If on Linux, disable `AppArmor` before running tests:
+ * ```sh
+ * echo 0 | sudo tee /proc/sys/kernel/apparmor_restrict_unprivileged_userns
```
marvinhagemeister: I'm not sure if it's a good idea to tell contributors on Linux to disable `AppArmor` globally on their machines. Contributing to a JS framework shouldn't come with the requirements to disable central security features in the OS you're using.

iuioiua: Very good point. Silly for me to do. Reverted.

iuioiua:
@marvinhagemeister, are you able to please take a look? It'd be great to have this bump in test performance.

marvinhagemeister (approved):
LGTM

PR #2904, opened 2025-05-10
refactor: inline `App` getters and setters

timreichen:


marvinhagemeister (changes requested):
Inlining these means exposing them as part of the public API. I'm not sure if the router class or the island registry is finalized enough to expose as part of the public API.

timreichen:
> Inlining these means exposing them as part of the public API. I'm not sure if the router class or the island registry is finalized enough to expose as part of the public API.

How about accessing it via private symbol?

marvinhagemeister:
> > Inlining these means exposing them as part of the public API. I'm not sure if the router class or the island registry is finalized enough to expose as part of the public API.
> 
> How about accessing it via private symbol?

I'd reject such refactorings as they it only comes down to personal taste which way to prefer. I'd rather spend the available energy elsewhere, where user's can benefit from the changes rather than changing coding styles.

marvinhagemeister:
Closing as we're not going to merge this.

timreichen:
> > > Inlining these means exposing them as part of the public API. I'm not sure if the router class or the island registry is finalized enough to expose as part of the public API.
> 
> > 
> 
> > How about accessing it via private symbol?
> 
> 
> 
> I'd reject such refactorings as they it only comes down to personal taste which way to prefer. I'd rather spend the available energy elsewhere, where user's can benefit from the changes rather than changing coding styles.

I disagree that this is only personal taste. Although functionally it might work the same, it is not obvious (at least it was not for me) why there are exported variables that are not defined until you read the rest of the code to find where they are assigned and what they do. This wastes time and makes it harder to contribute for new contributors.

iuioiua:
> I disagree that this is only personal taste. Although functionally it might work the same, it is not obvious (at least it was not for me) why there are exported variables that are not defined until you read the rest of the code to find where they are assigned and what they do. This wastes time and makes it harder to contribute for new contributors.

I agree

marvinhagemeister:
This page on MDN might help https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks#access_to_private_properties

iuioiua:
I think the need to read that page might be more to our point that this way of doing things might not be so clear to the reader up front. I like Tim's change because it's immediately clear to the reader what's going on and a lot more beginner friendly. The main point against this change, however, is that it would make the router, island registry and build cache properties public. And it's good to keep the number of symbols in the public API to a minimum.

PR #2906, opened 2025-05-10
refactor: make `App.prototype.handler()` sync

timreichen:


marvinhagemeister (changes requested):
The underlying `buildCache.fromSnapshot()` function is still async. By making `app.handler()` sync, this PR introduces a race condition where a request might be coming in before the snapshot is restored. In this case we go through the error handling logic and will show an error.

For production servers this is not an acceptable way to handle client traffic.

timreichen:
> The underlying `buildCache.fromSnapshot()` function is still async. By making `app.handler()` sync, this PR introduces a race condition where a request might be coming in before the snapshot is restored. In this case we go through the error handling logic and will show an error.
> 
> For production servers this is not an acceptable way to handle client traffic.

This PR also makes `fromSnapshot()` sync. Or are there more async calls that I missed?

marvinhagemeister:
> This PR also makes `fromSnapshot()` sync. Or are there more async calls that I missed?

Oh apologies, I missed that in the changes. In that case it makes sense.

marvinhagemeister (approved):
LGTM

PR #2908, opened 2025-05-10
refactor: fix `App.prototype.handler()` return type 

timreichen:


marvinhagemeister (changes requested):
I don't think we should leak this edge case to users. There is no other scenario that I'm aware of where the returned function could return a `Response` synchronously. We shouldn't expose that internal to users. Did you find more scenarios in which that could happen?

timreichen:
> I don't think we should leak this edge case to users. There is no other scenario that I'm aware of where the returned function could return a `Response` synchronously. We shouldn't expose that internal to users. Did you find more scenarios in which that could happen?

No, although middlewares allow both `Response` and `Promise<Response>` return values. So in a sense `Response | Promise<Response>` probably is expected as a return type even if all middlewares are awaited.
I don't see the exposure as a problem in that this is what is being returned.

marvinhagemeister:
Closing as we're not going to land this.

PR #2911, opened 2025-05-10
fix: do not skip app-wrapper for sibling routes

### LINKED ISSUE #2910: App wrapper not rendered for sibling route
This was a bit fiddly to reproduce. But my issue is basically the following:

- A route `routes/bar.tsx` has `skipAppWrapper: true` and `skipInheritedLayouts: true`
- A route `routes/foo.tsx` has no `config` set
- A route `routes/index.tsx` has no `config` set
- There is a `_app.tsx` and `_500.tsx` in `routes/`
- The path `/foo` is rendered without the `_app.tsx` wrapper

I came upon this bug when I has a route in my app inside `routes/api/cart` which was intended to return partials, and thus did not need `_app` or `_layout`. Instead though, this caused all routes except `/` to no longer get the app wrapper.

If I removed the config from `_500.tsx`, or removed the `_500.tsx`, I could no longer reproduce the issue. So I could not manage to make the reproduction any smaller than this.

### PR with failing unit test

I managed to create a unit test that fails with the same issue in #2911.

### Reproduction

The issue was a bit fiddly to reproduce, as it seems to depend on the order in which routes are read. `/index.tsx` is rendered correctly (has App wrapper), but the `/foo.tsx` route does not even though it should. I haven't found the solution to this, but it looks to me that [whether an app-wrapper exists or not can be permanently set to `false`](https://github.com/denoland/fresh/blob/main/src/plugins/fs_routes/mod.ts#L217-L227) for a certain order of route handling.

```tsx
// _500.tsx
export const config = {
  skipInheritedLayouts: true
};

export default () => {
  return <h2>500 error</h2>;
};
```

```tsx
// _app.tsx
export default (({ Component }) => {
  return <>app/<Component /></>;
};
```

```tsx
// bar.tsx
export const config = {
  skipAppWrapper: true,
  skipInheritedLayouts: true
};

export default () => {
  return <>bar</>;
};
```

```tsx
// foo.tsx
export default () => {
  return <>foo</>;
};
```

```tsx
// index.tsx
export default () => {
  return <>index</>;
};
```

#### Expected result

Route `/foo` has an app wrapper.

#### Actual result

Route `/foo` does not have ean app wrapper.

csvn:
This PR adds a unit test for the issue in #2910. I have not found a solution yet, since I'm not very familiar with the `fsRoutes` code, and what it is doing. But a thought a unit test with the failing code could help to find a bug fix.

**Edit:** I managed to fix the failing test by removing a couple of lines, but I'm not too familiar with the code so I'm not sure if it's correct. It seems to pass all the tests still...

Closes #2910

### DIFF
--- a/src/plugins/fs_routes/mod.ts
+++ b/src/plugins/fs_routes/mod.ts
@@ -221,12 +221,10 @@ export async function fsRoutes<State>(
 
       // _app template
       if (skipApp && mod.path === "/_app") {
-        hasApp = false;
         continue;
       } else if (!skipApp && mod.config?.skipAppWrapper) {
         skipApp = true;
         if (hasApp) {
-          hasApp = false;
           // _app component is always first
           components.shift();
         }
--- a/src/plugins/fs_routes/mod_test.tsx
+++ b/src/plugins/fs_routes/mod_test.tsx
@@ -558,6 +558,56 @@ Deno.test("fsRoutes - route overrides _app", async () => {
   expect(doc.body.firstChild?.textContent).toEqual("route");
 });
 
+Deno.test("fsRoutes - route overrides _app does not affect other routes", async () => {
+  const server = await createServer({
+    "routes/_500.tsx": {
+      config: { skipInheritedLayouts: true },
+      default: () => <>error!</>,
+    },
+    "routes/_app.tsx": {
+      default: (ctx) => (
+        <div>
+          app/<ctx.Component />
+        </div>
+      ),
+    },
+    "routes/_layout.tsx": {
+      default: (ctx) => (
+        <>
+          layout/<ctx.Component />
+        </>
+      ),
+    },
+    // By having "bar" come before "foo", this results in a bug
+    // where the app wrapper is missing from "foo".
+    "routes/bar.tsx": {
+      config: {
+        skipAppWrapper: true,
+        skipInheritedLayouts: true,
+      },
+      default: () => <>bar</>,
+    },
+    "routes/index.tsx": { default: () => <>index</> },
+    "routes/foo.tsx": { default: () => <>foo</> },
+  });
+
+  const [indexRes, fooRes, barRes] = await Promise.all([
+    server.get("/"),
+    server.get("/foo"),
+    server.get("/bar"),
+  ]);
+  const [index, foo, bar] = await Promise.all([
+    indexRes.text(),
+    fooRes.text(),
+    barRes.text(),
+  ]);
+  expect(parseHtml(index).body.firstChild?.textContent).toEqual(
+    "app/layout/index",
+  );
+  expect(parseHtml(foo).body.firstChild?.textContent).toEqual("app/layout/foo");
+  expect(parseHtml(bar).body.firstChild?.textContent).toEqual("bar");
+});
+
 Deno.test("fsRoutes - handler return data", async () => {
   const server = await createServer({
     "routes/index.tsx": {


src/plugins/fs_routes/mod_test.tsx
```diff
@@ -558,6 +558,56 @@ Deno.test("fsRoutes - route overrides _app", async () => {
...
+    },
+    // By having "bar" come before "foo", this results in a bug
+    // where the app wrapper is missing from "foo".
+    "routes/bar.tsx": {
+      config: {
+        skipAppWrapper: true,
+        skipInheritedLayouts: true,
+      },
+      default: () => <>bar</>,
+    },
```
csvn, line 589: If `route/bar.tsx` came after `routes/foo.tsx` in the record, the issue disappears. Since iteration over the files happens in key-order (for test method `createServer`) it seems to be some connection to the order FS will walk over the directory for this issue to occur in a real app.

src/plugins/fs_routes/mod.ts
```diff
@@ -221,12 +221,12 @@ export async function fsRoutes<State>(
...
       // _app template
       if (skipApp && mod.path === "/_app") {
-        hasApp = false;
+        // hasApp = false;
         continue;
       } else if (!skipApp && mod.config?.skipAppWrapper) {
         skipApp = true;
         if (hasApp) {
-          hasApp = false;
+          // hasApp = false;
```
csvn, line 229: I don't know if it's correct just to remove these lines. It looks wrong to me to set `hasApp` to `false` if we're looping over every route, since a latter route may have the app wrapper disabled always.

This change makes the test I added pass anyway...

src/plugins/fs_routes/mod.ts
```diff
@@ -221,12 +221,10 @@ export async function fsRoutes<State>(
...
       // _app template
       if (skipApp && mod.path === "/_app") {
-        hasApp = false;
         continue;
       } else if (!skipApp && mod.config?.skipAppWrapper) {
         skipApp = true;
         if (hasApp) {
-          hasApp = false;
           // _app component is always first
           components.shift();
```
csvn, line 229: I don't know if it's correct just to remove these lines. It looks wrong to me to set `hasApp` to `false` if we're looping over every route, since a latter route may have the app wrapper disabled always.

This change makes the test I added pass anyway...

csvn:
@marvinhagemeister I think this PR is ready for review. I don't see any failing tests, and it looks like the test I added now passes so this should likely fix #2910 too.

marvinhagemeister (approved):
Great catch! Seems like always setting `hasApp = false` was a bit heavy handed. Happy to hear that this was all that's needed to fix it 👍

marvinhagemeister:
Ah wait kinda wondering if there is a case where a route says to skip the app wrapper, but the next route in the list doesn't. In that case we need to ensure that the app wrapper template is still applied.

csvn:
> Ah wait kinda wondering if there is a case where a route says to skip the app wrapper, but the next route in the list doesn't. In that case we need to ensure that the app wrapper template is still applied.

I think that's exactly what I tried to accomplish with my change? To me `skipApp` is used for that, and `hasApp` is whether or not an `_app` file exists at all. I don't think "`_app` exists" should ever be changed after set to `true`.

- After my changes, `hasApp` is initialized to `false`, and only ever changes once to `true` if `/_app` is encountered.
- Nested for loops are running over two route modules, and `routeMod` seems to be all routes, and `mod` the current "special" route (e.g. `_app`, `_layout` etc)
  - `skipApp` (`let skipApp = !!routeMod.config?.skipAppWrapper;`) and `mod.config?.skipAppWrapper`, are these ones together not enough to determine if the app wrapper should be included?
- I don't really understand what the `components.shift()` part does 😅 

It at least looks off to me if `hasApp` is set to `false`, because if routes `1`, `2` and `3` are processed in that order, if `1` uses `skipAppWrapper: true`, then `hasApp` becomes `false` and `2` and `3` can never have the app-wrapper.

Not sure I understood your question. But I don't think I've dug into `fs_routes/mod.ts` enough to understand it more than that. 🤔

csvn:
Perhaps `hasApp` could be renamed `appRouteExists`, and `useAppWrapper`/`useAppComponent` instead of `skipApp`. Maybe that would make it clearer? I think it should likely be enough to only ever change `skipApp` in current `for` loop, because no other route should be able to make permanent changes to another route?

marvinhagemeister:
Ah you're totally right. I confused `hasApp` with `skipApp`, I guess a sign that more distinct names would be better 😅

csvn:
@marvinhagemeister there's been quite a few fixes and changes in the last few weeks. Would it be possible to release a new Alpha version soon? The last release was in January, and it would be nice to test out some of the recent bug-fixes with a new JSR release 🙏 😇 

(No pressure though 😁)

marvinhagemeister:
@csvn yep, released alpha 30 earlier today.

PR #2914, opened 2025-05-10
fix: allow relative paths for outDir and staticDir

### LINKED ISSUE #2913: Support relative paths for `build.outDir` and `staticDir`
IMO it should be possible to use relative paths for `build.outDir` and `staticDir`, which would be resolved against the `root` option:

```ts
// /src/main.ts
export const app = new App<State>({
  root: './src',
  build: {
    outDir: './.fresh'  // -> /src/.fresh
  },
  staticDir: 'public'  // -> /src/public
});
```

Currently, this does not work, as the paths are used as-is for `build.outDir` and `staticDir`. Currently, it requires quite verbose code and repetition to make these properties relative to `root`:

```ts
// /src/main.ts
import { join } from '@std/path/join';

export const app = new App<State>({
  root: './src',
  build: {
    outDir: join(import.meta.dirname!, './.fresh')  // -> /src/.fresh
  },
  staticDir: join(import.meta.dirname!, './public')
});
```

If not specifying anything for `build.outDir` and `staticDir`, they will use their default values and be relative against `root`, so it works for default values but not new or relative values.

csvn:
This PR consists of several commits, where the goal is to re-enable the test (fix for Windows), as well as allow relative paths for config options to address #2913.

1. Re-enables test and normalizes paths for Windows
2. Add unit test for `normalizeConfig` and the `root` option
3. Add unit test for `normalizeConfig` for `build.outDir` and `staticDir`, and allow the properties to be relative
4. Add JSDoc for fields in `FreshConfig`

I can split these into separate PR's if preferred, but all code changes are very in nearly the same places.

Closes #2913

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -49,13 +49,15 @@ export function parseRootPath(root: string, cwd: string): string {
     root = path.dirname(root);
   }
 
+  if (Deno.build.os === "windows") {
+    root = root.replaceAll("\\", "/");
+  }
+
   return root;
 }
 
 export function normalizeConfig(options: FreshConfig): ResolvedFreshConfig {
-  const root = options.root
-    ? parseRootPath(options.root, Deno.cwd())
-    : Deno.cwd();
+  const root = parseRootPath(options.root ?? ".", Deno.cwd());
 
   return {
     root,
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #2915, opened 2025-05-10
fix: remove unnecessary config property in init script

csvn:
From what I can see, `fsRoutes` uses the `FreshConfig.root` option by default, so this property does not need to be specified. I also added some JSDoc for the `dir` option in `fsRoutes()` as I understood it's purpose.


marvinhagemeister (approved):
Thanks!

PR #2916, opened 2025-05-10
fix: allow Data to be inferred from Handler function

csvn:
This PR adds support for single-function GET handler data to be inferred when using `define.page<typeof handler>()`.

```ts
export const handler = define.handlers((ctx) => {
  return page({ foo: "foo" });
});

export default define.page<typeof handler>((ctx) => {
  ctx.data; // Correctly resolves as { foo: string }
});
```

This PR also adds a test that checks that the types inferred are correct by type checking.


marvinhagemeister (approved):
Nice, thanks!

PR #2917, opened 2025-05-10
fix: update dependency versions

csvn:
I had an issue while using Fresh alpha where I got errors because I was using `v2.0.4` of `@preact/signals`. I had to ensure I was using `v1.x`, because `jsr:@fresh/core` had a dependency on the 1.x version. The errors said something about "effect() is already running", when the library was loaded twice, and it was hard to figure out what the root issue was.

Is there a way to set a version range with `deno.json` `imports`? I could not find support for range (`-`) or OR (`||`), that can be [done with `package.json`](https://github.com/angular/angular/blob/main/packages/core/package.json#L23).

Is there a reason not to use the `v2` version of Signals? If this PR does something dumb, I don't mind it being closed. I was just surprised a newly generated Fresh project had to use the "old" signals major.


csvn:
Oh, I see a lot of test errors. It seems the v2 signal version with it's `requestAnimationFrame` usage is not compatible with Fresh without some work, and a lot of values are `undefined` instead of their expected values 🤔

iuioiua:
I'd hold off on this PR until #2900 is merged.

csvn:
True, it's probably better to address the `@preact/signals` issue separately. I also found it weird with dependency versions being repeated, so making use of workspace for consolidated versions in #2900 makes sense.

PR #2922, opened 2025-05-12
refactor(core): replace `missingBuildHandler()` with throwing of new `SetupError`

iuioiua:
This PR changes the behaviour of `App.prototype.handler()` in that when run, it throws a new `SetupError` instead of succeeding and serving a page that guides the dev to finish setting up the server. This provide more immediate feedback.

This might be a prerequisite for #2896.


tests/fixture_precompile/valid/main.tsx
iuioiua, line 1: These changes might need to be reverted if checks in `App.prototype.handler()` are moved to `App.prototype.listen()`.

PR #2929, opened 2025-05-13
fix: make static folders with dot such as `.well-known` allowed

csvn:
This PR intends to address an error logged when opening a DevTools in a Chromium based browser while using Fresh in dev mode.

<details>

<summary><strong>See error</strong></summary>

```
Error: Processed file resolved outside of static dir C:\Users\<name>\fresh-site\src\static\.well-known\appspecific\com.chrome.devtools.json
    at MemoryBuildCache.readFile (https://jsr.io/@fresh/core/2.0.0-alpha.29/src/dev/dev_build_cache.ts:81:13)
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async freshStaticFiles (https://jsr.io/@fresh/core/2.0.0-alpha.29/src/middlewares/static_files.ts:30:18)
    at async FreshReqContext.fn (https://jsr.io/@fresh/core/2.0.0-alpha.29/src/middlewares/mod.ts:99:18)
    at async https://jsr.io/@fresh/core/2.0.0-alpha.29/src/dev/middlewares/error_overlay/middleware.tsx:15:14
    at async FreshReqContext.fn (https://jsr.io/@fresh/core/2.0.0-alpha.29/src/middlewares/mod.ts:99:18)
    at async fn (https://jsr.io/@fresh/core/2.0.0-alpha.29/src/middlewares/mod.ts:99:18)
    at async https://jsr.io/@fresh/core/2.0.0-alpha.29/src/app.ts:232:16
    at async mapped (ext:deno_http/00_serve.ts:407:18)
```

</details>

DevTools tries to load a `<root>/static/.well-known/appspecific/com.chrome.devtools.json` file, which incorrectly is flagged as "outside" the static directory, and thus generates errors instead of 404.

Maybe hidden folders (starting with `.`) in `/static` should be opt-in though? I'm not sure if anyone puts folders with `.` prefix inside `static/` without intending them to be accessible?

- `..` throws errors (outside staticDir)
- `.` returns "missing file" (unless `config.allowHiddenFolders` is true)

## Automatic workspace folders

I think a follow-up PR could be for the Dev `Builder` to also implement support for [Automatic workspace folders](https://chromium.googlesource.com/devtools/devtools-frontend/+/main/docs/ecosystem/automatic_workspace_folders.md), by returning JSON content automatically at `.well-known/appspecific/com.chrome.devtools.json`.

I think a good first step is to allow `.<folder-name>`, especially since `.well-known/` has [many valid use-cases](https://en.wikipedia.org/wiki/Well-known_URI), and as such is not unreasonable to want to add static files in.

### DIFF
--- a/src/build_cache.ts
+++ b/src/build_cache.ts
@@ -111,7 +111,7 @@ export class ProdBuildCache implements BuildCache {
       : path.join(base, pathname);
 
     // Check if path resolves outside of intended directory.
-    if (path.relative(base, filePath).startsWith(".")) {
+    if (path.relative(base, filePath).startsWith("..")) {
       return null;
     }
 
--- a/src/middlewares/static_files.ts
+++ b/src/middlewares/static_files.ts
@@ -10,7 +10,7 @@ import { trace, tracer } from "../otel.ts";
  * Fresh middleware to enable file-system based routing.
  * ```ts
  * // Enable Fresh static file serving
- * app.use(freshStaticFles());
+ * app.use(staticFiles());
  * ```
  */
 export function staticFiles<T>(): MiddlewareFn<T> {
--- a/src/build_cache_test.ts
+++ b/src/build_cache_test.ts
@@ -1,0 +1,78 @@
+import { expect } from "@std/expect";
+import * as path from "@std/path";
+import { ProdBuildCache, type StaticFile } from "./build_cache.ts";
+import type { ResolvedFreshConfig } from "./mod.ts";
+
+async function getContent(readResult: Promise<StaticFile | null>) {
+  const res = await readResult;
+  if (res === null) return null;
+  if (res.readable instanceof Uint8Array) throw new Error("not implemented");
+  return new Response(res.readable).text();
+}
+
+Deno.test({
+  name: "ProdBuildCache - should error if reading outside of staticDir",
+  fn: async () => {
+    const tmp = await Deno.makeTempDir();
+    const config: ResolvedFreshConfig = {
+      root: tmp,
+      mode: "production",
+      basePath: "/",
+      staticDir: path.join(tmp, "static"),
+      build: {
+        outDir: path.join(tmp, "dist"),
+      },
+    };
+    await Deno.mkdir(path.join(tmp, "static", ".well-known"), {
+      recursive: true,
+    });
+    await Deno.mkdir(path.join(tmp, "dist", "static"), {
+      recursive: true,
+    });
+    await Promise.all([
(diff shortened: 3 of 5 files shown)


src/build_cache.ts
```diff
@@ -111,7 +111,7 @@ export class ProdBuildCache implements BuildCache {
       : path.join(base, pathname);
 
     // Check if path resolves outside of intended directory.
-    if (path.relative(base, filePath).startsWith(".")) {
+    if (path.relative(base, filePath).startsWith("..")) {
```
csvn, line 114: I can't come up with a scenario where just `.` and not `..` can escape to a parent directory except for e.g. `./../`. But if this is deemed "risky", we could perhaps change it to:

```ts
    const rel = path.relative(base, filePath);
    if (rel.startsWith(".") && !rel.startsWith(".well-known/") {
```

The `.well-known` folder is the only example I can come up with currently that is required to start with `.`, though there are likely cases somewhere 😄

marvinhagemeister: yeah it's fine. The `../foo` case is the one to guard against. I guess I was just lazy when writing that code and checked for a single dot.

src/middlewares/static_files.ts
```diff
@@ -10,7 +10,7 @@ import { trace, tracer } from "../otel.ts";
  * Fresh middleware to enable file-system based routing.
  * ```ts
  * // Enable Fresh static file serving
- * app.use(freshStaticFles());
```
marvinhagemeister, line 13: Oops, good catch!

marvinhagemeister (approved):
Great idea, I didn't know that Chrome Devtools supported the `.well-known` directory. Thanks for making a PR and appreciate the addition of tests 👍

src/build_cache.ts
csvn: Good, then I wasn't misunderstanding something! 🙌

csvn:
> Thanks for making a PR and appreciate the addition of tests 👍

I think the tests could be structured a bit better, but was a bit lazy and combined them all into a single test 😅

> Great idea, I didn't know that Chrome Devtools supported the .well-known directory. 

I might try to create a PR for `MemoryBuildCache` (for `deno run -A dev.ts`) to return a valid `com.chrome.devtools.json` file. Then it's possible for edits in DevTools to directly apply to source files. I saw [Angular CLI](https://github.com/angular/angular-cli/pull/29757) adding support for this, and Chromium is likely used by a lot of Developers. But not a high prio right now 😃

PR #2931, opened 2025-05-13
BREAKING(core): move build cache checks out of `App.prototype.handler()`

iuioiua:
`App.prototype.handler()` is now more accurately named as it's now a handler rather than a handler factory. Checking for the build cache is something that should be checked upon listening.

Needed for #2896

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -92,7 +92,7 @@ async function listenOnFreePort(
   options: ListenOptions,
   handler: (
     request: Request,
-    info?: Deno.ServeHandlerInfo,
+    info: Deno.ServeHandlerInfo,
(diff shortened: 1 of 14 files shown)


marvinhagemeister (commented):
> Checking for the build cache is something that should be checked upon listening.

Are you saying that user's should do this check themselves?

```tsx
const handler = app.handler();
Deno.serve(handler)
```

How would we hook into the listen aspect here?

iuioiua:
Maybe we could move the build logic into a new `.build()` method and call that method within `.listen()`. If needed, the user can call it before `.handle()`.

iuioiua:
I'll revisit this once approaching `deno serve` support.

PR #2933, opened 2025-05-13
refactor(core): remove `FsAdapter.readFile()`

iuioiua:
I realise we complicate a lot of objects, classes and interfaces for testing purposes. By stubbing (and possibly) spying only where we need to, we can simplify these symbols and jump through far fewer mental hoops to reason about things.

Pre-requisite for #2934

### DIFF
--- a/src/dev/builder.ts
+++ b/src/dev/builder.ts
@@ -5,7 +5,6 @@ import {
   type ListenOptions,
   setBuildCache,
 } from "../app.ts";
-import { fsAdapter } from "../fs.ts";
 import * as path from "@std/path";
 import * as colors from "@std/fmt/colors";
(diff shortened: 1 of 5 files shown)


csvn:
I think perhaps `FsAdapter` exists so that it's possible for Fresh to eventually run on e.g. NodeJS, Cloudflare or other non-Deno environments.

iuioiua:
I think `FsAdapter` exists more for having testing and implementation share the same interface. IMO it's best not to complicate implementation for the sake of testing. You can see this in other spots of the codebase which means the reader has to jump through hoops to figure out what's going on. While it's possible it might be for runtime compatibility, I'm not aware of any such plans.

iuioiua:
Closing in favour of another upcoming PR.

PR #2946, opened 2025-05-14
fix(init): must use urls for asset copying

marvinhagemeister:
Second attempt to fix https://github.com/denoland/fresh/pull/2943 . Turns out that `import.meta.dirname` is `undefined` when served from JSR. I'm not able to reproduce that error locally though.

Fixes https://github.com/denoland/fresh/pull/2943

### DIFF
--- a/init/src/init.ts
+++ b/init/src/init.ts
@@ -395,7 +395,7 @@ ${GRADIENT_CSS}`;
   await writeFile(
     "static/favicon.ico",
     await Deno.readFile(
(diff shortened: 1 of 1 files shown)


csvn:
>  Turns out that import.meta.dirname is undefined when served from JSR.

Wow, that was news to me too. Is that intended behavior or is a bug? I'm more surprised that `import.meta.resolve` works compared to `dirname` 😅

Anyway, thanks for the quick fix!

Issue #2953, opened 2025-05-15
[v2.0] Object version of `handler` not supported for `_error`, `_500` and `_404`


csvn:
While migrating a page from Fresh v1 to v2-alpha, I noticed that the `handler` stopped working for what was previously `_404` in the new `_error`.

The issue seems to stem from using the objectMethod type of handler over just a `function` for `GET`:

```ts
export const handler = define.handlers({
  GET(ctx) {
    // Return 404 with no HTML body if URL looks to be file-like
    const is404 = ctx.error instanceof HttpError && ctx.error.status === 404;
    const isFileLike = fileExtensions.some((ext) => ctx.url.pathname.endsWith(ext));

    if (is404  && isFileLike) return new Response('', { status: 404, statusText: 'Not found' });

    return page();
  }
});

// When changed to the below style it started working:
export const handler = define.handlers((ctx) => {
  // Same method body
});
```

There are two "FIXME"'s in the code that is most likely the cause of this issue:

- https://github.com/denoland/fresh/blob/main/src/plugins/fs_routes/mod.ts#L252
- https://github.com/denoland/fresh/blob/main/src/plugins/fs_routes/mod.ts#L279

## Solution

A quick fix would be to log a warning until the issue is fixed.

- Should `_404` support any other handler types than `GET`?
- Does `_500`/`_error` support methods other than `GET`?

PR #2954, opened 2025-05-15
fix: log warning on invalid route configurations

csvn:
Related to #2953.

I found it confusing that there were unsupported route configurations, that fails silently without any errors or wanings. Though I think it is a good idea to not throw for this, as that would be worse DX. But logging a warning if the developer is migrating from Fresh v1, or if the developer is using code that will not work (e.g. copy pasting a handler from a route to `_middleware`), I think it's helpful where reasonable to help guide a bit more.

Does a small refactor first to hopefully improve typing to be more correct for `InternalRoute.handlers`, then adds warnings for the following:

- `_middleware` using Object form (`GET`, `POST` handlers) of the handler format
  - Helps find issues when moving handler from route to middleware
- `_layout` using `handlers`
  - E.g. if developer moves a route to a layout
- ~~`_404`, `_500` or `_error` using Object handler form (`GET`, `POST` handlers), which is not yet supported (See #2953)~~
  - No longer necessary as it is fixed by #2955

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -98,9 +98,10 @@ export type RouteHandler<Data, State> =
   | HandlerByMethod<Data, State>;
 
 export function isHandlerByMethod<D, S>(
-  handler: RouteHandler<D, S>,
+  handler: RouteHandler<D, S> | RouteHandler<D, S>[] | null,
 ): handler is HandlerByMethod<D, S> {
-  return handler !== null && typeof handler === "object";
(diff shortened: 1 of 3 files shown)


csvn:
I made this before #2955, so it's not as necessary anymore. It's not as helpful, but I think it might still have some value to log when a route has an invalid pattern.

marvinhagemeister (approved):
LGTM, thanks!

PR #2956, opened 2025-05-15
chore: Update links

matthewjnield:
This commit updates several links which are outdated and now redirect to different URLs on the updated Deno website and documentation.

There are a few dozen instances of URLs starting with `https:://deno.land/x` and `https:://deno.land/std`. I left these alone since they are dependencies. Would it make sense to update these from the versions hosted at `deno.land` to their JSR equivalents, since that is the new and preferred location of the Deno Standard Library packages as well as other JavaScript/TypeScript packages used by Deno projects?

For example, the latest version of the Deno Standard Library is now hosted at `https://jsr.io/@std` instead of  `https://deno.land/std`.

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The [documentation](https://fresh.deno.dev/docs/introduction) is available on
 
 ## 🚀 Getting started
 
-Install the latest [Deno CLI](https://deno.land/) version.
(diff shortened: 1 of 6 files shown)


marvinhagemeister (approved):
Thanks!

Issue #2937, opened 2025-05-14
Fresh with tailwindcss processes css files in static/assets


patill:
I have a fresh app which includes a pre built React vite app. So the routes/index.tsx file just uses Deno.readFile to get the static index.html file from the vite project. This html file imports itself a (compiled) tailwind css file. Problem is that fresh/tailwind wants to process the css file included, but this tailwind css file is already processed and should not be processed by tailwind anymore.

How can I tell the tailwind plugin/ fresh app not to consider css files in the /static/assets folder?

I tried to include the original vite app tailwind css file (not yet processed by the vite build process) into the /static/assets folder, and fresh will convert it. But the css file from the vite app is not compatible with fresh. The processed css file of the vite app contains a lot of @layer stuff which is missing if I put the non-processed css file in the fresh app. Some basic styles are working, but most styles are missing.

**How to reproduce?**

1. Make a new fresh app with tailwindcss.
2. Make a new (deno) vite app with `deno run -A npm:create-vite-extra --template deno-react-ts`.
3. Add Tailwind css to the vite app as told in the [tailwindcss docs](https://tailwindcss.com/docs/installation/using-vite), only replace npm installation with `deno add npm:tailwindcss npm:@tailwindcss/vite`.
4. Build the vite app, e.g. `deno task build`.
5. Copy the contains of the /dist folder of the vite app into the /static folder of the fresh app.
6. Modify the /routes/index.tsx file to contain only a handler:
 
```
export const handler: Handlers = {
  async GET(_req: Request, ctx) {
  const file = await Deno.readFile("static/index.html");
    return new Response(file, {
      headers: {
        "content-type": "text/html; charset=utf-8",
      },
    });
  },
};
```

7. Start the fresh app, e.g. `deno task start`. The app will start but there will be errors. When trying to build the app the build process will not finish.

When running the localhost server the error actually will show up only when opening the page in the browser. The error is `CssSyntaxError: <css input>:1:443: `@layer base` is used but no matching `@tailwind base` directive is present.`

patill:
In fact I found a workaround:

When omiting the `.css` postfix of the file the file will not be processed by tailwindcss.

Of course then you have the problem that the css file might not be processed by the browser because its mime type is not right.

So this didn't solve the problem.

patill:
Another update:

When running fresh with `deno task start` the compiler complains about the css file, but it then delivers it in its correct form.
The build process though stops due to a failure.

There should be a way to exclude some files from being processed.

[...]

PR #2959, opened 2025-05-16
feat(tailwindcss): add excludePaths option to skip specific directions #2937

SisyphusZheng:
**What does this PR do?**  
This PR adds an `excludePaths` configuration option to the Fresh Tailwind CSS plugin, allowing users to skip processing specific directories or files. This solves the conflict when hosting pre-built CSS files (e.g., from Vite) that have already been processed by Tailwind and should not be re-compiled by Fresh.

### **Problem Being Solved**  
#2937 Based on the scenario described in this issue, in certain use cases, users may not want Tailwind CSS to automatically process global CSS files. We can add a configuration option to allow skipping specific directories. Add the `excludePaths` option allows users to specify paths like `/static/assets/` to prevent Fresh from processing these files.

### **Changes Made**  
1. **Added Configuration Option**  
   - Extended `TailwindPluginOptions` in `types.ts` with `excludePaths?: string[]`.  
   - Example: `excludePaths: ["/static/assets/"]` skips all CSS files in the `static/assets` directory.

### DIFF
--- a/plugin-tailwindcss/src/mod.ts
+++ b/plugin-tailwindcss/src/mod.ts
@@ -12,7 +12,13 @@ export function tailwind<T>(
 
   builder.onTransformStaticFile(
     { pluginName: "tailwind", filter: /\.css$/ },
+    //a simple function to check if the path is in the excludePaths
+    //showcase: tailwind(builder, app, { excludePaths: ["/x/x"] });
     async (args) => {
+      if (options.excludePaths?.some((path) => args.path.includes(path))) {
+        return { content: args.text, map: undefined };
+      }
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Thanks for working on this! I feel like it would be useful to extend this option to work for every build plugin, not just tailwind. I've opened https://github.com/denoland/fresh/pull/2960 that makes this a general option and supports regex and glob patterns as well.

SisyphusZheng:
> Thanks for working on this! I feel like it would be useful to extend this option to work for every build plugin, not just tailwind. I've opened #2960 that makes this a general option and supports regex and glob patterns as well.

LGTM, i'll close this pr.

PR #2960, opened 2025-05-16
feat(core): support `exclude` option for transformers

### LINKED ISSUE #2937: Fresh with tailwindcss processes css files in static/assets

marvinhagemeister:
This PR adds general support for an `exclude` option to file transformers. This allows plugin to skip processing excluded files. The exclude array can be a `string`, a `RegExp` or a glob pattern.

Fixes https://github.com/denoland/fresh/issues/2937

### DIFF
--- a/plugin-tailwindcss/src/mod.ts
+++ b/plugin-tailwindcss/src/mod.ts
@@ -1,17 +1,17 @@
 import type { TailwindPluginOptions } from "./types.ts";
 import { initTailwind } from "./compiler.ts";
-import type { Builder } from "fresh/dev";
+import type { FreshBuilder } from "fresh/dev";
 import type { App } from "fresh";
 
 export function tailwind<T>(
-  builder: Builder,
+  builder: FreshBuilder,
   app: App<T>,
   options: TailwindPluginOptions = {},
 ): void {
   let processor: ReturnType<typeof initTailwind> | null;
 
   builder.onTransformStaticFile(
-    { pluginName: "tailwind", filter: /\.css$/ },
+    { pluginName: "tailwind", filter: /\.css$/, exclude: options.exclude },
     async (args) => {
       if (!processor) processor = initTailwind(app.config, options);
       const instance = await processor;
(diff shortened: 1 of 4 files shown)


SisyphusZheng:
May I inquire about the situation with test2.x? The fmt check appears to exhibit inconsistent behavior between the local environment and the other environment—using identical checks yields differing formatting requirements across these two setups. Interestingly, this issue has surfaced in the last three PRs, yet a review of the file modifications reveals no obvious changes or events that could have triggered this discrepancy.

marvinhagemeister:
@SisyphusZheng We had to unpublish Deno `2.3.2` because of the formatting changes, but forgot to do so for the place which the GitHub Deno action pulls from. It's resolved now.

PR #2961, opened 2025-05-16
ci: improve CI perf by adding dependency caching

csvn:
The `denoland/setup-deno` action recently added built-in caching (https://github.com/denoland/setup-deno/pull/89, https://github.com/denoland/setup-deno/pull/98). This makes it a lot simpler to add caching of the `DENO_DIR` during CI runs, without needing `@actions/cache` manual setup.

This should speed up CI runs by avoiding downloading dependencies between every run.


marvinhagemeister (approved):
LGTM, thanks for adding this!

PR #2964, opened 2025-05-16
fix: duplicate `@preact/signals` version

### LINKED ISSUE #2962: [v2.0] Bug: "Error: Cycle detected" when using `@preact/signals@2.x`
Can't use signals in Fresh v2 newly created project without downgrading the preact signals package.

This issue likely occurs because there ends up being two duplicate package versions, or a version mismatch between what Fresh and Islands use as signals version.

**See:**
- https://github.com/denoland/fresh/blob/main/deno.json#L53
- https://github.com/denoland/fresh/blob/main/init/src/init.ts#L8

## Reproduction steps

1. Generate new project
   ```sh
   deno run -Ar jsr:@fresh/init@2.0.0-alpha.33
   ```
2. Move `useSignal` count from prop to inside Island
   ```tsx
   import { useSignal } from "@preact/signals";
   import { Button } from "../components/Button.tsx";

   export default function Counter() {  
     const count = useSignal(3);
     return (
       <div class="flex gap-8 py-6">
         <Button onClick={() => count.value -= 1}>-1</Button>
         <p class="text-3xl tabular-nums">{count}</p>
         <Button onClick={() => count.value += 1}>+1</Button>
       </div>
     );
   }
   ```
3. Start dev server and open page 
4. Island fails to render with error:
   ```
   Uncaught (in promise) Error: Cycle detected
        at b.S (index.ts:778:3)
        at d3 (index.ts:49:36)
        at index.ts:172:2
        at index.ts:212:2
        at O (index.js:239:22)
        at E (render.js:42:2)
        at _render (reviver.ts:114:5)
   ```

## Workaround

Setting `@preact/signals` in `deno.json` to `npm:@preact/signals@^1` resolves the issue.

marvinhagemeister:
We initialized new projects with `@preact/signals@2.x` but Fresh itself was using `1.x`. There shall only ever be one version.

Kinda makes me wish Deno had some sort of `peerDependencies`.

Fixes https://github.com/denoland/fresh/issues/2962

### DIFF
--- a/www/deno.json
+++ b/www/deno.json
@@ -35,7 +35,7 @@
     "marked": "npm:marked@^14.1.2",
     "marked-mangle": "npm:marked-mangle@^1.1.9",
     "postcss": "npm:postcss@8.4.35",
-    "preact": "npm:preact@^10.24.1",
+    "preact": "npm:preact@^10.26.6",
     "preact-render-to-string": "npm:preact-render-to-string@^6.5.11",
     "prismjs": "npm:prismjs@^1.29.0",
     "tailwindcss": "npm:tailwindcss@^3.4.1"
--- a/www/deno.json
+++ b/www/deno.json
@@ -8,7 +8,7 @@
     "$ga4": "https://raw.githubusercontent.com/denoland/ga4/main/mod.ts",
     "@fresh/plugin-tailwind": "../plugin-tailwindcss/src/mod.ts",
     "@luca/esbuild-deno-loader": "jsr:@luca/esbuild-deno-loader@^0.11.0",
-    "@preact/signals": "npm:@preact/signals@^1.3.0",
+    "@preact/signals": "npm:@preact/signals@^2.0.4",
     "@std/crypto": "jsr:@std/crypto@1",
     "@std/datetime": "jsr:@std/datetime@0.225.0",
     "@std/encoding": "jsr:@std/encoding@1",
--- a/deno.json
+++ b/deno.json
@@ -50,7 +50,7 @@
     "$ga4": "https://raw.githubusercontent.com/denoland/ga4/main/mod.ts",
     "@luca/esbuild-deno-loader": "jsr:@luca/esbuild-deno-loader@^0.11.0",
     "@opentelemetry/api": "npm:@opentelemetry/api@^1.9.0",
-    "@preact/signals": "npm:@preact/signals@^1.2.3",
+    "@preact/signals": "npm:@preact/signals@^2.0.4",
     "esbuild": "npm:esbuild@0.23.1",
     "esbuild-wasm": "npm:esbuild-wasm@0.23.1",
     "@std/crypto": "jsr:@std/crypto@1",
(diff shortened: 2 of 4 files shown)


csvn:
> Kinda makes me wish Deno had some sort of peerDependencies.

Yes, and version ranges, e.g. `v1.x || v2.x`.

csvn:
I think tests will need to wait microtask/animation frame to match what Signals v2 had as breaking change. 🤔 

**Edit:** [@preact/signals@2.0.0](https://github.com/preactjs/signals/releases/tag/%40preact%2Fsignals%402.0.0)

> https://github.com/preactjs/signals/pull/604 [fea3e8d](https://github.com/preactjs/signals/commit/fea3e8da7a36944d87310678fad291aeacc55d8d) Thanks [@JoviDeCroock](https://github.com/JoviDeCroock)! - Defer all DOM updates by an animation frame, this should make it so
that any previously synchronous DOM update will be instead delayed by an
animation frame. This allows Preact to first perform its own render
cycle and then our direct DOM updates to occur. These will now
be performed in a batched way which is more performant as the browser
is prepared to handle these during the animation frame.

> 
> This does impact how Preact based signals are tested, when
you perform a signal update, you'll need to wrap it in act. In a way
this was always the case, as a signal update that resulted in
a Preact state update would require it to be wrapped in act, but
now this is the norm.

**Edit 2:** Perhaps using [`act()` from `preact/test-utils`](https://preactjs.com/guide/v10/unit-testing-with-enzyme/#triggering-state-updates-and-effects-with-act) can help, if there are signal changes in tests?

> To execute state updates and effects synchronously, use the `act` function from `preact/test-utils` to wrap the code that triggers the updates.

**Edit 3:** Nvm, seems it was another issue, multiple different signal versions that may have been the issue 😅

marvinhagemeister:
Upstream Deno issue for `peerDependencies` outside of npm packages https://github.com/denoland/deno/issues/26199

marvinhagemeister:
> I think tests will need to wait microtask/animation frame to match what Signals v2 had as breaking change. 🤔 

Nah, only the `www` e2e tests were failing. Turns out I forgot to increase the version number for signals in that.

www/deno.json
```diff
@@ -8,7 +8,7 @@
     "$ga4": "https://raw.githubusercontent.com/denoland/ga4/main/mod.ts",
     "@fresh/plugin-tailwind": "../plugin-tailwindcss/src/mod.ts",
     "@luca/esbuild-deno-loader": "jsr:@luca/esbuild-deno-loader@^0.11.0",
-    "@preact/signals": "npm:@preact/signals@^1.3.0",
+    "@preact/signals": "npm:@preact/signals@^2.0.4",
```
csvn, line 11: Is it not possible to remove this dependency from here instead? Will that not use the same version as the root `deno.json` instead via `workspaces`?

marvinhagemeister: Yeah there is a PR for that https://github.com/denoland/fresh/pull/2900 . Was kinda hoping we could have support for `workspace:*` syntax similar to pnpm in Deno by now, but it looks like that is getting pushed on the back burner.

PR #2966, opened 2025-05-16
Docs[Canary]: clean the twind part for canary version 

SisyphusZheng:
## What does this PR do?

This PR cleans all the Twind - related parts in the canary documentation. Specifically, it removes the Twind - related entries in the docs/toc.ts file, which streamlines the documentation and eliminates references to the Twind technology that might no longer be relevant or in use.

## Problem Being Solved
Outdated Information: Twind removed from fresh v2.  Keeping Twind - related content in the documentation can lead to confusion for new users who might be using the canary version of Fresh with Tailwind CSS.

## Changes Made
docs/toc.ts:
Removed the "migrating - to - tailwind" entry. 
Removed the "using - twind - v1" entry, which is directly related to using the Twind v1 version and is no longer needed in the canary documentation.
Removed the "using - fresh - canary - version" entry, which is no reason in canary docs


marvinhagemeister (approved):
Thanks!

PR #2972, opened 2025-05-18
tests: remove redundant `getStdOutput()` test utility

iuioiua:
Noticed while looking into an issue.


init/src/init_test.ts
```diff
@@ -245,9 +245,12 @@ Deno.test("init - errors on missing build cache in prod", async () => {
       stdout: "piped",
       stderr: "piped",
       cwd: dir,
+      env: {
+        NO_COLOR: "true",
```
iuioiua, line 249: Eliminates the need to strip ANSI codes.

iuioiua:
Closing as stale.

PR #2974, opened 2025-05-19
tests(init): remove `MockTTY`

iuioiua:
These changes simplify implementation logic of the init workflow and mock `prompt()` and `console()` functionality in a more direct manner. This makes things easier to reason about by jumping through fewer mental hoops.


marvinhagemeister (approved):
Nice, stubbing + `using` is a great pattern for this 👍

PR #2979, opened 2025-05-19
test(core): remove `FsAdapter`

iuioiua:
These changes simplify implementation logic and mock FS APIs in a more direct manner, making things easier to reason about.


src/plugins/fs_routes/mod.ts
```diff
@@ -62,60 +64,67 @@ export interface FsRoutesOptions {
...
-  ]);
+      const url = new URL(relative, "http://localhost/");
+      relRoutePaths.push(url.pathname.slice(1));
+    }
+  } catch (error) {
+    // `islandDir` or `routesDir` does not exist, so we can skip it
+    if (!(error instanceof Deno.errors.NotFound)) {
+      throw error;
+    }
+  }
```
iuioiua, line 127: Here, instead of checking whether the first argument passed to `walk()` is a directory, we just simply try to walk it. If it's not a directory, a `Deno.errors.NotFound` error is thrown and we ignore it. This has the same functionality as before, but just without having to use some `isDirectory()` function.

src/test_utils.ts
```diff
@@ -90,30 +88,4 @@ export function serveMiddleware<T>(
...
-          name: file, // FIXME?
-          path: file,
-        };
-        yield entry;
-      }
-    },
-    // deno-lint-ignore require-await
-    async isDirectory(dir) {
-      return Object.keys(files).some((file) => file.startsWith(dir + "/"));
-    },
```
iuioiua, line 112: Note: this mock is no longer exists, but that seems fine as tests pass.

src/dev/file_transformer_test.ts
```diff
@@ -52,9 +42,8 @@ function consumeResult(result: ProcessedFile[]) {
 }
 
 Deno.test("FileTransformer - transform sync", async () => {
-  const transformer = testTransformer({
-    "foo.txt": "foo",
-  });
+  const transformer = new FreshFileTransformer();
+  using _denoReadFileStub = stubDenoReadFile("foo");
```
Mrashes, line 46: first time seeing the using keyword, neat!

src/plugins/fs_routes/mod.ts
Mrashes: Is there any performance gain from avoiding the isDirectory function?

Mrashes (commented):
Overall change makes sense - as I'm not a contributor I don't feel confident in approving but good job on this!

src/plugins/fs_routes/mod.ts
iuioiua: Perhaps a slight performance benefit. But that's not the goal of the change. It just avoids the potential for hitting a (very unlikely) race condition by just attempting the operation and catching if it fails instead of checking if the operation can be performed then doing the operation if it can be. Does that make sense?

This was the exact topic of discussion for `fs.exists()` within the Standard Library. See the note in https://jsr.io/@std/fs/doc/~/exists.

iuioiua:
@marvinhagemeister PTAL

iuioiua:
While this makes testing and implementation much simpler, closing as stale.

PR #2982, opened 2025-05-20
fix: add back `LayoutConfig`

marvinhagemeister:
~~Makes it easier to upgrade from 1.x .~~

Initially, I assumed it would be just an alias to `RouteConfig`, but upon closer inspection that has some additional properties that make no sense for a layout.


iuioiua (approved):
LGTM

PR #2983, opened 2025-05-20
test: use disposable tmp dirs

csvn:
> [!TIP]
> Review without whitespace changes to have an easier time comparing changes

Replaces the patterns below with consistent `AsyncDisposable` and `await using` resource management for automatic cleanup.

- `await withTmpDir(async (dir) => { ... })` with `finally`
  - Makes code a bit flatter, and reduces callback-hell a bit
- `await Deno.makeTempDir()` + `Deno.remove()`
- `await Deno.makeTempDir()`

Helps be more consistent in removing temporary folders. Among others, ensures temp folders are removed in the tests below:

- `src/build_cache_test.ts`
- `src/dev/builder_test.ts`
- `src/dev/dev_build_cache_test.ts`

## TODO's

The `Deno.makeTempDir()` method is still used by `buildProd` as well, but this method is used in the top module scope, where it does not work correctly with `Deno.test` (it runs before the test methods). I removed this part from the PR and leave this for later to avoid making the PR too big.

- https://github.com/csvn/fresh/blob/test%2Fdisposable-tmp-dirs/tests/test_utils.tsx/#L64
- https://github.com/csvn/fresh/blob/test%2Fdisposable-tmp-dirs/www/main_test.ts/#L7

### DIFF
--- a/init/src/init_test.ts
+++ b/init/src/init_test.ts
@@ -8,6 +8,7 @@ import * as path from "@std/path";
 import { getStdOutput, withBrowser } from "../../tests/test_utils.tsx";
 import { waitForText } from "../../tests/test_utils.tsx";
 import { withChildProcessServer } from "../../tests/test_utils.tsx";
+import { withTmpDir as withTmpDirBase } from "../../src/test_utils.ts";
 import { stub } from "@std/testing/mock";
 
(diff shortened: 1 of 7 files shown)


marvinhagemeister (approved):
LGTM, thanks!

PR #2984, opened 2025-05-20
fix: convert island filePath to valid variable name

### LINKED ISSUE #2971: Frontend crash when using dash cased filenames for islands that use default export
Version: 2.0.0-alpha.34

```
Uncaught SyntaxError: Unexpected token '-'
```

because the compiled js looks like this:

```js
aceadca9fe0ebe8/fresh-runtime.js";import main-buttons from "/_fresh/js/2fc56a56443508c719f55aa91aceadca9fe0ebe8/main-buttons.js";impor
```

I have a feeling that the internal `app.island` should probably sanitize import names.

csvn:
I only added a unit test, but maybe it's testing internals too much? Should I add a more E2E test in `tests/` with real files?

Closes #2971

### DIFF
--- a/src/app_test.tsx
+++ b/src/app_test.tsx
@@ -501,3 +501,34 @@ Deno.test("FreshApp - throw when middleware returns no response", async () => {
   expect(res.status).toEqual(500);
   expect(text).toContain("Internal server error");
 });
+
+Deno.test("FreshApp - adding Island should convert to valid export names", () => {
+  const app = new App();
+  const islands = getIslandRegistry(app);
+
+  const component1 = () => <>OK</>;
+  const component2 = () => <>OK</>;
+  const component3 = () => <>OK</>;
+  app.island("/islands/foo.v2.tsx", "default", component1);
+  app.island("/islands/_bar-baz-...-$.tsx", "default", component2);
+  app.island("/islands/1_hello.tsx", "default", component3);
+
+  expect(islands.get(component1)!).toEqual({
+    file: "/islands/foo.v2.tsx",
+    name: "foo_v2",
+    exportName: "default",
+    fn: component1,
+  });
+  expect(islands.get(component2)!).toEqual({
+    file: "/islands/_bar-baz-...-$.tsx",
+    name: "_bar_baz_$",
+    exportName: "default",
+    fn: component2,
+  });
+  expect(islands.get(component3)!).toEqual({
+    file: "/islands/1_hello.tsx",
+    name: "_hello",
+    exportName: "default",
+    fn: component3,
+  });
+});
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -43,6 +43,8 @@ export function mergePaths(a: string, b: string) {
  * pathToExportName("/islands/nav-bar.tsx"); // "nav_bar"
  * pathToExportName("/islands/_.$bar.tsx");  // "_$bar"
  * pathToExportName("/islands/1.hello.tsx"); // "_hello"
+ * pathToExportName("/islands/collapse...repeat_-dash.tsx");
+ * // "collapse_repeat_dash"
  * ```
  */
 export function pathToExportName(filePath: string): string {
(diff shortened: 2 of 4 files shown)


src/app_test.tsx
```diff
@@ -501,3 +501,34 @@ Deno.test("FreshApp - throw when middleware returns no response", async () => {
...
+    exportName: "default",
+    fn: component2,
+  });
+  expect(islands.get(component3)!).toEqual({
+    file: "/islands/1_hello.tsx",
+    name: "_hello",
+    exportName: "default",
+    fn: component3,
+  });
+});
```
iuioiua, line 534: I think this is fine.
```suggestion
  const component = () => <>OK</>;
  app.island("/islands/foo.v2.tsx", "default", component);
  app.island("/islands/_bar-baz-...-$.tsx", "default", component);
  app.island("/islands/1_hello.tsx", "default", component);

  expect(islands.get(component1)!).toEqual({
    file: "/islands/foo.v2.tsx",
    name: "foo_v2",
    exportName: "default",
    fn: component,
  });
  expect(islands.get(component2)!).toEqual({
    file: "/islands/_bar-baz-...-$.tsx",
    name: "_bar_baz_$",
    exportName: "default",
    fn: component,
  });
  expect(islands.get(component3)!).toEqual({
    file: "/islands/1_hello.tsx",
    name: "_hello",
    exportName: "default",
    fn: component,
  });
});
```

src/utils_test.ts
iuioiua, line 1: Are you able to add a test for `/_{2,}/g`?

src/utils.ts
```diff
@@ -32,3 +32,22 @@ export function mergePaths(a: string, b: string) {
...
+ * pathToExportName("/islands/_.$bar.tsx");  // "_$bar"
+ * pathToExportName("/islands/1.hello.tsx"); // "_hello"
+ * ```
+ */
+export function pathToExportName(filePath: string): string {
+  const name = path.basename(filePath, path.extname(filePath));
+  // Regex for valid JS identifier characters
+  const regex = /^[^a-z_$]|[^a-z0-9_$]/gi;
+  return name.replaceAll(regex, "_").replaceAll(/_{2,}/g, "_");
+}
```
iuioiua:
```suggestion

const INVALID_JS_IDENTIFIER = /^[^a-z_$]|[^a-z0-9_$]/gi;
const DOUBLE_UNDERSCORE = /_{2,}/g

/**
 * Converts a file path to a valid JS export name.
 *
 * @example
 * ```ts
 * pathToExportName("/islands/foo.tsx");     // "foo"
 * pathToExportName("/islands/foo.v2.tsx");  // "foo_v2"
 * pathToExportName("/islands/nav-bar.tsx"); // "nav_bar"
 * pathToExportName("/islands/_.$bar.tsx");  // "_$bar"
 * pathToExportName("/islands/1.hello.tsx"); // "_hello"
 * ```
 */
export function pathToExportName(filePath: string): string {
  return path.basename(filePath, path.extname(filePath))
    .replaceAll(INVALID_JS_IDENTIFIER, "_")
    .replaceAll(DOUBLE_UNDERSCORE, "_");
}
```

src/utils.ts
```diff
@@ -32,3 +32,22 @@ export function mergePaths(a: string, b: string) {
...
+/**
+ * Converts a file path to a valid JS export name.
+ *
+ * @example
+ * ```ts
+ * pathToExportName("/islands/foo.tsx");     // "foo"
+ * pathToExportName("/islands/foo.v2.tsx");  // "foo_v2"
+ * pathToExportName("/islands/nav-bar.tsx"); // "nav_bar"
+ * pathToExportName("/islands/_.$bar.tsx");  // "_$bar"
+ * pathToExportName("/islands/1.hello.tsx"); // "_hello"
```
iuioiua, line 45: Able to add an example with `/_{2,}/g`?

iuioiua (changes requested):

src/app_test.tsx
csvn: Won't work. IslandsRegistry (`islands.get()`) is a Map, where the key is the component function. Each one has to be unique, or they overwrite each other. Using the same one for every island would result in only one island existing. Also, your suggested code has no reference to `component1`, `component2` and `component3`.

marvinhagemeister (approved):
LGTM, sweet this is great!

PR #2989, opened 2025-05-22
feat(core): `serverCache()` middleware

### LINKED ISSUE #8: Caching of pages

iuioiua:
Early reviews welcome.

Todo:
- [ ] Refine use in `www`
- [ ] Add test for respecting of `Vary` headers
- [ ] Add documentation
- [ ] Add to `init` workflow

Closes #8


src/middlewares/cache.ts
```diff
@@ -0,0 +1,92 @@
...
+ * The {@linkcode CacheOptions.include} option can be used to determine whether
+ * the current request should be cached or not based on the context argument.
+ *
+ * @param cache The cache to use.
+ * @param options Options for the cache middleware.
+ * @returns A middleware function that caches `GET` requests.
+ *
+ * @example Cache the whole site
+ *
+ * Here, `GET` requests to `/` will return the same response.
```
csvn: The example title and the description here kinda conflicts a bit. I think it can be less confusing if the description here is changed to something like:

```suggestion
 * @example Cache the whole site
 *
 * All `GET` requests will be cached, including File-System routes.
 * Thus requests to `/` will return the same response as long as
 * a cache entry for the `Request` exists.
```

It's also not explicit in the existing JSDoc that the cache is best-effort,
and is purely a performance optimization. Not a guarantee to always return
the same data. 🤔

src/middlewares/cache.ts
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * @example Cache specific routes
+ *
+ * ```ts
+ * import { App, cache } from "fresh";
+ *
+ * const webCache = await caches.open("my-cache");
+ *
+ * const app = new App()
+ *   .use("/cacheable", cache(webCache), () => new Response("Cached!"))
```
csvn: The `.use()` method does not support path nor multiple middlewares. This would have to be `.get()`.

src/middlewares/cache.ts
```diff
@@ -0,0 +1,92 @@
...
+ *   .use("/uncacheable", () => new Response("Not cached!"));
+ *
+ * await app.listen();
+ * ```
+ */
+export function cache<State>(
```
csvn: I wonder if a middleware called `cache` can be confused for a middleware that adds e.g. [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control), `Etag`, or other headers related to caching on the client.

Should this middleware eventually do both, or perhaps might it be better to name this something like `serverCache`, to indicate that we're returning cached `Response`'s?

src/middlewares/cache.ts
csvn, line 1: I think a lot of people use FS-based routes over `app.use()` to create pages for their site. But there's no example showing how that can be accomplished with this middleware without caching all routes (or via the `include` option).

It could be great to have an example of how to cache everything under e.g. `routes/docs/` with something like this:

```ts
/**
 * @example Cache everything under `/route/docs`
 *
 * Add the `cache()` middleware via file `/route/docs/_middleware.ts`.
 * Any GET requests under `/route/docs/*` will automatically be cached.
 *
 * ```ts
 * export default [
 *   cache(await caches.open("my-cache"))
 * ];
 * ```
 */
```

I think this might be one of the more realistic scenarios to use caching for.

src/middlewares/cache.ts
```diff
@@ -0,0 +1,92 @@
...
+  cache: Cache,
+  options?: CacheOptions<State>,
+): MiddlewareFn<State> {
+  return async (ctx) => {
+    if (
+      ctx.req.method !== "GET" || (options?.include && !options.include(ctx))
+    ) return ctx.next();
```
csvn: By checking `options.include` eagerly, it becomes impossible for a route to e.g. set a flag that it can be cached.

```ts
// /routes/foo.tsx
export default define.page(ctx => {
  ctx.state.cacheable = true;
  return <>{crypto.randomUUID()}</>;
});
```

```ts
// /routes/_middleware.ts
export default cache(await caches.open('c'), {
  // Does not work, always `undefined`
  include: (ctx) => ctx.state.cacheable ?? false
});
```

This becomes very limiting, as pretty much the only thing we can use to determine if a route is to be cached is the `ctx.url`, or any middleware state set before the `cache()` middleware. I think it might be better to instead use `include` for whether or not to add something to cache.

```ts
  if (options?.include(ctx) ?? false)) await cache.put(ctx.req, res.clone());
```

1. Check if request method is cacheable
2. Check if Response exists in cache, return it if it does
3. Add Response to cache if cacheable
4. Return Response

src/middlewares/cache.ts
iuioiua: Great suggestions. Thanks!

src/middlewares/cache.ts
iuioiua: That's a great point too. Changing to `serveCache()` for clarity.

iuioiua (commented):
Great suggestions! Thank you.

DontMash:
Why was this abandoned?

iuioiua:
I no longer have the time to contribute. Feel free to copy any code from here, if you'd like to create your own PR 🙂

DontMash:
> I no longer have the time to contribute. Feel free to copy any code from here, if you'd like to create your own PR 🙂

Based on what you've already done, what is still left to do? Is the Todo-list in the description up-to-date?

iuioiua:
It was a little while ago but yes, the todo list is up-to-date, IIUC.

PR #2990, opened 2025-05-22
feat[init]: Add "nodeModulesDir": "auto" to the deno.json in the freshv2 init script #2973 #2813

### LINKED ISSUE #2813: When using tailwind, the deno.json generates deprecated nodeModulesDir
1. Update deno to latest version with `deno upgrade`
2. Create a brand new fresh project with `deno run -A -r https://fresh.deno.dev` and select tailwind and vscode
3. cd into project
4. `deno task start`
5. terminal throws this error: Warning `"nodeModulesDir": true` is deprecated in Deno 2.0. Use `"nodeModulesDir": "auto"` instead
6. Simple fix: `"nodeModulesDir": true` should be `"nodeModulesDir": "auto"`
7. _Super awesome fix_: fresh devs could update the way fresh generates the deno.json file so I don't have to do this every time. 

P.S. I snooped around at the files in the fresh repo but I couldn't find where it incorrectly creates this part of the deno.json file. sorry for not putting up a merge request, i would have if i was better at this XD

P.P.S. Shout out to [Marvin ](https://github.com/marvinhagemeister) the goat.

### LINKED ISSUE #2973: Islands Not Working in Latest Fresh 2 with Preact 10.26.6
Even with a brand new generated application ex:`deno run -Ar jsr:@fresh/init@2.0.0-alpha.34` the generated code doesn't work (console error and interacting with buttons in the island don't do anything). 

Console error:
```
fresh-runtime.js:29 Uncaught (in promise) ReferenceError: n is not defined
    at u (fresh-runtime.js:29:53)
    at l.__b (fresh-runtime.js:96:7)
    at l.__b (fresh-runtime.js:222:11)
    at c2.__b (parse.ts:130:1)
    at index.ts:831:1
    at O (constants.js:2:8)
    at E (constants.js:2:8)
    at _render (reviver.ts:114:5)
```

Downgrading to preact 10.26.5 on my personal Fresh 2 application things start to work again.

SisyphusZheng:
The issue in #2973 appears to be caused by multiple cached versions of dependencies. After directly deleting the ~/.cache/deno directory, the problem was resolved. Enabling nodeModulesDir: "auto" can also avoid this issue. Upon checking the requirement in #2813, it suggests changing "nodeModules": "true" to the recommended "auto" in Fresh v1, while Fresh v2 does not have this setting. Considering these two issues, the configuration in Fresh v1, and future support for Fresh plugins (such as Tailwind v4), I believe adding "nodeModulesDir": "auto" to the init script would be beneficial in many aspects.

Also re-edit the test_utils to add functions for re-deleting the tmp directory and setting timeouts, due to permission issues on Github CI testing in Win env.

Closes #2973
Closes #2813


src/test_utils.ts
```diff
@@ -122,10 +122,48 @@ export async function withTmpDir(
...
+
   return {
     dir,
     async [Symbol.asyncDispose]() {
-      await Deno.remove(dir, { recursive: true });
+      try {
+        await safeRemoveWithTimeout(dir, 5000);
+      } catch (e) {
+        throw e;
+      }
```
csvn: I don't think we need to throw errors at all if we fail to remove temp dirs. Logging a warning should be enough IMO. For CI, it doesn't really matter if we clean up. We could perhaps even check for a env variable and skip cleaning up in CI.

I also think it's overkill to have a timeout for 5s before skipping removal, the current retry logic only waits ~500ms, and the deleted folders are not very big. I think a simplified version could look something more like this:

```ts
    async [Symbol.asyncDispose]() {
      if (Deno.env.get("CI") === "true") {
        // deno-lint-ignore no-console
        console.log(`Skipping cleanup of temp dir in CI: ${dir}`);
        return;
      }

      let retries = 3;
      while (retries > 0) {
        try {
          await Deno.remove(dir, { recursive: true });
          break;
        } catch (e) {
          if (!(e instanceof Deno.errors.PermissionDenied)) break;
        }
        retries -= 1;
        await delay(100);
      }

      // deno-lint-ignore no-console
      console.warn(`Failed to remove temp dir: ${dir}`);
    }
```

> I think aborting file deletion should be simpler though, added a feature request to support `signal` with `AbortSignal` to Deno CLI: https://github.com/denoland/deno/issues/29417. `AbortSignal.timeout(5000)` should make code like that much simpler

- Skip removing temp dirs in CI
- Only use 3 retries with a short delay
- Use `delay()` fn for delay to make code easier to read
- Use `console.warn` over `throw err` when removing folder fails after retries

SisyphusZheng: THX! I'm not really sure how to proceed, just wanna fix the ci prob in win env, I'll re-edit according to your suggestions.

csvn: I'm not a maintainer, so you don't have to (@marvinhagemeister might have other feedback). But I think it should help mitigate the CI issues on Windows better if we just skip the cleanup, as well as being faster 😃

GitHub actions runs inside docker images anyway, and tmp folders should be discarded between runs.

SisyphusZheng: > I'm not a maintainer, so you don't have to (@marvinhagemeister might have other feedback). But I think it should help mitigate the CI issues on Windows better if we just skip the cleanup, as well as being faster 😃
> 
> GitHub actions runs inside docker images anyway, and tmp folders should be discarded between runs.

I see that you've been involved in editing the relevant functions and Ur ideas are far better than mine. I only made a tmp patch for the tmp dir permission issue in the Win env on GitHub CI. It would undoubtedly be best to minimize the corresponding side effects.

IMOP, it would be best that someone to review and help my Pr. 🎇

SisyphusZheng:
--patch-1
As suggested by @cvsn, the code has been provided to fix the CI failure caused by permission denial of temporary files in the Windows environment on GitHub CI.

marvinhagemeister:
Looks like this was already merged as part of https://github.com/denoland/fresh/pull/3054

PR #2993, opened 2025-05-22
feat: add flag to FreshContext for partial requests

csvn:
Addresses a gap between [Fresh v1](https://github.com/denoland/fresh/blob/1.x/src/server/types.ts#L192) and Fresh v2, by adding `ctx.isPartial` to `FreshContext`.

The same route (e.g. `GET /api/autocomplete`) can be called from both an Island or a Partial. Returning JSON or HTML depending on the destination of the code.

Or is the intent that endpoints returning partials and data should be separated, and this pattern is not to be encouraged in Fresh v2? My use case in Fresh v1 was an `/routes/api/cart.tsx`, which can either just toggle a product from the cart (via cookie), but also update UI elements (e.g. cart icon) based on if called from partial or not.

image

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,3 +1,4 @@
 export const INTERNAL_PREFIX = "/_frsh";
 export const DEV_ERROR_OVERLAY_URL = `${INTERNAL_PREFIX}/error_overlay`;
 export const ALIVE_URL = `${INTERNAL_PREFIX}/alive`;
+export const PARTIAL_SEARCH_PARAM = "fresh-partial";
--- a/src/context.ts
+++ b/src/context.ts
@@ -14,7 +14,7 @@ import {
   RenderState,
   setRenderState,
 } from "./runtime/server/preact_hooks.tsx";
-import { DEV_ERROR_OVERLAY_URL } from "./constants.ts";
+import { DEV_ERROR_OVERLAY_URL, PARTIAL_SEARCH_PARAM } from "./constants.ts";
 import { BUILD_ID } from "./runtime/build_id.ts";
 import { tracer } from "./otel.ts";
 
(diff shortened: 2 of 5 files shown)


src/context.ts
```diff
@@ -126,6 +135,7 @@ export class FreshReqContext<State>
     this.info = info;
     this.params = params;
     this.config = config;
+    this.isPartial = url.searchParams.has(PARTIAL_SEARCH_PARAM);
```
csvn, line 138: Given that this state can be derived from `url`, can be replaced with a `get` method instead.

src/context.ts
```diff
@@ -90,14 +98,15 @@ export let getBuildCache: (ctx: FreshContext<unknown>) => BuildCache;
...
+  readonly config: ResolvedFreshConfig;
+  readonly url: URL;
+  readonly req: Request;
+  readonly params: Record<string, string>;
+  readonly state: State = {} as State;
   data: unknown = undefined;
   error: unknown | null = null;
-  info: Deno.ServeHandlerInfo | Deno.ServeHandlerInfo;
+  readonly info: Deno.ServeHandlerInfo | Deno.ServeHandlerInfo;
+  readonly isPartial: boolean;
```
csvn, line 109: Most of the properties could safely add `readonly` here to match the interface without TS errors. But I can revert this if preferred, but I think it can help minimize mistakes.

src/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const INTERNAL_PREFIX = "/_frsh";
 export const DEV_ERROR_OVERLAY_URL = `${INTERNAL_PREFIX}/error_overlay`;
 export const ALIVE_URL = `${INTERNAL_PREFIX}/alive`;
+export const PARTIAL_SEARCH_PARAM = "fresh-partial";
```
csvn, line 4: Files in `src/runtime` already imported from this file, so it seemed reasonable to move this constant here so we could avoid it being hard-coded in `src/context.ts`.

marvinhagemeister (approved):
Good catch, looks like we forgot about that property. Thanks for adding it 👍

PR #2994, opened 2025-05-22
fix: improve styling of "Edit on GitHub" link for better visibility in Dark mode

SisyphusZheng:
change style for dark mode
last:
image
now:
image

### DIFF
--- a/www/routes/docs/[...slug].tsx
+++ b/www/routes/docs/[...slug].tsx
@@ -221,7 +221,7 @@ export default define.page<typeof handler>(function DocsPage(props) {
                   <div class="px-4 md:px-0 flex justify-between my-6">
                     <a
                       href={`https://github.com/denoland/fresh/edit/main/${page.file}`}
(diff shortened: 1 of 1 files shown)


www/routes/docs/[...slug].tsx
Comment (csvn) [inline www/routes/docs/[...slug].tsx:1]: I don't get the same result as your PR description. This is how it looks for me on https://fresh.deno.dev/docs/introduction with both dark/light:

image

You don't have "Dark Reader" or some other extension running in the background?

Comment (SisyphusZheng) [reply www/routes/docs/[...slug].tsx:1]: None，Recheck in my phone, in safari and chrome, get same error style. Not sure why it is different with ur env.imageimage

Comment (csvn) [reply www/routes/docs/[...slug].tsx:1]: Ahh, I see now. This cannot be reproduced when in Dark mode as system default:

- System preference: Light, Site: Light ✅
- System preference: Light, Site: Dark ❌
- System preference: Dark, Site: Light ✅
- System preference: Dark, Site: Dark ✅

I got the same issue if I changed to Light as system default, and had the site in dark mode. I noticed this since your screenshots used a light toolbar, that that was likely required.

marvinhagemeister (approved):
Sweet, thanks!

PR #2999, opened 2025-05-28
Feat[WWW]: add hello bar in docs part

SisyphusZheng:
## Why was this PR initiated?

Currently, there is too little documentation on Fresh v2, and users often need to refer to the latest News of Fresh v2. Therefore, it may be meaningful to also place the latest News in the Docs section, so that users can realize it is the most important reference at present.

## What does this PR do? 
It separates the HelloBar and the Sidebar in Slug.tsx into independent components and islands, adds the HelloBar at the top of the docs section, and enables the Sidebar to rebound according to height during the dropdown process.

## How was the testing conducted? 
Through Deno task ok and local environment style checks.

## Preview
image
image

### DIFF
--- a/www/routes/docs/[...slug].tsx
+++ b/www/routes/docs/[...slug].tsx
@@ -17,6 +17,8 @@ import { TableOfContents } from "../../islands/TableOfContents.tsx";
 import SearchButton from "../../islands/SearchButton.tsx";
 import VersionSelect from "../../islands/VersionSelect.tsx";
 import { define } from "../../utils/state.ts";
+import HelloBar from "../../components/HelloBar.tsx";
+import DynamicSidebar from "../../islands/DynamicSidebar.tsx";
 
 interface Data {
   page: Page;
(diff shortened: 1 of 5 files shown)


www/routes/docs/[...slug].tsx
```diff
@@ -149,96 +151,84 @@ export default define.page<typeof handler>(function DocsPage(props) {
...
-                <div class="lg:order-1 min-w-0 max-w-3xl w-full">
-                  <h1 class="text-4xl text-foreground-primary tracking-tight font-bold md:mt-0 px-4 md:px-0 mb-4">
-                    {page.title}
-                  </h1>
-                  <div
-                    class="markdown-body mb-8"
-                    // deno-lint-ignore react-no-danger
-                    dangerouslySetInnerHTML={{ __html: html }}
-                  />
+          <DynamicSidebar page={page} categories={CATEGORIES[page.version]} />
```
Comment (Mrashes) [inline www/routes/docs/[...slug].tsx:184]: Why make categories a prop rather than have it live in the DynamicSidebar island?  Feels like we would benefit from reduced prop drilling here.

Mrashes (commented):
Personally prefer named exports but that's not biding.  Think this clean up makes a lot of sense.  Not a contributor so don't feel comfortable approving but well done!

www/routes/docs/[...slug].tsx
Comment (SisyphusZheng) [reply www/routes/docs/[...slug].tsx:184]: Your suggestions were very effective. I have made some modifications based on your two suggestions, but during the process, I found that I had to deal with similar issues in code written by others, and the incomplete work has affected the tests. Therefore, I hope to carry out further cleanup work later, or you can also join in the relevant cleanup. Thank you for your review.

bartlomieju:
Thanks for the contribution! The idea of showing the HelloBar on the docs page is reasonable, but the current approach has some issues that would need significant rework — mainly turning the sidebar into a client-side island just for a scroll animation effect (which could be done with pure CSS `sticky` positioning), and the lock file containing unrelated dependency changes.

If you'd like to revisit this, a simpler approach would be to just extract the HelloBar component and add it to the docs layout, using CSS for any scroll-based positioning rather than a JavaScript island. Happy to help review a revised version!

SisyphusZheng:
> Thanks for the contribution! The idea of showing the HelloBar on the docs page is reasonable, but the current approach has some issues that would need significant rework — mainly turning the sidebar into a client-side island just for a scroll animation effect (which could be done with pure CSS `sticky` positioning), and the lock file containing unrelated dependency changes.
> 
> If you'd like to revisit this, a simpler approach would be to just extract the HelloBar component and add it to the docs layout, using CSS for any scroll-based positioning rather than a JavaScript island. Happy to help review a revised version!

It's great to hear from you. May I ask if you have taken over the Fresh project? Please feel free to let me know if there is anything else I can assist with. I have always been very interested in the Fresh project.

PR #3003, opened 2025-05-30
Fix[WWW] Solve the Navigation Problem #3002

### LINKED ISSUE #3002: Navigating to `https://fresh.deno.dev/docs/concepts/partials` breaks the docs
Is this happening because fresh is detecting the partial code inside the code blocks? 😮 

Anyways, as you can see below, clicking on Partials from the side bar seemingly jumbles up markdown sections on all docs pages until one fully reloads.

https://github.com/user-attachments/assets/1a0a79a2-763a-497c-87bb-203c1c4b4920

SisyphusZheng:
## Problem:
The problem was reproduced as described in #3002. Chrome DevTools shows the following error:

props.js:137 Uncaught (in promise) InvalidCharacterError: Failed to execute 'setAttribute' on 'Element': 'docs-content""' is not a valid attribute name.
    at j (props.js:137:8)
    at V (index.js:528:5)
    at O (index.js:326:28)
    at I (children.js:97:16)
    at V (index.js:547:4)
    at O (index.js:326:28)
    at I (children.js:97:16)
    at V (index.js:547:4)
    at O (index.js:326:28)
    at I (children.js:97:16)

## Solution:
The issue was resolved by removing the double quotes around docs-content in the Markdown image description. After this change, the application resumed normal operation.

## Test:
Deno task ok and local env check

## Close:
close #3002

### DIFF
--- a/docs/latest/concepts/partials.md
+++ b/docs/latest/concepts/partials.md
@@ -76,7 +76,7 @@ Let's use a typical documentation page layout as an example. It often features a
 main content area and a sidebar of links to switch between pages of the
 documentation (marked green here).
 
-![A sketched layout of a typical documentation page with the sidebar on the left composed of green links and a main content area on the right. The main content area is labeled as Partial "docs-content"](/docs/fresh-partial-docs.png)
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Nice catch, thanks!

PR #3005, opened 2025-05-31
Fix[core] compiler option not recognized in monorepo problem#3004

### LINKED ISSUE #3004: `jsxImportSource` compiler option not recognized in monorepo setup
## Description

When running Fresh in a monorepo setup, the `jsxImportSource` compiler option is not properly recognized from the application's `deno.json`, even though it's correctly configured. Fresh seems to be looking for this option in the root `deno.json` instead.

## Steps to Reproduce

1. Set up a monorepo with Fresh application in a subdirectory
3. Run `deno task dev` from the application directory

## Expected Behavior

Fresh should recognize the `jsxImportSource` compiler option from the application's `deno.json`.

## Actual Behavior

Error occurs:
error: Uncaught (in promise) Error: Option compilerOptions > jsxImportSource not set in: /xx/xx/root/deno.json
## Workaround

Adding `"workspace": []` to the application's `deno.json` resolves the issue by disconnecting it from the monorepo workspace settings.

## Environment

- Deno version: 2.3.3 (stable, release, aarch64-apple-darwin)
- Fresh version: 2.0.0-alpha.34
- Operating System: macOS 15.5.0 (Darwin)

## Reproduction Repository

[repo](https://github.com/betamasi/fresh-monorepo-bug
)
## Additional Context

Full reproduction steps and configuration details can be found in the [repository's README](https://github.com/betamasi/fresh-monorepo-bug/blob/main/README.md).

SisyphusZheng:
Resolves issue where jsxImportSource from subproject's deno.json was not recognized, as Fresh was incorrectly using the root workspace config. 

Closes: #3004

### DIFF
--- a/src/dev/builder.ts
+++ b/src/dev/builder.ts
@@ -228,6 +228,8 @@ export async function readDenoConfigForCompilerOptions(
 ): Promise<{ config: DenoConfig; filePath: string }> {
   let dir = directory;
   const configs: { config: DenoConfig; filePath: string }[] = [];
+  
+  // Collect all config files from current directory up to workspace root
   outer: while (true) {
     for (const name of ["deno.json", "deno.jsonc"]) {
       const filePath = path.join(dir, name);
@@ -240,6 +242,8 @@ export async function readDenoConfigForCompilerOptions(
           config = JSON.parse(file);
         }
         configs.push({ config, filePath });
+        
+        // Stop at workspace root (config with workspace property)
         if (config.workspace) {
           break outer;
         }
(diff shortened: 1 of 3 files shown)


src/dev/builder.ts
```diff
@@ -261,30 +265,36 @@ export async function readDenoConfigForCompilerOptions(
...
+    mergedConfig.compilerOptions = {
+      ...mergedConfig.compilerOptions,
+      ...appConfig.config.compilerOptions,
+    };
+  }
+
+  // Always preserve workspace setting from the appropriate config
+  if (appConfig.config.workspace !== undefined) {
+    mergedConfig.workspace = appConfig.config.workspace;
+  } else if (rootConfig.config.workspace !== undefined) {
```
marvinhagemeister, line 290: This seems overly complicated. We have an array of config objects and we know that the first one is the app config (aka workspace member). So you can iterate over that array and check for every items if the necessary JSX compiler options are set. When you encounter such an config item, or when the config item as `workspace` set, abort.

SisyphusZheng: Cool thx, i will fix it in next week

marvinhagemeister:
Thanks for working on this. We needed this fix right now and went ahead by merging https://github.com/denoland/fresh/pull/3096 instead.

PR #3009, opened 2025-06-04
fix(update): `fresh/runtime` import specifier updated incorrectly

SisyphusZheng:
…ports

The original code is shown in the following figure, and there is an incorrect usage in it.


marvinhagemeister (approved):
Nice catch! Looks like a copy & paste error caused the bug

PR #3010, opened 2025-06-04
feat(update): More logs for update script #2968 

### LINKED ISSUE #2968: Suggestion: Fresh 2 migration needs additional logging
I'm running the fresh upgrader from 1 => 2.  Before full release this should have some more logging as after listing my project directory I didn't know if anything was happening.  Either some logging saying what is happening or some progress bar would be really appriciated.

I've only just run the tool, will document if I run into any other bumps.

SisyphusZheng:
Before:
340fffdbf558fa7a1bc1036972e0679

After:
Case1:
1adc44294f2001dc43cd411254dfc2d
Case2:
8cd3445877d5311ff8d9b9f66adb30e

In Test Env:
df5b20de6d1437e471c6dcfc1e55178

Close #2968

### DIFF
--- a/update/src/update.ts
+++ b/update/src/update.ts
@@ -179,30 +179,23 @@ export async function updateProject(dir: string) {
   await Promise.all(sfs.map(async (sourceFile) => {
     try {
       const wasModified = await updateFile(sourceFile);
-      completedFiles++;
       if (wasModified) {
         modifiedFilesList.push(sourceFile.getFilePath());
       }
+
+      return wasModified;
+    } catch (err) {
+      // deno-lint-ignore no-console
+      console.error(`Could not process ${sourceFile.getFilePath()}`);
+      throw err;
+    } finally {
+      completedFiles++;
 
       // update progress bar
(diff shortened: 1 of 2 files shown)


update/src/update.ts
```diff
@@ -127,23 +146,104 @@ export async function updateProject(dir: string) {
...
+
+      return wasModified;
     } catch (err) {
+      completedFiles++;
+
+      // update progress bar (even on failure)
+      const progress = createProgressBar(completedFiles, sfs.length);
+      Deno.stdout.writeSync(
+        new TextEncoder().encode(`\r${colors.blue(progress)} Complete!`),
+      );
```
marvinhagemeister: Nit: Move this into a `finally` block. That way we don't need to duplicate the code to clear the progress bar twice.

marvinhagemeister (commented):
This is great! Left one small nit and then it's ready to go in 👍

marvinhagemeister (approved):
LGTM, thanks for adding this 💯

Issue #3011, opened 2025-06-04
suggestion: Expansion of official middleware for Fresh


Octo8080X:
Fresh 2 is changing to an API design, similar to frameworks like Express and Hono.
In Fresh 1, the community and developers were free to develop middleware and plug-ins.

In the case of Hono, the availability of a variety of official middleware is considered one of the reasons for its popularity.

Why don't you introduce official middleware to Fresh, too?
At first, I think porting some of the middleware available in Hono would be a good starting point.  
basic, CSRF, cache, CORS ex.

The repository for managing them doesn't necessarily have to be part of the main Fresh repository.

For example, denoland/fresh-extensions.
I would be willing to dedicate my efforts to porting middleware if the community, including myself, shares the same understanding.

Would providing official middleware go against Fresh's philosophy or policy?

I would really appreciate hearing thoughts on this.

marvinhagemeister:
> Would providing official middleware go against Fresh's philosophy or policy?

No, go for it. Whilst we might not want every single middleware shipping with Fresh itself, the popular ones like cors, basic auth and other stuff Hono has is very reasonable.

Octo8080X:
@marvinhagemeister 
Thank you for your response!

I would like to contribute to Fresh by porting middleware such as CORS and Basic Auth.

I look forward to working with you.

PR #3013, opened 2025-06-06
Add CORS middleware 

Octo8080X:
## Why was this PR initiated?

I opened an issue and held discussions about officially providing middleware in Fresh by porting middleware from Hono.

https://github.com/denoland/fresh/issues/3011

## What does this PR do? 

I introduced the CORS middleware bundled with Fresh by porting the CORS middleware from Hono.

I appreciate your review.

### DIFF
--- a/src/middlewares/cors.ts
+++ b/src/middlewares/cors.ts
@@ -157,7 +157,6 @@ export const cors = (options?: CORSOptions) => {
       });
     }
 
-    console.log(`CORS Middleware: ${ctx.req.method} ${ctx.req.url}`);
     // For non-OPTIONS requests
     const actualResponse = await ctx.next();
 
--- a/src/middlewares/cors.ts
+++ b/src/middlewares/cors.ts
@@ -1,4 +1,5 @@
-import { FreshContext } from "fresh";
+import type { FreshContext } from "fresh";
+import type { MiddlewareFn } from "./mod.ts";
 
 type CORSOptions = {
   origin:
@@ -51,7 +52,7 @@ type CORSOptions = {
  * // ];
  * ```
  */
-export const cors = (options?: CORSOptions) => {
+export function cors<T>(options?: CORSOptions): MiddlewareFn<T> {
   const defaults: Partial<CORSOptions> = {
     origin: "*",
     allowMethods: ["GET", "HEAD", "PUT", "POST", "DELETE", "PATCH"],
@@ -181,4 +182,4 @@ export const cors = (options?: CORSOptions) => {
 
     return actualResponse;
   };
-};
+}
--- a/src/middlewares/cors.ts
+++ b/src/middlewares/cors.ts
@@ -52,16 +52,13 @@ export type CORSOptions = {
  * ```
  */
 export function cors<T>(options?: CORSOptions): MiddlewareFn<T> {
-  const defaults: Partial<CORSOptions> = {
+  const opts: CORSOptions = {
     origin: "*",
     allowMethods: ["GET", "HEAD", "PUT", "POST", "DELETE", "PATCH"],
     allowHeaders: [],
     exposeHeaders: [],
-  };
-  const opts = {
-    ...defaults,
     ...options,
-  } as CORSOptions;
+  };
 
   const findAllowOrigin = ((optsOrigin: CORSOptions["origin"]) => {
     if (typeof optsOrigin === "string") {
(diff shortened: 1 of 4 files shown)


src/middlewares/cors.ts
```diff
@@ -0,0 +1,185 @@
...
+        optsOrigin.includes(requestOrigin) ? requestOrigin : null;
+    }
+  })(opts.origin);
+
+  return async (ctx: FreshContext): Promise<Response> => {
+    const responseHeaders = new Headers();
+
+    function setResponseHeader(key: string, value: string) {
+      responseHeaders.set(key, value);
+    }
```
marvinhagemeister: Nit: Inline this function. A `Map.set()` doesn't need to be abstracted away.

src/middlewares/cors.ts
```diff
@@ -0,0 +1,185 @@
...
+    // For non-OPTIONS requests
+    const actualResponse = await ctx.next();
+
+    // Apply common CORS headers (ACAO, Credentials, ExposeHeaders)
+    responseHeaders.forEach((value, key) => {
+      // Do not overwrite Vary if it was set by downstream, merge instead
+      if (key.toLowerCase() !== "vary") {
+        actualResponse.headers.set(key, value);
+      }
+    });
```
marvinhagemeister: Doing a loop here is more expensive than a map lookup. The `Headers` class already takes care of casing:

```sh
> let a = new Headers()
undefined
> a.set("Vary", "foo")
undefined
> a.get("vary")
"foo"
> a.get("VARY")
"foo"
>
```

src/middlewares/cors_test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+  app.get("/api6/abc", (_ctx: FreshContext) => { // Added /api6/abc route
+    return Response.json({ success: true });
+  });
+
+  it("GET default", async () => {
+    const res = await new FakeServer(app.handler()).handler(
+      new Request("https://localhost/api/abc"),
+      CONN_INFO,
+    );
```
marvinhagemeister: Nit: There is no need for instantiating `FakeServer` in this whole test file, since we're not using any of the `.get()/.post()` methods that it provides.

And since we're using the same app for every test we can just do:

```ts
const handler = app.handler()

// ...later in a test
const res = await handler(new Request(...))
```

We don't need to pass `CONN_INFO` either as that is optional and will fall back to a default value. I don't see any test cases which rely on particular values being passed there.

src/middlewares/cors.ts
```diff
@@ -0,0 +1,185 @@
+import type { FreshContext } from "fresh";
+import type { MiddlewareFn } from "./mod.ts";
+
+type CORSOptions = {
```
marvinhagemeister: Nit: This is part of the public API, so we should export it in the top level package.

src/middlewares/cors.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    // Merge our calculated varyValues with any existing Vary from downstream response
+    const existingVary = actualResponse.headers.get("Vary");
+    if (existingVary) {
+      existingVary.split(/\s*,\s*/).forEach((v) => varyValues.add(v));
+    }
+
+    if (varyValues.size > 0) {
+      actualResponse.headers.set("Vary", Array.from(varyValues).join(", "));
+    }
```
marvinhagemeister: In the case where the downstream response has a `Vary` header, but we don't want to add any additional values in the cors middleware, we're still going to split it and set it back to the exact same value here. We should avoid that.

marvinhagemeister (commented):
This is great! Thanks for thinking of adding tests 💯 

Left some minor nitpicky comments, but I feel like this is close to landing!

src/middlewares/cors.ts
Octo8080X: I exported the contents of type and also updated src/mod.ts to export them as well.

src/middlewares/cors_test.ts
Octo8080X: I refactored the code to remove the use of FakeServer.

src/middlewares/cors.ts
Octo8080X: I inlined what was previously a function.

src/middlewares/cors.ts
Octo8080X: @marvinhagemeister 
Thank you for the advice.
I believe you're saying that "it would be better to delete it" — is my understanding correct?

src/middlewares/cors.ts
Octo8080X: I fundamentally refactored it to eliminate the need for looping. Thanks!

marvinhagemeister (approved):
Sweet, thanks for doing this 👍

jub0bs:
`ctx.req.method === "OPTIONS"` is too restrictive: not all OPTIONS requests are CORS-preflight requests. More things to look out for in https://jub0bs.com/posts/2023-02-08-fearless-cors/

PR #3014, opened 2025-06-07
fix[core] path checking and recursive call issues in MemoryBuildCache.readFile method

SisyphusZheng:
## Problem Description
When processing transformed files, the pathname variable is redefined inside the for loop. After the loop ends, the check uses the last pathname value from the loop, rather than the originally requested file pathname. 

This may result 

1. in the file content not being returned correctly even if the originally requested file has been properly processed and added to #processedFiles.
2. potential infinite recursion problem

## Solution
Use a new variable transformedPathname to store the pathname of the currently processed file, and check in each loop whether the pathname of the currently processed file is the same as the originally requested file pathname.

### DIFF
--- a/src/dev/dev_build_cache.ts
+++ b/src/dev/dev_build_cache.ts
@@ -99,12 +99,16 @@ export class MemoryBuildCache implements DevBuildCache {
             `Processed file resolved outside of static dir ${file.path}`,
           );
         }
-        const pathname = `/${relative}`;
+        const transformedPathname = `/${relative}`;
 
-        this.addProcessedFile(pathname, file.content, null);
-      }
-      if (this.#processedFiles.has(pathname)) {
-        return this.readFile(pathname);
+        this.addProcessedFile(transformedPathname, file.content, null);
+
+        if (
+          transformedPathname === pathname &&
+          this.#processedFiles.has(transformedPathname)
+        ) {
+          return this.readFile(transformedPathname);
+        }
       }
     } else {
       try {


src/dev/dev_build_cache.ts
```diff
@@ -99,12 +99,16 @@ export class MemoryBuildCache implements DevBuildCache {
...
-      }
-      if (this.#processedFiles.has(pathname)) {
-        return this.readFile(pathname);
+        this.addProcessedFile(transformedPathname, file.content, null);
+
+        if (
+          transformedPathname === pathname &&
+          this.#processedFiles.has(transformedPathname)
+        ) {
+          return this.readFile(transformedPathname);
```
marvinhagemeister, line 110: I'm a bit worried about moving this inside the loop. The `return` here will prematurely exit the function if the first file isn't transformed.

SisyphusZheng: i will recheck that later, thx for review

bartlomieju:
Thanks for looking into this, but the bug this PR aims to fix doesn't actually exist.

The `const pathname` on line 150 is **block-scoped** to the for loop body (that's how `const`/`let` work in JavaScript). After the loop ends, `pathname` on line 154 correctly refers to the original function parameter from line 93, not the loop variable. There's no variable leakage.

The infinite recursion concern is also not an issue — the recursive `readFile(pathname)` call immediately hits the `#processedFiles.get(pathname)` check at line 94 and returns the cached result on the first call.

This change would also introduce a regression: by moving the check inside the loop with an early return, it would skip adding remaining files from the `transformed` array to `#processedFiles` if the requested file happens to appear before other output files.

SisyphusZheng:
> Thanks for looking into this, but the bug this PR aims to fix doesn't actually exist.
> 
> The `const pathname` on line 150 is **block-scoped** to the for loop body (that's how `const`/`let` work in JavaScript). After the loop ends, `pathname` on line 154 correctly refers to the original function parameter from line 93, not the loop variable. There's no variable leakage.
> 
> The infinite recursion concern is also not an issue — the recursive `readFile(pathname)` call immediately hits the `#processedFiles.get(pathname)` check at line 94 and returns the cached result on the first call.
> 
> This change would also introduce a regression: by moving the check inside the loop with an early return, it would skip adding remaining files from the `transformed` array to `#processedFiles` if the requested file happens to appear before other output files.

Thx for response and ur hard work for fresh, rely wanna see the fresh v2 official version, anything I can do, just let me know, thx

PR #3015, opened 2025-06-07
fix[core]Re-point Tailwind CSS to the local repository.

SisyphusZheng:
This reference was incorrectly modified in the previous PR. As a source code repository, it perhaps shouldn't rely on the published files.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -81,7 +81,7 @@
 
     "ts-morph": "npm:ts-morph@^25.0.1",
 
-    "@fresh/plugin-tailwind": "jsr:@fresh/plugin-tailwind@^0.0.1-alpha.7",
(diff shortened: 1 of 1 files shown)


deno.json
```diff
@@ -81,7 +81,7 @@
 
     "ts-morph": "npm:ts-morph@^25.0.1",
 
-    "@fresh/plugin-tailwind": "jsr:@fresh/plugin-tailwind@^0.0.1-alpha.7",
+    "@fresh/plugin-tailwind": "./plugin-tailwindcss/src/mod.ts",
```
marvinhagemeister, line 84: This is fine the way it was before, because the tailwind plugin is a workspace member. I'm kinda neutral which way to go as from Deno's perspective there is no difference.

SisyphusZheng:
yep,thx for reply

PR #3016, opened 2025-06-07
fix[utils]restore the screenshot function that failed due to missing dependencies.

SisyphusZheng:
The relevant dependencies were lost in previous cleanups. This situation caused the screenshot function to fail. 

This PR properly cleans up the functional function and restores its operation.


marvinhagemeister (approved):
Nice, thanks!

PR #3018, opened 2025-06-10
Add CSRF middleware

Octo8080X:
## Why was this PR initiated?

I opened an issue and held discussions about officially providing middleware in Fresh by porting middleware from Hono.

https://github.com/denoland/fresh/issues/3011

## What does this PR do? 

I introduced the CSRF middleware bundled with Fresh by porting the CSRF middleware from Hono.

I appreciate your review.


src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,423 @@
...
+  let simplePostHandler = spy(rawSimplePostHandler);
+
+  beforeEach(() => {
+    simplePostHandler = spy(rawSimplePostHandler);
+  });
+
+  describe("simple usage", () => {
+    const app = new App();
+
+    //
```
marvinhagemeister: Nit: empty comment

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,423 @@
...
+    app.all("*", csrf());
+    app.get("/form", (_ctx: FreshContext) => new Response("<form></form>"));
```
marvinhagemeister: Nit: If you don't care about the context arg, you can leave it out. Also no need to type it explicitely.

```ts
app.get("/form", () => new Response("<form></form>"));
```

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,423 @@
...
+            body: JSON.stringify({ name: "fresh" }),
+          }),
+        );
+
+        expect(res.status).toBe(200);
+        expect(await res.text()).toBe("fresh");
+      });
+
+      it('should be 403 for "Application/x-www-form-urlencoded" cross origin', async () => {
+        const res = await app.handler()(
```
marvinhagemeister: Nit: Let's be consistent. Other tests re-use the `handler()` function

```suggestion
        const res = await handler(
```

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,423 @@
...
+         * <script>
+         * fetch('/form', {
+         *   method: 'POST',
+         *   headers: {
+         *     'content-type': 'application/x-www-form-urlencoded',
+         *   },
+         *   body: 'name=fresh',
+         * });
+         * </script>
+         */
```
marvinhagemeister: Nit: I'd delete these comments. I feel like they don't add any additional information that isn't already present a few lines later.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,94 @@
+import type { FreshContext } from "../context.ts";
+
+type IsAllowedOriginHandler = (
```
marvinhagemeister: This type is part of the public API through `CSRFOptions`. We need to export this too otherwise publishing will fail because of a private interface.

marvinhagemeister (commented):
This is great, left some minor nit comments. Let me know what you think 👍

src/middlewares/csrf_test.ts
Octo8080X: I’ve removed the part you pointed out—thank you!

src/middlewares/csrf_test.ts
Octo8080X: I overlooked that—thank you!

src/middlewares/csrf_test.ts
Octo8080X: That was a result of bringing over comments from the original source.
I’ve removed them—thank you!

src/middlewares/csrf.ts
Octo8080X: I had received your comments during the CORS work as well—thank you! I've exported it.

Octo8080X:
@marvinhagemeister 
> This is great, left some minor nit comments.

Thank you—I'm glad to hear you say that.
But I’ll always remember how great the original Hono is—it’s something I’ll keep close to heart.

I've updated the parts you mentioned—when you have a moment, I'd really appreciate it if you could take another look.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,94 @@
...
+  origin: string,
+  context: FreshContext,
+) => boolean;
+export interface CSRFOptions {
+  origin?: string | string[] | IsAllowedOriginHandler;
+}
+
+const isSafeMethodRe = /^(GET|HEAD)$/;
+const isRequestedByFormElementRe =
+  /^\b(application\/x-www-form-urlencoded|multipart\/form-data|text\/plain)\b/i;
```
iuioiua: This is a personal preference but I like to avoid regexes as much as possible. They can be difficult to understand and can be buggy. Even the simple ones. I'd suggest:

```ts
function isSafeMethod(method: string): boolean {
  return method === "GET" || method === "POST";
}
```

Of even, for brevity:
```ts
function isSafeMethod(method: string): boolean {
  return ["GET", "POST"].includes(method);
}
```

You could even inline this logic. No need for separate functions.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/**
+ * CSRF Protection Middleware for Fresh.
+ * Adapted from Hono's CORS middleware.
+ * Would like to express my gratitude to all the developers of Hono.
+ *
+ * @param {CSRFOptions} [options] - The options for the CSRF protection middleware.
+ * @param {string|string[]|(origin: string, context: Context) => boolean} [options.origin] - Specify origins.
+ * @returns {MiddlewareHandler} The middleware handler function.
```
iuioiua:
```suggestion
 * @param options The options for the CSRF protection middleware.
 * @returns The middleware handler function.
```
Type information isn't needed in JSDocs for TypeScript.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,94 @@
+import type { FreshContext } from "../context.ts";
+
+export type IsAllowedOriginHandler = (
+  origin: string,
+  context: FreshContext,
+) => boolean;
+export interface CSRFOptions {
+  origin?: string | string[] | IsAllowedOriginHandler;
+}
```
iuioiua: Are we able to document these types?

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,94 @@
...
+    ) {
+      return new Response(
+        "Forbidden: CSRF protection failed. The request is not allowed from this origin.",
+        {
+          status: 403,
+        },
+      );
+    }
+
+    return await ctx.next();
```
iuioiua:
```suggestion
  return (ctx: FreshContext): Promise<Response> => {
    if (
      !isSafeMethodRe.test(ctx.req.method) &&
      isRequestedByFormElementRe.test(
        ctx.req.headers.get("content-type") || "text/plain",
      ) &&
      !isAllowedOrigin(ctx.req.headers.get("origin"), ctx)
    ) {
      return new Response(
        "Forbidden: CSRF protection failed. The request is not allowed from this origin.",
        {
          status: 403,
        },
      );
    }

    return ctx.next();
```
Nit

iuioiua (changes requested):

src/middlewares/csrf.ts
marvinhagemeister: +1 for going with simple if checks here
-1 on using an array allocation + iterating over it with `.includes()`

src/middlewares/csrf.ts
marvinhagemeister: Doesn't matter really which way to go.

src/middlewares/csrf.ts
Octo8080X: I chose the simpler approach.
You mentioned it was fine to use it in-lined, but it was a bit too lengthy, so I made it into a method instead.

src/middlewares/csrf.ts
Octo8080X: I added some documentation about the typings. Thank you!

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+  context: FreshContext,
+) => boolean;
+
+/**
+ * Options for configuring the CSRF protection middleware.
+ *
+ * @property origin - Specifies the allowed origin(s) for requests. Can be a string, an array of strings, or a custom handler function.
+ *   - string: A single allowed origin.
+ *   - string[]: Multiple allowed origins.
+ *   - IsAllowedOriginHandler: A function to determine if an origin is allowed.
```
iuioiua: This should be defined within the interface ([example](https://github.com/denoland/std/blob/0ad72c39d773c05ae8d61958c65dc36756f5f1da/http/file_server.ts#L150-L156)).

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export interface CSRFOptions {
```
iuioiua:
```suggestion
export interface CsrfOptions {
```
Names for public interfaces in the Deno runtime and Standard Library APIs use camelcasing, even for abbreviations. E.g. `Deno.serveHttp`, `Deno.listenTls`, etc. Ref: https://github.com/denoland/std/pull/2582

src/middlewares/csrf_test.ts
iuioiua, line 1: Tests should be wrapped in `Deno.test()`. See other tests in the codebase.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+    origin: string | undefined | null,
+    ctx: FreshContext,
+  ) => {
+    if (origin === undefined || origin === null) {
+      return false;
+    }
+    return handler(origin, ctx);
+  };
+
+  return async (ctx: FreshContext): Promise<Response> => {
```
iuioiua:
```suggestion
  return (ctx: FreshContext): Promise<Response> => {
```
Again, doesn't seem to need to be async here.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+    ) {
+      return new Response(
+        "Forbidden: CSRF protection failed. The request is not allowed from this origin.",
+        {
+          status: 403,
+        },
+      );
+    }
+
+    return await ctx.next();
```
iuioiua: Ditto.
```suggestion
    return ctx.next();
```

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+        ctx.req.headers.get("content-type") || "text/plain",
+      ) &&
+      !isAllowedOrigin(ctx.req.headers.get("origin"), ctx)
+    ) {
+      return new Response(
+        "Forbidden: CSRF protection failed. The request is not allowed from this origin.",
+        {
+          status: 403,
+        },
+      );
```
iuioiua:
```suggestion
      throw new HttpError(403, "CSRF protection failed. The request is not allowed from this origin.");
```
Allows the user to handle the error in their own way.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+  })(options?.origin);
+  const isAllowedOrigin = (
+    origin: string | undefined | null,
+    ctx: FreshContext,
+  ) => {
+    if (origin === undefined || origin === null) {
+      return false;
+    }
+    return handler(origin, ctx);
+  };
```
iuioiua: This is hard to follow, to me, at least. Would this work?
```suggestion
  const isAllowedOrigin = (
    origin: string | undefined | null,
    ctx: FreshContext,
  ) => {
    if (!origin) return false;
    const optsOrigin = options?.origin;
    if (!optsOrigin) {
      return origin === new URL(ctx.req.url).origin;
    }
    if (typeof optsOrigin === "string") {
      return origin === optsOrigin;
    }
    if (typeof optsOrigin === "function") {
      return optsOrigin(origin, ctx);
    }
    return Array.isArray(optsOrigin) && optsOrigin.includes(origin);
  };
```

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,117 @@
...
+ * // Function
+ * // It is strongly recommended that the protocol be verified to ensure a match to `$`.
+ * // You should *never* do a forward match.
+ * app.use(
+ *   '*',
+ *   csrf({
+ *     origin: (origin) => /https:\/\/(\w+\.)?myapp\.example\.com$/.test(origin),
+ *   })
+ * )
+ * ```
```
iuioiua: Perhaps these would be better done as separate `@example`s.

iuioiua (changes requested):

src/middlewares/csrf.ts
marvinhagemeister: I like the suggestion, but I would replace `new URL(ctx.req.url).origin` with `ctx.url.origin`. No need to allocate another `URL` instance when there already is one under `ctx.url`

src/middlewares/csrf.ts
marvinhagemeister: This makes no difference

src/middlewares/csrf.ts
marvinhagemeister: Makes no difference

src/middlewares/csrf.ts
iuioiua: Ah, missed that I did that. Yep!

src/middlewares/csrf.ts
iuioiua: Yeah, to be clear, this is a nit. Though, adding `async` when not needed needlessly wraps the return value in a promise.

src/middlewares/csrf_test.ts
Octo8080X: I updated it to use Deno.test and #step.
Thank you!

src/middlewares/csrf.ts
Octo8080X: I switched to using HttpError.

src/middlewares/csrf.ts
Octo8080X: Based on your suggestion, I made some corrections.
I’m also using ctx.url.origin as you recommended.

src/middlewares/csrf.ts
Octo8080X: I made some fixes.
Do they align with what you intended?

src/middlewares/csrf.ts
iuioiua: Pinging this again.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+  const contentTypeLower = contentType.toLowerCase();
+
+  return contentTypeLower === "application/x-www-form-urlencoded" ||
+    contentTypeLower === "multipart/form-data" ||
+    contentTypeLower === "text/plain";
+}
+
+/**
+ * CSRF Protection Middleware for Fresh.
+ * Adapted from Hono's CORS middleware.
```
iuioiua: Maybe worth linking to Hono's documentation or API reference for this middleware?

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+ * Would like to express my gratitude to all the developers of Hono.
```
iuioiua:
```suggestion
```
I don't think a personal note is needed in documentation for a symbol.

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param options Options for the CSRF protection middleware.
+ * @returns The middleware handler function.
+ *
+ * @example Specifying origins with using `origin` option: default
```
iuioiua:
```suggestion
 * @example Basic usage (with defaults)
```
Maybe add a clear and concise explanation of the defaults right above the snippet?

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+ * ```ts
+ * const app = new App<State>()
+ *
+ * app.use(csrf())
+ * ```
+ *
+ * @example Specifying origins with using `origin` option: string
```
iuioiua:
```suggestion
 * @example Specifying static origins
```

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @example Specifying origins with using `origin` option: function
+ * It is strongly recommended that the protocol be verified to ensure a match to `$`.
+ * You should *never* do a forward match.
+ *
+ * ```ts
+ * app.use(
+ *   '*',
+ *   csrf({
+ *     origin: (origin) => /https:\/\/(\w+\.)?myapp\.example\.com$/.test(origin),
```
iuioiua: This makes me think that the option should accept regex instead of a function 🤔

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,121 @@
...
+ * app.use(
+ *   csrf({
+ *     origin: ['https://myapp.example.com', 'http://development.myapp.example.com'],
+ *   })
+ * )
+ * ```
+ *
+ * @example Specifying origins with using `origin` option: function
+ * It is strongly recommended that the protocol be verified to ensure a match to `$`.
+ * You should *never* do a forward match.
```
iuioiua:
```suggestion
 * @example Specifying more complex origins
```
I think any explanation should be clear, concise and lie between the example title and the snippet.

iuioiua (changes requested):
Nits but looking better!

src/middlewares/csrf.ts
Octo8080X: I made some changes.
It’s not much different from the static origins example — what do you think?
```ts
 *   csrf({
 *     origin: (origin) => ['https://myapp.example.com', 'http://development.myapp.example.com'].includes(origin),
 *   })
```

Octo8080X:
@iuioiua 

I've made revisions based on all the comments you provided.
Could you please review it when you have a chance? Thank you!

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,363 @@
...
+import { csrf } from "./csrf.ts";
+import { expect } from "@std/expect";
+import { spy } from "@std/testing/mock";
+
+const rawSimplePostHandler = async (ctx: FreshContext) => {
+  if (ctx.req.headers.get("content-type") === "application/json") {
+    return new Response((await ctx.req.json())["name"]);
+  } else {
+    return new Response((await ctx.req.formData()).get("name"));
+  }
```
iuioiua:
```suggestion
  return ctx.req.headers.get("content-type") === "application/json"
    ? new Response((await ctx.req.json())["name"])
    : new Response((await ctx.req.formData()).get("name"));
```

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,363 @@
...
+};
+
+Deno.test("CSRF by Middleware", async (t) => {
+  let simplePostHandler = spy(rawSimplePostHandler);
+
+  //beforeEach(() => {
+  //  console.log("Resetting simplePostHandler spy");
+  //  simplePostHandler = spy(rawSimplePostHandler);
+  //});
```
iuioiua:
```suggestion
```

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,115 @@
...
+ *     origin: (origin) => ['https://myapp.example.com', 'http://development.myapp.example.com'].includes(origin),
+ *   })
+ * )
+ * ```
+ */
+export function csrf(
+  options?: CsrfOptions,
+): (ctx: FreshContext) => Promise<Response> {
+  const isAllowedOrigin = (
+    origin: string | undefined | null,
```
iuioiua:
```suggestion
    origin?: string | null,
```

src/middlewares/csrf.ts
```diff
@@ -0,0 +1,115 @@
...
+ * @param origin - The origin of the incoming request.
+ * @param context - The FreshContext object for the current request.
+ * @returns A boolean indicating if the origin is allowed.
+ */
+export type IsAllowedOriginHandler = (
+  origin: string,
+  context: FreshContext,
+) => boolean;
+
+/** Options for configuring the CSRF protection middleware. **/
```
iuioiua:
```suggestion
/** Options for {@linkcode csrf}. **/
```

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,363 @@
...
+    app.post("/form", simplePostHandler);
+    app.put("/form", () => new Response("OK"));
+    app.delete("/form", () => new Response("OK"));
+    app.patch("/form", () => new Response("OK"));
+    app.head("/form", () => new Response("OK"));
+
+    const handler = app.handler();
+
+    await t.step("GET /form - should be 200 for any request", async () => {
+      simplePostHandler = spy(rawSimplePostHandler);
```
iuioiua: Is this actually needed? What happens if we remove this line? Ditto for the other test steps.

src/middlewares/csrf_test.ts
```diff
@@ -0,0 +1,363 @@
...
+const rawSimplePostHandler = async (ctx: FreshContext) => {
+  if (ctx.req.headers.get("content-type") === "application/json") {
+    return new Response((await ctx.req.json())["name"]);
+  } else {
+    return new Response((await ctx.req.formData()).get("name"));
+  }
+};
+
+Deno.test("CSRF by Middleware", async (t) => {
+  let simplePostHandler = spy(rawSimplePostHandler);
```
iuioiua:
```suggestion
  using simplePostHandler = spy(rawSimplePostHandler);
```
Might not better but maybe best to do.

iuioiua (approved):
All just non-blocking nits. LGTM!

src/middlewares/csrf_test.ts
Octo8080X: I had written it out a bit formally, but as you pointed out, it was unnecessary. Thank you!

src/middlewares/csrf_test.ts
Octo8080X: spy doesn’t support the using keyword, so using let seems to be the best solution. Thanks!

marvinhagemeister:
Digging a bit more into the overall topic of CSRF, a friend pointed me to this excellent writeup by the go team: https://github.com/golang/go/issues/73626#issue-3046320918

Couple of thoughts compared to this PR:
- We're missing the `OPTIONS` method
- Not sure how `Content-Type` checks help. Do you know more about that?
- We don't support the newer `Sec-Fetch-Site` header

I feel like we should follow GO's approach.

GO implementation: https://github.com/golang/go/blob/5dac42363ba8281a3f4f08e03af2292b763adc38/src/net/http/csrf.go#L122-L163
The GO tests might also be useful to take a look at https://github.com/golang/go/blob/master/src/net/http/csrf_test.go

marvinhagemeister (approved):
LGTM, thanks for adding this!

PR #3020, opened 2025-06-11
fix: import current version of fresh

iuioiua:
Previously, all packages except `@fresh/core` and even some tests might've imported `jsr:@fresh/core@^2.0.0-alpha.29` rather than the latest version of Fresh. This PR fixes that.


deno.json
```diff
@@ -83,9 +83,6 @@
 
     "@fresh/plugin-tailwind": "jsr:@fresh/plugin-tailwind@^0.0.1-alpha.7",
     "@std/front-matter": "jsr:@std/front-matter@^1.0.5",
-    "fresh/compat": "./src/compat/mod.ts",
-    "fresh/dev": "./src/dev/mod.ts",
-    "fresh/runtime": "./src/runtime/shared.ts",
```
iuioiua, line 88: These imports are redundant because of the `fresh` import that's already defined.

marvinhagemeister (approved):
LGTM

marvinhagemeister (approved):

PR #3023, opened 2025-06-12
fix(update): correct return and argument types for `defineFn()`

iuioiua:
Previously, the return union type of the inner function and the function itself contained a `Request` instead of a `Response`.


marvinhagemeister (approved):
Thanks!

PR #3027, opened 2025-06-12
feat:use std lib to replace the custom processbar 

### LINKED ISSUE #3021: suggestion: use the Standard Library's progress bar instead of own implementation
`@std/update` uses its own progress bar.

https://github.com/denoland/fresh/blob/cb402b983985f0b26be59fc97ca19188aee38ef7/update/src/update.ts#L21-L29

We could instead use [`@std/cli/unstable-progress-bar`](https://jsr.io/@std/cli/doc/unstable-progress-bar), which is better tested and will be improved over time.

This is a good first issue.

SisyphusZheng:
- The constructor of the process-bar in the currently used std/cli differs from that of the latest version, so the relevant dependencies have been updated.

- The custom processbar has been replaced with the std lib, and customization has been carried out to meet the current scenario.

In test env, seem like: 
bb2868ed3911071ce9915a9c8662786

close #3021

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -42,7 +42,7 @@
   },
   "imports": {
     "@deno/doc": "jsr:@deno/doc@^0.172.0",
-    "@std/cli": "jsr:@std/cli@^1.0.17",
+    "@std/cli": "jsr:@std/cli@^1.0.19",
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Thanks!

PR #3029, opened 2025-06-12
test: add fresh_version_check test

### LINKED ISSUE #3024: suggestion: check that `fresh` import is correct
I think there should be a check that the `fresh` import uses the latest version, that is the version defined in the `version` field of the root `deno.json` file. This would prevent the need for fixes like the one in #3020.

SisyphusZheng:
add a test file to auto check the fresh_version, based on deno.json --version.
40ab0eb376190e1533307e5422bc293

test blocked by #3020
close #3024 

--patch-1 
update based on reivew

Simplified version consistency test that only checks if the version field matches the fresh import version in deno.json, and current implementation provides clear visual difference representation.

41875d5bb2b944dbb25f484a814963c

--patch-2
update based on reivew
make it simple, and fix a little bug to make it run fine


iuioiua (commented):
Respectfully, this seems very over-engineered. My original idea was to just check that, in the root `deno.json` file, the `version` field matched the version of the `fresh` import within that same file. This can be accomplished in just a few lines of code.

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,175 @@
...
+      files.push(filePath);
+    } catch {
+      // File doesn't exist or is not accessible, ignore
+    }
+  }
+
+  return files;
+}
+
+Deno.test("Fresh version consistency", async () => {
```
iuioiua: I very much like the idea of this check being wrapped in a test. Nice!

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,26 @@
+import { resolve } from "jsr:@std/path";
+import { assertEquals } from "jsr:@std/assert";
+
+Deno.test("Fresh version consistency", async () => {
+  const config = JSON.parse(await Deno.readTextFile(resolve("deno.json")));
```
iuioiua:
```suggestion
import { assertEquals } from "jsr:@std/assert";
import config from "../deno.json" with { type: "json" };

Deno.test("Fresh version consistency", async () => {
```

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,26 @@
...
+  // get the expected version from the deno.json file
+  const expectedVersion = config.version;
+  if (!expectedVersion) {
+    throw new Error("No version field found in deno.json");
+  }
```
iuioiua:
```suggestion
```
The `version` field will always be defined, unless some major structural changes happen to the codebase. So, safe to remove.

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // get the version from the fresh import
+  const freshImport = config.imports?.fresh;
+  if (!freshImport) {
+    throw new Error("No fresh import found in deno.json");
+  }
```
iuioiua:
```suggestion
```
Ditto, this can be removed. Safe to assume `fresh` will be a defined import.

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  const match = freshImport.match(/@fresh\/core@\^?([^/]+)/);
+  const actualVersion = match?.[1];
```
iuioiua:
```suggestion
  const [_, actualVersion] = freshImport.split("@^");
```
Simpler

iuioiua (changes requested):

tests/fresh_versions_test.ts
```diff
@@ -0,0 +1,26 @@
...
+Deno.test("Fresh version consistency", async () => {
+  const config = JSON.parse(await Deno.readTextFile(resolve("deno.json")));
+  // get the expected version from the deno.json file
+  const expectedVersion = config.version;
+  if (!expectedVersion) {
+    throw new Error("No version field found in deno.json");
+  }
+
+  // get the version from the fresh import
+  const freshImport = config.imports?.fresh;
```
iuioiua:
```suggestion
  const freshImport = config.imports.fresh;
```
Ditto, safe to assume `imports` field will always be defined.

iuioiua (approved):
Cool!

bartlomieju (approved):

PR #3030, opened 2025-06-13
chore: change `patch` Deno config file option to `links`

iuioiua:
Ref: https://github.com/denoland/deno/pull/29677


marvinhagemeister (approved):
LGTM

PR #3032, opened 2025-06-13
fefactor: refactor sreenshot and add test

SisyphusZheng:
- Refactor screenshot functionality into testable utility functions
- Add comprehensive test suite to catch regressions early
- Ensure functionality doesn't break unexpectedly without notification

--patch-1
based on review, make throw error naturally


www/utils/screenshot.ts
```diff
@@ -1,32 +1,63 @@
...
+    validateUrl(url);
+    await captureScreenshot(url, id);
+    // deno-lint-ignore no-console
+    console.log(`Screenshot saved as ${id}1x.jpg and ${id}2x.jpg`);
+  } catch (error) {
+    const message = error instanceof Error ? error.message : String(error);
+    // deno-lint-ignore no-console
+    console.log(`Error: ${message}`);
+    Deno.exit(1);
+  }
```
iuioiua: This `catch` block seems redundant. Why not just let the program throw naturally?

www/utils/screenshot.ts
```diff
@@ -1,32 +1,63 @@
...
-const image2x = await Image.decode(raw);
-await Deno.writeFile(
-  `./www/static/showcase/${id}2x.jpg`,
-  await image2x.encodeJPEG(80),
-);
+    await Deno.writeFile(path2x, await image2x.encodeJPEG(80));
+    const image1x = image2x.resize(image2x.width / 2, Image.RESIZE_AUTO);
+    await Deno.writeFile(path1x, await image1x.encodeJPEG(80));
+  } finally {
+    await browser.close();
```
iuioiua, line 45: Is this line and the `try`/`catch` blocks needed? Does the browser close even if the script fails anyway?

iuioiua (changes requested):

www/utils/screenshot.ts
marvinhagemeister: It's fine. The browser would be killed by the parent process regardless, but there is nothing wrong with allowing child processes to gracefully shutdown and run some cleanup code themselves.

www/utils/screenshot.ts
marvinhagemeister: +1 on erroring naturally.

marvinhagemeister (approved):
Sweet, thanks!

PR #3034, opened 2025-06-13
fix(core): move to new esbuild deno plugin

marvinhagemeister:
Move to the new esbuild plugin that leverages `@deno/loader` under the hood. This is based on the exact same rust crates that Deno itself uses to do resolution, instead of doing it's own resolution.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -51,7 +51,6 @@
     "preact": "npm:preact@^10.26.6",
     "preact-render-to-string": "npm:preact-render-to-string@^6.5.11",
     "$ga4": "https://raw.githubusercontent.com/denoland/ga4/main/mod.ts",
-    "@luca/esbuild-deno-loader": "jsr:@luca/esbuild-deno-loader@^0.11.0",
     "@opentelemetry/api": "npm:@opentelemetry/api@^1.9.0",
     "@preact/signals": "npm:@preact/signals@^2.0.4",
     "esbuild": "npm:esbuild@0.25.4",
--- a/deno.json
+++ b/deno.json
@@ -42,6 +42,7 @@
   },
   "imports": {
     "@deno/doc": "jsr:@deno/doc@^0.172.0",
+    "@deno/esbuild-plugin": "jsr:@deno/esbuild-plugin@^1.0.0",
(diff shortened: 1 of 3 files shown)


deno.json
csvn, line 1: Nit: Shouldn't `@luca/esbuild-deno-loader` be removed on L54? It's unused in the project now from what I can see.

src/dev/esbuild.ts
```diff
@@ -1,4 +1,4 @@
-import { denoPlugins } from "@luca/esbuild-deno-loader";
+import { denoPlugin } from "@deno/esbuild-plugin";
```
csvn, line 1: Awesome! Looking forward even more though to being able to run `deno bundle` for Fresh and it working out of the box 😁

deno.json
marvinhagemeister: You're right, there is. some work in progress here. I forgot to mark this as a draft.

src/dev/esbuild.ts
marvinhagemeister: Yeah we've been chatting about that but we'd need plugin support in `deno bundle` for fresh. Let's see what the future holds. For now I'm just excited to replace the prior esbuild loader with something more closer with how Deno operates 🎉

csvn: Yeah, I think `@deno/loader` was a great move, as well as the plugins [`@deno/rolldown-loader`](https://jsr.io/@deno/rolldown-plugin) (vite support too) + [`@deno/esbuild-plugin`](https://jsr.io/@deno/esbuild-plugin). That likely covers ~90% of all bundling scenarios, as well as avoiding `node_modules` when using Deno more often!

csvn: > Yeah we've been chatting about that but we'd need plugin support in deno bundle for fresh.

I didn't think of plugins, which indeed makes it harder to support. But perhaps it could be solved later on by supporting it via a programmatic `Deno.bundle` API? That would make it similar to how `$ deno serve` + `Deno.serve` both has CLI + programmatic API.

iuioiua:
Would we be able to get a new alpha version published to try this out?

PR #2392, opened 2024-04-02
feat: Support startServer for unix domain sockets

mcandeia:

PR #3035, opened 2025-06-14
feat: add IP filter middleware

### LINKED ISSUE #3011: suggestion: Expansion of official middleware for Fresh

Octo8080X:
Add an IP restriction middleware for Fresh, ported from Hono's IP restriction middleware.

Supports deny lists, allow lists, and CIDR subnet matching. Deny rules take precedence over allow rules.

```ts
app.use(ipFilter({
  denyList: ["192.168.1.10", "10.0.0.0/8"],
  allowList: ["192.168.1.0/24"],
}));
```

Closes #3011


src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,395 @@
...
+  address: string;
+  addressType?: AddressType;
+};
+
+export interface ConnInfo {
+  /**
+   * Remote information
+   */
+  remote: NetAddrInfo;
+}
```
iuioiua: These types already exist within the Deno runtime API. See https://docs.deno.com/api/deno/network.

src/middlewares/ip_restriction_test.ts
iuioiua, line 1: Each test should be wrapped in a `Deno.test()` call.

src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,395 @@
...
+      }
+      return blockError();
+    }
+    if (allowMatcher(remoteData)) {
+      const res = await ctx.next();
+      return res;
+    }
+
+    if (allowLength === 0) {
+      return await ctx.next();
```
iuioiua:
```suggestion
      return ctx.next();
```

src/middlewares/ip_restriction.ts
iuioiua, line 1: There's a lot of IP address logic here that might be valuable in other domains. I recommend creating an issue for the Standard Library for any useful utilities to [`@std/net`](https://jsr.io/@std/net).

iuioiua (changes requested):
First pass

src/middlewares/ip_restriction.ts
marvinhagemeister: makes no difference

src/middlewares/ip_restriction_test.ts
Octo8080X: @iuioiua san. 

Thanks for your comment!
Sorry if I’m mistaken, but I thought when writing in a BDD style, Deno.test would be replaced by describe. Was I incorrect about that?

https://jsr.io/@std/testing/doc/bdd#:~:text=()%2C%2018)%3B%0A%7D)%3B-,Mixed%20test%20grouping,-Both%20nested%20test

src/middlewares/ip_restriction.ts
Octo8080X: The part you’re pointing out is indeed a generic utility — in Hono, it’s placed in utils as well.
I’ll go ahead and create an issue for now.

src/middlewares/ip_restriction_test.ts
Octo8080X: I looked at the other code as well.
I understand now that using Deno.test instead of a BDD style is the standard.
I’ll prepare a fix.

src/middlewares/ip_restriction.ts
Octo8080X: @iuioiua 

I have created an issue.Thanks.
https://github.com/denoland/std/issues/6722

Octo8080X:
@iuioiua 
Hello.
The processing logic for IP addresses has been refactored and isolated into @std/net. 
The middleware has been subsequently reconstructed utilizing this new module. Your review is requested.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
+import type { Context } from "../context.ts";
+import type { Middleware } from "./mod.ts";
+import { isIPv4, isIPv6, matchSubnets } from "@std/net/unstable-ip";
+
+export type AddressType = "IPv6" | "IPv4" | "none";
```
iuioiua: I suggest using [`Deno.NetworkInterfaceInfo.family`](https://docs.deno.com/api/deno/~/Deno.NetworkInterfaceInfo.family) instead of this new type.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function findType(addr: string): AddressType {
+  if (isIPv4(addr)) {
+    return "IPv4";
+  } else if (isIPv6(addr)) {
+    return "IPv6";
+  }
+  return "none";
+}
```
iuioiua: See first comment.

```suggestion
function findType(addr: string): Deno.NetworkInterfaceInfo.family | undefined {
  if (isIPv4(addr)) {
    return "IPv4";
  } else if (isIPv6(addr)) {
    return "IPv6";
  }
  return undefined;
}
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+    return matchSubnets(addr, ipList);
+  };
+}
+
+function blockError(): Response {
+  return new Response("Forbidden", {
+    status: 403,
+    headers: {
+      "Content-Type": "text/plain",
+    },
```
iuioiua: Not needed. The `content-type` header by default is `"text/plain;charset=UTF-8"` anyway.
```suggestion
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
+import type { Context } from "../context.ts";
+import type { Middleware } from "./mod.ts";
+import { isIPv4, isIPv6, matchSubnets } from "@std/net/unstable-ip";
+
+export type AddressType = "IPv6" | "IPv4" | "none";
+export interface IPRestrictionRules {
+  denyList?: string[];
+  allowList?: string[];
+}
```
iuioiua: This interface and its properties should have JSDocs.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+  return async function ipRestriction<State>(ctx: Context<State>) {
+    if (ctx.info.remoteAddr.transport !== "tcp") {
+      throw new TypeError(
+        "Unsupported transport protocol. Only TCP is supported.",
+      );
+    }
+
+    if (!ctx.info.remoteAddr.hostname) {
+      return blockError();
+    }
```
iuioiua: IIUC, this conditional will never be `true` because `ctx.info.remoteAddr.hostname` is always defined.

```suggestion
```

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+function blockError(): Response {
+  return new Response("Forbidden", {
+    status: 403,
+    headers: {
+      "Content-Type": "text/plain",
+    },
+  });
+}
```
iuioiua: Not needed. See `onError` comment within `ipRestriction`.
```suggestion
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (type == "none") {
+      return blockError();
+    }
+
+    if (denyMatcher(addr)) {
+      if (options?.onError) {
+        return options.onError({ addr, type }, ctx);
+      }
+      return blockError();
+    }
```
iuioiua: Does a few things:
1. Combines `if` statements.
2. Makes it more immediate clear of what's happening by using `matchSubnets` inline.
```suggestion
    if (type === undefined || matchSubnets(addr, denyList)) {
      return onError({ addr, type }, ctx);
    }
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const denyMatcher = buildMatcher(denyList);
+  const allowMatcher = buildMatcher(allowList);
+
+  return async function ipRestriction<State>(ctx: Context<State>) {
+    if (ctx.info.remoteAddr.transport !== "tcp") {
+      throw new TypeError(
+        "Unsupported transport protocol. Only TCP is supported.",
+      );
+    }
```
iuioiua: This isn't exactly true. UDP transport should be supported, so long as Fresh allows it.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ *   },
+ * };
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}, customOnError))
+ * ```
+ */
+export function ipRestriction<State>(
+  { denyList = [], allowList = [] }: IPRestrictionRules,
+  options?: IpRestrictionOptions,
+): Middleware<State> {
+  const allowLength = allowList.length;
```
iuioiua: Nit. Doing this provides no value.
```suggestion
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    if (allowLength === 0) {
+      return await ctx.next();
+    } else {
+      if (options?.onError) {
+        return await options.onError({ addr, type }, ctx);
+      }
+      return blockError();
+    }
```
iuioiua: A little cleaner and makes it more immediately clear what's happening by inlining `matchSubnets`.
```suggestion
    if (matchSubnets(addr, allowList) || allowList.length === 0) {
      return ctx.next();
    }

    return onError({ addr, type }, ctx);
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+      );
+    }
+
+    if (!ctx.info.remoteAddr.hostname) {
+      return blockError();
+    }
+
+    const addr = ctx.info.remoteAddr.hostname == "localhost"
+      ? "172.0.0.1"
+      : ctx.info.remoteAddr.hostname;
```
iuioiua: Not sure if this should be here. I think the hostname can also be `0.0.0.0` for local traffic. Should we account for that too or should we get the user to allow these local addresses themselves? I'd lean towards the latter.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ * ```
+ */
+export function ipRestriction<State>(
+  { denyList = [], allowList = [] }: IPRestrictionRules,
+  options?: IpRestrictionOptions,
+): Middleware<State> {
+  const allowLength = allowList.length;
+
+  const denyMatcher = buildMatcher(denyList);
+  const allowMatcher = buildMatcher(allowList);
```
iuioiua: Don't need these. See below.

```suggestion
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  return async function ipRestriction<State>(ctx: Context<State>) {
```
iuioiua: Nit. Nothing wrong with this being an `async` function. Rather, there's just no reason for it to be.
```suggestion
  return function ipRestriction<State>(ctx: Context<State>) {
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ * const customOnError: IpRestrictionOptions = {
+ *   onError: () => {
+ *     return new Response("custom onError", { status: 401 });
+ *   },
+ * };
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}, customOnError))
+ * ```
+ */
+export function ipRestriction<State>(
+  { denyList = [], allowList = [] }: IPRestrictionRules,
```
iuioiua: Maybe better. Not sure.
```suggestion
  rules: IPRestrictionRules,
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export interface IpRestrictionOptions {
+  onError?: <State>(
+    remote: { addr: string; type: AddressType },
+    ctx: Context<State>,
+  ) => Response | Promise<Response>;
+}
+
+/**
+ * IP Restriction Middleware for Fresh.
```
iuioiua: Nit. Also, is that a sufficient description of this middleware?
```suggestion
 * IP restriction middleware for Fresh.
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+    return "IPv4";
+  } else if (isIPv6(addr)) {
+    return "IPv6";
+  }
+  return "none";
+}
+
+export interface IpRestrictionOptions {
+  onError?: <State>(
+    remote: { addr: string; type: AddressType },
```
iuioiua:
```suggestion
    remote: { addr: string; type: Deno.NetworkInterfaceInfo.family },
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ * @returns The middleware handler function.
+ *
+ * @example Basic usage (with defaults)
+ * ```ts
+ * const app = new App<State>()
+ *
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}))
+ * ```
+ *
+ * @example custom error handling
```
iuioiua: Nit
```suggestion
 * @example Custom error handling
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ * ```ts
+ * const customOnError: IpRestrictionOptions = {
+ *   onError: () => {
```
iuioiua: I think we should show how the `onError`s parameters can be used here.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+export interface IPRestrictionRules {
+  denyList?: string[];
+  allowList?: string[];
+}
+
+function buildMatcher(ipList: string[]) {
+  return (addr: string) => {
+    return matchSubnets(addr, ipList);
+  };
+}
```
iuioiua: Not needed if we go ahead with using `matchSubnets` inline, as stated below.
```suggestion
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export function ipRestriction<State>(
+  { denyList = [], allowList = [] }: IPRestrictionRules,
+  options?: IpRestrictionOptions,
+): Middleware<State> {
+  const allowLength = allowList.length;
+
+  const denyMatcher = buildMatcher(denyList);
+  const allowMatcher = buildMatcher(allowList);
+
```
iuioiua: Something like this will avoid the need for `if (options?.onError)` conditionals and being inline makes the behaviour of `onError` more immediately understood.
```suggestion
  
  const onError = options?.onError ?? (() =>
    new Response("Forbidden", {
      status: 403,
      headers: {
        "Content-Type": "text/plain",
      },
    }));
```

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,116 @@
...
+ * ```ts
+ * const customOnError: IpRestrictionOptions = {
+ *   onError: () => {
+ *     return new Response("custom onError", { status: 401 });
+ *   },
+ * };
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}, customOnError))
+ * ```
+ */
+export function ipRestriction<State>(
```
iuioiua: I'm not strongly for or against this name... Are we sure about this name? Any other alternatives?

iuioiua (changes requested):

packages/fresh/src/middlewares/ip_restriction.ts
Octo8080X: Thank you. The configuration has been modified to permit the passage of both TCP and URP protocols.

packages/fresh/src/middlewares/ip_restriction.ts
Octo8080X: @iuioiua 
How about 'IpRule' or 'IpFilter' instead of 'IP Restriction'? They sound pretty good.

packages/fresh/src/middlewares/ip_restriction.ts
Octo8080X: I've made the adjustment to make it inline. Thanks!

packages/fresh/src/middlewares/ip_restriction.ts
Octo8080X: I've gone ahead and added the usage examples.

packages/fresh/src/middlewares/ip_restriction.ts
iuioiua: `ipFilter` sounds okay to me.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    if (matchSubnets(addr, rules.denyList || [])) {
+      if (options?.onError) {
+        return options.onError({ addr, type }, ctx);
+      }
+      return blockError();
+    }
+
+    if (matchSubnets(addr, rules.allowList || [])) {
+      const res = await ctx.next();
```
iuioiua:
```suggestion
      return ctx.next();
```
Again, awaiting `ctx.next()` adds no value.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+      return res;
+    }
+
+    if ((rules.allowList || []).length === 0) {
+      return await ctx.next();
```
iuioiua:
```suggestion
      return ctx.next();
```
Ditto

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      if (options?.onError) {
+        return await options.onError({ addr, type }, ctx);
```
iuioiua:
```suggestion
        return options.onError({ addr, type }, ctx);
```
Ditto

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+ *   },
+ * };
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}, customOnError))
+ * ```
+ */
+export function ipRestriction<State>(
+  rules: IPRestrictionRules,
+  options?: IpRestrictionOptions,
+): Middleware<State> {
+  return async function ipRestriction<State>(ctx: Context<State>) {
```
iuioiua:
```suggestion
): Middleware<State> {
  const onBlock = options.onError ?? () => new Response("Forbidden", { status: 403 });

  return async function ipRestriction<State>(ctx: Context<State>) {
```
Reminder: avoids having to write `if (options?.onError)` blocks and makes things clear more immediately.

packages/fresh/src/middlewares/ip_restriction.ts
iuioiua: Reminder

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+        "Unsupported transport protocol. TCP & UDP is supported.",
+      );
+    }
+
+    const addr = ctx.info.remoteAddr.hostname;
+    const type = findType(addr);
+
+    if (type == undefined) {
+      return blockError();
+    }
```
iuioiua: In which case is this conditional `true`? Genuine question. If it's never `true`, this conditional can be removed completely, which I suspect is the case.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+): Middleware<State> {
+  return async function ipRestriction<State>(ctx: Context<State>) {
+    if (
+      ctx.info.remoteAddr.transport !== "udp" &&
+      ctx.info.remoteAddr.transport !== "tcp"
+    ) {
+      throw new TypeError(
+        "Unsupported transport protocol. TCP & UDP is supported.",
+      );
+    }
```
iuioiua, line 106: A good question to ask is "does Fresh allow other traffic besides UDP and TCP?" If the answer is no, this section can be removed, as the conditional will never evaluate to `true`.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    const addr = ctx.info.remoteAddr.hostname;
+    const type = findType(addr);
```
iuioiua: IMO, `findType` logic should be inline here. No reason for it to be a separate function if it's not being reused. That's just me.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    if (type == undefined) {
+      return blockError();
+    }
+
+    if (matchSubnets(addr, rules.denyList || [])) {
+      if (options?.onError) {
+        return options.onError({ addr, type }, ctx);
+      }
+      return blockError();
```
iuioiua: Ditto. See comment about `onError` being defined within the exported function.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+      const res = await ctx.next();
+      return res;
+    }
+
+    if ((rules.allowList || []).length === 0) {
+      return await ctx.next();
+    } else {
+      if (options?.onError) {
+        return await options.onError({ addr, type }, ctx);
+      }
```
iuioiua: Ditto. See comment about `onError` being defined within the exported function.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,122 @@
...
+      type: Deno.NetworkInterfaceInfo["family"] | undefined;
+    },
+    ctx: Context<State>,
+  ) => Response | Promise<Response>;
+}
+
+/**
+ * IP restriction Middleware for Fresh.
+ *
+ * @param rules rules `{ denyList: string[], allowList: string[] }`.
```
iuioiua:
```suggestion
 * @param rules Deny and allow rules object.
```
How you had it isn't how `@param` JSDoc tags should be written. They need to be written in natural human language describing what the variable is.

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * @param rules rules `{ denyList: string[], allowList: string[] }`.
+ * @param options Options for the IP Restriction middleware.
+ * @returns The middleware handler function.
+ *
+ * @example Basic usage (with defaults)
+ * ```ts
+ * const app = new App<State>()
+ *
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}))
```
iuioiua:
```suggestion
 * app.use(ipRestriction({
 *   denyList: ["192.168.1.10", "2001:db8::1"],
 * }));
```
Nit

packages/fresh/src/middlewares/ip_restriction.ts
```diff
@@ -0,0 +1,122 @@
...
+ * const customOnError: IpRestrictionOptions = {
+ *   onError: (remote, ctx) => {
+ *     console.log(
+ *       `Request URL: ${ctx.url}, Blocked IP: ${remote.addr} of type ${remote.type}`,
+ *     );
+ *
+ *     return new Response("custom onError", { status: 401 });
+ *   },
+ * };
+ * app.use(ipRestriction({denyList: ["192.168.1.10", "2001:db8::1"]}, customOnError))
```
iuioiua:
```suggestion
 * app.use(ipRestriction({
 *   denyList: ["192.168.1.10", "2001:db8::1"]
 * }, customOnError));
```
Nit

iuioiua (changes requested):
Getting closer to being ready, IMO. Few nits. HMU on Discord (same username) if you want to chat about any of this.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: I'll just leave `addr` and `type`  as they are, since I'll be using them later.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: The 'transport' can be one of five options: "tcp", "udp", "unix", "unixpacket", or "vsock". Since 'hostname' is only available for "tcp" and "udp", I believe this handling process is necessary.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: The docs say A is a string and could be either IPv6 or IPv4, so I added that part just to be safe. Do you think it's unnecessary?
https://docs.deno.com/api/deno/~/Deno.NetAddr.hostname

deno.lock
iuioiua, line 1: Should these changes be reverted?

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export interface ipFilterOptions {
+  onError?: <State>(
+    remote: {
+      addr: string;
+      type: Deno.NetworkInterfaceInfo["family"] | undefined;
+    },
+    ctx: Context<State>,
+  ) => Response | Promise<Response>;
+}
```
iuioiua: This interface and its properties need documentation.

packages/fresh/src/middlewares/ip_filter.ts
iuioiua: I think whether you're using `addr` and `type` is unrelated to my point 😛

See my suggestion below.

packages/fresh/src/middlewares/ip_filter.ts
iuioiua: To clarify, I'm wondering whether Fresh does any preprocessing of traffic that prevents non-UDP and non-TCP from being served... But according to #2392, UNIX sockets are supported (I assume it hasn't changed).

iuioiua: Actually, I'm thinking it's worth removing this block anyway. I think we should filter IPs regardless of whether it's UDP or TCP traffic or not.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+      ctx.info.remoteAddr.transport !== "udp" &&
+      ctx.info.remoteAddr.transport !== "tcp"
+    ) {
+      throw new TypeError(
+        "Unsupported transport protocol. TCP & UDP is supported.",
+      );
+    }
+
+    const addr = ctx.info.remoteAddr.hostname;
+    const type = findType(addr);
```
iuioiua:
```suggestion
    const type = IPv4(addr) ? "IPv4" : "IPv6";
```

packages/fresh/src/middlewares/ip_filter.ts
iuioiua: Yes, I do think it's unnecessary because `type` can only ever be `IPv4` or `IPv6`.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+  const onBlock = options?.onError ?? (() => blockError());
+
+  return function ipFilter<State>(ctx: Context<State>) {
+    if (
+      ctx.info.remoteAddr.transport !== "udp" &&
+      ctx.info.remoteAddr.transport !== "tcp"
+    ) {
+      throw new TypeError(
+        "Unsupported transport protocol. TCP & UDP is supported.",
+      );
```
iuioiua:
```suggestion
      return ctx.next();
```

packages/fresh/src/middlewares/ip_filter.ts
iuioiua: Sorry, to correct my stance, yes, we should return early if not UDP or TCP traffic, because an IP address isn't being used. But I don't think throwing is the correct behaviour. Instead, we should just pass through. See my suggestion above.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+ * app.use(ipFilter({
+ *   denyList: ["192.168.1.10", "2001:db8::1"]
+ * }, customOnError));
+ * ```
+ */
+export function ipFilter<State>(
+  rules: IpFilterRules,
+  options?: ipFilterOptions,
+): Middleware<State> {
+  const onBlock = options?.onError ?? (() => blockError());
```
iuioiua:
```suggestion
  const onBlock = options?.onError ?? (() => new Response("Forbidden", { status: 403 }));
```
So the reader doesn't have to scroll back up to see what `onBlock` does by default (since it's simple anyway).

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    if (matchSubnets(addr, rules.allowList || [])) {
+      return ctx.next();
+    }
+
+    if ((rules.allowList || []).length === 0) {
+      return ctx.next();
+    } else {
+      return onBlock({ addr, type }, ctx);
```
iuioiua:
```suggestion
    return onBlock({ addr, type }, ctx);
```
I believe this first `if` statement is already covered by the preceding allow block.

packages/fresh/src/middlewares/ip_filter.ts
```diff
@@ -0,0 +1,121 @@
...
+    remote: {
+      addr: string;
+      type: Deno.NetworkInterfaceInfo["family"] | undefined;
+    },
+    ctx: Context<State>,
+  ) => Response | Promise<Response>;
+}
+
+/**
+ * IP restriction Middleware for Fresh.
```
iuioiua, line 62: You should explain this filter. I.e. the precedence of rules. So deny rules take precedence over allow rules and traffic not matching any rule is denied by default (implicit deny).

iuioiua (commented):
Getting closer.

deno.lock
Octo8080X: I think ip_filter is necessary because it relies on @std/net.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: ve set it up so that traffic passes through if it's neither UDP nor TCP.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: I tidied up `blockError`.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: this section is needs.

This is because there were cases where the permission list was not explicitly set.
However, as pointed out, some parts seemed unnecessary, so I have reorganized the content.

Thank you

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: I seem to have been overthinking it unnecessarily. I would like to adopt your proposal.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: We have addressed this matter.

packages/fresh/src/middlewares/ip_filter.ts
Octo8080X: I mentioned this matter.

Thankyou

iuioiua (approved):
LGTM 👍🏾

Octo8080X:
@marvinhagemeister 

Thank you for the review on this PR a while back. It’s been about two months since iuioiua-san kindly provided feedback, and I truly appreciate the support. Please let me know if there is anything additional I can do to help get this merged—I'm ready to jump back in and make any necessary updates immediately.

PR #3036, opened 2025-06-14
chore: add tests for fresh/compat imports

### LINKED ISSUE #3025: suggestion: add tests for `fresh/compat` imports
Ensuring that `fresh/compat` imports work well will be important for everyone migrating from Fresh 1 to 2. See #3023.

SisyphusZheng:
- `fresh/compat - defineFn mixed return types` - Tests all supported return types (Response, VNode, null) with both sync and async handlers, including FreshContext parameter usage

-  `fresh/compat - multiple compat imports` - Tests importing multiple compat functions simultaneously

close #3025 

--patch-1
try to use deno/std Api to replace

--patch-2
make it simple and clean

--patch-3
Convert update_test.ts to .tsx for JSX support
Use static imports instead of dynamic imports for better performance
Switch from @std/assert to @std/expect for consistency
Use ReturnType<> for automatic type inference instead of manual type definitions

Sincerely grateful to @iuioiua for the guidance and patience.

### DIFF
--- a/update/src/update_test.ts
+++ b/update/src/update_test.ts
@@ -8,6 +8,11 @@ import {
 import { expect } from "@std/expect";
 import { walk } from "@std/fs/walk";
 import { withTmpDir } from "../../src/test_utils.ts";
+import type { FreshContext } from "../../src/context.ts";
+import { h, VNode } from "preact";
+import { assertType, IsExact, IsNullable } from "@std/testing/types";
+import { assertEquals, assertInstanceOf } from "@std/assert";
+import { spy } from "@std/testing/mock";
 
 async function writeFiles(dir: string, files: Record<string, string>) {
   const entries = Object.entries(files);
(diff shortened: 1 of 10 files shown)


update/src/update_test.ts
iuioiua, line 1: You might not have to write new files. Instead, try using [`@std/testing/types`](https://jsr.io/@std/testing/doc/types).

SisyphusZheng:
--patch-2
Simplified compat tests by removing redundant spy calls and unused type assertions while maintaining full migration compatibility validation.

update/src/update_test.ts
```diff
@@ -8,6 +8,10 @@ import {
 import { expect } from "@std/expect";
 import { walk } from "@std/fs/walk";
 import { withTmpDir } from "../../src/test_utils.ts";
+import type { FreshContext } from "../../src/context.ts";
+import { h, type VNode } from "preact";
+import { assertType, type IsExact } from "@std/testing/types";
+import { assertEquals, assertInstanceOf } from "@std/assert";
```
iuioiua: I think we should align with the rest of the codebase and instead use `@std/expect`.

update/src/update_test.ts
```diff
@@ -703,3 +707,37 @@ Deno.test("update - island files", async () => {
     `import { IS_BROWSER } from "fresh/runtime";`,
   );
 });
+
+type CompatHandler<T = unknown> = (
+  ctx: FreshContext<T>,
+) => Response | VNode | Promise<Response | VNode | null> | null;
+
+Deno.test("compat - defineFn works", async () => {
+  const { defineRoute, defineLayout, defineApp } = await import(
```
iuioiua: No reason why these functions can't be imported top-level.

update/src/update_test.ts
```diff
@@ -703,3 +707,37 @@ Deno.test("update - island files", async () => {
...
+    "../../src/compat.ts"
+  );
+
+  const handlers = {
+    response: defineRoute(() => new Response("test")),
+    vnode: defineLayout(() => h("span", null, "test")),
```
iuioiua: Maybe this should be a `.tsx` file and use a component like how it'd typically be used.
```suggestion
    vnode: defineLayout(() => <span>test</span>),
```

iuioiua (changes requested):

update/src/update_test.ts
marvinhagemeister: yeah, but that can happen in another PR. It's like that in `main` too.

SisyphusZheng:
## --patch-3
- Convert update_test.ts to .tsx for JSX support
- Use static imports instead of dynamic imports for better performance
- Switch from @std/assert to @std/expect for consistency
- Use ReturnType<> for automatic type inference instead of manual type definitions

Addresses code review feedback from @iuioiua

update/src/update_test.tsx
```diff
@@ -703,3 +708,25 @@ Deno.test("update - island files", async () => {
...
+    response: defineRoute(() => new Response("test")),
+    vnode: defineLayout(() => <span>test</span>),
+    null: defineApp(() => null),
+  };
+
+  assertType<IsExact<typeof handlers.response, DefineResult>>(true);
+
+  const ctx = {} as FreshContext<unknown>;
+  expect(handlers.response(ctx)).toBeInstanceOf(Response);
+  expect(handlers.null(ctx)).toEqual(null);
```
iuioiua:
```suggestion
  const ctx = {} as FreshContext<unknown>;
  expect(defineRoute(() => new Response("test"))(ctx)).toBeInstanceOf(Response);
  expect(defineRoute(() => <span>test</span>)(ctx)).toBeInstanceOf(Object);
  expect(defineRoute(() => null)(ctx)).toEqual(null);
```
This is fine if `defineRoute`, `defineApp` and `defineLayout` are all equivalent, which I think is a safe assumption. These tests cover when the argument is a `Response`, `VNode` or one of the types that constitute the `ComponentChildren` union type.

update/src/update_test.tsx
```diff
@@ -703,3 +708,25 @@ Deno.test("update - island files", async () => {
...
+});
+
+Deno.test("compat - functions equivalent", () => {
+  assertType<IsExact<typeof defineApp, typeof defineRoute>>(true);
+  assertType<IsExact<typeof defineRoute, typeof defineLayout>>(true);
+
+  const handler = defineApp(() => new Response("ok"));
+  expect(handler({} as FreshContext<unknown>)).toBeInstanceOf(Response);
+});
```
iuioiua: IMO, I don't think this is needed. This test would only ever fail if `defineApp`, `defineRoute` and `defineLayout` no longer equaled `defineFn`, and I don't see that ever changing. If we still want to test that, the first 2 lines will suffice.

update/src/update_test.tsx
```diff
@@ -8,6 +8,11 @@ import {
 import { expect } from "@std/expect";
 import { walk } from "@std/fs/walk";
 import { withTmpDir } from "../../src/test_utils.ts";
+import type { FreshContext } from "../../src/context.ts";
+import { assertType, type IsExact } from "@std/testing/types";
+import { defineApp, defineLayout, defineRoute } from "../../src/compat.ts";
+
+type DefineResult = ReturnType<typeof defineRoute>;
```
iuioiua:
```suggestion
```
Maybe not needed due to my other suggestions.

iuioiua (changes requested):

SisyphusZheng:
--patch-4
based on review, Removed redundant compatibility tests and type definitions, simplified test logic.

iuioiua (approved):
I forgot to mention, I think this should exist in `src/compat_test.tsx`. Once done, LGTM. I think this is a good start for testing `fresh/compat`. But I think better tests can be added that, say, compare how the current `defineRoute` behave with that of Fresh 1.x, and so on.

SisyphusZheng:
> I forgot to mention, I think this should exist in `src/compat_test.tsx`. Once done, LGTM. I think this is a good start for testing `fresh/compat`. But I think better tests can be added that, say, compare how the current `defineRoute` behave with that of Fresh 1.x, and so on.

I will first move it to src/compat_test.tsx, then make further improvements based on your suggestions. Meanwhile, I will also fix the issues caused by using assert in another PR.

iuioiua:
FYI, I've already created #3046.

marvinhagemeister (approved):
LGTM, Thanks!

PR #3037, opened 2025-06-14
test(init): consolidate tests

iuioiua:
Reduces the time taken to perform these tests by a couple of seconds and reuses already created projects instead of creating new ones.


iuioiua:
@marvinhagemeister, this might help a little with flaky tests like https://github.com/denoland/fresh/actions/runs/16156257216/job/45599269830?pr=3020

marvinhagemeister:
-1 on switching from individual `Deno.test()` calls to `t.step()`. With the former you can run an individual test with `Deno.test.only()` which is not possible with steps.

iuioiua:
I feel like that's a relatively small cost compared to the benefit of having faster and more streamlined tests. Nevertheless, I'll close.

PR #3038, opened 2025-06-14
chore:clean wrong import

SisyphusZheng:
chore:clean wrong import for otel and assert, just move to deno.json

closed by #3047 #3046

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -42,6 +42,7 @@
   },
   "imports": {
     "@deno/doc": "jsr:@deno/doc@^0.172.0",
+    "@deno/otel": "jsr:@deno/otel@^0.0.2",
     "@std/cli": "jsr:@std/cli@^1.0.19",
     "@std/collections": "jsr:@std/collections@^1.0.11",
     "@std/http": "jsr:@std/http@^1.0.15",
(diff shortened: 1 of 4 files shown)


deno.json
```diff
@@ -55,7 +56,7 @@
     "@preact/signals": "npm:@preact/signals@^2.0.4",
     "esbuild": "npm:esbuild@0.25.4",
     "esbuild-wasm": "npm:esbuild-wasm@0.25.4",
-    "@std/assert": "jsr:@std/assert@^1.0.16",
+    "@std/assert": "jsr:@std/assert@^1.0.13",
```
iuioiua, line 59: Why downgrade?

SisyphusZheng:
Here's an error with an incorrect import version that isn't actually being used. My package is imported through JSR. To resolve similar problem, I submitted PR #3041.

www/utils/screenshot_test.ts
iuioiua, line 1: Really, if we want to align with the rest of the codebase, we should be using `@std/expect`. That'd omit for having `@std/assert` in the root `deno.json`.

SisyphusZheng: agree with you. but the import（assert） has already been merged. If needed, I can go back to clean it up and the util script as well. for this PR, I only intend to remove these two imports.”

iuioiua: > f needed, I can go back to clean it up and the util script as well.

Yeah, IMO, that's the right move. Better to be consistent.

SisyphusZheng:
closed by #3046 #3047

PR #3041, opened 2025-06-15
feat: add unused imports check test

SisyphusZheng:
#why 
Here's an error with an incorrect import version that isn't actually being used. My package is imported through JSR(#3032). To resolve this issue, I submitted PR(#3038),to prevent similar errors, I submitted this PR..

Add a test to automatically detect unused imports in `deno.json` configuration file.

## Purpose
- Ensures only necessary dependencies are included in the project

## related
related #3032 
related #3038 

## What this PR does
- Adds `tests/imports_check_test.ts` that scans all TypeScript files
- clean wrong imports

### Successed:
a17d5e3d5c51a5d0061cc58ef1b38e4
## failed -> error ci
98c1049a7ab91834c77393b268ca6f4

# patch-1
##why error?
The import usage check failed to detect that @deno/doc and @std/collections are actually used indirectly through remote module imports. The script incorrectly marked them as unused.

```bash
// tools/check_docs.ts contains:
import { checkDocs } from "https://github.com/denoland/std/raw/refs/heads/main/_tools/check_docs.ts";
```

## how to fix?
Add a whitelist mechanism to handle known indirect dependencies, now the script will Now this script will automatically identify items in the whitelist as indirect references.

412ba3e6383866c9cb8f4309275cb0c

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -41,9 +41,7 @@
     "exclude": ["**/*_test.*", "src/__OLD/**", "*.todo"]
   },
   "imports": {
-    "@deno/doc": "jsr:@deno/doc@^0.172.0",
     "@std/cli": "jsr:@std/cli@^1.0.19",
-    "@std/collections": "jsr:@std/collections@^1.0.11",
     "@std/http": "jsr:@std/http@^1.0.15",
     "@std/uuid": "jsr:@std/uuid@^1.0.7",
     "fresh": "jsr:@fresh/core@^2.0.0-alpha.29",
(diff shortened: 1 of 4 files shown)


bartlomieju:
Closing this PR — the motivation is valid but there are a few issues with the current approach:

1. **Unrelated changes** — the `@std/assert` downgrade, `deno.lock` churn, and `test_utils.tsx` import change should not be part of this PR
2. **Fragile detection** — string-matching across concatenated file contents will miss re-exports, dynamic imports, and transitive usage. The `INDIRECT_DEPENDENCIES` whitelist will silently go stale
3. **Noisy test output** — 8 `console.log` calls with lint suppressions; tests should be quiet on success

If you'd like to take another pass:

- Consider adding a script in `tools/` rather than a test that runs on every CI pass
- Use Deno's module graph API (`deno info --json`) to resolve actual dependency usage instead of string matching — this handles re-exports and transitive deps correctly
- Keep the PR focused — one concern per PR, no unrelated dependency changes

Thanks for the contribution!

Issue #3040, opened 2025-06-14
Feature expansion .use(path, middleware)


Octo8080X:
## Is your feature request related to a problem?

It’s difficult to apply middleware only to a specific path or directory.
With the current implementation, .use(middleware) is supported, but .use(path, middleware) isn’t available.

You can work around this to some extent with:

```javascript
.all(path, middleware)
```
However, when the request URL is /admin:

```javascript
app.all("/admin", middleware) // responds
app.get("/admin", handler) // doesn't respond
```
Ideally, we’d like to use:

```javascript
app.use("/admin", middleware)
```
to apply the middleware, and then:

```javascript
app.get("/admin", handler)
```
should be able to match afterwards.
This would be the expected behavior.

## Is your feature request related to a problem? Please describe.

I also tried using `.all(path, middleware)`, but as mentioned above, this approach does not resolve the issue.
We need .use(path, middleware).
Or is there any way to work around this?

I tried placing a _middleware.ts file, but it ends up applying the middleware to requests that shouldn’t be processed by it.
Here’s an example of what’s going on:
```ts
//routes\_middleware.ts
import {define} from '../utils.ts'

export default define.middleware(async (ctx) => {
  console.log(`Request received: ${ctx.req.method} ${ctx.req.url}`);
  return await ctx.next()
});
```
The log output when it was run:
```
Request received: GET http://localhost:8000/ <= OK
Request received: POST http://localhost:8000/api/abc <= OK
Request received: GET http://localhost:8000/service-worker.js <= NG
```

marvinhagemeister:
+1 I like this proposal

PR #3044, opened 2025-06-16
Add Feature expansion .use(path, middleware)

Octo8080X:
target: https://github.com/denoland/fresh/issues/3040

Currently in progress.


src/app.ts
```diff
@@ -212,19 +226,48 @@ export class App<State> {
...
+      // When mounting at the root, add middleware directly to the host app
+      for (const middleware of middlewares) {
+        this.#router._middlewares.push(middleware);
+      }
 
-    for (let i = 0; i < routes.length; i++) {
-      const route = routes[i];
+      // Add routes as-is since we're mounting at the root
+      for (let i = 0; i < routes.length; i++) {
+        const route = routes[i];
```
iuioiua, line 237:
```suggestion
      for (const route of routes) {
```
Nit. And ditto for other loops.

src/router.ts
```diff
@@ -2,7 +2,7 @@ export type Method = "HEAD" | "GET" | "POST" | "PATCH" | "PUT" | "DELETE";
 
 export interface Route<T> {
   path: string | URLPattern;
-  method: Method | "ALL";
+  method: Method | "ALL" | "MIDDLEWARE";
```
iuioiua: This seems odd to me. Is there an alternative?

src/app.ts
```diff
@@ -212,19 +226,48 @@ export class App<State> {
...
-      this.#router.add(route.method, merged, combined);
+        // When there are middlewares, we need to add middleware handlers to each route
+        // Extract middleware handlers into a flat array ensuring we get the right type
+        const middlewareHandlers: Array<MiddlewareFn<State>> = [];
+        for (const middleware of middlewares) {
+          for (const handler of middleware.handlers) {
+            // We need to safely cast here to handle the type properly
+            middlewareHandlers.push(handler as unknown as MiddlewareFn<State>);
+          }
+        }
```
iuioiua:
```suggestion
        const middlewareHandlers = middlewares.flatMap((middleware) =>
          middleware.handlers
        );
```

src/app.ts
```diff
@@ -212,19 +226,48 @@ export class App<State> {
...
+
+        // Create a new array of handlers with the correct type
+        const combinedHandlers: Array<MiddlewareFn<State>> = [
+          ...middlewareHandlers,
+          ...route.handlers.map((h) => h as unknown as MiddlewareFn<State>),
+        ];
```
iuioiua:
```suggestion
        const combinedHandlers = [
          ...middlewareHandlers,
          ...route.handlers,
        ];
```

src/app.ts
```diff
@@ -172,8 +172,22 @@ export class App<State> {
...
+      pathOrMiddleware instanceof URLPattern
+    ) {
+      // First argument is a path
+      if (!middleware) {
+        throw new Error("Middleware is required when path is provided");
+      }
+      this.#router.addMiddleware(pathOrMiddleware, middleware);
+    } else {
+      this.#router.addMiddleware("/*", pathOrMiddleware);
+    }
```
iuioiua:
```suggestion
    if (typeof pathOrMiddleware === "function") {
      this.#router.addMiddleware("/*", pathOrMiddleware);
      return this;
    }

    if (!middleware) {
      throw new TypeError("Middleware is required when path is provided");
    }
    this.#router.addMiddleware(pathOrMiddleware, middleware);
```
Maybe a little tidier?

src/router.ts
```diff
@@ -30,13 +30,32 @@ export const IS_PATTERN = /[*:{}+?()]/;
...
+        handlers: [handler],
+        method: "MIDDLEWARE",
+      });
+    } else {
+      this._middlewares.push({
+        path: pathname,
+        handlers: [handler],
+        method: "MIDDLEWARE",
+      });
+    }
```
iuioiua:
```suggestion
    this._middlewares.push({
      path: isPattern(pathname) ? new URLPattern({ pathname }) : pathname,
      handlers: [handler],
      method: "MIDDLEWARE",
    });
```
We can factor out that conditional into an `isPattern()` which can be reused on lines 60-61.

iuioiua (changes requested):
Quick pass

src/app.ts
marvinhagemeister: It's fine as is. Which for-loop to use doesn't matter. Although it's unlikely to matter here, the traditional for-loop is even faster as it doesn't need to invoke the iterator stuff compared to `for-of`.

src/router.ts
Octo8080X: I think there’s a way to use “ALL” instead of “MIDDLEWARE”.
Or we could allow undefined, but I feel that would have too large of an impact.

src/app.ts
Octo8080X: Thank you.
With the suggested approach, pathOrMiddleware can't be narrowed down to string | URLPattern, which causes a type error—so I decided to keep it as it is for now.

src/router.ts
Octo8080X: I decided to go with "ALL".

src/router.ts
Octo8080X: Thanks!
I extracted it as isURLPattern.

src/router.ts
```diff
@@ -30,29 +30,37 @@ export const IS_PATTERN = /[*:{}+?()]/;
...
+  add(
+    method: Method | "ALL",
+    pathname: string | URLPattern,
+    handlers: T[],
+  ): void {
+    this._routes.push({
+      path: (
+          typeof pathname === "string" && pathname !== "/*" &&
+          !this.isURLPattern(pathname)
+        )
```
iuioiua: Nit: To clarify, my idea was to have everything inside this condition be factored out. Not just the `URLPattern` bit. Also, I don't see a reason why it should be a method when it doesn't use any property from this class.

Octo8080X: Apologies — I didn’t fully grasp your intent at first.
I’ve now moved the logic out of the class and reorganized the content.

src/app.ts
```diff
@@ -172,8 +172,23 @@ export class App<State> {
...
+  ): this;
+  use(
+    pathOrMiddleware: string | URLPattern | MiddlewareFn<State>,
+    middleware?: MiddlewareFn<State>,
+  ): this {
+    if (typeof pathOrMiddleware === "function") {
+      this.#router.addMiddleware("/*", pathOrMiddleware);
+    } else {
+      if (!middleware) {
+        throw new Error("Middleware is required when path is provided");
```
iuioiua:
```suggestion
        throw new TypeError("Middleware is required when path is provided");
```
Nit

src/app.ts
```diff
@@ -212,19 +227,44 @@ export class App<State> {
...
 
-      const merged = typeof route.path === "string"
-        ? mergePaths(path, route.path)
-        : route.path;
-      const combined = isSelf
-        ? route.handlers
-        : middlewares.concat(route.handlers);
-      this.#router.add(route.method, merged, combined);
+        // Create a new array of handlers with the correct type
+        const combinedHandlers: Array<MiddlewareFn<State>> = [
```
iuioiua:
```suggestion
        const combinedHandlers = [
```
Nit: this might not be needed

src/app_test.tsx
iuioiua, line 1: Nice tests!

src/router.ts
```diff
@@ -28,31 +28,36 @@ export interface Router<T> {
...
-        path: pathname,
-        handlers,
-        method,
-      });
-    }
+  add(
+    method: Method | "ALL",
+    pathname: string | URLPattern,
+    handlers: T[],
+  ): void {
```
iuioiua:
```suggestion
  ) {
```
Nit: the Deno team consider defining `void` and `Promise<void>` as explicit return types as needless boilerplate.

iuioiua (approved):
LGTM! Just a few non-blocking nits.

src/app_test.tsx
Octo8080X: Thanks!

marvinhagemeister:
Thanks for working on a PR. We've refactored the internals of Fresh quite a bite to have sort of an internal route tree when building up the middlewares. This was needed to get other features in that we needed, but does unfortunately mean that the functionality proposed in this PR has already been implemented, see https://github.com/denoland/fresh/pull/3086

PR #3046, opened 2025-06-16
test(www): use `@std/expect` instead of `@std/assert`

iuioiua:
Aligns screenshot tests to the rest of the codebase by using `@std/expect` instead of `@std/assert`.


marvinhagemeister (approved):
LGTM. Once, conflicts are resolved this can land.

PR #3047, opened 2025-06-17
chore(www): remove `@deno/otel` package

iuioiua:
This seems to be no longer needed thanks to https://github.com/denoland/deno/pull/27541.


marvinhagemeister (approved):
LGTM, thanks!

Issue #3045, opened 2025-06-16
Support: Dialogs + history + partials


marvinhagemeister:
See https://discord.com/channels/684898665143206084/991511118524715139/1384264907184275476

> I've hit a roadblock with Fresh 2 due to the partial popstate handling. Partials are great for navigation but for a work project I require a system that controls dialogs via URL (pushState and popState). 
>
> Currently, when a popstate event is fired, the fresh runtime causes a page reload. I cannot get rid of the state/data in the event. 
Is there a chance you could make the history.reload() behavior opt-out some way?

bartlomieju:
This is already supported. The `popstate` handler in `packages/fresh/src/runtime/client/partials.ts` checks for `e.state.fClientNav` and returns early if it's not set:

```js
if (!e.state.fClientNav) return;
```

So if your code uses `history.pushState({}, ...)` without `fClientNav: true` in the state object, Fresh will not intercept the popstate event — no reload, no partial fetch. Fresh only acts on history entries it created itself.

PR #3048, opened 2025-06-17
fix: avoid reload for user-code history changes (#3045)

ch2i5:
The Fresh partials popstate handler causes page reloads when user code makes use of navigation history state. This can lead to a loss of user data in a browsing context. 
To fix this behavior, FreshHistoryState now has a flag which validates partials functionality/explicitly opts-in to any partial related location management.

### DIFF
--- a/src/runtime/client/partials.ts
+++ b/src/runtime/client/partials.ts
@@ -27,7 +27,11 @@ export const PARTIAL_ATTR = "f-partial";
 
 class NoPartialsError extends Error {}
 
+// Fresh partials history updates set the fClientNav flag
+// and prevent reloads in the popstate handler when
+// user-code triggers history navigation events.
 export interface FreshHistoryState {
+  fClientNav: boolean;
   index: number;
   scrollX: number;
   scrollY: number;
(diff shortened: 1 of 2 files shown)


src/runtime/client/partials.ts
```diff
@@ -28,6 +28,7 @@ export const PARTIAL_ATTR = "f-partial";
 class NoPartialsError extends Error {}
 
 export interface FreshHistoryState {
+  fClientNav: boolean;
```
iuioiua, line 34: Perhaps a little explanation of this field here might help.

src/runtime/client/partials.ts
```diff
@@ -164,6 +167,11 @@ addEventListener("popstate", async (e) => {
     }
     return;
   }
+  // Do nothing if Fresh navigation is not explicitly opted-in.
+  // Other applications might manage scrollRestoration individually.
+  if (e.state.fClientNav !== true) {
+    return;
+  }
```
iuioiua:
```suggestion
  if (!e.state.fClientNav) return;
```

iuioiua (changes requested):
Nits

iuioiua (approved):

marvinhagemeister (commented):
Nice, would love a test case for this! You can check out the test cases for partials for inspiration.

ch2i5:
hi @marvinhagemeister, this got probably lost behind more important tasks but I wanted to ask if my commits suffice and could be merged+deployed by next week as I have a product demo coming up.

marvinhagemeister (approved):
LGTM, thanks for fixing this 👍

PR #3049, opened 2025-06-18
refactor(src): remove redundant coloring from `createCodeFrame()`

iuioiua:
`src/runtime/server/preact_hooks.tsx` is the only place that `createCodeFrame()` is used which stripped the ANSI codes anyway.


marvinhagemeister (approved):
LGTM

PR #2903, opened 2025-05-09
feat: update the tailwind plugin to suuport tailwindv4 #2819

SisyphusZheng:
Inspired by #2819, #2829, and https://github.com/pakornv/fresh-plugin-tailwindcss, I made minimal modifications to the plugin-tailwind, I originally intended to update based on #2829, but I was informed by the system that I couldn't push.

 Due to the breaking changes in Tailwind v4, some PostCSS plugins are no longer required. Additionally, I updated the syntax in the /www directory. The /www directory now uses the updated plugin supported by Tailwind v4.I think this has made initial progress.

# What This PR Solves
The Tailwind plugin has been updated to support Tailwind v4, and it is currently functioning well both in the repository's www directory and in the initialized project which for test.

### 1. Dependency Optimization
- Removed `cssnano` dependency, using Tailwind v4's built-in optimization(Based on @csvn  's suggestion)
- Removed `autoprefixer` dependency
- Updated `tailwind-plugin` version number
### 2. Build Configuration Adjustments
- Added `--node-modules-dir` parameter to `/www` task for Tailwind v4 support (Based on @marvinhagemeister 's suggestion)
- Added `--allow-scripts` parameter to enable `@tailwindcss/oxide` execution
### 3. Code Adaptation
- Modified Tailwind-related code in `/www` to support Tailwind v4
- Updated dependency configuration in `deno.json`
### 4. Feature Restoration
- Added `imagescript` and `astral` dependencies to restore screenshot functionality lost in previous PR updates
- Fixed the issue where the `tailwind-plugin` dependency was incorrectly pointing to the JSR package instead of the local repository source code in previous PR.

## --patch-2
- Upgraded Tailwind CSS to the latest version.
- Removed unnecessary dependencies.
- Enabled CSS minification by default when Fresh is in production mode.
- Updated docs/canary/examples/migration-guide.md to support the new Tailwind plugin format.
- Updated init/src/init.ts to support the new Tailwind version and plugin version, use new `@import "tailwindcss"`, and generate the new plugin format.
- Updated init/src/init_test.ts: tests now expect `css` to match `/tailwindcss/` instead of `/@tailwind/`.
- Removed plugin-tailwindcss/src/types.ts as it is no longer needed, and updated plugin-tailwindcss/src/mod.ts and www/dev.ts to support the new format.

Co-authored-by: Asher Gomez <ashersaupingomez@gmail.com>

## --patch-3
[...]

PR #3054, opened 2025-06-22
feat:support tailwindcss v4

### LINKED ISSUE #2819: Tailwind 4
Any Guidance on updating to / using Tailwind 4 with Fresh?

### LINKED ISSUE #3056: suggestion:support tailwindcss v4
It's time to work on supporting Tailwind V4 with the new esbuild Loader plugin built.

SisyphusZheng:
update /www and tailwind-plugin to support tailwind v4.

Thanks to the merge of #3034, the work of integrating and advancing #2903 is promoted.

related #3034 
related #2903 

close #3056
close #2819

--patch-1 
based on review,

1. del compiler and make it more clean and simple,
2.  fix init tailwind part,
3. fix test tailwind part

--patch-2

1. fix the docs 
2. fix the docs of plugin and version

--patch-3
1. Addressed various maintenance tasks (chores)
2. Updated the Tailwind CSS plugin to align with this release. Generated a fresh project using the updated init configuration and 
3. deployed it via Deno Deploy EA. Deployment is functioning correctly.

--patch-4
Move postcss imports to Tailwind section in init script

--patch-5
Refactored Tailwind plugin type definitions by copying PluginOptions properties from @tailwindcss/postcss and adding Fresh-specific exclude option, while exporting types from package entrypoint for better accessibility.

Test Project: https://fresh.fresh-press.deno.net/
Test Project Repository: https://github.com/SisyphusZheng/fresh-tailwindv4
Test Plugin: https://jsr.io/@freshpress/fresh-plugin-tailwind@1.0.0


plugin-tailwindcss/src/mod.ts
```diff
@@ -1,19 +1,19 @@
...
 import type { App } from "fresh";
+import type { TailwindPluginOptions } from "./types.ts";
 
 export function tailwind<T>(
   builder: FreshBuilder,
   app: App<T>,
   options: TailwindPluginOptions = {},
 ): void {
-  let processor: ReturnType<typeof initTailwind> | null;
+  let processor: ReturnType<typeof initTailwind> | null = null;
```
iuioiua:
```suggestion
  const instance = await postcss(twPostcss({
    optimize: app.config.mode === "production",
    ...options,
  }));
```
Might this work? If so, we can remove lines 16-17 and `plugin-tailwindcss/src/compiler.ts`. It'd require `tailwind()`  to become async and be `await`ed where used, including documentation.

deno.json
```diff
@@ -1,4 +1,5 @@
 {
+  "nodeModulesDir": "auto",
```
iuioiua, line 2: Should we also add this to `deno.json` in the init script?

plugin-tailwindcss/src/mod.ts
iuioiua: I also realise we weren't passing `options` from `tailwind()` to `initTailwind()`. This snippet fixes that.

SisyphusZheng: thx so much

deno.json
SisyphusZheng: yep thx

iuioiua (changes requested):
Instances of `tailwind()` not being `await`ed still remain in the codebase. E.g. https://github.com/SisyphusZheng/fresh/blob/aac6207306cfcda2405e2761958136fb637cf470/docs/canary/examples/migration-guide.md?plain=1#L62. I suggest searching for "`tailwind(`".

SisyphusZheng:
> Instances of `tailwind()` not being `await`ed still remain in the codebase. E.g. https://github.com/SisyphusZheng/fresh/blob/aac6207306cfcda2405e2761958136fb637cf470/docs/canary/examples/migration-guide.md?plain=1#L62. I suggest searching for "`tailwind(`". 

thx for pointing out this issue. This is actually part of the documentation and does not affect the code. I hope the documentation can be revised later. 

Of course, I will also double-check if there is any code that hasn't been actually fixed.

iuioiua:
But it matters that the documentation is correct according to the implementation. Otherwise, users will be mislead on how to use `tailwind()`.

SisyphusZheng:
> But it matters that the documentation is correct according to the implementation. Otherwise, users will be mislead on how to use `tailwind()`.

I understand the importance of this issue now and will make the modification. Thank you.

www/deno.json
```diff
@@ -1,6 +1,6 @@
 {
   "tasks": {
-    "start": "deno run -A --watch=static/,routes/,../src,../docs dev.ts",
+    "start": "deno run -A --allow-scripts --watch=static/,routes/,../src,../docs dev.ts",
```
iuioiua: Is `--allow-scripts` required now? Genuine question. I haven't checked. If so, we should consider whether other tasks such as `build` need it too, including those created from the init script.

iuioiua (approved):
LGTM!

www/deno.json
SisyphusZheng: The oxide script requires --allow-script during deno install, and I can't figure out a way to avoid this warning for the time being.

plugin-tailwindcss/src/types.ts
```diff
@@ -1,48 +1,6 @@
...
-   * to share target browsers with Babel, ESLint and Stylelint
-   */
-  overrideBrowserslist?: string | string[];
-
-  /** do not raise error on unknown browser version in `Browserslist` config. */
-  ignoreUnknownVersions?: boolean;
-}
+import type { OnTransformOptions } from "fresh/dev";
 
 export interface TailwindPluginOptions {
```
iuioiua:
```suggestion
export interface TailwindPluginOptions extends PluginOptions {
```
From postcss

SisyphusZheng: Thx for rechek,  btw, i already check the https://postcss.org/api/, PostCSS only has `PluginOptions` as a parameter type for `PluginCreator`, not as an exportable interface, and this is Fresh-specific, IMO, may not suitable for this

iuioiua: I think a few things need to be addressed:
1. `TailwindPluginOptions` should be exported from `plugin-tailwindcss/src/mod.ts` (package entrypoint). Otherwise, that interface is inaccessible. IMO, the interface should be moved and `plugin-tailwindcss/src/types.ts` removed, but that might conflict with how Marvin wants things structured.
2. `PluginOptions` is defined in `@tailwindcss/postcss`, not `postcss`. But that's okay, we can just copy the `PluginOptions` interface properties, `base` (not sure if it's applicable for us) and `optimize`, to `TailwindPluginOptions`.
3. These interface properties should be documented. E.g. `optimize` defaults to `true` if Fresh is in production mode and `false` otherwise.

plugin-tailwindcss/src/mod.ts
```diff
@@ -1,20 +1,26 @@
...
 import type { App } from "fresh";
+import twPostcss from "@tailwindcss/postcss";
+import postcss from "postcss";
+import type { TailwindPluginOptions } from "./types.ts";
+
+// Re-export types for public API
+export type { TailwindPluginOptions } from "./types.ts";
 
-export function tailwind<T>(
+export async function tailwind<T>(
```
marvinhagemeister: Let's keep the sync variant of before so that we don't have to change this to be `async`.

SisyphusZheng: yep

plugin-tailwindcss/src/mod.ts
```diff
@@ -1,20 +1,26 @@
...
+    ...tailwindOptions,
+  }));
 
   builder.onTransformStaticFile(
-    { pluginName: "tailwind", filter: /\.css$/, exclude: options.exclude },
+    { pluginName: "tailwind", filter: /\.css$/, exclude },
     async (args) => {
-      if (!processor) processor = initTailwind(app.config, options);
-      const instance = await processor;
       const res = await instance.process(args.text, {
```
iuioiua, line 24: Might this now throw since `instance` could be an unresolved promise?

iuioiua: Ignore... I thought `instance` was a promise.

iuioiua:
Are you able to resolve the merge conflicts? @marvinhagemeister, are you able to take a look? It seems a lot of people would love to have this added.

marvinhagemeister:
Resolved the merge conflicts, but it looks like the PR is not yet ready to be merged. Several aspects of the site in `www` are missing styling and look broken. I've followed the update instructions on this page https://tailwindcss.com/docs/upgrade-guide and applied them, but looks like there is more going on.

main:

Screenshot 2025-07-21 at 09 16 49

This PR:

Screenshot 2025-07-21 at 09 16 21

marvinhagemeister (approved):
LGTM, Thanks for updating the plugin! I've went ahead and updated the code in `www` for v4 too.

PR #3064, opened 2025-06-27
fix(core): use `PageProps` instead of `FreshContext` in `define*()` compat functions

iuioiua:
In Fresh 1, define apps used `RouteContext`, which align with `PageProps` in Fresh 2, rather than `FreshContext`. E.g. before this PR, one would get a type error if they tried to use `ctx.Component` within `defineApp()` and now they shouldn't.


marvinhagemeister (approved):
LGTM, thanks!

PR #3067, opened 2025-06-28
fix[docs]:Update Fresh canary usage instructions in docs

### LINKED ISSUE #3058: Problems using the canary version of Fresh 2
I try to follow the instructions to use the [canary version of Fresh](https://fresh.deno.dev/docs/examples/using-fresh-canary-version).

I cannot get it to run.

I replaced the jsr Fresh 2 entry in `deno.json` with a derived string from the page like `https://raw.githubusercontent.com/denoland/fresh/e996b5e2fece42dc2c235ab050f7604f366e74ff/`.

So the fresh import line in deno.json is like:

`"fresh": "https://raw.githubusercontent.com/denoland/fresh/e996b5e2fece42dc2c235ab050f7604f366e74ff/"`

This results that the line `import { Builder } from "fresh/dev";` in `dev.ts` issues an error:

`Relative import path "fresh/dev" not prefixed with / or ./ or ../ and not in import map from "file:///home/../../dev.ts"`

I cannot even initiate a new Fresh project according to the instructions.

The instructions on the above mentioned page (at the bottom) say that this command should work:

`deno run -A -r https://raw.githubusercontent.com/denoland/fresh/main/init.ts`

Which results in "error: module not found", because the folder structure of the Fresh sources has changed.

But even e.g. `deno run -A -r https://raw.githubusercontent.com/denoland/fresh/e996b5e2fece42dc2c235ab050f7604f366e74ff/init/src/init.ts` which points to an exisiting init.ts file, does not work.

It issues the error "Relative import path "@str/path" not prefixed with / or ./ or ../" from line 2 of the file `init.ts`.

What is the correct way to use the current sources?

The github address given in the instructions issues a 400 errror when used in the browser like `https://raw.githubusercontent.com/denoland/fresh/e996b5e2fece42dc2c235ab050f7604f366e74ff/`. When ading a correct folder and file name like `https://raw.githubusercontent.com/denoland/fresh/e996b5e2fece42dc2c235ab050f7604f366e74ff/init/src/init.ts` it opens the file in the browser.

SisyphusZheng:
Revise documentation to clarify how to use the latest Fresh 2 canary, specific commits, and forks via deno.json. Add instructions for creating new projects using JSR and local sources, and update references to current package versions. Include links to the Fresh v2 blog and migration guide for further reading.

close #3058

### DIFF
--- a/docs/latest/examples/using-fresh-canary-version.md
+++ b/docs/latest/examples/using-fresh-canary-version.md
@@ -9,7 +9,7 @@ script. This page has you covered.
 
 ## Canary Fresh in `deno.json`
 
-### Latest alpha version 
+### Latest alpha version
 
 The easiest way to use Fresh 2 canary is with the update command:
 
@@ -17,7 +17,8 @@ The easiest way to use Fresh 2 canary is with the update command:
 deno run -A -r jsr:@fresh/update@2.0.0-alpha.35 .
 ```
 
-This will automatically update your `deno.json` to use the specified canary version.
+This will automatically update your `deno.json` to use the specified canary
+version.
 
 ### Specific commit
 
@@ -55,7 +56,7 @@ For testing your own fork or PR:
 
 ## Creating a new project
 
-### Using JSR 
+### Using JSR
 
 ```sh Terminal
 deno run -A -r jsr:@fresh/init@2.0.0-alpha.35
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks!

PR #3070, opened 2025-06-30
fix[test]:ensures more stable CI runs on Windows environments.

SisyphusZheng:
This pull request includes a small but important change to the `withTmpDir` function in `src/test_utils.ts`. The change adds error handling to ensure that a `PermissionDenied` error during directory removal does not interrupt the execution.


marvinhagemeister (approved):
LGTM, thanks!

PR #3080, opened 2025-07-08
feat: expose `HttpError` for clients

### LINKED ISSUE #2995: suggestion: Add ./src/error.ts to deno.json exports
Currently, you have to import HttpError from jsr:@fresh/core. This is fine on the server side but doesn’t work client side. In my project I use custom error classes which extend HttpError. The problem is that I need to use these error classes in the browser too. In order for that to work I import HttpError using https://jsr.io/@fresh/core/2.0.0-alpha.34/src/error.ts.

That being said, it would be much nicer to import it using jsr:fresh/core/error or jsr:fresh/core/runtime.

Hence, I’d like to suggest adding ./src/error.ts to the exports in the deno.json. Would that be possible?

marvinhagemeister:
This PR exposes the `HttpError` class for clients. 

To do that I needed to make the error message optional to avoid bloating up the bundle with lots of status strings.

Fixes https://github.com/denoland/fresh/issues/2995


bartlomieju:
LGTM! Clean approach to avoid pulling STATUS_TEXT into the client bundle. Merging main in to get it up to date.

bartlomieju (approved):

PR #3081, opened 2025-07-08
feat: support overwriting default 404 handler

### LINKED ISSUE #1918: way to disable 404 route

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/1918

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -242,7 +242,13 @@ export class App<State> {
     return this;
   }
 
-  handler(): (
+  /**
+   * Create handler function for `Deno.serve` or to be used in
+   * testing.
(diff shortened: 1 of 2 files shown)


src/app.ts
```diff
@@ -242,7 +242,13 @@ export class App<State> {
     return this;
   }
 
-  handler(): (
+  /**
+   * Create handler function for `Deno.serve` or to be used in
+   * testing.
+   * @param {() => Promise<Response>} [nextFn] overwrite default 404 handler
+   * @returns {Deno.ServeHandler}
```
iuioiua, line 249: Nit: JSDoc types don't need to be defined since we're already using TypeScript

PR #3122, opened 2025-07-23
major: build islands + scan routeDir without importing user code

### LINKED ISSUE #1843: Can't get environment variables when prebuilding by Github worflow
This is what I wrote in utils.ts:
```typescript
export const supabase = createClient<Database>(
  Deno.env.get("SUPABASE_URL") as string,
  Deno.env.get("SUPABASE_KEY") as string,
);
```
After I push the commit, I got this error message on GIthub workflow page:
```typescirpt
Build step >
Run deno task build
Task build deno run -A dev.ts build
.
.
.
.
error: Uncaught (in promise) Error: supabaseUrl is required.
    at new p (https://esm.sh/v132/@supabase/supabase-js@2.4.0/esnext/supabase-js.mjs:2:2168)
    at nt (https://esm.sh/v132/@supabase/supabase-js@2.4.0/esnext/supabase-js.mjs:2:5286)
    at file:///home/runner/work/a-pp-greeting-cards-platform/a-pp-greeting-cards-platform/utils/utils.ts:20:25
Error: Process completed with exit code 1.
```

And I've setup the env var on deno deploy dashboard and this works on local .env file.

### LINKED ISSUE #2240: `kv.listenQueue` prevents build mode from working

### LINKED ISSUE #2592: [fresh@2.0] Islands under `routes/(_islands)/...` won't be detected as islands
From [`this thread`](https://discord.com/channels/684898665143206084/991511118524715139/1260586916764586036):

when placing islands under e.g. `routes/(_islands)/MyIsland.tsx`, `MyIsland` won't be picked up as an island like it did in 1.x. Placing it under `islands/MyIsland.tsx` does work.

### LINKED ISSUE #2640: deno serve --parallel
Is it possible to use the deno serve with fresh? When I try to run deno serve --parallel dev.ts it spun 8 different ports. This is not ideal behavior. deno serve automatically handles the multi worker listening on single port. Is there any way to leverage that for fresh?

marvinhagemeister:
This fixes a bunch of long standing issues in our tracker where a build wouldn't finish because of importing user code. It's common for backend apps to spawn some open ended tasks like subscribing to Deno KV or other things. This addresses another problem of having to set up ENV vars just for the build, even though they are technically not required to build islands.

This also gives a roughly ~30% performance boost on cold starts (at least on `deno.com`). Unfortunately, this does require some breaking changes. But I believe the performance improvements + removing the footgun of loading user code during build is well worth it. This work is also a precursor in case we want to exploring moving to vite at some point in the future.

## Breaking changes

1. Build options have moved to the builder.
```diff
  // main.ts
- const app = new App({ build: { outDir: "dist/" } })
+ const app = new App()

  // dev.ts
- const builder = new Builder();
+ const builder = new Builder({ outDir: "dist/" });
```

2. Another breaking change is how file based routes are loaded. Since they're prepared and loaded by the builder the API changed. 

```diff
  // main.ts
  const app = new App()
    .use(staticFiles())
  
- await fsRoutes(app, {
-   loadIsland: (path) => import(`./islands/${path}`),
-   loadRoute: (path) => import(`./routes/${path}`),
- }); 
+ app.fsRoutes()
```
You can optionally change the `islandDir` or `routeDir` by passing that to the `Builder`:

```ts
// dev.ts
const builder = new Builder({
  islandDir: "/path/to/islands", 
  routeDir: "/path/to/routes"
});
```

3. Import `app` lazily in `dev.ts`. Passing `app` to `builder.build()` is not necessary anymore.

```diff
- import { app } from "./main.ts";

  const builder = new Builder({ target: "safari12" });
  
  if (Deno.args.includes("build")) {
-    await builder.build(app);
+    await builder.build();
  } else {
-   await builder.listen(app);
+   await builder.listen(() => import("./main.ts"));
  }
```

4. The production run command has changed. It's now using `deno serve`
```diff
  // deno.json
  {
    "tasks": {
      "dev": "deno run -A --watch=static/,routes/ dev.ts",
      "dev": "deno run -A dev.ts build",
-     "start": "deno run -A main.ts",
+     "start": "deno serve -A _fresh/server.js",
    }
  }
```
Remove this bit in `main.ts`:

```diff
  // main.ts
  
- if (import.meta.main) {
-   await app.listen()
- }
```

Fixes https://github.com/denoland/fresh/issues/2240
Fixes https://github.com/denoland/fresh/issues/2640
Fixes https://github.com/denoland/fresh/issues/2592
Fixes https://github.com/denoland/fresh/issues/1843


src/fs_routes.ts
```diff
@@ -0,0 +1,280 @@
...
+  routePattern: string;
+}
+
+// deno-lint-ignore no-explicit-any
+function isFreshFile<State>(mod: any): mod is FreshFsMod<State> {
+  return mod !== null && typeof mod === "object" &&
+      typeof mod.default === "function" ||
+    typeof mod.config === "object" || typeof mod.handlers === "object" ||
+    typeof mod.handlers === "function" || typeof mod.handler === "object" ||
+    typeof mod.handler === "function";
```
csvn: This is not a new issue, but I think the null/object check here only applies to the first condition (`mod.default`), since `&&` has higher precedence than `||`. I think all the `||` conditions should be wrapped in a parenthesis for this to be correct, or perhaps to make this easier to read:

```ts
  if (mod === null || typeof mod !== "object") return false;
  return typeof mod.default === "function" ||
    typeof mod.config === "object" || typeof mod.handlers === "object" ||
    typeof mod.handlers === "function" || typeof mod.handler === "object" ||
    typeof mod.handler === "function";
```

src/config.ts
csvn, line 1: The `FreshConfig.staticDir` property seems unused, except for tests. Maybe remove it? It's not in sync with the property value in `Builder` like `root` is (`ctx.config.root` seems to be possibly useful to have in middleware/routes, `ctx.config.staticDir` less so).

init/src/init.ts
```diff
@@ -371,10 +371,8 @@ const exampleLoggerMiddleware = define.middleware((ctx) => {
...
 
-await fsRoutes(app, {
-  loadIsland: (path) => import(\`./islands/\${path}\`),
-  loadRoute: (path) => import(\`./routes/\${path}\`),
-});
+// Include file-system based routes here
+app.fsRoutes();
 
 if (import.meta.main) {
   await app.listen();
```
csvn: Since it's now intended to use `deno serve -A _fresh/server.js` and not `deno run -A main.ts`, should this entrypoint code not also be removed?

init/src/init.ts
```diff
@@ -506,7 +503,7 @@ if (Deno.args.includes("build")) {
       check: "deno fmt --check . && deno lint . && deno check",
       dev: "deno run -A --watch=static/,routes/ dev.ts",
       build: "deno run -A dev.ts build",
-      start: "deno run -A main.ts",
+      start: "deno serve -A _fresh/server.js",
```
csvn, line 502: I really love the `app.fsRoutes()` changes. But it feels a little sad to no longer use `main.ts` as the entrypoint. Especially since the outDir (`_fresh`, which in my project is renamed to `.fresh`) becomes hard-coded in two places. It also looks like the `root` property in `_fresh/server.js` may not always be the same as `Builder.config.root` (`server.js` assumes parent folder, `..`, is the root, which may not be true if `outDir: "../_fresh"` is used?).

It would have been nice if it was possible to keep `main.ts` as the entrypoint, and import the build instead.

```ts
export const app = new App()
  .use(staticFiles())
  .fsRoutes();

// Handles setting up build cache from configured build settings
if (import.meta.main) await app.loadBuild(import('./dev.ts'));

export default {
  fetch: app.handler()
}
```
```ts
export const builder = new Builder({ outDir: ".fresh" });

tailwind(builder);

if (import.meta.main) {
  if (Deno.args.includes("build")) {
    await builder.build();
  } else {
    await builder.listen(() => import("./main.ts"));
  }
}
```

Though I realize this might make it conceptually harder to understand the relationship between dev/main given that they can then import each other and both be entrypoints.

This could be addressed by making build/dev servers CLI commands, and converting `dev.ts` to a configuration instead (essentially default export with current build options + init/listen callbacks). Similar to how many other frameworks use e.g. `vite` + `vite build` and config file for start/dev and build commands, instead of a `dev.ts` file with programmatic API and config inside the code.

csvn (commented):
Sorry for my unsolicited feedback here on this in-progress PR, but was excited given the interesting and breaking changes. Even if I have vague misgivings about the entrypoint, this PR looks really awesome and promising and well worth the breaking changes! 🎉

init/src/init.ts
marvinhagemeister: Thanks for sharing your thoughts on this. I really appreciate having another pair of eyes on this.

The problem with referencing `dev.ts` from `main.ts` is that it pulls in all the dev dependencies on production. Deno scans the AST for import statements or expression, but doesn't check whether they are guarded as you'd need code evaluation to check if you really pass the guard condition. Deno errors on the side of caution and will include all reachable dependencies. This would mean that you'd pull in `esbuild` and other heavy things during production which we want to avoid.

Regarding going with a CLI: I thought about that for a while but this would break running it on Deno Deploy (the "classic" one, on the newer Deploy EA no problems). In Deploy classic the entrypoint _must_ be a file and cannot be a free command or even task. This limits choices a bit for now.

So this kinda leaves us with two options. Hard code it in `main.ts` like:
```ts
if (import.meta.main) await app.loadBuild(import('./_fresh/snapshot.js'));
```
or hard code it in `deno.json` in a task. Between those two options I prefer the task method. I'm open to changing the directory name `_fresh` to `.fresh` in general. Starting with an `_` character is a bit unusual and I think all other frameworks start with a dot too. Now would be the time to change that before a final v2 release is out.

iuioiua:
Shall I close #2874?

init/src/init.ts
csvn: > The problem with referencing dev.ts from main.ts is that it pulls in all the dev dependencies on production

Of course, forgot about this when I first made the suggestion 😅
At least the CLI suggestion seemed feasible, though breaking in current classic deploy. That seemed to have been a suitable option otherwise to keep dev dependencies away from production code (e.g. `deno run fresh/build`).

> Hard code it in `main.ts` like: ...or hard code it in `deno.json` in a task. Between those two options I prefer the task method.

Yeah, of these options the current way seems better. It's less boilerplate overall, and less API surface (if we discount `fresh/do-not-use`).

I may have been a bit thrown off as well by not "seeing" the actual code that is the entrypoint, as that's hidden inside the framework, compared to v1 and v2-alpha versions. It's previously been more explicit in the how the server runs in prod given the `app.listen()` part when `main.ts` runs as the entrypoint, and I was thinking if that could be preserved.

> I'm open to changing the directory name _fresh to .fresh in general. Starting with an _ character is a bit unusual and I think all other frameworks start with a dot too. Now would be the time to change that before a final v2 release is out.

I would like that. I changed it mostly cause how I felt most other generated folders would prefer `.` over `_` as prefix (e.g. `.angular`, `.cache` and even `.deno` inside `node_modules` are some examples). I was just being pedantic 😀

marvinhagemeister: I've played around a bit just now with your idea of doing something like:
```ts
if (import.meta.main) {
  await app.loadBuild(() => import('./_fresh/snapshot.js')); // <-- red squiggly lines
}
```
It's a neat alternative to changing the task too. One thing I didn't consider is that this gives red squiggly lines in the editor when you create a new project or clone it without having it built first. That seems like a deal braker.

I totally related to having everything in the user's code being simpler to understand. But it might just not work well together with generated files/assets.

> I would like that. I changed it mostly cause how I felt most other generated folders would prefer . over _ as prefix (e.g. .angular, .cache and even .deno inside node_modules are some examples)

Sounds like a plan. I can take a look at that after this PR lands. Don't want to make the diff any larger.

src/fs_routes.ts
marvinhagemeister: Good catch, thanks for pointing this out. This looks indeed wrong and your suggestion makes it more clear!

init/src/init.ts
marvinhagemeister: Something just crossed my mind: With moving to `deno serve` is one can _still_ import the server yourself if desired:

```ts
import serve from "./_fresh/server.js";

Deno.serve(serve.fetch)
```

csvn: > One thing I didn't consider is that this gives red squiggly lines in the editor when you create a new project or clone it without having it built first. That seems like a deal braker.

Agree, it feels a bit weird importing something from files that may not exist 😅

> Sounds like a plan. I can take a look at that after this PR lands. Don't want to make the diff any larger.

Of course, that makes total sense 👍

src/config.ts
marvinhagemeister: Good catch, yeah it's unused now. I'll remove the `staticDir` config.

PR #3135, opened 2025-07-25
feat: add separate tailwindcss v3 plugin

marvinhagemeister:
Makes it easier to keep using tailwindcss v3 with newer Fresh versions.

### DIFF
--- a/docs/canary/plugins/tailwindcss.md
+++ b/docs/canary/plugins/tailwindcss.md
@@ -26,11 +26,10 @@ that.
 
 ```diff dev.ts
   import { Builder } from "fresh/dev";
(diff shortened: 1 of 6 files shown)


docs/canary/plugins/tailwindcss.md
```diff
@@ -26,11 +26,10 @@ that.
 
 ```diff dev.ts
   import { Builder } from "fresh/dev";
-  import { app } from "./main.ts";
 + import { tailwind } from "@fresh/plugin-tailwind";
```
marvinhagemeister, line 29: Drive by fix. Noticed that these docs were out of date.

PR #3140, opened 2025-07-25
feat: allow passing temporal objects to islands

nnmrts:
Didn't add any checks for Temporal yet because

1. I don't know how to even approach that (are there similar examples?)
2. I don't know if this PR will even get accepted before Temporal is widely supported

Btw: Is it possible to do the same I did here in a plugin?

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -103,6 +103,9 @@
       "include": ["no-console"]
     }
   },
+  "unstable": [
+    "temporal"
+  ],
   "fmt": {
     "exclude": ["./www/static/**/*.svg"]
   }
--- a/src/jsonify/__snapshots__/round_trip_test.ts.snap
+++ b/src/jsonify/__snapshots__/round_trip_test.ts.snap
@@ -50,6 +50,22 @@ snapshot[`round trip - [ 1, <1 empty item>, 3 ] 1`] = `"[[1,-7,2],1,3]"`;
 
 snapshot[`round trip - /foo["]/ 1`] = `'[["RegExp","foo[\\\\"]", ""]]'`;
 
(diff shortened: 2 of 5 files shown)


deno.json
```diff
@@ -103,6 +103,9 @@
       "include": ["no-console"]
     }
   },
+  "unstable": [
+    "temporal"
+  ],
```
marvinhagemeister, line 108: This is probably a deal breaker for now until the `Temporal` API is marked stable. Forcing all Fresh users to use unstable features might be too much of an ask.

nnmrts: Reasonable. I didn't know changing the config of this repo would force users of fresh to enable Temporal as well. Anyways, there probably is some way to approach all this without even needing that config change, for now I was just too lazy to think more about this after seeing all the failing tests. Even adding a `typeof Temporal !== "undefined"` everywhere resulted in dozens of test fails and I wasn't sure why and how to fix that.

marvinhagemeister: The CI is failing because we test with Chrome and Chrome doesn't support the `Temporal` API yet. See https://caniuse.com/temporal

nnmrts: I know that, what I mean is, even when I avoided the astral crashes, I at least got type errors (if not also test errors at least in the round trip test, not sure anymore) which I wasn't sure how to fix so I figured I'll just open this draft PR in the form it *would* be if Temporal was (more) stable (for example if Chrome ships support for Temporal and astral updates to support that), without any weird workarounds or monkey patches. Sorry for the confusion. 😬

lucacasonato: @marvinhagemeister just FYI, unstable temporal in this repo will not have any effect on users of fresh - `unstable` is not something that we allow deps to set.

bartlomieju:
Closing in favor of #3701

PR #3143, opened 2025-07-26
Closed parentheses in `about.tsk`

guzman109:


marvinhagemeister (approved):
Thanks!

marvinhagemeister:
Looks like a semicolon is missing at the end of these lines too

PR #3145, opened 2025-07-26
fix: Set the minimum height of the initial project page generated by init to the screen height

SisyphusZheng:
This pull request makes a small enhancement to the `Home` page component in `init/src/init.ts` by ensuring the page takes up the full height of the screen.

* [`init/src/init.ts`](diffhunk://#diff-2d3062a5b277212719aa2542000aa6e3a62e90b7217994970cf4a6dfe01237eaL414-R414): Added the `min-h-screen` class to the outermost `div` to ensure the page has a minimum height equal to the screen height.

image


init/src/init.ts
```diff
@@ -411,7 +411,7 @@ export default define.page(function Home() {
   const count = useSignal(3);
 
   return (
-    <div class="px-4 py-8 mx-auto fresh-gradient">
+    <div class="px-4 py-8 mx-auto fresh-gradient min-h-screen">
```
marvinhagemeister, line 417: Nice, can you also add the CSS for that to the fallback code for when people create a project without tailwindcss? The code for that is in the variable `NO_TAILWIND_STYLES`

marvinhagemeister: Ah not quite. The fallback css code is _only_ loaded when you don't pick the "use tailwindcss" option. When you _do_ pick tailwind this will not be used, so it should be a tailwind class name regardless. Tailwind doesn't have a `full-height` class as far as I'm aware.

SisyphusZheng: I finally figured it out. LOL

marvinhagemeister (approved):
Sweet, thanks!

PR #3147, opened 2025-07-28
add correct ctx.state example to init and docs

knotbin:
- Correct outdated, non-functional and confusing "Removal of <Head> component" section of migration guide in the docs to be real functional code that uses Fresh 2 and define helpers correctly
- Added example of ctx.state being used for a head component (title) because it's something almost every single project will need to know how to do.
- Small change to define docs use of define.page to be more in line with its use commonly
- Removed mention of deno-puppeteer - a non-official repository that hasn't been updated in three years - needing to be installed from `CONTRIBUTING.md`


docs/canary/advanced/define.md
```diff
@@ -52,11 +52,11 @@ export const handler = define.handlers({
   },
 });
 
-// When you type `props.data.*` you'll get autocompletion
-export default define.page<typeof handler>((props) => {
+// When you type `ctx.state.*` you'll get autocompletion
+export default define.page((ctx) => {
```
marvinhagemeister: With this change `props.data` will be untyped. `ctx.state.foo` doesn't exist in this example.

docs/canary/examples/migration-guide.md
```diff
@@ -125,33 +125,46 @@ The `<Head>` component was used in Fresh 1.x to add additional tags to the
...
+// utils.ts
+export interface State {
+  title?: string;
+}
+export const define = createDefine<State>();
+
+// routes/about.tsx
+import { define } from "../utils.ts";
+
+export default define.page(function AboutPage(ctx) {
```
marvinhagemeister, line 144: Need to check if we support this style of declaring handlers. Not sure if we do.

knotbin: This is how the init template project declares its pages (even before my changes)

docs/canary/advanced/define.md
knotbin: Yes you're right, I was getting mixed up with the previous example on that page sorry

docs/canary/examples/migration-guide.md
marvinhagemeister: Can you point to where that's the case? I can't find something like that anywhere. I've never seen this pattern around fresh:

```ts
export default define.page(function AboutPage(ctx) {
  GET(ctx) { // <-- This looks like an error. Where is the `GET` function coming from?
    //...snip
  }
}
```

knotbin: Yes you're 100% right I just updated that. Sorry for the misunderstanding I thought your comment was about the `export default define.page(function AboutPage(ctx) {` line above the one that caused the error.

marvinhagemeister (approved):
Sweet, thanks!

PR #3162, opened 2025-07-31
feat(docs): add file/language type icons to code blocks

csvn:
I originally wanted to make it clearer the distinction between "files" and e.g. `` ```tsx foo.tsx `` compared to `` ```sh Terminal `` or `` ```txt Project structure``. I originally started just trying to add a 📝 icon or something. But instead I ended showing file types of all markdown languages.

I took the file type icons for the SVG's from https://github.com/vscode-icons/vscode-icons, which has a MIT license.

I think the icons make it a even more clear what the language of the code snippet is.

I have a follow-up PR for improving code block titles and making examples more consistent in #3163.

## Comparisons

<details>

<summary><strong>Open &mdash; Before | After</strong></summary>

### TS + Shell icons - Hover tooltips and alt text

image

### TS + Shell icons 2

image

### Shell + Dockerfile icons

image

### Nested blocks + Git icon

image

</details>


www/utils/markdown.ts
```diff
@@ -139,9 +210,9 @@ class DefaultRenderer extends Marked.Renderer {
       const icon = `<svg class="icon"><use href="/icons.svg#${type}" /></svg>`;
       return `<blockquote class="admonition ${type}">\n<span class="admonition-header">${icon}${
         label[type]
-      }</span>${Marked.parser(tokens)}</blockquote>\n`;
+      }</span>${this.parser.parse(tokens)}</blockquote>\n`;
     }
-    return `<blockquote>\n${Marked.parser(tokens)}</blockquote>\n`;
+    return `<blockquote>\n${this.parser.parse(tokens)}</blockquote>\n`;
```
csvn, line 217: Change this to ensure we re-use the current renderer recursively, instead of the general one.

This helps ensure code blocks even in block quotes get's rendered consistently. Though this also surfaced a "regression", though it seems to be explicitly styled this way. I can fix it if needed so there's no added border compared to before.

image

www/utils/markdown.ts
```diff
@@ -95,11 +159,18 @@ class DefaultRenderer extends Marked.Renderer {
       title = match[2] ?? "";
     }
 
+    const icon = LOGOS.find((l) => l.file.test(title) || l.lang.test(lang));
+
     let out = `<div class="fenced-code">`;
 
-    if (title) {
+    if (title || icon) {
```
csvn, line 168: Unsure if we want blocks with no title to show here. I think it looks nicer, but a bit on the fence.

I intend to look into adding file/description to blocks in #3163 though, which will make this less of an issue.

www/utils/markdown.ts
```diff
@@ -14,6 +15,69 @@ Marked.marked.use(mangle());
 
 const ADMISSION_REG = /^\[(info|warn|tip)\]:\s/;
 
+const LOGOS = [
+  {
+    lang: /^tsx?$/,
+    file: /\.tsx?$/,
+    src: asset("/logos/typescript.svg"),
+    text: "Typescript",
+  },
```
csvn, line 24: I added file types for everything there was code blocks for in the current code.

The reason I have both file/lang properties is because I wanted ` ```diff main.tsx ` to be shown as TS and not Diff. So `file` is for checking for filename or extension matching on the "title". `lang` is fallback to the code block language to match.

www/static/markdown.css
```diff
@@ -812,11 +812,20 @@ ol.nested li:before {
...
   font-size: 0.8125rem;
   line-height: 1;
+  grid-auto-flow: column;
+  align-items: center;
+  gap: 0.625rem;
+}
+.fenced-code-title > img {
+  background-color: transparent;
+  margin: -0.2rem 0;
+  width: 1.25rem;
```
csvn, line 828: I generally don't like negative margins a lot (can be hard to understand), but it was to mitigate the increased height of block headers. I found icons reasonably readable at ~20px height/width, but this made the headers ~7px larger given that the text before was only 13px.

This margin helps mitigate the issue a bit (without affecting padding for text-only).

docs/latest/concepts/static-files.md
```diff
@@ -29,7 +29,7 @@ version of this path that contains a build ID for cache busting. When the asset
 is requested at this "locked" path, it will be served with a cache lifetime of
 one year.
 
-```jsx routes/page.tsx
+```tsx routes/page.tsx
```
csvn, line 32: Found a couple of cases like this where `js`/`jsx` (or `bash` and `sh`) was used over `ts`/`tsx` for no apparent reason (conflicting with file name). Addressed the cases I found, so all the types I added should cover the ones used.

www/utils/markdown.ts
marvinhagemeister: Nice, I remember having run into this, but didn't have time to look into it myself.

www/utils/markdown.ts
marvinhagemeister: I think it's fine to show even if no title is present.

www/static/markdown.css
marvinhagemeister: Playing around a bit, I feel like adding a bit of spacing around the icons looks great! I don't mind that they are a little larger.

marvinhagemeister (approved):
Nice, this looks really cool! Makes the code snippets feel more "lively". Love this!

PR #3163, opened 2025-07-31
fix(docs): improve and increase consistency for code blocks

csvn:
Follow-up from #3162.

This PR does various tweaks, mostly related to code blocks.

- Adds missing titles to code blocks, and updates titles to be more consistent
  - Adds filenames to most ` ```tsx` blocks with e.g. `main.ts`, `dev.ts` or a suitable example (e.g. `routes/greeting.tsx`) based on the code
  - Adds filenames to ` ```diff` blocks if there's a relevant file, to make the icon e.g. TS over Diff icon
  - Code blocks with Fresh code (e.g. interfaces) uses ` ```tsx fresh 🍋`
  - Changes most ` ```sh` blocks with example terminal commands to ` ```sh Terminal`
- Updates ` ```txt Project Structure` or ` ```diff Project Structure` blocks to be more consistent
  - Changes `sh` or `txt` to `txt-files` which improves highlighting and makes icon consistent
  - Renames `<project dir>` to `<project root>` to match builder `root` option terminology
  - Adds `<project root>` root folder to most/all overview folder overviews
  - Makes the structure blocks more consistently displayed
- Tried to make some "bare" examples more complete

## Comparison

Here's a small curated set of before/after changes.

### Deployment

- `deno` highlighted in terminal commands
- `.gitignore` syntax highlighting works without using `sh` as language

image

### File routing

Updated icon, comments same as before, but root and hierarchy have new colors (`operator`/`function`).

image

### DIFF
--- a/docs/canary/concepts/static-files.md
+++ b/docs/canary/concepts/static-files.md
@@ -32,7 +32,7 @@ By default, Fresh adds caching headers for the `src` and `srcset` attributes on
 
 ```tsx main.tsx
 // Caching headers will be automatically added
-app.get("/user", () => <img src="/user.png" />);
+app.get("/user", (ctx) => ctx.render(<img src="/user.png" />));
 ```
 
 You can always opt out of this behaviour per tag, by adding the
@@ -40,7 +40,10 @@ You can always opt out of this behaviour per tag, by adding the
 
 ```tsx main.tsx
 // Opt-out of automatic caching headers
-app.get("/user", () => <img src="/user.png" data-fresh-disable-lock />);
+app.get(
+  "/user",
+  (ctx) => ctx.render(<img src="/user.png" data-fresh-disable-lock />),
+);
 ```
 
 ## Adding caching headers manually
--- a/docs/canary/concepts/static-files.md
+++ b/docs/canary/concepts/static-files.md
@@ -9,7 +9,7 @@ disk for optimal performance with
 [`ETag`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag)
 headers.
 
-```ts
+```ts main.ts
 import { staticFiles } from "fresh";
 
 const app = new App()
@@ -30,17 +30,17 @@ const builder = new Builder({ staticDir: "path/to/staticDir" });
 By default, Fresh adds caching headers for the `src` and `srcset` attributes on
 `<img>` and `<source>` tags.
 
-```tsx
+```tsx main.tsx
 // Caching headers will be automatically added
-<img src="/user.png" />;
+app.get("/user", () => <img src="/user.png" />);
 ```
 
 You can always opt out of this behaviour per tag, by adding the
 `data-fresh-disable-lock` attribute.
 
-```tsx
+```tsx main.tsx
 // Opt-out of automatic caching headers
-<img src="/user.png" data-fresh-disable-lock />;
+app.get("/user", () => <img src="/user.png" data-fresh-disable-lock />);
 ```
 
 ## Adding caching headers manually
@@ -48,9 +48,11 @@ You can always opt out of this behaviour per tag, by adding the
 Use the `asset()` function to add caching headers manually. It will be served
 with a cache lifetime of one year.
 
-```tsx
+```tsx routes/about.tsx
 import { asset } from "fresh/runtime";
 
-// Adding caching headers manually
-<a href={asset("/brochure.pdf")}>View brochure</a>;
+export default function About() {
+  // Adding caching headers manually
+  return <a href={asset("/brochure.pdf")}>View brochure</a>;
+}
 ```
--- a/docs/canary/deployment/index.md
+++ b/docs/canary/deployment/index.md
@@ -17,7 +17,7 @@ which contains the optimized assets.
 > [info]: The `_fresh` folder should not be committed to git. Exclude it via
 > `.gitignore`.
 >
-> ```sh .gitignore
+> ```gitignore .gitignore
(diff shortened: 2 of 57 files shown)


www/utils/prism.ts
```diff
@@ -6,5 +6,41 @@ import "prismjs/components/prism-diff.js";
...
+ * ```txt-files
+ * <project root>
+ * ├── routes/             # File system based routes
+ * │   ├── _app.tsx        # Renders the outer <html> content structure
+ * │   └── index.tsx       # Renders /
+ * ├── dev.ts     # Run this during development
+ * └── main.ts    # Run this for production
+ * ```
+ */
+Prism.languages["txt-files"] = {
```
csvn, line 37: I didn't like that the myriad of "Project structure" code blocks was using `sh` (for comments coloring) or `txt`. Since #3162, this makes the icons vary and it looks odd when using `sh` for the icon.

The configuration here, together with the styling changes, both improves readability of the code blocks, but also makes it more consistent.

- Lowers line-height to make the gap between "lines" less obvious
- Highlights the project root (renamed from project dir, to match `BuildOptions.root` terminology)
- Makes the lines (`├──`, `└──` or `│   `) colored the same as operators
- Makes `#` until the end of the line comments

## Before/After

image

image

www/utils/prism.ts
```diff
@@ -6,5 +6,41 @@ import "prismjs/components/prism-diff.js";
...
 import "prismjs/components/prism-yaml.js";
+import "prismjs/components/prism-ignore.js";
+
+/** Extends `sh` with `deno` as a function token in Shell/Bash languages */
+Prism.languages.sh.deno = {
+  pattern: /(^|[\s;|&]|[<>]\()(?:deno)(?=$|[)\s;|&])/,
+  lookbehind: true,
+  alias: "function",
+};
+Prism.languages.bash.deno = Prism.languages.sh.deno;
```
csvn, line 17: Very minor fix, just copied the existing `function` code from `sh`/`bash` and replaced tools with `deno`. Just a small win to get nicer highlighting for code snippets.

image

www/utils/prism.ts
```diff
@@ -6,5 +6,41 @@ import "prismjs/components/prism-diff.js";
 import "prismjs/components/prism-json.js";
 import "prismjs/components/prism-bash.js";
 import "prismjs/components/prism-yaml.js";
+import "prismjs/components/prism-ignore.js";
```
csvn, line 9: Makes the `gitignore` language look much nicer, and avoids using `sh` for ignore syntax (for comments).

image

docs/latest/concepts/deployment.md
```diff
@@ -8,8 +8,8 @@ deployed to any system or platform that can run a Deno based web server.
 
 Here are instructions for specific providers / systems:
 
-- [Deno Deploy](#deno-deploy)
-- [Docker](#docker)
+- [Deno Deploy](#-deno-deploy)
+- [Docker](#-docker)
```
csvn, line 12: Dunno if this is supposed to be like this, but the anchors did not work for me before.

csvn:
I browsed through every page in the docs for latest/canary that I could find.

Except for some pages that have blocks for API examples (see image below), I have added and tweaked all other titles I could find.

image

www/utils/prism.ts
marvinhagemeister: Nice, yeah using `sh` language here was more a band aid to get comments dimmed.

docs/latest/concepts/deployment.md
marvinhagemeister: mh yeah that looks odd. We should adapt our code to get rid of the leading dash. Can happen in another PR.

docs/canary/concepts/static-files.md
```diff
@@ -30,27 +30,29 @@ const builder = new Builder({ staticDir: "path/to/staticDir" });
 By default, Fresh adds caching headers for the `src` and `srcset` attributes on
 `<img>` and `<source>` tags.
 
-```tsx
+```tsx main.tsx
 // Caching headers will be automatically added
-<img src="/user.png" />;
+app.get("/user", () => <img src="/user.png" />);
```
marvinhagemeister: I don't thing straight up returning JSX works just yet. At the minimum it should be:

```tsx
app.get("/user", ctx => ctx.render(<img src="/user.png" />));
```

It's a bit sad that this makes it a bit verbose.

That said seeing this kinda makes me wonder if we should support returning JSX out of the box. Seams like a neat feature.

docs/canary/deployment/index.md
```diff
@@ -17,7 +17,7 @@ which contains the optimized assets.
 > [info]: The `_fresh` folder should not be committed to git. Exclude it via
 > `.gitignore`.
 >
-> ```sh .gitignore
+> ```gitignore .gitignore
```
marvinhagemeister, line 20: nice!

docs/canary/concepts/static-files.md
csvn: Oops, forgot about that; updated. Mostly used to the fs-based code, where a page can return JSX directly 😅

I dunno if these examples became much better. I just originally found it odd that the code snippet (`;`) looked more like HTML than TS, and wanted to make the example more realistic.

csvn: I also notice now that I updated the v1.x docs differently than the v2 one.

image

marvinhagemeister (approved):
Sweet, this is great! Really appreciate the new syntax highlighting for gitignore and sh files. Makes the docs look much nicer!

PR #3167, opened 2025-08-01
fix: correct anchors by using heading text only

csvn:
Addresses the issue mentioned in https://github.com/denoland/fresh/pull/3163#discussion_r2246708003.

The issue was that the `raw` markdown was being used for e.g. `## My title`, thus the slug became `-my-title` instead of `my-title`. Slugs were working correctly for `codespan` in headers, e.g. ``## `someCode` ``.

I decided to hard error if unexpected heading syntax is used over sluggifying unexpected content (can become super weird otherwise, e.g. `## <button ...>My heading</button>`). I can revert to fallbacking to `raw` if preferred, but perhaps with at least trimming leading `-` just in case.

### DIFF
--- a/docs/toc.ts
+++ b/docs/toc.ts
@@ -85,8 +85,8 @@ const toc: RawTableOfContents = {
           ["daisyui", "daisyUI", "link:canary"],
           ["modifying-the-head", "Modifying the <head>", "link:latest"],
           ["creating-a-crud-api", "Creating a CRUD API", "link:latest"],
+          ["markdown", "Rendering Markdown", "link:canary"],
           ["rendering-raw-html", "Rendering raw HTML", "link:canary"],
-          ["markdown", "Markdown", "link:canary"],
           [
             "sharing-state-between-islands",
             "Sharing state between islands",
--- a/www/utils/markdown.ts
+++ b/www/utils/markdown.ts
@@ -3,6 +3,7 @@ import denoJson from "../../deno.json" with { type: "json" };
 export { extractYaml as frontMatter } from "@std/front-matter";
 
 import * as Marked from "marked";
+import { HttpError } from "fresh";
 import { asset } from "fresh/runtime";
 import { escape as escapeHtml } from "@std/html";
 import { mangle } from "marked-mangle";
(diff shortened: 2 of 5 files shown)


docs/toc.ts
```diff
@@ -85,8 +85,8 @@ const toc: RawTableOfContents = {
           ["daisyui", "daisyUI", "link:canary"],
           ["modifying-the-head", "Modifying the <head>", "link:latest"],
           ["creating-a-crud-api", "Creating a CRUD API", "link:latest"],
+          ["markdown", "Rendering Markdown", "link:canary"],
           ["rendering-raw-html", "Rendering raw HTML", "link:canary"],
```
csvn, line 89: Thought this looked more consistently named, and is similar to v1 docs.

image

www/utils/markdown.ts
```diff
@@ -231,9 +239,15 @@ export function renderMarkdown(
...
+    const html = opts.inline
+      ? Marked.parseInline(input, markedOpts) as string
+      : Marked.parse(input, markedOpts) as string;
 
-  return { headings: renderer.headings, html };
+    return { headings: renderer.headings, html };
+  } catch (err) {
+    throw new HttpError(500, "Markdown parsing error", {
+      cause: err,
+    });
```
csvn, line 251: Given that it's possible that Marked will throw for unexpected markdown syntax, it's probably a good idea to handle errors here and at least map to a `HttpError`.

marvinhagemeister: FWIW: All Fresh apps have a `try/catch` block at the root of the app so that we'll always return a `Response` object.

https://github.com/denoland/fresh/blob/6a41b7b1943ed3a88b112b3485e35b2743286649/src/app.ts#L424-L438

marvinhagemeister (approved):
Sweet, thanks for correcting these!

www/utils/markdown.ts
csvn: I just wanted to ensure that `props.error` is `HttpError` and not the `Error` from the Markdown parser in https://github.com/csvn/fresh/blob/main/www/routes/_error.tsx/#L33-L45. Might make it easer to handle it here in the future.

PR #3168, opened 2025-08-01
test: check that _app and _layout are ignored by dev err frame

csvn:
Encountered this issue in #3167, but already fixed in #3165. I noticed `/_frsh/error_overlay` was using `_app`/`_layout`, which seemed unnecessary.

Since this was already fixed in #3165, I just added a test to check that this is not a regression in the future.

### DIFF
--- a/src/dev/builder_test.ts
+++ b/src/dev/builder_test.ts
@@ -1,7 +1,8 @@
-import { expect } from "@std/expect";
+import { expect, fn } from "@std/expect";
 import * as path from "@std/path";
 import { Builder, specToName } from "./builder.ts";
 import { App } from "../app.ts";
+import { DEV_ERROR_OVERLAY_URL } from "../constants.ts";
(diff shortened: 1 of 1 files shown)


src/dev/builder_test.ts
```diff
@@ -388,6 +389,45 @@ Deno.test({
...
+
+        const text = await res.text();
+        expect(text).toMatch(/__ok__/);
+        expect(appLayoutSpy).toHaveBeenCalledTimes(0);
+
+        controller.abort();
+      },
+    });
+
+    // expect(text).toContain("<h1>ok</h1>");
```
marvinhagemeister: nit: commented out code

marvinhagemeister: oh wait I'm to blame as this was copy and pasted from my test 😅

marvinhagemeister (approved):
Sweet, thanks! Always good to have more tests.

PR #3170, opened 2025-08-01
feat: allow ctx.render to skip wrappers

csvn:
Allow `ctx.render()` to bypass app wrapper and layouts.

- I originally intended this as a solution for middleware (`app.use()`) to have the same capability as routes (`app.route()`) to solve the `devErrorOverlay()` including wrappers.
- The docs mention it's possible to "opt-out" during `ctx.render()`
  > Every [ctx.render()](https://fresh.deno.dev/docs/canary/concepts/context#render-1) call will include the app wrapper component by default, unless opted out.
  > \- https://fresh.deno.dev/docs/canary/advanced/app-wrapper

I don't need this myself currently, but it seemed a bit weird that middleware was not able to bypass app wrapper or layouts, because it's not possible to pass a `RouteConfig`. If this API is not something you want to support, I could remove the code changes in favor of just updating docs (above) and adding the tests.

### Open questions

Should it not be possible for "index" to use a layout? E.g. calls to `/foo` should include the layout:

```ts
const app = new App()
  .layout("/foo", ({ Component }) => (<>foo/<Component /></>))
  .get("/foo", (ctx) => ctx.render(<>index</>));

// Should this not be "foo/index"? Currently renders "index"
```

Currently, I ignored the test for this, since it was not what I expected. I tried to simulate the following folder layout, but with imperative code:

```
<root>/routes
└── /foo
    ├── _layout.tsx
    └── index.tsx
```

### DIFF
--- a/src/app_test.tsx
+++ b/src/app_test.tsx
@@ -697,3 +697,83 @@ Deno.test("App - .get/post/patch/put/delete/head/all() - lazy", async () => {
   res = await server.head("/");
   expect(await res.text()).toEqual("ok");
 });
+
+Deno.test("App - .appWrapper()", async () => {
+  const app = new App()
+    .appWrapper(({ Component }) => (
+      <>
+        app/<Component />
+      </>
+    ))
+    .get("/", (ctx) => ctx.render(<>index</>));
+
+  const server = new FakeServer(app.handler());
+
+  const res = await server.get("/");
+  expect(await res.text()).toContain("<body>app/index<");
+});
+
+Deno.test.ignore("App - .layout()", async () => {
+  const app = new App()
+    .layout("/", ({ Component }) => (
+      <>
+        layout/<Component />
+      </>
+    ))
+    .layout("/foo", ({ Component }) => (
+      <>
+        foo/<Component />
+      </>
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
Nice, looks like I mistakenly thought we already supported it, but we didn't actually had the code for that. Supporting opting out of these in `ctx.render()` is much more consistent with how the rest of Fresh works 👍 

Regarding the index/layout question. There is likely a scenario where the implementation of file system based routes and layouts leaks onto `App`. On the file system you can have this setup:

```
<root>/routes
└── /foo
    ├── _layout.tsx
    └── index.tsx
```

vs

```
<root>/routes
└── /foo
|   └─── _layout.tsx
└─ foo.tsx
```

There is probably a bug somewhere lurking. I agree with you that I would expect this code snippet to work:

```ts
const app = new App()
  .layout("/foo", ({ Component }) => (<>foo/<Component /></>))
  .get("/foo", (ctx) => ctx.render(<>index</>));
```

marvinhagemeister (approved):
Nice, this makes the API much more consistent. Thanks for adding this!

PR #3175, opened 2025-08-01
fix!: rename the _fresh build folder to .fresh (BREAKING)

csvn:
As [mentioned in a prior discussion](https://github.com/denoland/fresh/pull/3122#discussion_r2226988418), it seemed like renaming `_fresh` build folder to `.fresh` would align more with generated files than the current `_`. This PR implements that suggestion, to make the breaking change potentially sooner rather than later.

### Prior arts

- `node_modules/.deno` - Deno installed packages
- `.vite` - Cache of prebuilt packages and files
- `.angular` - Cache of built files and metadata
- `.github/` or `.vscode/` folders

The `.` prefix in general seems to be used either for generated files, files that should be considered "hidden", or configuration files (e.g. `.editorconfig` or `.gitignore`). This seems to fit well with `outDir`, and would make naming more consistent IMO.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,4 +34,4 @@ jobs:
         with:
           project: "fresh"
           entrypoint: "server.js"
-          root: "./www/_fresh/"
+          root: "./www/.fresh/"
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-_fresh/
+.fresh/
 .vite/
 vendor/
 node_modules/
--- a/deno.json
+++ b/deno.json
@@ -19,7 +19,7 @@
     "release": "deno run -A tools/release.ts"
   },
   "exclude": [
-    "**/_fresh/*",
+    "**/.fresh/*",
     "**/tmp/*",
     "*/tests_OLD/**",
     "**/vite.config.ts.*",
--- a/docs/latest/advanced/builder.md
+++ b/docs/latest/advanced/builder.md
@@ -38,7 +38,7 @@ const builder = new Builder({
   // The path to your server entry point. (Default: `<root>/main.ts`)
   serverEntry?: string;
   // Where to write generated files when doing a production build.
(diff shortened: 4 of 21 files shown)


fry69:
I strongly object to this. `.fresh` is a hidden folder, those usually reserved for configuration files or ephemeral caches.

Generated output should be visible in a clearly marked/named folder. Common folder names are:
- `dist`
- `out`
- `build`
- `public` (that may clash with `static` a bit)

The examples also are not fitting IMHO:
- `node_modules`/.deno` is a special case, it's wild west inside that hierarchy
- `.vite`/`.angular` are cache/staging folders, not generated output. Cache/staging folders are ephemeral, those can be deleted without any side effect, unlike generated output folders, which might break a running site
- `.github`/`.vscode` are well-known special purpose configuration folders

If no agreement on a new folder name for the generated output can be reached, please keep the `_fresh` name, it is at least not a hidden folder.

fibibot (commented):
The rename touches `deno.json`, `.gitignore`, `packages/init`, the deploy workflow, docs, and the migration tool consistently. One thing worth confirming intent on: the asset URL prefix is still `_fresh` — `packages/fresh/src/dev/builder.ts:333` (`/_fresh/js/${BUILD_ID}/`), `packages/fresh/src/dev/dev_build_cache.ts:581`, and `packages/fresh/src/middlewares/static_files.ts:87`. After this PR the on-disk folder is `.fresh` but the URL path served to the browser is still `_fresh/js/...`. `packages/fresh/tests/islands_test.tsx:801` still asserts the `_fresh/js/` URL too. Intentional (URL path independent of build folder), or should they move together?

- nit: `packages/update/src/update.ts:140` only matches `tasks.preview === "deno serve -A _fresh/server.js"` as an exact string — a user who customised the preview task (e.g. added `--port`) won't be migrated.
- nit: the migration tool doesn't touch the user project's `.gitignore`, so existing projects will need a manual `_fresh/` → `.fresh/` swap there.

CI green on b24d008.

PR #3176, opened 2025-08-01
ci: reduce workflow failures on forks

csvn:
Publishing packages workflow and deploying to Deno Deploy step runs even for forks. This is annoying, since it creates some spam every push for any open PR's.

This PR tries to address this, by skipping certain steps or workflows unless the current GitHub org is `denoland`, which should be pretty safe long term.

## Failed workflow spam

image

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,6 +29,8 @@ jobs:
 
       - name: Upload to Deno Deploy
         uses: denoland/deployctl@v1
+        # Skip publishing for forks
+        if: github.repository_owner == 'denoland'
         with:
           project: "fresh"
           entrypoint: "./www/_fresh/server.js"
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,6 +29,8 @@ jobs:
 
       - name: Upload to Deno Deploy
         uses: denoland/deployctl@v1
+        # Skip publishing for forks
+        if: github.repository_owner == 'denoland'
```
csvn, line 33: Is the purpose here clear enough without a comment as well?

marvinhagemeister: yeah, seems fine to me.

marvinhagemeister (approved):
Thanks! Admittedly I never looked at that tab before. Great idea on ensuring that we're on top of that!

.github/workflows/deploy.yml
csvn: Didn't manage to remove the comment since it's already merged. Thanks though 😁

csvn:
Sweet! Just synced my fork after merging this, and everything's green! 👍 

image

PR #3177, opened 2025-08-02
style: fix horizontal scrollbar for some screen sizes

csvn:
On the Fresh docs, I get a vertical scrollbar for a small range of viewport size of the browser. This is annoying me, since I get it when I place two windows next to each other on a 21:9 screen.

The issue seems to stem from a button using `p-1 -m-1`. But the negative margin is there to address the increased space due to making the button easier to click (most likely for touch/mobile). But this makes the button overflow when there is no right-margin (`2xl:mr-0`).

image

## Solution

The simplest and easiest fix here is just to increase the margin to cover for this case. `2xl:mr-1` seems to do the trick. But it might be a good idea to have a little more gap towards the edge of the screen, e.g. `2xl-mr-2`.

### DIFF
--- a/www/components/NavigationBar.tsx
+++ b/www/components/NavigationBar.tsx
@@ -22,7 +22,7 @@ export default function NavigationBar(
   const isDocs = props.active == "/docs";
   return (
     <nav class={"flex " + (props.class ?? "")} f-client-nav={false}>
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #3180, opened 2025-08-02
test(docs): type check docs code examples

csvn:
This PR adds the initial seeds of type checking the code in all code blocks for the Markdown Docs.

- Adds a unit test, disabled due to the many remaining errors, that can help find incorrect code
- Fixes a bunch of syntax errors, type errors, and other code issues found in the docs

## Future improvements

A lot of remaining issues is mostly related to missing imports, e.g. `App`, `staticFiles`, `HttpError` etc. There's a couple of different directions to solve this in future PR's.

- Add explicit imports to all docs examples (may become unnecessarily verbose)
- Add code to docs code blocks, but with a hint to remove it when rendering the markdown as HTML
   ```tsx
   import { App } from "fresh";
   // #docs-start

   // Everything above marker above is trimmed away
   const app = new App();
   ```
- Add a `types.d.ts` with "global" types used during type checking, so globals, imports and so on can be handled globally

I do think that some examples probably should be more explicit with imports, but it would be overkill for a lot of places such as `app.md` and `context.md`.

### DIFF
--- a/docs/canary/concepts/app.md
+++ b/docs/canary/concepts/app.md
@@ -20,7 +20,7 @@ app.listen();
 All items are applied from top to bottom. This means that when you defined a
 middleware _after_ a `.get()` handler, it won't be included.
 
-```ts main.ts
+```tsx main.tsx
 const app = new App()
   .use((ctx) => {
     // Will be called for all middlewares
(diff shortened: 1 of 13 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #3185, opened 2025-08-04
Revert "feat: Add experimental vite plugin (#3184)"

marvinhagemeister:
This reverts commit a1571b76edf22f755ce7df7debea9854109f67a2.

It broke fresh.deno.dev :S

PR #3187, opened 2025-08-04
feat: add vite plugin

marvinhagemeister:
Second attempt after we had to revert https://github.com/denoland/fresh/pull/3185


SisyphusZheng:
that's what I really wanted, cool and thx

PR #3189, opened 2025-08-05
fix: add `deno compile` entry

### LINKED ISSUE #3123: All routes are missing after `deno compile`
Steps to reproduce:

```sh
deno compile -A --include static/ --include _fresh --output fresh_server main.ts
mv ./fresh_server /tmp/
cd /tmp/
./fresh_server
```

When you open the page in the browser, all routes will return a 404 error.

Please note you have to move the executable to somewhere else other than the project directory. If the executable still stays in the project directory, the routes won't be missed. Seems like the executable is still linking some JS/TS files outside the executable.

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/3123

### DIFF
--- a/docs/canary/deployment/deno-compile.md
+++ b/docs/canary/deployment/deno-compile.md
@@ -1,0 +1,15 @@
+---
+description: "Generate a self contained executable with deno compile."
+---
+
+You can create a self-contained executable out of your app with the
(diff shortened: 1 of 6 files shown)


csvn:
Nice! Great to have until https://github.com/denoland/deno/issues/27210 is implemented to support compiling for `deno serve` directly.

PR #3197, opened 2025-08-06
chore: restructure repository into packages/ folder

marvinhagemeister:
We're having a growing number of packages in this repository. This PR moves them into a `packages/` directory like it's commonly done in other projects.


csvn:
After seeing #3194 and the growing number of plugins, I had the exact same thought. I think the new layout makes it much easier to find docs, and navigate the project for new contributors. Really nice change! 💯

PR #3198, opened 2025-08-06
feat: add `define.layout` helper

marvinhagemeister:


csvn:
I just noticed this while quickly checking this PR, but it doesn't seem like `define.page` or `define.layout` accept async component?`

```tsx
export default define.page(async (props) => {
  return <>Hello {props.world}</>;
};
```

There is no error with the above, but it only works since `ComponentChild` from Preact has `object` among it's types. I think maybe it there should be some type that transforms `FunctionComponent` to possibly return `MaybeLazy<ComponentChildren>`. Maybe some mapped type can ensure to return `MaybeLazy<ReturnType<FunctionComponent>>` instead?

csvn:
Sorry, maybe I should create an issue instead 😅

PR #3223, opened 2025-08-18
Fix typo csrf.md

crowlsyong:
safguard --> safeguard


marvinhagemeister (approved):
Whoops, good catch! Thanks for making a PR to fix this!

PR #3227, opened 2025-08-19
www: port fresh.deno.dev to `vite`

marvinhagemeister:
This PR ports fresh.deno.dev to use our vite plugin.


csvn:
Just asking due to curiosity, feel free to ignore if it's hard to answer.

Is changing `www` to use the `vite` plugin to make dogfooding the plugin easier, or is the plan to make the `vite` plugin the primary/main way to build Fresh apps? I think I read that the plan is to support both for now, but it's interesting if there are plans for Fresh v2 or beyond.

marvinhagemeister:
> Is changing `www` to use the `vite` plugin to make dogfooding the plugin easier, or is the plan to make the `vite` plugin the primary/main way to build Fresh apps? I think I read that the plan is to support both for now, but it's interesting if there are plans for Fresh v2 or beyond.

Yes, the plan is to make `vite` the default in the future once our plugin is stable. Dogfooding by moving `www` to vite helps in catching early bugs. There are no plans to deprecate the existing `dev.ts` way to build fresh apps for now. I want to keep supporting that for as long as reasonable to avoid churn for existing projects.

csvn:
Awesome, thanks for the response. 💪

PR #3256, opened 2025-08-25
fix(vite): wrong resolution with query param

marvinhagemeister:


marvinhagemeister:
Replaced by https://github.com/denoland/fresh/pull/3257

PR #3268, opened 2025-08-27
chore: release @fresh/plugin-vite@0.9.6

marvinhagemeister:


fry69:
I think this can get closed, since #3273 does the same and got merged.

PR #3288, opened 2025-09-01
chore: exclude tests from publish

marvinhagemeister:


fry69:
If I am not mistaken, this does the opposite from the title? It **includes** tests during publish?

marvinhagemeister:
Oh no, you're right. I guess I shouldn't code right after waking up 🤦

PR #3299, opened 2025-09-01
feat(vite): check valid imports per environment

### LINKED ISSUE #3296: Enforced server-only imports
I have an issue where I do not want certain files to ever appear in the Fresh bundle on the frontend, and importing anything from those files should be considered a build error. I cannot find a documented or undocumented way to achieve this.

marvinhagemeister:
Fixes https://github.com/denoland/fresh/issues/3296

### DIFF
--- a/packages/plugin-vite/src/plugins/verify_imports.ts
+++ b/packages/plugin-vite/src/plugins/verify_imports.ts
@@ -1,0 +1,168 @@
+import type { Plugin } from "vite";
+import * as cl from "@std/fmt/colors";
+import type { PluginContext } from "rollup";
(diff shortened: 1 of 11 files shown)


packages/plugin-vite/src/plugins/verify_imports.ts
```diff
@@ -0,0 +1,167 @@
+import type { Plugin } from "vite";
+import * as cl from "@std/fmt/colors";
+import type { PluginContext } from "rollup";
```
fry69, line 3: Shouldn't this type get imported from vite instead of rollup? Or am I missing something here?

PR #3312, opened 2025-09-02
docs: mention that <Head> is not dynamic

fry69:
Adds a small note that `<Head>` is not dynamic by default


marvinhagemeister (approved):
Good idea!

PR #3346, opened 2025-09-06
[fix]WWW: Vite SSR remote import error by using JSR package

SisyphusZheng:
### Problem

Vite SSR in Node.js throws the following error:

```
[ERR_UNSUPPORTED_ESM_URL_SCHEME] Only file and data URLs are supported by the default ESM loader. Received protocol 'https'
```

This is caused by attempting to import modules via a remote `https://` URL, which is supported by Deno but not by Node.js/Vite.

### Solution

Added the JSR package for `@fresh/plugin-vite` using:

```sh
deno add jsr:@fresh/plugin-vite
```

Updated `deno.json` imports:

```json
"imports": {
  "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@^0.9.13",
  "@tailwindcss/vite": "npm:@tailwindcss/vite@^4.1.12",
  "vite": "npm:vite@^7.1.4",
  "vite-plugin-inspect": "npm:vite-plugin-inspect@^11.3.2"
}
```

### Result

- The SSR error is resolved.
- The project now works correctly with Vite SSR in Node.js environments.

---


marvinhagemeister:
I don't understand how the proposed changes get rid of the mentioned error message.

SisyphusZheng:
> ```json
> "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@^0.9.13",
> ```

Previously, when I directly forked the project, I couldn't successfully run the dev command in www. An error occurred, and I was required to add a specific file. However, I forgot to take a screenshot of the issue, and now it seems impossible to reproduce it. If I don't encounter this problem again within three days, I will close this

SisyphusZheng:
> I don't understand how the proposed changes get rid of the mentioned error message.

Previously, when I directly forked the project, I couldn't successfully run the dev command in [www](http://www/). An error occurred, and I was required to add a specific file. However, I forgot to take a screenshot of the issue, and now it seems impossible to reproduce it. If I don't encounter this problem again within three days, I will close this

bartlomieju:
Closing since the issue is not reproducible. The `www/` workspace already has access to `@fresh/plugin-vite` via the monorepo workspace -- adding it as an explicit JSR dependency would shadow the local package with a stale published version.

SisyphusZheng:
> Closing since the issue is not reproducible. The www/ workspace already has access to @fresh/plugin-vite via the monorepo workspace -- adding it as an explicit JSR dependency would shadow the local package with a stale published version.

agree thx

PR #3340, opened 2025-09-05
[docs]Add images to getting started guide

SisyphusZheng:
Replaced placeholder TODOs with actual screenshots in the getting started documentation and added three new image files to illustrate the initialization, about page, and countdown examples.

PR #3347, opened 2025-09-06
[fix]Rename geeting-started-2-about.PNG to geeting-started-2-about.png

SisyphusZheng:
#3340 Inconsistent renaming caused images to fail normal detection. Since Git cannot recognize this case-sensitive modification, I will submit two PR.


fry69:
While you are at it, maybe also change `geeting` to `getting`? (same with the other image)

SisyphusZheng:
> While you are at it, maybe also change `geeting` to `getting`? (same with the other image)

oh ur right thx

PR #3349, opened 2025-09-06
docs: include missing imports in the getting started tutorial

jalejomg:
The Countdown code in the tutorial doesn't work when copied and pasted because the imports for useSignal and useEffect are missing. Adding the missing imports will help those following the tutorial to have a better experience.


marvinhagemeister (approved):
Thanks!

PR #3370, opened 2025-09-09
fix(vite): allow optional property access to Deno global

### LINKED ISSUE #3296: Enforced server-only imports

CertainLach:
Ideally, those expressions should be replaced with undefined for better code optimizations, yet I haven't figured it out the best way for this.

Fixes: https://github.com/denoland/fresh/issues/3296#issuecomment-3270452525

### DIFF
--- a/packages/plugin-vite/src/plugins/patches/deno_global.ts
+++ b/packages/plugin-vite/src/plugins/patches/deno_global.ts
@@ -18,6 +18,13 @@ export function denoGlobal({ types: t }: { types: typeof types }): PluginObj {
           ) {
             return;
           }
+          // Ignore `Deno?.thing` expression. Those will result in undefined, and that might be used for shims.
+          if (
+            t.isOptionalMemberExpression(path.parent) &&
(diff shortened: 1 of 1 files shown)


CertainLach:
> Ideally, those expressions should be replaced with undefined for better code optimizations, yet I haven't figured it out the best way for this.

Vite doesn't seem to optimize `undefined?.value` to `undefined` for some reason, so I decided not to replace `Deno?.prop` with `undefined?.prop`, seems like the whole expression needs to be rewritten instead

marvinhagemeister:
Keeping the expression as is, is the right way to go.

marvinhagemeister (approved):
LGTM, thanks!

CertainLach:
> Keeping the expression as is, is the right way to go.

However this prevents dead code elimination, not ideal

PR #3374, opened 2025-09-09
fix(vite): exclude packages from deno cache

CertainLach:
When disabling `node_modules` directory in Deno, prefresh plugin transforms files in ~/.cache/deno/npm/registry.npmjs.org/ directory, causing an arcane error to occur: https://github.com/preactjs/prefresh/issues/124

After disabling this transform my site now works on updated fresh version

I'm not sure if is the correct way to exclude this, but I don't see any longer pattern that would work here

The cache is stored in $XDG_CACHE_HOME/deno/npm/$REGISTRY_HOST, where only deno/npm part seems to be static

### DIFF
--- a/packages/plugin-vite/src/mod.ts
+++ b/packages/plugin-vite/src/mod.ts
@@ -174,7 +174,7 @@ export function fresh(config?: FreshViteConfig): Plugin[] {
     ...devServer(),
     prefresh({
       include: [/\.[cm]?[tj]sx?$/],
-      exclude: [/node_modules/],
+      exclude: [/node_modules/, /\/deno\/npm\//],
       parserPlugins: [
         "importMeta",
         "explicitResourceManagement",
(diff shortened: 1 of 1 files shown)


packages/plugin-vite/src/mod.ts
```diff
@@ -166,7 +166,7 @@ export function fresh(config?: FreshViteConfig): Plugin[] {
     ...devServer(),
     prefresh({
       include: [/\.[cm]?[tj]sx?$/],
-      exclude: [/node_modules/],
+      exclude: [/node_modules|deno\/npm/],
```
CertainLach: See the pull request description, plus I'm not sure if that would work on windows with alternative (`\`) path separator.

As far as I can see, vite normalizes the paths, but I can't test that.

CertainLach:
Extended matched part `deno/npm` => `/deno/npm/` to reduce risk of possible path collision

marvinhagemeister (approved):
Good catch, thanks!

PR #3385, opened 2025-09-10
docs: update testing with routes and islands

### LINKED ISSUE #3383: docs: Example in the "Testing with file routes and islands" section of Testing is incomplete
The example `my-app.test.ts` contains the following:
```lang=ts
import { createBuilder } from "vite";
const builder = await createBuilder({
  // config omitted for brevity
});
await builder.build();
```

This doesn't compile. Using a freshly initialized fresh 2.0 project, creating a test with the same content as the example (adjusted for the paths) causes a compilation error:

```
$ deno test -A tests/
Check file:///[...]/fresh-project/tests/test.test.ts
TS2554 [ERROR]: Expected 1 arguments, but got 0.
await builder.build();
              ~~~~~
    at file:///[...]/fresh-project/tests/test.test.ts:10:15

    An argument for 'environment' was not provided.
      build(environment: BuildEnvironment): Promise<RollupOutput | RollupOutput[] | RollupWatcher>;
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        at file:///[...]/fresh-project/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/index.d.ts:2181:9

error: Type checking failed.

  info: The program failed type-checking, but it still might work correctly.
  hint: Re-run with --no-check to skip type-checking.
```

Unsurprisingly, running with `--no-check` fails as vite attempts to access its properties.

It seems to be a left-over of when the legacy builder was used (https://fresh.deno.dev/docs/advanced/builder). That example uses a similar argument-less call to `builder`.

fry69:
closes #3383

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,3 @@ coverage/
 vite-inspect/
 .vite-inspect/
 demo/dist/
-
-# Local development
-CLAUDE.md
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ coverage/
 vite-inspect/
 .vite-inspect/
 demo/dist/
+
+# Local development
+CLAUDE.md
--- a/docs/latest/testing/index.md
+++ b/docs/latest/testing/index.md
@@ -147,15 +147,23 @@ behavior:
 ### Server-side rendering of islands
 
 You can test that your islands render correctly on the server using the same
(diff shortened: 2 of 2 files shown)


marvinhagemeister (approved):
Thanks!

PR #3394, opened 2025-09-11
fix: basePath handling for static assets

### LINKED ISSUE #3391: Fresh basePath or Vite base property not applied correctly.
I want to mount a Fresh app to an Hono instance.

Doing this works:
```
import { Hono } from "jsr:@hono/hono";
import UI from "./_fresh/server.js";

const app = new Hono();
app.mount('/', UI.fetch); 

Deno.serve(app.fetch);
```

But i want to mount the fresh app to another path:

```
app.mount('/ui', UI.fetch); 
```

Opening http://localhost:8000/ui from browser correctly loads index.html, but all other files/assets return a 404.
That is a relative path issue.
I tried setting "basepath" in FreshConfig, and also "base" in vite.config.ts.
Either way i could not make it work.
I think Fresh is overriding/ignoring the vite's base property.

Trying with the initial fresh counter app.
Opening http://localhost:8000/ui/api/World from browser works and returns "Hello, World!".
Using Deno 2.5.0 and Fresh 2.0.

fry69:
closes #3391 

---
Generated PR message:

Fix base path handling for assets in plugin-vite

Fixes issue where Fresh apps mounted at non-root paths (e.g., /ui) in frameworks like Hono would fail to load assets correctly.

  Problem

  When mounting a Fresh app with app.mount('/ui', UI.fetch), the HTML would load but all assets (CSS, JS, images) would return 404 because asset paths
  were hardcoded to start with /assets/ instead of respecting the Vite base configuration.

  Solution

  - Modified packages/plugin-vite/src/plugins/server_entry.ts to read Vite's base config and apply it to asset paths
  - Added getAssetPath() helper function to construct proper asset URLs with base path
  - Updated both CSS and asset file registration to use the base path

  Testing

  Added comprehensive test case "vite build - base path asset handling" that:
  - Builds demo with custom base path /my-app/
  - Verifies generated server.js contains correctly prefixed asset paths
  - Follows existing test patterns and utilities

  Usage

  Users can now properly mount Fresh apps by configuring both:
```ts
  // vite.config.ts
  export default defineConfig({
    base: "/ui/",
    plugins: [fresh()],
  });

  // main.ts  
  const app = new App({ basePath: "/ui" });

  // hono-app.ts
  app.mount('/ui', UI.fetch); // Assets now work correctly
```

  Backward compatible - no changes needed for apps mounted at root path.


marvinhagemeister (approved):
Sweet, thanks!

PR #3397, opened 2025-09-11
fix: avoid showing "Copied" if copying fails

csvn:
I noticed that in denoland/fresh#3393 it looks like it's possible that the "Copied"/Check icon will be shown even if copying to clipboard fails.

This PR changes it so that the Check icon only shows if copy succeeds.

### DIFF
--- a/www/assets/styles.css
+++ b/www/assets/styles.css
@@ -94,16 +94,7 @@ html[data-theme="dark"] ::selection {
   background-repeat: no-repeat;
 }
 
-.group:not([data-copied]) .group-not-copied {
-  display: block;
-}
-.group[data-copied] .group-not-copied {
-  display: none;
-}
-
-.group[data-copied] .group-copied {
-  display: block;
-}
+.group[data-copied] .group-not-copied,
 .group:not([data-copied]) .group-copied {
   display: none;
 }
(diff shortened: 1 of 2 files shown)


www/assets/styles.css
```diff
@@ -94,16 +94,7 @@ html[data-theme="dark"] ::selection {
...
-.group:not([data-copied]) .group-not-copied {
-  display: block;
-}
-.group[data-copied] .group-not-copied {
-  display: none;
-}
-
-.group[data-copied] .group-copied {
-  display: block;
-}
```
csvn, line 106: I think it's probably fine here to leave the "original" `display` value instead of forcing `block` here? Isn't it better to set `block` or `grid` or w/e on the original element with Tailwind?

csvn (commented):
Due to denoland/deno#30695 I wasn't able to test this out locally, so not 100% sure if the CSS works as intended for Markdown blocks.

www/assets/styles.css
csvn: Managed to test this now locally. Looks correct to me without `display: block`. Seems like an improvement to me, in case e.g. icons want another display mode such as `inline-block`.

marvinhagemeister (approved):
Thanks!

PR #3399, opened 2025-09-12
fix: show help text and refresh help formatting

csvn:
- Show help text when `-h` or `--help` is passed to `@fresh/init`
- Print current version for `@fresh/init` during `--help`
  image
- Refresh `--help` text with colors similar to `deno -h`; + "Fresh branding" style for the `@fresh/init` text
- Print `@fresh/core` version when `@fresh/init` runs
  image

I think showing versions is helpful for verifying/debugging what Fresh version will be installed during init before all files are generated.

#### Before

image

#### After

image

#### Before

image

#### After

image

### DIFF
--- a/packages/init/src/init.ts
+++ b/packages/init/src/init.ts
@@ -2,6 +2,7 @@
 import * as colors from "@std/fmt/colors";
 import * as path from "@std/path";
 import * as semver from "@std/semver";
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #3401, opened 2025-09-12
fix(vite): remove Deno global check

marvinhagemeister:
Ultimately, it causes more problems than it solves.


csvn:
> Ultimately, it casues more problems than it solves.

Maybe a lint rule would be safer? I made #3179 because I feel like it would be easier to add more lint rules in TS rather than Deno core for Fresh. (Though the PR might need some polish)

marvinhagemeister:
Yeah a lint rule might be better. Although it's tricky to determine whether a module is used in an island context or not.

csvn:
> Yeah a lint rule might be better. Although it's tricky to determine whether a module is used in an island context or not.

Ahh... right. We can't determine any imported `components/*`, only Islands (`/islands/*`, `routes/**/(islands)/*`) directly. 🤔

Issue #3402, opened 2025-09-12
Unable to deploy Fresh to Cloudflare Workers


4x25:
Hi, I'm trying to deploy a Fresh project to Cloudflare Workers, but I've been unsuccessful.

Here are the steps I followed:

1.  I created a new project using the following command:
    ```
    deno run -Ar jsr:@fresh/init
    ```

2.  I followed the [Cloudflare Workers | Fresh docs](https://fresh.deno.dev/docs/deployment/cloudflare-workers) and added the required code to `vite.config.ts` file.

3.  I added a `wrangler.jsonc` file with the following content:
    ```jsonc
    {
      "$schema": "node_modules/wrangler/config-schema.json",
      "name": "fresh2-example",
      "main": "./main.ts",
      "compatibility_date": "2025-04-01",
      "compatibility_flags": ["nodejs_compat"],
      "observability": {
        "enabled": true
      },
      "upload_source_maps": true
    }
    ```

4.  I added the following code to the end of my `main.ts` file:
    ```typescript
    export default {
      fetch: app.handler(),
    };
    ```

5.  I pushed the project to a GitHub repository, which you can find here: [mzl980425/fresh2-example](https://github.com/mzl980425/fresh2-example). Then, I created a new Worker on Cloudflare and connected it to this repository. The worker's build configuration is as follows:
    *   **Build command:** `npx deno task build`
    *   **Deploy command:** `npx wrangler deploy`

6.  After the deployment was complete, I found that static assets are served correctly (e.g., `https://xxx.workers.dev/favicon.ico`, `https://xxx.workers.dev/logo.svg`). However, all other page and API routes return a "NOT FOUND" error, including the root page (`https://xxx.workers.dev/`) and API routes like (`https://xxx.workers.dev/api2/world`).

Is there something wrong with my code or configuration? Any help would be greatly appreciated.

Thanks

marvinhagemeister:
Thanks for sharing the detailed steps of what you tried. It looks like step 4 is wrong. The proper entry point is autogenerated by fresh and is available under `_fresh/server.js` after having run `deno task build`. Directly importing from `main.ts` will not work because it's missing the autogenerated code that sets up the route manifest and much more.

4x25:
I reverted step 4, and updated my `wrangler.jsonc` file, changing the `main` entry to `"./_fresh/server.js"`.

Then I run `deno task build`, the following error is in the console:

[...]

PR #3403, opened 2025-09-12
docs: mention entry in cloudflare worker docs

marvinhagemeister:
Related https://github.com/denoland/fresh/issues/3402


fry69:
It seems this is not enough? See -> https://github.com/denoland/fresh/issues/3402#issuecomment-3284278472

PR #3405, opened 2025-09-12
docs: update cloudflare wokers docs

marvinhagemeister:


fry69:
Are you sure this trick with the `server.js` file importing something from `_fresh` will work?

With a clean slate there is nothing in `_fresh` and the build will fail.

PR #3420, opened 2025-09-12
feat: allow initalizing with "src" dir

csvn:
I have existing v1 Fresh projects that I was planning to upgrade to v2. But I was unsure about how to migrate, since I was using a `src/` folder. By reading the source code, I found that with Vite, the `root` property can be used to select the root folder for all code.

This PR re-introduces the docs example for how to use a "src" folder from Fresh v1, but also adds a `--src-dir` option during init such that a project can be initialized with a `src` dir.

image

### DIFF
--- a/docs/latest/examples/use-src-dir.md
+++ b/docs/latest/examples/use-src-dir.md
@@ -1,0 +1,63 @@
+---
+description: |
+  Change the source directory to effectively manage your project.
+---
+
+To reduce the number of files in the root project directory, it's possible to
+configure a `"src"` directory when initalizing Fresh.
+
+```sh Terminal
+# Move code into "src" folder
+deno run -Ar jsr:@fresh/init --src-dir
+```
+
+It's also possible to use a different folder name:
+
+```sh Terminal
+deno run -Ar jsr:@fresh/init --src-dir=app
+```
+
+## Files structure
+
+When initializing the project with `--src-dir`, the structure will look roughly
+like below:
+
+```txt-files Project structure
+<project root>
+├── src/
+│   ├── components/
+│   │   └── Button.tsx
+│   ├── islands/
+│   │   └── Counter.tsx
+│   ├── routes/
+│   │   └── index.tsx
+│   ├── static/
+│   ├── client.ts
+│   └── main.ts
+├── deno.json
+└── vite.config.ts
+```
+
+## Updating an existing project
+
+To migrate an existing Fresh project using Vite, to use a `src` directory follow
+the steps below.
+
+1. Move all code related to Fresh into the `src` folder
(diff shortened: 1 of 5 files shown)


docs/latest/examples/use-src-dir.md
```diff
@@ -0,0 +1,63 @@
...
+   ```diff
+    export default defineConfig({
+   +  root: "src",
+      plugins: [fresh()],
+    });
+   ```
+3. Update the `"start"` task in `deno.json` to the new `_fresh` location
+   ```diff
+   - "start": "deno serve -A _fresh/server.js",
+   + "start": "deno serve -A src/_fresh/server.js",
```
fry69, line 60: Is it a good idea to put generated artifacts into the `src/` folder? Even when those get `gitignored`, they do not belong there IMHO.

marvinhagemeister (commented):
I like this a lot! Wondering if we can find a way to put the `_fresh` directory at the top. It feels wrong to put it into `src/`

bartlomieju (commented):
Nice feature! A few things I noticed:

1. **Bug: `--watch` paths not updated** — The `dev` task generates:
   ```
   deno run -A --watch=static/,routes/ src/dev.ts
   ```
   But with a src dir, the watched paths should be `src/static/,src/routes/` (or relative from the src dir), otherwise file watching won't work.

2. **Typo**: "initalizing" → "initializing" in `use-src-dir.md`

3. **`check` task not updated**: `deno check` might need the src dir path too, depending on how Fresh resolves it.

lunadogbot (commented):
The ubuntu-only CI failure is the `typos` check (`.github/workflows/ci.yml` runs `crate-ci/typos` only on `ubuntu` + `v2.x`) — "initalizing" appears in the PR title, in `docs/latest/examples/use-src-dir.md:6`, and in the `--src-dir` block of `HELP_TEXT` at `packages/init/src/init.ts:65` if the help text is exercised. Fixing the spelling everywhere should turn that job green and resolves @bartlomieju's earlier typo nit.

Also worth a heads-up before rebasing: `packages/init/src/init.ts` on main moved `nodeModulesDir` from `"auto"` → `"manual"` after this branch was cut. The PR still writes `"auto"` at `init.ts:576`. Naive conflict resolution would silently revert main's change — preserve `"manual"` on rebase.

The `--watch=static/,routes/` concern from the prior review is moot since `--src-dir` is Vite-only and Vite overrides `dev`/`build` to `vite`/`vite build` a few lines later.

- nit: `viteConfig += \`  root: "${src.slice(0, -1)}",\n\`` doesn't escape — `--src-dir='a"b'` produces invalid TS. `JSON.stringify(src.slice(0, -1))` is the one-liner fix.

Holding approval until CI is green.

Issue #1705, opened 2023-08-25
Support easily using a CSP across more than one route at a time


kevingorski:
Per [this Discord thread](https://discord.com/channels/684898665143206084/991511118524715139/1144709728686059701):

> KG: Is there a way to apply a CSP to more than a single route at a time?
> MH: It should be doable via a middleware and add these headers. I'm not too happy about the hook we have for that at the moment. Would be great to have a GH issue and come up with a better way (likely middlewares)

adamgreg:
It doesn't seem to be possible to set this via a middleware currently, but I have a (somewhat unpleasant) workaround, in case it helps anyone else too lazy to configure their routes individually 😜 .

In `main.ts`, before running `start()`, I modify the route modules in the manifest, making it as if they all exported a config with `csp: true`:

```ts
// Modify routes here to enable CSP for every route without exception.
// @ts-ignore: Manifest route types must be changed by this operation
manifest.routes = Object.fromEntries(
  Object.entries(manifest.routes).map(([k, v]) => {
    const config = { ...((v as { config?: RouteConfig }).config), csp: true };
    return [k, { ...v, config }];
  }),
);
```

Then call the `useCSP()` hook in the layout / app wrapper. It took me some trial & error to find directives that worked. In particular the way Fresh handles islands caused me some trouble - it uses a script tag (including a nonce that gets added to the header by Fresh), but within that script it ESM imports that I had to allow using `'strict-dynamic'`. Also watch out for the quote marks that are required inside the directive strings. Here's what works for me:

```ts
// Allow static content from this host
csp.directives.defaultSrc = ["'self'"];

// Allow inline styles
csp.directives.styleSrc = ["'self'", "'unsafe-inline'"];

// Allow ESM imports in scripts (scripts tags have a nonce, added later)
csp.directives.scriptSrc = ["'strict-dynamic'"];
```

marvinhagemeister:
Resolved by https://github.com/denoland/fresh/pull/3421

PR #3421, opened 2025-09-12
Add CSP and header middleware

emilkrebs:
Added a CSP and header middleware to simplify the HTTP Security Response Header creation.

Fixes:
- https://github.com/denoland/fresh/issues/1705

Example:

```ts
// main.ts

export const app = new App<State>()
    .use(csp({
        reportTo: '/api/csp-reports',
    }))
    .use(headers({
        'Cross-Origin-Resource-Policy': 'same-site',
        'Cross-Origin-Embedder-Policy': 'require-corp',
        'Cross-Origin-Opener-Policy': 'same-origin',
        'Permissions-Policy':
            'geolocation=(), camera=(), microphone=(), payment=(), usb=(), bluetooth=(), magnetometer=(), gyroscope=(), accelerometer=()',
        'Referrer-Policy': 'strict-origin-when-cross-origin',
        'X-Content-Type-Options': 'nosniff',
        'X-Frame-Options': 'DENY',
        'Strict-Transport-Security': 'max-age=63072000; includeSubDomains; preload',
    }))
```

Endpoint example:
```ts
// routes/api/csp-handler.ts

import { Context } from 'fresh';
import { State } from '../../main.ts';

export const handler = {
    async POST(ctx: Context<State>) {
        try {
            const text = await ctx.req.text();
            const body = JSON.parse(text);

            const logEntry = {
                timestamp: new Date().toISOString(),
                report: body,
                userAgent: ctx.req.headers.get('user-agent') || 'unknown',
            };
            await Deno.writeFile('csp-reports.log', new TextEncoder().encode(JSON.stringify(logEntry) + '\n'), { append: true });

            return new Response(JSON.stringify({ status: "ok" }), { status: 200 });
        } catch (error) {
            console.error("Error handling CSP report:", error);
            return new Response(JSON.stringify({ status: "error", error: String(error) }), { status: 500 });
        }
    },
};

```

Implementing these headers improved the security report at [Security Headers](securityheaders.com) from F to A and helps to prevent security vulnerabilities like Cross-Site Scripting and Clickjacking.

**Before**
image
**After**
image


packages/fresh/src/middlewares/headers.ts
```diff
@@ -0,0 +1,23 @@
...
+  headers?: Record<string, string>,
+): Middleware<State> {
+  return async (ctx) => {
+    const res = await ctx.next();
+    for (const [key, value] of Object.entries(headers || {})) {
+      res.headers.set(key, value);
+    }
+    return res;
+  };
+}
```
marvinhagemeister: Let's drop this file entirely. It's not used for the csp code.

packages/fresh/src/middlewares/csp.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @example
+ * app.use(csp({
+ *   reportOnly: true,
+ *   reportTo: '/api/csp-reports',
+ *   csp: [
+ *       "script-src 'self' 'unsafe-inline'",
+ *   ],
+ * }));
+ */
+export function csp<State>(options: CSPOptions = {}): Middleware<State> {
```
marvinhagemeister, line 26: Sweet, would you be down for adding some tests?

packages/fresh/src/middlewares/headers.ts
emilkrebs: Got it! The idea kinda was to make repeated header creation a bit easier and better structured, so you don't have to use `res.headers.set(key, value);`.

marvinhagemeister (approved):
Nice, thanks! I've took liberty and added some docs as well as add en export to the main entrypoint so that this new middleware can be used 👍

PR #3423, opened 2025-09-13
fix: basePath handling for assets

### LINKED ISSUE #3391: Fresh basePath or Vite base property not applied correctly.

fry69:
trying to fix #3391 again, this gets a bit hairy, it forces propagating the basePath through the runtime.

Well, this code may not win a beauty contest, but so far it seems work and handle the mentioned cases.

Please review, check, flame me and revert my previous basePath PRs if this path is too much of a rabbit hole 😄


bartlomieju:
A few things I noticed:

1. **Public API change to `asset()` and `assetSrcSet()`** — these are exported from `shared.ts` and adding a `basePath` parameter changes the public signature. Users calling `asset("/foo.css")` won't break (it's optional), but it's a leaky abstraction — why should callers need to know about `basePath`? Internally Fresh always has access to `basePath` through render state, and `BUILD_ID` is already set as a module-level variable. Could `basePath` be handled the same way instead of threading it through parameters?

2. **`applyBasePath` can double-prefix** — there's no guard against `path` already starting with `basePath`. If someone does `asset("/ui/foo.css")` with `basePath="/ui"`, they'd get `/ui/ui/foo.css`.

3. **Character validation regex is incomplete** — `/@#?&=\s/` blocks some characters but allows others that are problematic in URL paths (e.g., `%`, `{`, `}`, `<`, `>`). Consider validating against a stricter URL path character set, or just doing `new URL(basePath, "https://localhost")` and checking it roundtrips cleanly.

4. **Island CSS paths** — in `RemainingHead()`, island CSS paths (`island.css[i]`) get `applyBasePath` applied. Are these paths guaranteed to be root-relative without the base? If they already include the base from the build step, this would double-prefix them.

bartlomieju (approved):
Thanks!

bartlomieju:
Actually I'm gonna close this one since the original issue is now fixed.

PR #3428, opened 2025-09-14
fix(tests): resolve resource leak in www/main_test.ts

fry69:
The `buildVite` function returns a disposable object which was not being properly disposed of, leading to a resource leak warning in the test suite.

This change wraps the tests in `www/main_test.ts` in an async function and uses `await using` to ensure that the temporary directory created by `buildVite` is cleaned up after the tests complete.


www/main_test.ts
```diff
@@ -6,88 +6,93 @@ import { expect } from "@std/expect";
 import { retry } from "@std/async/retry";
 import { buildVite } from "../packages/plugin-vite/tests/test_utils.ts";
 
-const result = await buildVite(import.meta.dirname!);
+async function runTests() {
```
iuioiua: Try [`Deno.test.beforeAll`](https://docs.deno.com/runtime/fundamentals/testing/#available-hooks) instead.

fry69: I do not know if using more Deno 2.5.0 isms is the way, people are already complaining that the Fresh 2 does not work with Deno LTS (2.2.14), see here -> https://github.com/denoland/deno/issues/30712

marvinhagemeister: It's fine imo for our own test suite. User's don't typically run our test suite.

fry69:
Well, the test pass now, but I think the PR title is a bit hyperbole now with these in place:
```ts
Deno.test({
  name: "shows version selector",
  sanitizeOps: false,
  sanitizeResources: false,
}, async () => {
```
But since suppressing resource leak warnings in tests seem to be extremely common in this codebase, it is okay to do so?
```shell
$ git grep "sanitizeResources: false" | wc
      75     225    5193
```

fry69:
In other news: I think I may have tripped over this issue -> https://github.com/denoland/deno/issues/30675

I'll leave this PR open for now as kind of an issue, if case others trip over this too.

www/main_test.ts
```diff
@@ -24,70 +41,71 @@ Deno.test("CORS should not set on GET /fresh-badge.svg", async () => {
...
-          cwd: result.tmp,
-          args: ["serve", "-A", "--port", "0", "_fresh/server.js"],
-        },
-        async (address) => {
-          await withBrowser(async (page) => {
-            await page.goto(`${address}/docs`);
-            await page.waitForSelector("#version");
+  sanitizeOps: false,
+  sanitizeResources: false,
+}, async () => {
```
marvinhagemeister: Let's reduce indentation noise by moving the test function into the options object `fn: async () => {`.

fry69: Done.

marvinhagemeister (approved):
LGTM

fry69:
Note to self: I should read more existing code.

PR #3430, opened 2025-09-14
fix(@fresh/core): Handle route components when returning Promise without explicit `async`/`await`

argshook:
During migration from fresh v1 to fresh v2 i noticed that page handlers in v2 are not rendering implicit promises.

it appears that there's a difference how the runtime of fresh v2 treats route handlers like these:

```typescript
define.page(() => {
  return <div>hello</div>;
})
```

and

```typescript
define.page(async () => { // <-- async here, no other changes
  return <div>hello</div>;
})
```

Linter complains that async is not needed and might even auto-remove it if IDE or some commit-hook is configured to do so.

For simple routes it's not an issue. But i had something like this:

```tsx
export default define.page((ctx) =>       // <-- no async here
  getTrainerList()                        //  <-- returns a Promise
    .match((trainers) => {

      // ... more logic.
      // the happy path returns JSX for fresh

      return (
        <Hydrant dehydratedState={dehydrate(queryClient)}>
          <TrainersList appUrl={ctx.state.appUrl} />
        </Hydrant>
      );
    }, (error) => {
      // the sad path returns 404

      console.error(`Error rendering ${ctx.route}`);
      console.error(error);
      throw new HttpError(404);
    })
);
```

but this doesn't work! i was digging through my code until i played with adding async:

```diff
-export default define.page((ctx) => // ...
+export default define.page(async (ctx) => // ...
```

this works. but linter complains that async is not needed. Which is annoying and too easy to use auto-fix.

However, without `async` the runtime of fresh v2 doesn't handle the promise and the route is not rendered at all.

This PR adds a test and fixes the issue by always awaiting the return value of the handler.

Please review, i'm somewhat hand-wavy with the implementation 🙊

### DIFF
--- a/packages/fresh/src/define.ts
+++ b/packages/fresh/src/define.ts
@@ -1,7 +1,7 @@
 import type { AnyComponent } from "preact";
 import type { HandlerByMethod, HandlerFn, RouteHandler } from "./handlers.ts";
 import type { Middleware } from "./middlewares/mod.ts";
-import type { PageProps } from "./render.ts";
+import { isAsyncAnyComponent, type PageProps } from "./render.ts";
 
 /**
  * A set of define functions that enable better type inference and code
@@ -173,6 +173,11 @@ export function createDefine<State>(): Define<State> {
       return handlers;
     },
     page(render) {
+      if (isAsyncAnyComponent(render)) return render;
+      if (typeof render === "function") {
+        // deno-lint-ignore no-explicit-any
+        return async (ctx) => await (render as any)(ctx);
+      }
       return render;
     },
     layout(render) {
--- a/packages/fresh/src/fs_routes.ts
+++ b/packages/fresh/src/fs_routes.ts
@@ -39,11 +39,15 @@ export interface FsRouteFile<State> {
   css: string[];
 }
 
-// deno-lint-ignore no-explicit-any
-function isFreshFile<State>(mod: any): mod is FreshFsMod<State> {
+function isFreshFile<State>(
+  // deno-lint-ignore no-explicit-any
+  mod: any,
+  commandType: CommandType,
+): mod is FreshFsMod<State> {
   if (mod === null || typeof mod !== "object") return false;
 
   return typeof mod.default === "function" ||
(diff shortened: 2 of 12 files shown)


fry69:
Regarding the PR title: I think the convention is to put the package name from the `packages/` folder into the braces, for `fresh` itself I think `(core)` is appropriate, or just leave the braces out? Maybe it is also a good idea to rename the `fresh` package to `core`?

Also I think making test pass in CI is good idea if you aim for inclusion. That can be daunting in some cases...

argshook:
Hey, thanks for checking!

> Regarding the PR title: I think the convention is to put the package name from the `packages/` folder into the braces, for `fresh` itself I think `(core)` is appropriate, or just leave the braces out? Maybe it is also a good idea to rename the `fresh` package to `core`?

package name is `fresh` so i used `fix(fresh)`. Not sure what is the agreed convention, just looked at the last PRs and tried to follow. Looks like it's conventional commits approach.

Let me know what the proper name should be, although don't think it's crucial.

> Also I think making test pass in CI is good idea if you aim for inclusion. That can be daunting in some cases...

Absolutely, however i'm having trouble running tests locally on `main` (pre-PR code), still checking. Not yet convinced my change is causing them to fail, but am looking.

edit: i am now convinced it's my code making tests fail 😎 checking!

fry69:
> package name is fresh

The folder name is `fresh`, but the package name is `@fresh/core` if you look inside the `deno.json` file. That is the confusing part I am aiming at.

argshook:
In fresh 1.7.3 returning a promise from route handler callback worked because there was this: https://github.com/denoland/fresh/blob/1.7.3/src/server/defines.ts#L21

in fresh 2 the define helper is only for type inference, there's no runtime code, so nothing async/awaits anymore.

i thought maybe instead of tweaking the rendered (which is sensitive to such changes), i'll instead adapt the same thing from 1.7.3.

Unfortunately one or the other thing breaks somewhere in the pipeline.

Closing, i'll manage to live without this 😁

PR #3446, opened 2025-09-16
fix: handle websocket connections using experimental deno:request event

### LINKED ISSUE #3350: Fresh V2 Beta WebSocket upgrade request never reaches server behind vite development environment
Hey there,

I've recently just migrated Fresh V2 from Alpha to Beta and now I'm noticing that WebSocket connections hangs and never arrives in a vite development environment. I've verified that the endpoint is most definitely reachable by sending a GET request, but however if the browser attempts a WebSocket connection, it seems to never even reach the server.

I'm guessing this has to do something with vite. No issues with production, building the project and starting has no issues with WebSocket connectivity.

```js
//Client Connection: new WebSocket("ws://127.0.0.1:5174/somewebsocket")
//Server never sees WebSocket upgrade request.

app.get("/somewebsocket", (ctx) => {
     const { socket, response } = Deno.upgradeWebSocket(ctx.req)

    //More logic here.

    return response
})
```

Thanks!

CertainLach:
Fixes: #3350

Requires deno with this PR: https://github.com/denoland/deno/pull/30746

Context:
- https://github.com/denoland/fresh/issues/3350#issuecomment-3294489536
- https://github.com/denoland/fresh/issues/3350#issuecomment-3294522965

### DIFF
--- a/packages/plugin-vite/src/plugins/dev_server.ts
+++ b/packages/plugin-vite/src/plugins/dev_server.ts
@@ -15,6 +15,22 @@ export function devServer(): Plugin[] {
       configureServer(server) {
         const IGNORE_URLS = /^\/(@(vite|fs|id)|\.vite)\//;
 
+        server.httpServer?.on("deno:request", (ev) => {
+          const { headers } = ev.req;
+          if (
+            headers.get("upgrade") !== "websocket" ||
+            // Vite uses websockets with protocol set to vite-hmr, vite-ping
+            headers.get("sec-websocket-protocol")?.startsWith("vite-")
+          ) {
+            return;
+          }
+          // When pendingResponse is set, rest of the handlers won't fire for any request, so websocket requests
+          // are going straight to the fresh server.
+          ev.pendingResponse = server.ssrLoadModule("fresh:server_entry").then(
+            (mod) => mod.default.fetch(ev.req) as Promise<Response>,
+          );
+        });
+
         server.middlewares.use(async (nodeReq, nodeRes, next) => {
           const serverCfg = server.config.server;
 


packages/plugin-vite/src/plugins/dev_server.ts
```diff
@@ -15,6 +15,22 @@ export function devServer(): Plugin[] {
       configureServer(server) {
         const IGNORE_URLS = /^\/(@(vite|fs|id)|\.vite)\//;
 
+        server.httpServer?.on("deno:request", (ev) => {
+          const { headers } = ev.req;
+          if (
+            headers.get("upgrade") !== "websocket" ||
+            // Vite uses websockets with protocol set to vite-hmr, vite-ping
+            headers.get("sec-websocket-protocol")?.startsWith("vite-")
+          ) {
```
CertainLach, line 24: I wish every non-vite request was going through this route, as the current implementation looks too fragile

It is however only required and used for websocket for now.

DontMash:
@marvinhagemeister 
I'm really interested in this fix and would also like to help.
You mentioned (https://github.com/denoland/fresh/issues/3554#issuecomment-3423562068) that this change will likely not be merged since 

> it deviates from the web platform

What would be your approach if intercepting the request is not the correct way.

marvinhagemeister:
@DontMash Create a Fresh cli that calls vite as a middleware. That's the proper solution. This PR can't be merged because it depends on another PR being merged into Deno itself. The PR for Deno is very unlikely to be merged.

DontMash:
> Create a Fresh cli that calls vite as a middleware.
What do you mean by _cli_ and how to use _vite as a middleware_?

Should `App` provide a _websocket_ command to define a websocket route and use _vite as a middleware_?

As far as I know, the vite middleware can only be accessed when creating a server or configuring a plugin.
That is why I also assumed intercepting the request in the plugin is the best solution.

CertainLach:
> Should App provide a websocket command to define a websocket route and use vite as a middleware?

You don't need to do that with requests being Deno requests, which is default in Fresh

The problem is that vite starts a node.js http server instead, and you can't properly convert node requests back to deno requests, because node.js handles websockets differently

It is possible however to create a vite dev middleware for fresh (I.e `app.use(viteMiddleware(viteConfig))`), that performs compilation, and responds with built client files from vite's output MemFS (or however it is called there).

DontMash:
> You don't need to do that with requests being Deno requests, which is default in Fresh

I know, but you still need to handle the upgrade yourself. The question was also closer related to meaning of _cli_ in this context

> The problem is that vite starts a node.js http server instead, and you can't properly convert node requests back to deno requests, because node.js handles websockets differently

I also tried to convert Nodejs websocket requests to Deno myself.

> It is possible however to create a vite dev middleware for fresh (I.e app.use(viteMiddleware(viteConfig))), that performs compilation, and responds with built client files from vite's output MemFS (or however it is called there).

This would mean that just the dev_server would not be part of the plugin anymore, but rather be implemented as a middleware for fresh?

CertainLach:
> This would mean that just the dev_server would not be part of the plugin anymorer, but rather be implemented as a middleware for fresh?

Exactly. It is not quite a middleware, because of how fresh handles build cache updates, but I already did that for webpack and the resulting setup works very well for me, it should work similarly for vite.

image

bartlomieju:
Closing — this depends on denoland/deno#30746 which was never merged and is unlikely to be (deviates from the web platform, per maintainer feedback).

The proper fix is for `Deno.upgradeWebSocket()` to work with `node:http` upgrade events, tracked in denoland/deno#33220. In the meantime, #3737 provides a working workaround using a second `Deno.serve()` + TCP proxy.

PR #3452, opened 2025-09-16
chore: fix update message

### LINKED ISSUE #3451: Adjust code for "fresh update notification" to show correct update code
Currently when there is a fresh update and I'm running with an older version, I get this message: 

Image

The problem being that it shows the old update script, which will downgrade me to the latest 1.x version. 
It should detect the current "branch" and suggest the appropriate update script. 
I think this deno task is now standard in newer fresh projects: 
`"update": "deno run -A -r jsr:@fresh/update ."` so we could suggest that to the user :)

fry69:
fix #3451


marvinhagemeister (approved):
Thanks!

PR #3425, opened 2025-09-13
fix(init): correct temporary directory creation in tests

fry69:
The tests for the `init` package were failing with a `PermissionDenied` error in Jules container environment when creating temporary directories. This was because the tests were attempting to create the temporary directories in the root of the project, where permissions are often restricted.

This change corrects the issue by removing the local `withTmpDir` helper function in `packages/init/src/init_test.ts` and instead using the one from `../../fresh/src/test_utils.ts`. The shared helper function correctly creates temporary directories in the operating system's default temporary directory, which resolves the permission issue.

Sample error message from Jules before this fix:

```
 ERRORS 

init - new project => ./packages/init/src/init_test.ts:99:6
error: PermissionDenied: Permission denied (os error 13): tmpdir
  const dir = await Deno.makeTempDir(options);
                         ^
    at Object.makeTempDir (ext:deno_fs/30_fs.js:168:10)
    at withTmpDir (file:///app/packages/fresh/src/test_utils.ts:136:26)
    at withTmpDir (file:///app/packages/init/src/init_test.ts:41:10)
    at file:///app/packages/init/src/init_test.ts:100:27

init - create project dir => ./packages/init/src/init_test.ts:107:6
error: PermissionDenied: Permission denied (os error 13): tmpdir
  const dir = await Deno.makeTempDir(options);
              ^
    at async withTmpDir (file:///app/packages/fresh/src/test_utils.ts:136:15)
    at async file:///app/packages/init/src/init_test.ts:108:21

init - with tailwind => ./packages/init/src/init_test.ts:122:6
error: PermissionDenied: Permission denied (os error 13): tmpdir
  const dir = await Deno.makeTempDir(options);
                         ^
    at Object.makeTempDir (ext:deno_fs/30_fs.js:168:10)
    at withTmpDir (file:///app/packages/fresh/src/test_utils.ts:136:26)
    at withTmpDir (file:///app/packages/init/src/init_test.ts:41:10)
    at file:///app/packages/init/src/init_test.ts:123:27

init - with vscode => ./packages/init/src/init_test.ts:140:6
error: PermissionDenied: Permission denied (os error 13): tmpdir
  const dir = await Deno.makeTempDir(options);
              ^
    at async withTmpDir (file:///app/packages/fresh/src/test_utils.ts:136:15)
    at async file:///app/packages/init/src/init_test.ts:141:21

init - fmt, lint, and type check project => ./packages/init/src/init_test.ts:153:6
[...]

PR #3453, opened 2025-09-16
chore: minor housekeeping

fry69:
-  remove leftover `www/dev.ts` from Vite conversion
- `plugin-vite` tests use only local folders on windows (like #3425)

Reason: Leftover `tmp_vite*` folders are piling up in my filesystem inside the repository


marvinhagemeister (approved):
LGTM

marvinhagemeister:
CI says "no" though. Not sure why.

```sh
[
  'error: Failed loading https://registry.npmjs.org/fdir for package "fdir"',
  "",
  "Caused by:",
  '    npm package not found in cache: "fdir", --cached-only is specified.'
]
```

fry69:
This does not actually work as I thought it would (a recurring theme).

The local folders seem to be necessary also on Linux and on macOS. I am reverting now and looking for a solution to clean the mess up after failed test leave dangling `tmp_vite` folders around.

fry69:
I kind of gave up on this, as `deno.test.afterAll()` seems also not to be doing what I thought it would do. Unsure if that is a bug or intentional. See my attempt here -> https://github.com/denoland/fresh/compare/main...fry69:fresh:fry69/fix-vite-tmp-folders

My hypothesis is that `deno.test.afterAll()` does not seem to fire in parallel mode, probably because test fail during setup. But this is just a guess.

FWIW, I leave this PR open for comments.

bartlomieju:
Changes look good -- the `www/dev.ts` removal is clean and the temp dir fix is the right approach (matching #3425). But CI is red on macOS and Linux (4 of 6 jobs fail, only Windows passes). Might be worth checking if that's a pre-existing/flaky failure or related to the temp dir change.

bartlomieju:
Closing this one out — the temp dir approach didn't pan out (as noted in the comments), and CI has been red since. If the `www/dev.ts` removal is still relevant we can do that as a one-liner separately. Thanks for investigating the temp folder cleanup!

PR #3459, opened 2025-09-17
docs: fix markdown docs

fry69:
This fixes the markdown docs.

Notes:
- JSX inside `main.ts` will not work, `main.tsx` is not supported, so a real route is needed
- added an outside example tutorial (please comment if that is okay)
- added https://github.com/denoland/fresh/tree/main/www as an example


docs/latest/examples/markdown.md
```diff
@@ -11,10 +11,10 @@ browser.
 First, let's install the [`@deno/gfm`](https://jsr.io/@deno/gfm) package that
 can transform markdown to html.
 
-1. Run `deno install jsr:@deno/gfm`
-2. Create a markdown file like `content.md`:
+1. Run `deno install --allow-scripts jsr:@deno/gfm`
```
marvinhagemeister: Nit JSR packages don't have `postinstall` scripts like npm. The `--allow-scripts` flag does nothing here.

fry69: Yeah, but I try to align to docs to a freshly installed demo app with `@fresh/init`. When I run that command without `--allow-scripts`, I get warnings about scripts for Tailwind, at least. Also the Vite script seems to run for the first time when run with `--allow-scripts`, so I almost habitually include that everywhere now.

I'll remove that option.

bartlomieju:
A couple of things I noticed:

- **`route/markdown.tsx` should be `routes/markdown.tsx`** — Fresh uses the plural `routes/` directory. The `../utils.ts` import also depends on this being correct.
- **External tutorial link** — the link to `jackfiszr.eu` is a third-party blog that could go stale. Worth considering whether the team wants to maintain external links in the official docs, or if just pointing to the `www/` source code is sufficient.

bartlomieju (approved):

PR #3469, opened 2025-09-19
test: some refactoring (WIP)

fry69:
This PR aims to centralize test helper function and anything that requires cross package imports inside the Fresh monorepo under a new not exported `@fresh/internal` package.

See also the this reverse engineered [design document](https://github.com/fry69/fresh/blob/fry69/refactor-tests-5/design/current-test-suite.md) for a bit more background.

Please comment if there is interest, if this refactor is desired.


bartlomieju:
Thanks for putting this together and for the design doc! We appreciate the effort to clean up the test infrastructure.

After reviewing, we don't think the churn is justified here — the current relative imports for test utils work fine, and introducing a new `@fresh/internal` package adds maintenance overhead (workspace config, import map changes across 37 files) without a user-facing benefit. The cross-package import ergonomics issue is minor enough that we'd rather keep things simple.

Closing for now, but thanks again for the contribution!

PR #3471, opened 2025-09-20
docs: add deno.json update section to migration guide

### LINKED ISSUE #3470: Packages missing in migration guide
I followed the migration guide and found 2 things missing:
1. There is no mention of installing vite anywhere in the migration guide. I kept getting this error: 

> Task dev vite
vite: command not found

2. Same with `@fresh/plugin-vite`. I got this error:

> TypeError: Relative import path "@fresh/plugin-vite" not prefixed with / or ./ or ../ and not in import map from "file:///Users/yousefshanawany/Developer/budget-cast/vite.config.ts.timestamp-1758360557534-01a1997d4846e8.mjs"

Was super confused until I fixed it manually.

fry69:
fix #3470 

This adds a short blurb about `deno.json` focusing on imports.

Background: Without `npm:vite` and `jsr:@fresh/plugin-vite` installed it can lead to confusing errors like this one:
```shell
$ deno task dev
vite: command not found
```

Which might lead to users install Vite globally, which is not needed (and will not fix this problem).

### DIFF
--- a/docs/latest/examples/migration-guide.md
+++ b/docs/latest/examples/migration-guide.md
@@ -155,7 +155,7 @@ accordingly.
   }
 ```
 
-## Update `deno.json`
+## Update imports
 
 Add `jsr:@fresh/plugin-vite` and `npm:vite` to `imports` in your `deno.json`
 file, either manually or with this command:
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good idea!

PR #3477, opened 2025-09-22
docs: fix misleading main.tsx filenames

### LINKED ISSUE #3476: Docs: Is it main.ts or main.tsx
The very first page of the docs is a little unclear about whether the main file is `main.tsx` or `main.ts`.

It's a bit a of problem because as it stands, JSX can't be used within that particular file, as in the first example, but the configuration is not looking for main.tsx and fails with an error if you (naively) change the extension.

fry69:
fix #3476

This removes misleading `main.tsx` filenames from the documentation examples.

Also repetitive/slightly misleading `main.ts` filenames got removed unless they are complete copy/paste ready examples including proper `import` statements.

Usually one complete example with a filename on top of the doc section  should be enough. Otherwise ambiguity is intentional, as there are multiple places possible to add code snippets.


mabasic:
Can we get this merged, the documentation is terrible at best. So many conflicting stuff even when comparing with a new fresh project

fry69:
> Can we get this merged, the documentation is terrible at best. So many conflicting stuff even when comparing with a new fresh project

The way forward seems to be to enable `main.tsx` again, see here -> https://github.com/denoland/fresh/issues/3476#issuecomment-3317472444

This PR is just for illustration until that got fixed.

brettchalupa:
@mabasic can you please elaborate more on:

> the documentation is terrible at best.

What issues are you having with the docs that the community can contribute to make them better?

mabasic:
> @mabasic can you please elaborate more on:
> 
> > the documentation is terrible at best.
> 
> What issues are you having with the docs that the community can contribute to make them better?

I should have wrote it all down, but I was trying to get my app working asap.

These are just the things that I can remember:
- after using the migrate script, the app was in state that was very different than the new fresh app. define.page is in new fresh app, but migrate left it without that and a few other things.
- tailwind was broken and had to manually upgrade it and move stuff around to get it working
- there is now static folder and assets folder. No clue what goes where, but funny story. I had a font in assets/fonts. When I start the app using dev the app does not see the fonts, so I moved them to static and now it worked, but when I went to compile the app it complained that fonts were not found, so I copied them back to assets. But then when I start that app it complains that fonts are missing so I also copied them to static/assets/fonts and assets/assests/fonts and now it work even if I have the same font copied to 4 places just to satisfy the app
- as I understand, there are now two ways of doing fresh: file based and "route" based. In the docs there is omly shown 'route' based where you put everything in main.ts. But also in docs there is code that you need to put jsx inside main.ts file which won't work.
- on errors page it says that you can add .notFound method, but then again you cant return jsx or any component from that. I have tried multiple times multiple ways.
- in migration guide it says that you need to add onError catch all method and then handle 404 from there. It does not work at all. I cant even get a console log from there, it does not trigger. I am guessing that is because I am mixing file based and route based routes. Nowhere it is written that you cant mix and match
- layouts in docs arw written one way (route bases). It kind of works but there is some stuff that errors out. I used _layout in the end
- in docs about app wrapper. The usage is that you call it from main.ts, but it just does not work again. Even in new fresh app it is in _app.
- also a bug, when I change the code hot reload / fast refresh sometimes works, but sometimes it starts complaining about cors and I have to stop the server and start it again (dev task) to get it working. This happens very frequently.

In conclusion, it seems like the docs were written for "route" based routing, the new app is written in file based routinf, and you probably can't mix and match both.

I will publish my app soon open source so can link it here.

P.S. The docker image is a good atarting point but it does not work in a monorepo. Have been trying for two days to get it working because some compilerOptions are in root deno, and some in package deno, and if you copy just the paxkage dorectory in docker, you get a bunch of errors xD

fry69:
@mabasic Thank you for your detailed response.

Please copy your list of complaints and observations into a separate new issue. In this PR discussion they might get overlooked and forgotten.

bartlomieju (commented):
Looks good overall! One small issue:

In `concepts/islands.md`, the filename changed from `main.tsx` to `main.ts`, but the code block contains JSX (`<MyIsland />`):

```diff
-```tsx main.tsx
+```tsx main.ts
```

This should either drop the filename entirely (like the other partial examples) or keep it as `main.tsx`. A file with JSX would be `.tsx`.

bartlomieju (approved):

PR #3486, opened 2025-09-24
feat: add Deno runtime version check warnings

fry69:
This adds warnings if an outdated or canary Deno version gets used.

### DIFF
--- a/packages/fresh/src/app.ts
+++ b/packages/fresh/src/app.ts
@@ -29,6 +29,7 @@ import {
   newRouteCmd,
 } from "./commands.ts";
 import { MockBuildCache } from "./test_utils.ts";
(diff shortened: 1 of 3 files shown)


marvinhagemeister (commented):
Whilst this approach works from a technical perspective, I'm not sure of this is a good idea to do from a product perspective.

It means we'd do a request to an external resource whenever Fresh starts up. As a developer I certainly wouldn't want any app I write to make requests on it's own that I told it to do.

Deno itself has an update checker and will remind folks when there is a new version once per day. We shouldn't give the impression that Fresh only ever works with the very latest version of Deno either. It's unfortunately that we ran into the fs close bug due to changes in tailwindcss, but I'm not sure if we should guard against that on the Fresh side of things.

fry69:
ACK

PR #3491, opened 2025-09-25
fix: add codesign hook for esbuild on Darwin platforms

### LINKED ISSUE #3378: Some tests are failing when run locally
I tested the Fresh 2 release (commit `fb341e4ab72768bcbe1d7185a515eb704493adcf`, tag `2.0.0`) on three different machines and get some consistent failing error when running tests.

Setup across all machines:
- non-root user
- repository cloned with `git clone https://github.com/denoland/fresh.git`
- current stable Deno 2.4.5
- ran `deno install --allow-scripts`
- ran tests with `deno test -A`

Machines:
- Laptop macOS 15.6.1 (24G90) (Apple Silicon)
	- Result: `FAILED | 445 passed (14 steps) | 7 failed | 18 ignored (1m48s)`
- Local bare metal server Ubuntu-25.04 (x86_64)
	- Result: `FAILED | 404 passed (14 steps) | 8 failed | 16 ignored (1m32s)`
- Remote VM box Ubuntu-24.04 (x86_64)
	- Result: `FAILED | 405 passed (14 steps) | 7 failed | 16 ignored (2m11s)`

Results are in this gist: https://gist.github.com/fry69/13940c6790018201a4cd009825f9dae7
(slightly abbreviated for readability, can provide more details if needed)

TL;DR: a few consistent `Could not find server address` errors and the dreaded `Text file busy (os error 26)` on Ubuntu with Vite (see also https://github.com/denoland/deno/issues/30424)

fry69:
fix #3378 

Signing esbuild binaries seem to fix the problem with failing tests on my machine (M1 Air/macOS 15.7):

```shell
$ deno test -A 
[...]
ok | 476 passed (14 steps) | 0 failed | 17 ignored (2m26s)
```

```shell
$ deno test -A --parallel
[...]
ok | 476 passed (14 steps) | 0 failed | 17 ignored (1m42s)
```

Before:

```shell
$ deno test -A packages/plugin-vite/tests/dev_server_test.ts
[...]
FAILED | 21 passed | 7 failed | 1 ignored (11s)
```

Below is an analysis why this is happening.

Caveat: I am not sure if this fix is correct, or if this hiding symptoms for a deeper root cause, that this manual `codesign` step should not be necessary. But it works for me, now, and that makes me happy.

For the curious, here are my notes while debugging this problem with the help of GPT-5 Codex -> https://gist.github.com/fry69/dca16c1164defd93a62adecaa0a2b2dd

---

Investigated the mac-only failures and found that macOS was killing the Vite/esbuild helper when a second dev server spun up; the crash logs showed `SIGKILL (Code Signature Invalid)` for the esbuild binary. These tests keep one long-lived server running while launching others, so the unsigned/adhoc esbuild binary in the npm cache got revalidated and terminated only on this platform, which is why CI (Linux) kept passing.

## actions taken
- Added `ensureEsbuildSigned()` in test_utils.ts to locate `npm:esbuild/bin/esbuild`, run `codesign --force --sign - …` once per process on Darwin, and surface clear guidance if `codesign` isn’t available.
- Hooked that helper into `launchDevServer`, `spawnDevServer`, and `buildVite` so every Vite invocation now runs against a freshly re-signed binary.
- Kept the rest of the test utilities unchanged, cleaned up the temporary debug artifacts, and ran formatter.

## verification 
- `deno fmt packages/plugin-vite/tests/test_utils.ts`
- `deno test -A packages/plugin-vite/tests/dev_server_test.ts` (all 28 relevant cases now pass locally on macOS)

## notes & follow-ups
- The new logic depends on the macOS `codesign` tool (installed with the Xcode Command Line Tools). If it’s missing, developers will get a direct error telling them to install it or sign the binary manually.
- No CI impact is expected; the helper early-outs on non-Darwin systems.
- Root cause mapped: macOS code-signing enforcement → esbuild killed mid-launch → Vite servers never reported their address → tests crashed. The fix ensures the binary stays valid before any server starts.

### DIFF
--- a/packages/plugin-vite/tests/test_utils.ts
+++ b/packages/plugin-vite/tests/test_utils.ts
@@ -4,6 +4,58 @@ import { walk } from "@std/fs/walk";
 import { withTmpDir } from "../../fresh/src/test_utils.ts";
 import { withChildProcessServer } from "../../fresh/tests/test_utils.tsx";
 
+let ensureEsbuildSignedPromise: Promise<void> | null = null;
+
+async function ensureEsbuildSigned() {
+  if (Deno.build.os !== "darwin") {
+    return;
+  }
+
+  if (ensureEsbuildSignedPromise) {
+    return await ensureEsbuildSignedPromise;
+  }
+
+  ensureEsbuildSignedPromise = (async () => {
+    let esbuildPath: string;
+    try {
+      esbuildPath = path.fromFileUrl(
+        import.meta.resolve("npm:esbuild/bin/esbuild"),
+      );
+    } catch (err) {
+      throw new Error(
+        `Failed to resolve esbuild binary location: ${
+          err instanceof Error ? err.message : String(err)
+        }`,
+      );
+    }
+
+    try {
+      const command = new Deno.Command("codesign", {
+        args: ["--force", "--sign", "-", esbuildPath],
+        stdout: "null",
+        stderr: "piped",
+      });
+      const { code, stderr } = await command.output();
+      if (code !== 0) {
+        const message = new TextDecoder().decode(stderr).trim();
+        throw new Error(
+          `codesign exited with code ${code}${message ? ": " + message : ""}`,
+        );
+      }
+    } catch (err) {
+      ensureEsbuildSignedPromise = null;
+      if (err instanceof Deno.errors.NotFound) {
+        throw new Error(
+          "codesign tool was not found. Install Xcode command-line tools or codesign the esbuild binary manually.",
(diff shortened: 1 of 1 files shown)


fry69:
Fixed in #3493 so this is no longer needed, but I think it is a good idea to keep this pull request open as a draft for future reference, as the underlying problem may surface again.

Otherwise, feel free to close this PR at any time.

bartlomieju:
Gonna close this one now since it seems the problem hasn't resurfaced.

PR #3493, opened 2025-09-25
fix: set nodeModulesDir to manual

fry69:
Hopefully a less invasive fix for #3378


marvinhagemeister:
Upstream issue for the remaining type problem https://github.com/denoland/deno/issues/30850

marvinhagemeister (approved):
Thanks!

PR #3495, opened 2025-09-25
ci: fix publish workflow

fry69:
Looks like the `publish.yml` workflow also needs a `deno install` step, the current one fails with missing types, see here -> https://github.com/denoland/fresh/actions/runs/18017720086/job/51266850091#step:4:6

I have no clue why the `deploy.yml` workflow failed with `vite: command not found` (see here -> https://github.com/denoland/fresh/actions/runs/18017526863/job/51266206092?pr=3493#step:4:6).
That should work IMHO.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,5 +23,8 @@ jobs:
         with:
           cache: true
 
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Good catch!

PR #3496, opened 2025-09-26
Fix wrong variable name in migration guide

BrunoBernardino:
You have `export const handlers = ` when it needs to be `export const handler = `.

I'm unsure if there's any build process I need to run for this.


marvinhagemeister (approved):
Good catch, thanks!

PR #3497, opened 2025-09-26
test: make sure free ports get used

fry69:
This fixes issues I have with running tests on one of my servers, where port 8000 on `localhost` is occupied.

Note: The current `init` test will continue to fail, but that will hopefully get replaced by #3524 which does not have this problem.

Before:
```
[
  "error: Uncaught (in promise) AddrInUse: Address already in use (os error 98)",
  "    at listen (ext:deno_net/01_net.js:594:35)",
  "    at serveInner (ext:deno_http/00_serve.ts:675:16)",
  "    at Object.serve (ext:deno_http/00_serve.ts:591:10)",
  "    at ext:deno_http/00_serve.ts:819:10",
  "    at ext:runtime_main/js/99_main.js:708:13",
  "    at eventLoopTick (ext:core/01_core.js:179:7)"
]
[...]
Builder - write prod routePattern => ./packages/fresh/src/dev/builder_test.ts:284:6
Builder - file:// islands => ./packages/fresh/src/dev/builder_test.ts:319:6
Builder - serves static files in subdir in prod => ./packages/fresh/src/dev/builder_test.ts:474:6
init - can start built project => ./packages/init/src/init_test.ts:237:6

FAILED | 471 passed (14 steps) | 4 failed | 18 ignored
```

After:
```
ok | 475 passed (14 steps) | 0 failed | 18 ignored
```


packages/init/src/init_test.ts
```diff
@@ -57,6 +57,11 @@ async function patchProject(dir: string): Promise<void> {
   // assert with this stricter rule, before adding it to initialized projects
   json.lint.rules.include = ["verbatim-module-syntax"];
 
+  if (json.tasks && typeof json.tasks.start === "string") {
+    json.tasks.start =
+      "deno serve -A --host=127.0.0.1 --port=0 _fresh/server.js";
+  }
```
marvinhagemeister: We shouldn't do this here as some folks use the `start` script during production inside docker containers. When Fresh spawns on a random port the container port forwarding will not work as expected as ports need to be explicitly be forwarded when booting a docker container.

fry69: I'll remove this from this PR, since this old `init` hopefully gets replaced by #3524

marvinhagemeister (approved):
LGTM

PR #3498, opened 2025-09-26
static files dir is now configurable.

theTechGoose:
I wanted to be able to configure the static directory for the static files middleware. i'm just replacing /static path in the file with a parameter in the outer function of the closure.

### DIFF
--- a/packages/fresh/src/middlewares/static_files.ts
+++ b/packages/fresh/src/middlewares/static_files.ts
@@ -11,7 +11,7 @@ import { getBuildCache } from "../context.ts";
  * app.use(staticFiles());
  * ```
  */
-export function staticFiles<T>(): Middleware<T> {
+export function staticFiles<T>(serveDir = "/static"): Middleware<T> {
   return async function freshServeStaticFiles(ctx) {
     const { req, url, config } = ctx;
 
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
The static file directory can already be configured with vite's [`publicDir`](https://vite.dev/config/shared-options.html#publicdir) config. Did you ran into issues with that?

I'm asking because I don't think the proposed changes here will do anything since the build cache is not aware of the new path. So the changes will likely lead to broken apps if the newly added option is used.

theTechGoose:
Mmmm yeah that's a way better solution! I didn't have issues with it I just didn't see it as an option. I'm having a heck of a time trying to integrate this with my Deno mono-repo. Are there some docs anywhere about how to make this work across workspace member boundaries?

--
Raphael Castro
TED Team
Monster Reservations Group
C: 843.855.7133
www.monsterrg.com ( http://www.monsterrg.com/ )

marvinhagemeister:
It should work out of the box. It sounds like that's not working for you though and you're running into issues. Would you like to share them with us? Any details you found out help.

bartlomieju:
Closing this PR — the approach of doing `pathname.replace("/static", serveDir)` at serve time doesn't work correctly because:

1. `buildCache.readFile` looks up files by their URL pathname in the pre-built cache, not by filesystem path. The static directory mapping happens at build time, not in the serving middleware.
2. The string replace would also incorrectly match URL paths like `/static-page`.

If you need to serve static files from a different directory (e.g., `public/` instead of `static/`), that should be configured at the build/plugin level — for example via Vite's `publicDir` option.

PR #3503, opened 2025-09-27
docs: update readme for fresh 2

treejadey:
The current version of the README is misleading, because:

1. The build instructions result in a Fresh v1.0 project.
2. Since Fresh v2.0, Fresh no longer does JIT rendering on the edge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,6 @@ simplicity.
 
 Some stand-out features:
 
-- Just-in-time rendering on the edge.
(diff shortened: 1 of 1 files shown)


treejadey:
Please note that I think we should also add some additional features to the list, and might potentially want to redo the instructions for deployment with Deno Deploy

marvinhagemeister (approved):
Thanks!

PR #3510, opened 2025-10-01
docs: troubleshoot VS Code extension

fry69:
This adds a short blurb how to install the VS Code extension to the troubleshooting guide.

Possible concerns:
- maybe redundant as there are already official VS Code Deno docs
- could be shortened to just refer to the main docs or completely removed


marvinhagemeister (approved):
Good idea, thanks!

PR #3522, opened 2025-10-05
fix: prevent island names from shadowing JavaScript globals

### LINKED ISSUE #3521: Island files named after global objects cause runtime failures
Found in the Discord:

> Hi, why when i call an island Map.tsx everything breaks

After investigating this a bit with VS Code agent + Sonnet 4.5, it came up with this analysis:

---

## Problem

When a Fresh island file is named after a JavaScript global object (e.g., `Map.tsx`, `Set.tsx`, `Array.tsx`, `Promise.tsx`), the application fails at runtime due to variable shadowing in the generated client code.

## Root Cause

The `UniqueNamer` class (`/packages/fresh/src/utils.ts:53-75`) checks island names against `JS_RESERVED` keywords (e.g., `class`, `function`, `if`) but does not include JavaScript global objects (e.g., `Map`, `Set`, `Array`, `Promise`, `Date`, `RegExp`, `Error`, `Symbol`).

When generating client-side boot code (`/packages/fresh/src/runtime/server/preact_hooks.ts:620-622`), Fresh creates:

```javascript
import Map from "/_fresh/js/[build-id]/Map.js";
boot({Map}, ...);
```

In the object shorthand `{Map}`, the value `Map` resolves to the **global `Map` constructor** instead of the imported island component. The client-side hydration then receives the native constructor instead of the React/Preact component, causing failure.

## Affected Names

Any island filename matching JavaScript global objects:
- `Map`, `Set`, `WeakMap`, `WeakSet`
- `Array`, `Object`, `String`, `Number`, `Boolean`, `Symbol`, `BigInt`
- `Promise`, `Date`, `RegExp`, `Error`, `Function`
- `JSON`, `Math`, `Reflect`, `Proxy`
- `Intl`, `Atomics`, `DataView`, `ArrayBuffer`
- Plus all typed array constructors (`Uint8Array`, `Int32Array`, etc.)

## Fix Required

Add JavaScript global objects to the `JS_RESERVED` set in utils.ts, or modify the code generation in preact_hooks.ts to avoid variable name collisions (e.g., use explicit object properties or suffix imported names).

fry69:
fix #3521 

Generated PR message below

---

## Problem

Island files named after JavaScript global objects (e.g., Map.tsx, `Set.tsx`, `Promise.tsx`) cause runtime errors. When `UniqueNamer` generates identifiers from these filenames, the resulting variable names shadow global constructors, breaking instantiation:

```javascript
import Map from "/islands/Map.tsx";
boot({Map}, props);  // Map resolves to global constructor, not the island
```

This results in `TypeError: Map is not a constructor` when the framework attempts to use `new Map()` internally.

## Solution

Added JavaScript global objects to the `JS_RESERVED` set in `UniqueNamer`. Names matching reserved keywords or globals are now automatically prefixed with underscore:

```javascript
import _Map_1 from "/islands/Map.tsx";
boot({_Map_1}, props);  // No shadowing
```

## Changes

- **utils.ts**: Extended `JS_RESERVED` with:
  - JavaScript built-ins (Map, Set, Array, Promise, Object, etc.)
  - Web APIs (fetch, console, URL, WebSocket, etc.)
  - Deno-specific globals (Deno namespace)
  - Node.js compatibility globals (process, Buffer, global)

- **Integration tests**: Added test cases for both Builder and Vite build systems verifying islands named after globals work correctly

## Impact

Single-location fix in `UniqueNamer` applies to all code generation (islands, routes, etc.). Island names are prefixed when needed, preserving debuggability while preventing shadowing bugs.


packages/fresh/src/utils.ts
```diff
@@ -76,6 +76,9 @@ export class UniqueNamer {
 }
 
 // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words
+// Note: JavaScript global objects (Map, Set, Array, Promise, etc.) are NOT included here
+// because they are valid island names. Code generation uses safe prefixed variable names
+// (__FRSH_ISLAND_N) to avoid shadowing globals.
```
marvinhagemeister: Island names are only an internal thing. It's not something that is exposed to users, so it would be totally fine to include those globals here. Would likely make the changes a lot smaller too.

fry69: The approach in the fix has the upside that the filenames get preserved as is. They do show up in some user facing places. Here is what the agent says:

---

User-visible contexts where `island.name` appears:**

1. **HTML comment markers**: `<!--frsh:island:Map:0-->` (visible in View Source and DOM inspection)
2. **ISLAND_REGISTRY keys**: `ISLAND_REGISTRY.get("Map")` (used at runtime for island lookup)
3. **Error messages**: `"Could not find chunk for island Map"` (builder.ts:338)
4. **Partial state JSON**: `{"islands": [{"name": "Map", ...}]}` (preact_hooks.ts:587)

The island name comes from `UniqueNamer.getUniqueName()` which checks `JS_RESERVED`. If we add "Map" to `JS_RESERVED`, it becomes `"_Map"` in all these contexts.

**Comparison:**

| Approach | Island name | Import variable | User sees |
|----------|-------------|-----------------|-----------|
| Add to JS_RESERVED | `"_Map"` | `_Map` | `_Map` everywhere |
| Safe variables (current PR) | `"Map"` | `__FRSH_ISLAND_0` | `Map` in all user-facing contexts |

fry69: But I can see that this might balloon to a bigger issue when this internal names approach get picked up and generalized, instead of using it just to fix this issue.

Viewing it from that perspective this fix has a bit code smell, I agree 100% on that point. I'll look into this.

marvinhagemeister: 1. HTML comment markers are internal. They're are generated by Fresh, not by users
2. Island registry keys are internal too
3. Error messages is a bit of an exception, but I feel like the generated names are good enough for users to deduce what the actual island is. If we would store the original name we would unnecessarily bloat up bundle size by shipping strings (strings are always the worst offenders for that).
4. Partial state json is internal only too.

fry69: Thanks for the clarification, I let the agent rewrite this fix to use `JS_RESERVED` instead.

marvinhagemeister (approved):
LGTM, thanks!

PR #3523, opened 2025-10-05
fix: add lint ignore comments for console

fry69:
This adds:
- `// deno-lint-ignore-file no-console` to two generated files so lint does not complain
- turns a special character in a comment to a more commonly used one (LLMs really hate these)


fry69:
On second thought and after a quick discussion this is not needed.

PR #3517, opened 2025-10-03
feat: template based init script

fry69:
# Refactor: File-based template system for init

Replaces inline string templates with file-based templates in `packages/init-templates/`.

## Changes

- **Template storage**: Moved from inline strings to `assets/template/` directory
  - 4 complete templates: vite, vite-tailwind, builder, builder-tailwind
  - 3 additive variants: docker, vscode, vscode-tailwind
- **Base templates**: Common files in `assets/base/{vite,builder}`, synced via `sync_templates.ts`
- **Architecture**: Separated CLI (prompts, validation) from template engine (file operations)
- **Variable substitution**: Text based e.g. `__FRESH_VERSION__` -> Latest Fresh version
- **Tests**: 45 tests (unit, integration, CLI, sync), using `@std/expect`

See [README.md](https://github.com/fry69/fresh/blob/fry69/refactor-init-template/packages/init-templates/README.md) and [DESIGN.md](https://github.com/fry69/fresh/blob/fry69/refactor-init-template/packages/init-templates/DESIGN.md) for more details.

Since this is a rather large PR due to all the templates involved, I recommend take a look via the file browser branch view here -> https://github.com/fry69/fresh/tree/fry69/refactor-init-template/packages/init-templates

PR #3524, opened 2025-10-05
feat: template based init script

fry69:
Supersedes earlier attempt -> https://github.com/denoland/fresh/pull/3517

This is now ready for review.

The major upside and motivation for this template based approach is that it is now possible to view what gets generated in the repository as actual valid files including syntax highlighting and formatting.

Previously those files where just strings inside the old `init` script, which is IMHO a nightmare for maintainability.

The approach for this template system is loosely based on Vite's `create-vite`, which also includes (redundant) templates inside the repository, see here -> https://github.com/vitejs/vite/tree/main/packages/create-vite

Included in the tests are compatibility checks against the current `@fresh/init@2.0.9`, those can get removed at a later point when the templates start to diverge. They are only included for demonstration purposes, that this new implementation generates 100% identical output.

This means that the first PR after this got included, which modifies the templates so the output differs from the old `init` script, also needs to include:
```shell
rm packages/init/tests/compatibility_test.ts
```

Since this is a rather large PR, it may be more approachable to view the branch directly -> https://github.com/fry69/fresh/tree/fry69/refactor-init-template-rename/packages/init

Generated PR message below:

---

Replaces inline string-based project generation with file-based templates for improved maintainability and extensibility.

### Architecture

**Template structure:**
- `assets/template/` - Complete project templates (vite, vite-tailwind, builder, builder-tailwind)
- `assets/variants/` - Additive overlays (docker, vscode, vscode-tailwind)
- `assets/base/` - Common files synced across template variants

Templates use `__VARIABLE__` syntax for substitution (project name, dependency versions). Files with `__` prefix are renamed to `.` prefix during generation (e.g., `__gitignore` → .gitignore).

**Template selection logic:**
```
Base: vite|vite-tailwind|builder|builder-tailwind
Variants: +docker, +vscode, +vscode-tailwind
```

### Maintenance workflow

Edit shared files in `assets/base/vite/` or `assets/base/builder/`, then run `deno task sync` to propagate changes to template variants. Template-specific files (deno.json, vite.config.ts, styles.css) are never overwritten.

### Version resolution

Matches existing behavior: only Fresh core version is fetched from JSR, all other dependencies use fixed versions updated by release scripts.

### Testing

55 tests covering:
- Unit tests for core functionality
- Integration tests (project generation, builds, dev servers)
- CLI argument parsing
- Compatibility tests verifying output matches v2.0.9 (pre-refactor baseline)
- Template sync workflow

### Breaking changes

None. CLI interface and programmatic API remain identical. Package publishes to same `@fresh/init` scope at JSR.

### Documentation

- README.md - User-facing usage documentation
- DESIGN.md - Maintainer documentation (architecture, template maintenance, development setup)


marvinhagemeister (approved):
LGTM

PR #3529, opened 2025-10-06
fix: use free port in builder globals test

fry69:
Fix "port in use" problem re-introduced in builder globals test.


marvinhagemeister (approved):
LGTM

PR #3532, opened 2025-10-06
test(init): fix address in use

fry69:
Reverting back to the old init script also re-introduced the "address in use" problem.

This does not have any effect: `env: { PORT: "0" }`

Adding arguments to `deno task start` also have no effect.

The only solution I can come up with is to ignore the task and call `deno serve` with proper arguments directly for the build tests.


marvinhagemeister (approved):
LGTM

PR #3556, opened 2025-10-21
fix(vite): ignore tmp, timestamp, test, and dotfiles

### LINKED ISSUE #3585: Dev server crashes due to tmp and timestamp files
When initializing a new Fresh 2 project with `deno run -Ar jsr:@fresh/init`,
 the dev server frequently crashes due to various files created by the dev
process. This updates the vite plugin to ignore these files so that the dev
server doesn't crash.

### Repro Steps

1. Start a new Fresh 2 project: `deno run -Ar jsr:@fresh/init`
2. Boot up the dev server: `deno task dev`
3. Make changes

Dev server crashes with this type of error:

```
8:41:17 PM [vite] (client) hmr update /assets/styles.css (x2)
GET http://localhost:5173/
Shared value hello
GET http://localhost:5173/
Shared value hello
error: Uncaught NotFound: No such file or directory (os error 2) about ["path/to/project/db/schema.ts.tmp.12196.1761007289175"]
    at new FsWatcher (ext:runtime/40_fs_events.js:23:17)
    at Object.watchFs (ext:runtime/40_fs_events.js:73:10)
    at ext:deno_node/_fs/_fs_watch.ts:57:21
    at callback (ext:deno_web/02_timers.js:58:7)
    at eventLoopTick (ext:core/01_core.js:214:13)
```

## System

- Operating System: Fedora 42
- Deno 2.5.4

brettchalupa:
Fixes #3585 by ignoring files that were causing Vite to crash, as well as test files, which Vite doesn't need to watch

### DIFF
--- a/packages/plugin-vite/src/mod.ts
+++ b/packages/plugin-vite/src/mod.ts
@@ -62,6 +62,19 @@ export function fresh(config?: FreshViteConfig): Plugin[] {
         isDev = env.command === "serve";
 
         return {
+          server: {
+            watch: {
+              // Don't watch test files, temp files, or hidden files during development
+              // to prevent dev server crashes
+              ignored: [
+                "**/*_test.ts*",
+                "**/*.test.ts*",
+                "**/*.tmp.*",
+                "**/.timestamp-*",
+                "**/.*",
+              ],
+            },
+          },
           esbuild: {
             jsx: "automatic",
             jsxImportSource: "preact",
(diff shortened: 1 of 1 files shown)


brettchalupa:
@marvinhagemeister the issue this PR fixes is quite important, IMO. Every Fresh project's dev server I create crashes without this change. Would you be willing to review it when you're able?

marvinhagemeister:
It shouldn't crash. It kinda points at an issue in Deno. I want to get to the bottom of that. In the meantime you can always extend the vite config directly with the changes from this PR.

brettchalupa:
@marvinhagemeister yup! That's where I got the diff for this PR from. In every Fresh 2 project I've had to set this. What's interesting/strange is that I can't get it to happen on macOS. But it happens reliably for at least 6 Fresh 2 projects on Fedora Linux. I'll do some more debugging/investigating to try to pinpoint it some more.

brettchalupa:
Let me close this then for now and we can keep contained to discussion in #3585. I'll also post my diff there so if others run into the problem, they can have a quick solution.

PR #3560, opened 2025-10-21
chore: update README Deno Deploy docs for EA

### LINKED ISSUE #3537: Outdated Deno Deploy instructions
The instructions for Deno Deploy in the `README.md` file still seems to assume Fresh 1, they point to `main.ts`.

In general the instructions differ with https://fresh.deno.dev/docs/deployment/deno-deploy and both lack clarification between the new project wizard (which will create a new repository, it seems) and linking an existing GitHub project (which has a different workflow).

This can be confusing for newcomers. Even more confusing when the "New Deploy" changes everything again (slightly), which is even harder to incorporate in current documentation.

Which leads to these questions:
- Should both Deploys get mentioned?
- Or only classic (as it is now)?
- Or should the New Deploy prominently get pushed?

brettchalupa:
Fixes #3537

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,11 +52,10 @@ To deploy the project to the live internet, you can use
 [Deno Deploy](https://deno.com/deploy):
 
 1. Push your project to GitHub.
-2. [Create a Deno Deploy project](https://dash.deno.com/new).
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Thanks!

PR #3561, opened 2025-10-21
fix: disable static file caching in development

brettchalupa:
When running Fresh 2 in development, the ETag caching was leading to
incorrect assets being served. I'd often get stale assets from days ago
or separate apps. This problem was especially bad with multiple apps
that run on the same port. For example, I may work on App A on the
default port 5173. Then boot up App B. I'd get assets from App A when I
wouldn't want them, leading to a frustrating dev experience.


marvinhagemeister (approved):
Good catch, thanks!

PR #3562, opened 2025-10-21
feat: add `@/` path alias to generated projects

### LINKED ISSUE #3485: Set up `@/` alias by default
Now that we recommend it in our Deno docs we can officially follow that guideline in Fresh

https://github.com/denoland/docs/pull/2584

brettchalupa:
Sets up `@/` as the default path alias in newly generated Fresh
projects, following the official Deno documentation guidelines. This
allows developers to use `@/` for absolute imports from the project root
instead of relative paths. This is especially convenient to remove
import that require deeply traversing up the directory tree.

The `@/` alias is now included in the imports section of deno.json for
all new projects, enabling cleaner import statements like:

```
import { Button } from "@/components/Button.tsx";
```

Fixes #3485

### DIFF
--- a/docs/latest/getting-started/index.md
+++ b/docs/latest/getting-started/index.md
@@ -36,6 +36,36 @@ project folder should look like this:
 └── vite.config.ts  # Vite configuration file
 ```
 
+## Path aliases
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Sweet, thanks!

PR #3578, opened 2025-10-25
docs: add a notice about main.tsx, how to switch from main.ts

revelt:
Screenshot 2025-10-25 at 22 17 27

The code example's title instructs to use `main.tsx` but the project starter uses `main.ts`. Let's add a warning to tell users to configure vite after renaming this entry file.


fry69:
See also #3476 #3477

revelt:
I see, ok, let's close. Thank you

PR #3592, opened 2025-10-31
Update daisyUI installation instructions and file path

falvarador:
The directory where the stylesheet is located is updated.


marvinhagemeister (approved):
Thanks!

PR #3594, opened 2025-11-04
docs: add example for session management

### LINKED ISSUE #1216: proposal(docs): add an example of how to manage session cookies

DontMash:
Resolves: #1216


docs/latest/examples/session-management.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Solution
+
+We can use a middleware to get the `Request` from the `Context` and create a
+`Response` for it. Using the previously mentioned methods we retrieve the
+cookies from the `Request`. If there is no _session_ cookie already available,
+we will create a _session_ cookie and add it to the `Response`.
+
+```ts main.ts
+import { getCookie, setCookie } from "@std/http";
```
csvn: Nit: Spelling error


```suggestion
import { getCookies, setCookie } from "@std/http";
```

docs/latest/examples/session-management.md
```diff
@@ -0,0 +1,94 @@
...
+  if (!session) {
+    const uuid = crypto.randomUUID();
+    setCookie(response.headers, {
+      name: "session",
+      value: uuid,
+    });
+  }
+  return response;
+});
+```
```
csvn, line 56: I wonder if this example may be _too_ basic, as I see a couple of drawbacks:

- If no cookie exists currently, there's no way retrieve the `session` (UUID) as it's just returned in the header after returning from route/endpoint/middleware
- The `session` cookie must be parsed by potentially many other routes/middleware; putting it in `ctx.state.session` means it can be retrieved once and then re-used

My suggestion would be to do the following:

- Add a `State` interface with a `session` property
- Update the `state` with current `session` ID before running other routes/middleware
- Show how another route/middleware can use `ctx.state.session`

```ts main.ts
import { getCookie, setCookie } from "@std/http";

export interface State {
  session: string;
}

/** session middleware */
app.use(async (ctx: Context<State>) => {
  // Retrieve current session or generate a new one
  const session = getCookies(ctx.req.headers)["session"];
  ctx.state.session = session ?? crypto.randomUUID();

  // Run routes/middleware
  const response = await ctx.next();

  // Update session cookie if necessary
  if (!session) {
    setCookie(response.headers, {
      name: "session",
      value: ctx.state.session,
    });
  }

  return response;
});

// Use `ctx.state.session` in routes/middleware
app.get("/", (ctx: Context<State>) => new Response(
  `Your session ID is: ${ctx.state.session}`
));
```

It's a little more verbose, but to me it's a lot more realistic and usable. Now the `session` cookie actually becomes usable server-side for new sessions, and easier to access for existing sessions.

docs/latest/examples/session-management.md
DontMash: Fixed

docs/latest/examples/session-management.md
DontMash: I agree; Thank you for your addition!

My idea was to showcase the **most basic** implementation, since it would also be possible to persist the session in many different ways aside of the `State`.

DontMash: I added your suggestion :+1:

bartlomieju (approved):

PR #3595, opened 2025-11-04
fix(vite): use manifest entry filename for server entry import

cuipeiyu:
## Summary
When `rollupOptions.output.entryFileNames` is customized in vite config (e.g. `[hash].mjs`), the generated `server.js` was hardcoding `server-entry.mjs` as the import path, causing the server to fail to start.

This fix reads the actual entry filename from the Vite manifest and uses it in the generated `server.js` import, falling back to `server-entry.mjs` when no custom naming is configured.

## Changes
- Read `isEntry` file path from the Vite manifest in `writeBundle`
- Use the manifest filename (with fallback) when generating `server.js`
- Add test verifying custom `entryFileNames` produces correct `server.js` import

## Test plan
- [x] Existing build tests pass
- [x] New test: `vite build - custom rollup entryFileNames in server.js` verifies the generated `server.js` uses the hashed filename instead of hardcoded `server-entry.mjs`


marvinhagemeister (commented):
Nice, let's add a test for that

marvinhagemeister (commented):
Nice catch! Let's add a test case for this

bartlomieju (approved):

Issue #3597, opened 2025-11-04
JSR: Improve score with improved documentation, examples and package settings


csvn:
Given that Fresh is used by Deno and a very well used and visible package, it would be great if all packages under the `@fresh` scope on JSR had high scores.

<details>

<summary><strong>Show current scores 👇 </strong></summary>

<img width="1578" height="1251" alt="Image" src="https://github.com/user-attachments/assets/a5bdaead-d830-44b8-9d97-bc9609d2aca5" />

<img width="1555" height="935" alt="Image" src="https://github.com/user-attachments/assets/9822f097-2837-444d-98e2-f3aa0c91d504" />

</details>

I've created this issue as an umbrella issue to track various adjustments that should be quick to do to help improve scoring for all packages.

## JSR Maintainer Actions

Requires admin/member of `@fresh` to update.

### Mark `deno` as compatible runtime

- [ ] [`@fresh/build-id`](https://jsr.io/@fresh/build-id)
- [ ] [`@fresh/plugin-tailwind-v3`](https://jsr.io/@fresh/plugin-tailwind-v3)

### Add description to the package

- [ ] [`@fresh/build-id`](https://jsr.io/@fresh/build-id)
  - Suggestion: Internal Fresh dependency, do not use directly
- [ ] [`@fresh/core`](https://jsr.io/@fresh/core)
  - Suggestion: Fresh is a small, fast and extensible full stack web framework built on Web Standards. It’s designed for building high-quality, performant, and personalized web applications
  - (Text taken from from README.md)
- [ ] [`@fresh/init`](https://jsr.io/@fresh/init)
  - Suggestion: A CLI tool to bootstrap a new Fresh project
  - (Text taken from from README.md)
- [ ] [`@fresh/plugin-tailwind`](https://jsr.io/@fresh/plugin-tailwind)
  - Suggestion: A plugin for Fresh to use Tailwind CSS
- [ ] [`@fresh/plugin-tailwind-v3`](https://jsr.io/@fresh/plugin-tailwind-v3)
  - Suggestion: A plugin for Fresh to use Tailwind CSS v3
- [ ] [`@fresh/update`](https://jsr.io/@fresh/update)
  - Suggestion: A CLI tool to help upgrade an existing Fresh project
  - (Text taken from from README.md)

### Examples in `README.md` or module doc

- [x] [`@fresh/build-id`](https://jsr.io/@fresh/build-id): #3600
[...]

PR #3598, opened 2025-11-04
docs(core): add small usage & getting started example

csvn:
This PR just fills out the `README.md` with minimal usage example, such that they can get a small taste of how the framework can be used. This also helps improve the score in JSR.

Part of #3597

### DIFF
--- a/packages/fresh/README.md
+++ b/packages/fresh/README.md
@@ -5,3 +5,27 @@ Standards. It’s designed for building high-quality, performant, and personaliz
 web applications.
 
 [Learn more about Fresh](https://fresh.deno.dev/)
+
+## Usage
+
+Generate a new Fresh project with `@fresh/init`:
+
+```sh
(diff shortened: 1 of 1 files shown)


packages/fresh/README.md
```diff
@@ -5,3 +5,27 @@ Standards. It’s designed for building high-quality, performant, and personaliz
...
+
+```tsx
+import { App } from "fresh";
+
+const app = new App()
+  .get("/", () => new Response("hello world"))
+  .get("/jsx", (ctx) => ctx.render(<h1>render JSX!</h1>));
+
+app.listen();
+```
```
csvn, line 28: This example was just lifted from https://fresh.deno.dev/docs/introduction, to show a very small example for users.

Though I think the init command above satisfies the "show example code in README.md", so this section could be removed to avoid duplication.

marvinhagemeister (approved):
Thanks!

PR #3600, opened 2025-11-04
docs(build-id): add code block example

csvn:
Adds a module doc and code block to help satisfy JSR's scoring. Suggestions for alternatives or improvements very welcome.

Part of #3597

### DIFF
--- a/packages/build-id/mod.ts
+++ b/packages/build-id/mod.ts
@@ -1,3 +1,17 @@
+/**
+ * Don't use this package directly. It's considered internal for Fresh.
+ *
+ * @example
+ * ```ts
+ * // `BUILD_ID` is based on the `DENO_DEPLOYMENT_ID` environment variable.
+ * ```
+ *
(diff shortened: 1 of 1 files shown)


packages/build-id/mod.ts
```diff
@@ -1,3 +1,17 @@
...
+ *
+ * @example
+ * ```ts
+ * // `BUILD_ID` is based on the `DENO_DEPLOYMENT_ID` environment variable.
+ * ```
+ *
+ * @see https://fresh.deno.dev/docs/deployment/docker
+ *
+ * @module
+ * @private
```
csvn, line 12: I don't know if JSR respects `@private`, or if it even makes sense to make a "public" API marked as `@private` because consumers of Fresh should not use it 🤔

marvinhagemeister: I don't know either. Let's see what happens!

marvinhagemeister (approved):
LGTM, thanks!

PR #3601, opened 2025-11-04
docs(core): correct example code block formatting

csvn:
The example for the CSP middleware looks a bit broken currently on JSR, due to the `@example` markdown is not wrapped in a code block. 

image

## Changes

- Improves formatting a bit, to be less compact
- Wraps the example in a code block ` ```ts `
- Adds a title to the `@example` - "Basic usage" - To avoid the example being labeled as "Example1" on JSR

### DIFF
--- a/packages/fresh/src/middlewares/csp.ts
+++ b/packages/fresh/src/middlewares/csp.ts
@@ -12,16 +12,21 @@ export interface CSPOptions {
   csp?: string[];
 }
 
-/** Middleware to set Content-Security-Policy headers
+/**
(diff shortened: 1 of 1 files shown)


csvn:
Hmm... giving titles to examples seems to break indentation of examples when using VSCode, which is kinda bad... https://github.com/microsoft/TypeScript/issues/43955.

Which is more important, nicer documentation rendering on JSR, or correct rendering of tooltip in VSCode?

marvinhagemeister (approved):
LGTM, we should make our LSP better to handle example code blocks.

PR #3603, opened 2025-11-04
docs(examples): add module and public exports JSDoc for all entrypoints

csvn:
Addresses scoring of `@fresh/examples` of JSR for several metrics:

- ✅ Adds module doc for all entrypoints
- ✅ Adds descriptions to all public exports for `@fresh/examples` (0% -> 100%)

I could not find if `@fresh/examples` are part of any E2E tests to test using JSR package as mounted app or Island, or if this is just a small demo for users. So I just added small generic drescriptions.

Part of #3597

### DIFF
--- a/packages/examples/src/app1.tsx
+++ b/packages/examples/src/app1.tsx
@@ -1,6 +1,13 @@
+/**
+ * Module containing a simple example Fresh App
+ *
+ * @module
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Thanks!

PR #3613, opened 2025-11-11
feat: Add `.json()/.text()/.html()/.stream()` context helpers

marvinhagemeister:
Played around with Hono in the past days and the `.text()/.json()/.html()` helpers are kinda neat, even if you can always construct that yourself manually. Also added a `.stream()` helper to create streams without having to deal with `ReadableStream` and encodings yourself.

### DIFF
--- a/packages/fresh/src/context.ts
+++ b/packages/fresh/src/context.ts
@@ -27,7 +27,8 @@ import {
   renderRouteComponent,
 } from "./render.ts";
 import { renderToString } from "preact-render-to-string";
-import { isAsyncIterable, isIterable, isThenable } from "./utils.ts";
+
+const ENCODER = new TextEncoder();
 
 export interface Island {
   file: string;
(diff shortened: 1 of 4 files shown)


lucacasonato (commented):
This all looks good except the stream one, because it doesn't propagate cancellation into the iterator. Use ReadableStream.from instead?

Also I think the usage examples should be "in context" showing that you need to return the return value.

packages/fresh/src/context.ts
```diff
@@ -376,4 +373,165 @@ export class Context<State> {
...
+   * ```tsx
+   * ctx.text("Hello World!");
+   * ```
+   */
+  text(content: string, init?: ResponseInit): Response {
+    const headers = getHeadersFromInit(init);
+    headers.set("Content-Type", "text/plain; charset=utf-8");
+
+    return new Response(content, { ...init, headers });
+  }
```
lucacasonato, line 386: new Response already sets this header if I recall correctly.

marvinhagemeister: Yup, can confirm just looked at our sources.

PR #3620, opened 2025-11-14
feat(docs): improve copy button appearance on light mode

KiritaniAyaka:
| Before| After |
| :---: | :---: |
| before | after |

The appearance on dark mode doesn't change.


marvinhagemeister (approved):
Thanks!

PR #3621, opened 2025-11-15
feat(init): use recommended vscode settings for tailwind

KiritaniAyaka:
The recommended way to configure custom directives from Tailwind is to use `files.associations`.

Check recommended settings from Tailwind CSS IntelliSense Extension: https://github.com/tailwindlabs/tailwindcss-intellisense#recommended-vs-code-settings

This allows us to stop maintaining future changes to custom directives from TailwindCSS, instead letting the TailwindCSS extension decide how to handle them.


marvinhagemeister (approved):
Thanks!

PR #3625, opened 2025-11-18
fix: lockfile update

marvinhagemeister:
Looks like esm.sh changed something again


marvinhagemeister:
hmm seems like it has resolved itself.

PR #3630, opened 2025-11-18
perf: make updater faster by streaming instead of reading into memory

### LINKED ISSUE #3483: Fresh updater excessive memory use
Running the `deno run -Ar jsr:@fresh/update` seems to run quickly, but hangs on the last file conversion causing memory usage to increase to the system max and Windows to become unresponsive. Sometimes the process finishes, sometimes there's a bluescreen. 

Issue replicated on 2 Windows machines (32 gb and 48 gb), as well as Ubuntu 24.04 in a Docker container (23 gb ram).

Below is the log from the Ubuntu container:
> deno run -Ar jsr:@fresh/update
>  🍋 Fresh Updater 
> 
> Note: Breaking changes may require additional manual updates.
> 
> Where is the project directory? .
> 🚀 Starting Fresh 1 to Fresh 2 migration...
> 📝 Updating configuration files...
> ✅ Configuration updated successfully
> 🗑️ Deleted fresh.gen.ts (no longer needed)
> 🔍 Scanning for source files...
> 📁 Found 64 files to process
> 🔄 Processing files...
> [02:52] [##################################################] [479/479 files]
> 
> 📊 Migration Summary:
>    Total files processed: 64
>    Successfully modified: 61
>    Unmodified (no changes needed): 3

Image

ottenhoff:
Fix system crash (OOM) caused by updater spending time in folders like docs/

- Skip hidden and vendor directories (node_modules, vendor, docs, etc.) to avoid OOM
- free ASTs immediately
- ts-morph dependency bumped to ^27.0.2

Closes https://github.com/denoland/fresh/issues/3483


packages/update/src/update.ts
```diff
@@ -170,17 +175,27 @@ export async function updateProject(dir: string) {
...
+    const entry of walk(dir, {
+      includeDirs: false,
+      includeFiles: true,
+      exts: ["js", "jsx", "ts", "tsx"],
+      skip: [SKIP_DIRS],
+    })
+  ) {
+    totalFiles++;
+    if (!HIDE_FILES.test(entry.path)) userFileCount++;
+  }
```
marvinhagemeister: Let's store the found paths, so that we don't need to walk the file system again a second time.

ottenhoff: Agreed. I just pushed a new commit.

packages/update/src/update.ts
```diff
@@ -12,6 +13,10 @@ export const PREACT_SIGNALS_VERSION = "2.5.0";
 
 // Function to filter out node_modules and vendor directories from logs
 const HIDE_FILES = /[\\/]+(node_modules|vendor)[\\/]+/;
+// Directories we should completely skip when walking the tree. This keeps us
+// from loading vendored dependencies (often thousands of files) into ts-morph.
+const SKIP_DIRS =
```
marvinhagemeister, line 16: What's the difference between `HIDE_FILES` and `SKIP_DIRS`? Aren't both used for the same thing? Let's unify that.

ottenhoff: yup, done.

bartlomieju (commented):
Good fix for a real OOM problem. The core idea — stream files one-by-one via `walk` + `forget()`/`removeSourceFile()` instead of loading everything via `addSourceFilesAtPaths` — is solid. A few things to address:

### 1. `userFileCount` vs `totalFiles` are always equal (logic bug)

Since `walk` is called with `skip: [SKIP_DIRS]`, no file from a skipped directory will ever appear in `filesToProcess`. That means the `if (!SKIP_DIRS.test(entry.path)) userFileCount++` check will always be true for every entry, so `userFileCount === totalFiles` always. The distinction between the two is dead code after the refactor.

Same applies to the `modifiedFilesToShow` filter — it's filtering against paths that can never match since `walk` already excluded them.

You can simplify by dropping `userFileCount` entirely and just using `filesToProcess.length`.

### 2. `docs` in SKIP_DIRS is too aggressive

\`\`\`
node_modules|vendor|docs|.git|.next|.turbo|_fresh|dist|build|target|.cache
\`\`\`

A user's Fresh 1.x project could easily have a `docs/` directory containing actual `.tsx` routes or components that need migration. Unlike `node_modules`/`vendor`/`dist`/`build` which contain generated or third-party code, `docs` is a common name for user-authored content. I'd remove `docs` from this list.

### 3. The old code comment said "sequential" but was actually parallel

The original code had:
```ts
// process files sequentially to show proper progress
await Promise.all(sfs.map(async (sourceFile) => { ... }));
```

That was parallel via `Promise.all`, not sequential. The new `for` loop actually is sequential — good fix, but worth noting this was also a latent bug (progress bar was racing).

### 4. Minor: `build` and `target` in skip list

`build` is a pretty common directory name that could appear in user projects (e.g., `routes/build/index.tsx`). Consider whether matching it only at the project root level would be safer, or just remove it. `target` (Rust convention) is unlikely to contain TS/TSX files so that's fine.

### 5. Test change looks correct

The updated expectations for `node_modules/foo/bar.ts` and `vendor/foo/bar.ts` to retain old imports (not be transformed) is the right behavior — these files should now be skipped entirely.

Overall this is a good improvement. The main things to fix are #1 (dead code from the refactor) and #2 (`docs` being too aggressive).

Issue #3606, opened 2025-11-06
Vite Plugin doesn't transform server code


lishaduck:
This is probably working as intended, but when I tried to switch the logo to use an `.svg` import (was going to try to integrate [npm:vite-imagetools](https://www.npmjs.com/package/vite-imagetools)), Deno throws a _very_ long error:

```log
Task dev vite

  VITE v7.1.12  ready in 709 ms

  ➜  Local:   http://127.0.0.1:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
Error: Cannot find module '@assets/logo.svg' imported from '/Users/dukese01/Developer/webmaster-25-26/routes/index.tsx'
    at fetchModule (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:34205:34)
    at RunnableDevEnvironment.fetchModule (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:35039:10)
    at fetchModule (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:35008:16)
    at handleInvoke (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:28409:27)
    at EventEmitter.listenerForInvokeHandler (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:28468:19)
    at EventEmitter.emit (ext:deno_node/_events.mjs:436:20)
    at Object.send (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/chunks/config.js:18469:39)
    at Object.invoke (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/module-runner.js:532:31)
    at Object.invoke (/Users/dukese01/Developer/webmaster-25-26/node_modules/.deno/vite@7.1.12/node_modules/vite/dist/node/module-runner.js:581:31)
[...]

PR #3634, opened 2025-11-20
refactor: complete move to Vite Environment API

lishaduck:
I was perusing the code whilst investigating #3606 and noticed that, despite the release post claiming that Fresh had ["fully embrace[d]"](https://deno.com/blog/fresh-and-vite#:~:text=fully%20embraces) the Vite Environment API, it was still using `ssrLoadModule`. Once I started looking, I noticed a bunch of other code that wasn't future-proof (env-wise, not rolldown-wise, which'll bring it's own swath of deprecations[^1]), so I updated them as well.

The newly added source map tests from #3624 currently fail again because the ModuleRunner API at the heart of the Environment API relies on [`process.setSourceMapsEnabled`](https://docs.deno.com/api/node/process/~/Process.setSourceMapsEnabled), which is [currently set to no-op](https://github.com/denoland/deno/pull/22993) in Deno.

[^1]: Separately tried to try out rolldown-vite but it seems that Deno doesn't support aliasing dependencies whatsoever? (I know there's not official support, but I was going to try to edit the lockfile to hack around it; however, it seems that there's not alias support in there at all?)

### DIFF
--- a/packages/plugin-vite/src/plugins/verify_imports.ts
+++ b/packages/plugin-vite/src/plugins/verify_imports.ts
@@ -15,7 +15,7 @@ export interface ImportCheckDiagnostic {
 /** A check whether or not an import is valid or not for an environment */
 export type ImportCheck = (
   id: string,
-  env: "ssr" | "client",
+  env: string,
 ) => ImportCheckDiagnostic | void;
 
 export interface CheckImportOptions {
@@ -26,7 +26,7 @@ export function checkImports(pluginOptions: CheckImportOptions): Plugin {
   function check(
     options: CheckImportOptions,
     id: string,
-    env: "ssr" | "client",
+    env: "server" | "client",
   ): ImportCheckDiagnostic | undefined {
     for (let i = 0; i < options.checks.length; i++) {
       const check = options.checks[i];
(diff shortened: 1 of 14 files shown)


packages/plugin-vite/src/plugins/verify_imports.ts
```diff
@@ -61,32 +61,36 @@ export function checkImports(pluginOptions: CheckImportOptions): Plugin {
...
             }
 
             seen.add(key);
           }
         } else {
-          const key = `${env}::${id}::${importer}`;
+          const key = `${this.environment.config.consumer}::${id}::${importer}`;
           if (!seen.has(key)) {
-            result = check(pluginOptions, id, env);
+            result = check(pluginOptions, id, this.environment.config.consumer);
```
lishaduck, line 93: This code might be better off using `this.environment.name` for the keys.

lishaduck:
> which is [currently set to no-op](https://github.com/denoland/deno/pull/22993) in Deno

Looked closer, the actual code has a comment saying source maps are always on, so not sure what's up with that?

marvinhagemeister (approved):
Thanks! I'll bring the missing process method up to the CLI team

EDIT: nvm made a patch myself https://github.com/denoland/deno/pull/31358

lishaduck:
canary CI ran with https://github.com/denoland/deno/pull/31358 and still failed, not sure what's up. I think it probably makes sense to split out the ModuleRunner commit, merge the others, and figure out the source map failures separately.

lishaduck:
> I think it probably makes sense to split out the ModuleRunner commit [..] and figure out the source map failures separately.

Filed #3645 for further followup

PR #3645, opened 2025-12-06
feat: use modulerunner for ssr

lishaduck:
Extracted from/stacked upon #3634.


packages/plugin-vite/src/plugins/dev_server.ts
```diff
@@ -79,16 +90,20 @@ export function devServer(): Plugin[] {
...
           try {
-            const mod = await server.ssrLoadModule("fresh:server_entry");
-            const req = createRequest(nodeReq, nodeRes);
-            mod.setErrorInterceptor((err: unknown) => {
+            const mod = await server.environments.ssr.runner.import<unknown>(
+              "fresh:server_entry",
+            ) as FetchHandler;
+            mod.setErrorInterceptor?.((err: unknown) => {
               if (err instanceof Error) {
                 server.ssrFixStacktrace(err);
```
lishaduck, line 101: If this works to fix it, then that's great, but in theory it shouldn't be needed 🤔 

> `server.ssrFixStacktrace` and `server.ssrRewriteStacktrace` do not have to be called when using the Module Runner APIs. The stack traces will be updated unless `sourcemapInterceptor` is set to `false`. — [SSR Using ModuleRunner API](https://vite.dev/changes/ssr-using-modulerunner#:~:text=server%2EssrFixStacktrace%20and%20server%2EssrRewriteStacktrace%20do%20not%20have%20to%20be%20called%20when%20using%20the%20Module%20Runner%20APIs%2E%20The%20stack%20traces%20will%20be%20updated%20unless%20sourcemapInterceptor%20is%20set%20to%20false%2E)

I assumed it was because `sourcemapInterceptor` was indeed `false` because Deno had nooped `process.setSourceMapsEnabled`, but fixing that didn't seem to fix it. 

Relevant Vite code: https://github.com/vitejs/vite/blob/99897d27b44dd73307fa03e2f11f0baa1a1dc939/packages/vite/src/node/ssr/runtime/serverModuleRunner.ts#L58-L72

fibibot (commented):
Direction matches the rest of the env-API migration from #3634 — `ssrLoadModule` is the old shim for the SSR env.

One concern: `if (!isRunnableDevEnvironment(server.environments.ssr)) return;` exits the middleware without calling `next()` or writing a response — if the SSR env is ever non-runnable, the request hangs. For Fresh's default config this branch is unreachable, but it's a sharp edge; either fall through to `next()` or assert with a thrown error so failures are loud.

Holding approval until CI is green — `pr-title` is red because the workflow ran with the original title "Use ModuleRunner for SSR" before the lowercase rename. An empty push or rebase should re-trigger and clear it.

- nit: the hand-rolled `FetchHandler` types `setErrorInterceptor?` optional, forcing `?.` at the call site. `server_entry.ts` always exports it — making the field required would surface a real regression if that export ever changes.

PR #3646, opened 2025-12-07
Vite 8

lishaduck:
I have no intent on getting this working further, I just wanted to see how compatible Fresh would be already.

Currently @deno/loader integration is completely broken so... uh, the answer is: Fresh is completely broken under Vite 8.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -94,7 +94,7 @@
     "marked": "npm:marked@^15.0.11",
     "marked-mangle": "npm:marked-mangle@^1.1.9",
     "prismjs": "npm:prismjs@^1.29.0",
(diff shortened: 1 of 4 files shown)


lishaduck:
~~Cross-linking https://github.com/denoland/deno-rolldown-plugin.
We'll probably need it finished to replace `deno-esbuild-loader`.~~

A) we use `@deno/esbuild-plugin` now
B) we only use it in the (deprecated?) Builder API AFAICT
C) we should still fix it cuz it's just useful in general

bartlomieju:
This should be resolved with https://github.com/denoland/fresh/pull/3734 I believe

PR #3659, opened 2025-12-24
fix: handle paths with whitespaces in staticFiles

### LINKED ISSUE #3657: can't load static file if path contains blank (%20)
# Bug Description
Recently, I encountered an issue in my fresh project where static files fail to load if their path contains spaces – for example, accessing `localhost:5173/image%20.png` would not load the image. This issue went unnoticed initially, causing broken image rendering on my website without any obvious error cues.

I spent six hours troubleshooting this problem:
1. Learned to debug Deno applications in VSCode
2. Configured VSCode to debug Vite and fresh-vite-plugin
3. Attempted to debug fresh-core, only to find it was evaluated via the vite-plugin (making direct debugging impossible)

# Root Causes Identified
1. <a href="https://github.com/denoland/fresh/blob/2.2.0/packages/plugin-vite/src/plugins/dev_server.ts#L47" target="_blank" rel="noopener noreferrer">plugin-vite</a> not decode URI components using `decodeURIComponent()`, which blocks static resource loading at the primary layer (this appears to be the main mechanism for static file retrieval).
2. The `staticFile` component imported via `{ staticFiles } from "fresh"` (see <a href="https://github.com/denoland/fresh/blob/2.2.0/packages/fresh/src/build_cache.ts#L71" target="_blank" rel="noopener noreferrer">staticFile</a>) is effectively non-functional: it requires manual registration, yet the API is not publicly exposed – making it impossible to configure or inject custom logic.

# Custom Solution Code
```typescript
const MIME_TYPES: Record<string, string> = {
  txt: "text/plain",
  html: "text/html",
  css: "text/css",
  js: "application/javascript",
  json: "application/json",
  png: "image/png",
  jpg: "image/jpeg",
  jpeg: "image/jpeg",
  gif: "image/gif",
  webp: "image/webp",
  ico: "image/x-icon",
  svg: "image/svg+xml",
  pdf: "application/pdf",
  mp4: "video/mp4",
};

app.use(async (ctx) => {
  const { req, url, config } = ctx;

  // Decode URI component to handle spaces in file paths
  let pathname = decodeURIComponent(url.pathname);
  
  // Normalize base path if configured
  if (config.basePath) {
    pathname = pathname !== config.basePath
      ? pathname.slice(config.basePath.length)
      : "/";
  }

  // 安全校验：防止路径遍历攻击（避免访问上级目录文件）Resolve to static directory
  pathname = `/static/${pathname}`;

  // Security: Prevent path traversal attacks (e.g., ../../../../)
  const safePath = path.resolve(Deno.cwd(), pathname.slice(1));
  const rootDir = path.resolve(Deno.cwd());

  if (!safePath.startsWith(rootDir)) {
    return new Response("Forbidden", { status: 403 });
  }

  // Check if file exists
  let fileInfo: Deno.FileInfo | undefined;
  try {
    fileInfo = await Deno.stat(safePath);
  } catch (_error) {
    return await ctx.next();
  }
  if (!fileInfo || !fileInfo.isFile) {
    return await ctx.next();
  }

  // Serve the file with proper headers
  try {
    const file = await Deno.open(safePath, { read: true });
    const ext = path.extname(safePath).toLowerCase().replace(".", "");
    const contentType = MIME_TYPES[ext];
    if (!contentType) {
      return await ctx.next();
    }

    const headers = new Headers({
      "Content-Type": contentType,
      "Content-Length": fileInfo.size.toString(),
      "Vary": "If-None-Match",
      "Last-Modified": fileInfo.mtime?.toUTCString() || new Date().toUTCString(),
    });

    return new Response(file.readable, { headers });
  } catch (_error) {
    return await ctx.next();
  }
});
```

# version

```shell
$ deno -v
deno 2.6.0
$ deno task dev -v
Task dev vite "-v"
vite/7.3.0 linux-x64 node-v24.2.0
$ cat  deno.json |grep fresh
    "start": "deno serve -A _fresh/server.js",
    "update": "deno run -A -r jsr:@fresh/update ."
        "fresh",
    "**/_fresh/*"
    "fresh": "jsr:@fresh/core@^2.2.0",
    "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@^1.0.8",
```

LY1806620741:
Fix [#3657](https://github.com/denoland/fresh/issues/3657)

Static files with special characters in their names (spaces, `#`, unicode, etc.) fail to load due to path encoding issues.

## Root cause

`prepareStaticFile()` uses `new URL(pathname, base)` to normalize paths, but `new URL` doesn't properly encode characters that are already percent-encoded in the filesystem path. For example, a file literally named `test %20encodeUri` becomes `test%20%20encodeUri` (the space and literal `%20` are indistinguishable), instead of the correct `test%20%2520encodeUri`.

### `fresh-build` (`dev_build_cache.ts`)
- Add `systemPathToUrlEncoded()` that splits the path into components and encodes each with `encodeURIComponent()`, correctly handling characters that look like percent-encoding in filenames.
- Use it in `prepareStaticFile()` instead of `new URL()`.

### `fresh` static middleware (`static_files.ts`)
- Remove `decodeURIComponent()` from the pathname lookup — the build cache now stores URL-encoded paths, so the incoming `url.pathname` (already encoded) matches directly.

### `fresh-vite-plugin` (`dev_server.ts`)
- Add `decodeURIComponent()` when resolving URL paths to filesystem paths in the dev server, so files with special characters can be found on disk.

### Tests
- Update static file tests to use `systemPathToUrlEncoded()` when building the mock cache, matching the real build flow.
- Add integration tests for files with spaces in both dev and production modes.

### DIFF
--- a/deno.lock
+++ b/deno.lock
@@ -3794,7 +3794,7 @@
     "https://deno.land/x/imagescript@1.3.0/v2/ops/resize.mjs": "814e78ebce8eaf8f1f918688db7b52a141405e06a36ed4b25d04413d69e7d17b",
     "https://deno.land/x/imagescript@1.3.0/v2/ops/rotate.mjs": "a1b65616717bd2eed8db406affea3263b4674dada46b56441ef38167a187455d",
     "https://deno.land/x/imagescript@1.3.0/v2/util/mem.mjs": "4968d400dae069b4bf0ef4767c1802fd2cc7d15d90eda4cfadf5b4cd19b96c6d",
-    "https://esm.sh/@docsearch/js@3.5.2/es2020/js.mjs": "9b278cf3c0b26feded7d8efeac8e2b50f76bbafcf173a95002944bcc3482830a",
+    "https://esm.sh/@docsearch/js@3.5.2/es2020/js.mjs": "964600b3c133bccfa6a5ffa240e8272a08eeff22f5fea6993aa085cfc9e4d750",
     "https://esm.sh/@docsearch/js@3.5.2?target=es2020": "4bad084f771a1923fe042ece62a9078f482f8642cb0b1acb890905e58586fee7",
     "https://raw.githubusercontent.com/denoland/ga4/main/mod.ts": "36f72ba1c90b5ebdb811427f367cd95fa6772d2de2fb45d6e57550501ee6d476",
     "https://raw.githubusercontent.com/denoland/std/refs/heads/main/_tools/check_docs.ts": "e0bf3d0ddbea93e769f0ad24f16ef2c42311c489165c9e0de9d0876d0411ffdc",
--- a/packages/fresh/src/dev/dev_build_cache.ts
+++ b/packages/fresh/src/dev/dev_build_cache.ts
@@ -528,6 +528,23 @@ ${serializedFsRoutes}
 `.replaceAll(/\n[\n]+/g, "\n\n");
 }
 
+export function systemPathToUrlEncoded(systemPath: string) {
+  const normalizedPath = systemPath.replaceAll(WINDOWS_SEPARATOR, "/");
+
+  const pathComponents = normalizedPath.split("/").filter((comp) => comp);
+
+  const encodedComponents = pathComponents.map((comp) =>
+    encodeURIComponent(comp)
+  );
+
+  const encodePath = "/" + encodedComponents.join("/");
+
+  return new URL(
+    encodePath,
+    "http://localhost",
+  ).pathname;
+}
+
 export async function prepareStaticFile(
   item: PendingStaticFile,
   outDir: string,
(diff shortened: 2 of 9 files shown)


marvinhagemeister:
I'm afraid I don't speak that language. Can you translate that too english?

LY1806620741:
> I'm afraid I don't speak that language. Can you translate that too english?

Ok. It is written on the meaning of bilingualism, just look at the English part.

Fixed an issue where URL encoding errors caused some static files to be unreadable. For example, the file name contains spaces.

LY1806620741:
> I'm afraid I don't speak that language. Can you translate that too english?

This PR fixes critical path encoding bugs in both fresh-build and fresh-vite-plugin that caused static assets with special characters (like spaces) to fail loading or be served incorrectly.
1. Changes in fresh-build
Fixed incorrect space encoding in prepareStaticFile():
Issue: The function was incorrectly converting encoded URIs like test%20encodeUri into test%20%20encodeUri (injecting an extra literal space) instead of the correct double-encoded format test%20%2520encodeUri.
Fix: Corrected the logic to ensure proper URI component encoding without introducing malformed spaces.
Resolved double-encoding conflict:
Issue: There was redundant encoding logic between the staticFiles configuration and the freshServeStaticFiles() function, resulting in paths being encoded twice (e.g., % becoming %25).
Fix: Removed the duplicate encoding step to ensure paths are encoded exactly once.
2. Changes in fresh-vite-plugin
Fixed path conversion in devServer:
Issue: When the devServer attempted to read system files via urlpath, the path conversion logic failed to handle encoding correctly, leading to file not found errors for assets with special characters.
Fix: Updated the urlpath resolution logic to correctly decode/encode paths when accessing the file system.

PR #3669, opened 2026-01-24
fix: typo in Subscribe page handler reference

DouglasdeMoura:
This PR removes an extra `s` on the type definition of the handler for the page.


docs/latest/advanced/forms.md
marvinhagemeister, line 25: Looks like this needs to be updated too.

bartlomieju (approved):

PR #3671, opened 2026-01-29
fix: change redirect default status from 302 to 307

### LINKED ISSUE #2632: Bug: POST redirects with basepath wrong method.
From discord https://discord.com/channels/684898665143206084/991511118524715139/1279100135606849619

> When using deno fresh alone with the basepath config parameter set (let's say "/basepath"), all requests made to "/" are redirect to "/basepath" with a 302 Found status (this is correct I think).
But when the request is a POST, the redirection is made with a GET, I think the init method is not reused.

ddmoney420:
## Summary
Changes the default redirect status code from 302 to 307 to preserve the original HTTP method on redirect.

This fixes POST redirects being converted to GET requests when using basepath configuration.

## Problem
Per RFC 7231:
- **302 Found**: User agents MAY change the request method (browsers convert POST→GET)
- **307 Temporary Redirect**: Method and body MUST NOT be changed

The current default of 302 breaks form submissions when redirecting, as POST requests become GET requests.

## Solution
Change the default status in `ctx.redirect()` from 302 to 307.

Users who explicitly need 302 behavior can still pass it:
```ts
ctx.redirect("/path", 302)
```

## Test plan
- Updated existing redirect tests to expect 307 as default
- Added test verifying explicit 302 still works for backward compatibility

Fixes #2632

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/fresh/src/context.ts
+++ b/packages/fresh/src/context.ts
@@ -165,7 +165,7 @@ export class Context<State> {
    * ctx.redirect("//evil.com/");
    * ```
    */
-  redirect(pathOrUrl: string, status = 302): Response {
+  redirect(pathOrUrl: string, status = 307): Response {
     let location = pathOrUrl;
 
     // Disallow protocol relative URLs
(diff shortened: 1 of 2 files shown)


ddmoney420:
Closing this PR as part of cleaning up my contributions. I spread too thin across projects. Apologies for the noise.

PR #3678, opened 2026-02-11
chore: release 2.2.1

marvinhagemeister:


jeveloper:
Am I mistaken or the package failed to publish?

dgaubert:
@jeveloper nope, you are right. Seems to be an expired or missed token

image

jeveloper:
much love to Deno team, I guess fresh slid into lack of attention and thats okay. communication in the future would be preferred, even if 6 months of abandon time is about to happen :)

planning to deploy on Deno soon

PR #3682, opened 2026-02-21
Updates the examples on the homepage

jasonmccallister:
The example island and partials are using the old import styles, this PR updates those examples. 

Running `deno task ok` prompted for `deno install -r` - hence the updated lock file.


bartlomieju (approved):
Thanks

PR #3696, opened 2026-03-21
feat:Actually ship no JS by default

### LINKED ISSUE #3662: No JavaScript?
Hi!

As I was preparing a presentation about Fresh I wanted to demonstrate the following claim on the Fresh homepage:

> Fresh routes are dynamically server-rendered [Preact](https://preactjs.com/) components, so there's zero JavaScript shipped to the browser by default.

This claim is repeated in several places, including Fresh's `README.md`

> Zero runtime overhead: no JS is shipped to the client by default.

And I ran into an issue...

Given the following code:

`vite.config.ts`
```ts

export default defineConfig({
  plugins: [fresh({
    // https://github.com/denoland/fresh/issues/3476
    serverEntry: "./main.tsx",
  })],
});
```

`main.tsx`
```tsx
import { App, staticFiles } from "fresh";

export const app = new App();

app.use(staticFiles());

app.get("/", (ctx) => {
  const time = new Date().toLocaleString();
  return ctx.render(<p>Freshly server-rendered {time}</p>);
});
```

Renders the following HTML (line breaks & indentation added for clarity):
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
  </head>
  <body>
    <p>Freshly server-rendered 05/01/2026, 20:56:39</p>
    <script type="module" nonce="944c3613b7b1427d9c781403e6154b9b">import { boot } from "/assets/client-entry-Cz-rWMt7.js";boot({},"[[]]");</script>
  </body>
</html>
```

Image

As you can see, **it does load JavaScript!** A whopping 35kb of it. Can't be Preact, that's only 3kb according to their website. 🤔 

What JavaScript is Fresh loading, and why does it load any JavaScript for this extremely simple page?

Ionaru:
In many places, Fresh marketing claims the framework ships no JS by default, **this was false**.

This PR reworks core parts of the framework to check whether JS is needed on the client with these criteria:
1. Use of islands
2. Use of Partials (`f-client-nav`)

If either of those those criteria are met, JavaScript will be loaded as normal, but if both are false, then the `client-entry` script is omitted from the page.

Fixes https://github.com/denoland/fresh/issues/3662


deno.lock
bartlomieju, line 1: These seem completely unrelated to the issue at hand. Please revert

bartlomieju (changes requested):
See comment about the lockfile

Ionaru:
I could not run `deno install --frozen` without updating the lockfile.
Probably an issue on the main branch, but I'll revert that.

bartlomieju (commented):
Nice work — this fixes a real gap between Fresh's marketing ("zero JS by default") and reality. The approach of checking `islands.size > 0 || clientNavEnabled` is clean, and the HMR split is a good call so dev mode still gets live reload on static pages.

A few things to consider:

**Logic issue in `FreshRuntimeScript`**: After the PR, when `needsClientRuntime` is false, the function falls through past the `if (RENDER_STATE!.needsClientRuntime)` block to the HMR/production-static returns at the bottom. But when `needsClientRuntime` is *true*, the full-document branch already returns inside the block — so that's fine. However, the partial response path (line ~582) returns early and doesn't check `needsClientRuntime` at all. That seems correct for partials, but worth a comment noting that partials always assume the client runtime is already loaded by the parent page.

**`f-client-nav` detection only checks `"true"` string**: The diff hook checks `(vnode.props)[CLIENT_NAV_ATTR] === "true"`, but earlier in the same file (line ~145-146), the attribute is coerced via `String(vnode.props[CLIENT_NAV_ATTR])`. So passing `f-client-nav` as a boolean `true` (JSX `f-client-nav` or `f-client-nav={true}`) gets stringified to `"true"` before this check runs — that works. But the test uses `<html f-client-nav>` which in Preact/JSX becomes `true` (boolean). Worth verifying this ordering is guaranteed (diff hook runs after vnode hook), and maybe add a brief comment explaining the dependency.

**`f-client-nav={false}` test**: The test asserts no boot script, which is correct. But `f-client-nav={false}` still gets stringified to `"false"` and rendered into the HTML as `f-client-nav="false"`. The detection correctly skips it since `"false" !== "true"`. Good.

**Minor**: The `isJsMediaType` exhaustive return at the end of `deno.ts` — this is unrelated to the feature but fine as a drive-by fix for the type checker.

Overall this looks good to merge once the ordering guarantee between the vnode hook and diff hook is confirmed/documented.

packages/fresh/src/runtime/server/preact_hooks.ts
```diff
@@ -375,6 +382,17 @@ options[OptionsType.DIFF] = (vnode) => {
           break;
```
bartlomieju, line 439: The diff hook check here depends on the vnode hook (line ~145) having already stringified the boolean `f-client-nav` prop to `"true"`. If for some reason the diff hook fires before the vnode hook on the same VNode, this would miss boolean `true` values. Preact's current behavior guarantees vnode runs first, but a comment noting this dependency would help future maintainers.

Ionaru:
Thank you for the review! I'll process your feedback when I've got some time in the next couple of days.

Ionaru:
@bartlomieju Comments added and tests fixed :)

bartlomieju (approved):
Nice work, thank you

jeveloper:
yahooooo

:))

PR #3733, opened 2026-03-28
chore: inline ga4 module and remove raw GitHub import

bartlomieju:
## Summary
- Inlined the `$ga4` module (from `https://raw.githubusercontent.com/denoland/ga4/main/mod.ts`) into `www/utils/ga4.ts`
- Replaced old `deno.land/std` URL imports with `@std/http` (`getCookies`, `STATUS_TEXT`)
- Inlined the trivial `snakeCase` helper (was from `deno.land/x/case`)
- Removed `$ga4` import map entry from `deno.json`

The ga4 package was never published to jsr or npm, so inlining was the cleanest way to eliminate the last raw URL-based dependency.

## Test plan
- [x] `deno task ok` passes (fmt, lint, types, 505 tests)

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -61,7 +61,6 @@
     "mime-db": "npm:mime-db@^1.54.0",
     "preact": "npm:preact@^10.28.2",
     "preact-render-to-string": "npm:preact-render-to-string@^6.6.5",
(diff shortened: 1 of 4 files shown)


crowlKats (approved):
LGTM

PR #3697, opened 2026-03-24
fix: improve CJS-to-ESM transform for better npm compatibility

bartlomieju:
## Summary

Fixes several issues in the Fresh Vite plugin's CommonJS-to-ESM Babel transform (`commonjs.ts`) that caused npm packages to break during builds while working fine in dev mode.

- **Fix `.mts`/`.cts` extension detection** — `.cts` was incorrectly in the ESM branch (should be `.mts`), causing TypeScript CJS files to skip the transform
- **Handle `require.resolve()`** — injects `createRequire(import.meta.url)` polyfill so `require.resolve()` calls work in ESM output (#3619)
- **Polyfill `__dirname`/`__filename`** — injects `fileURLToPath`/`dirname` based polyfill when these CJS globals are used
- **Use `var` instead of `let` for `_default`** — avoids TDZ errors when Rollup reorders declarations in bundled output (#3653)
- **Track `module.exports.X` as named exports** — the MemberExpression visitor now handles both `exports.X` and `module.exports.X` patterns

Addresses: #3619, #3653

## Test plan

- [x] All 49 `commonjs_test.ts` tests pass (42 existing + 7 new)
- [x] New test cases cover each fix:
  - `require.resolve` injects `createRequire`
  - `require.resolve` + `require()` combined (single injection)
  - `.mts` files treated as ESM
  - `.cts` files treated as CJS
  - `__dirname`/`__filename` polyfill injection
  - `module.exports.X` tracked as named export
  - Primitive `module.exports` with namespace re-export guard

## Note on remaining CJS issues

This PR does **not** fix CJS issues caused by `require()`-to-`import` hoisting changing package.json conditional export resolution (#3505, #3478, #3449). Those need resolver-level changes. See #3593 for native addon externalization (also out of scope).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3653, opened 2025-12-15
CJS dependencies cause fatal issues when building/running that go unnoticed in dev mode


CAYdenberg:
Reproduction is here: https://github.com/CAYdenberg/fresh-2-cjs-build-repro

- deno task dev - **works**
- deno task build then deno task start - **does not work**, results in 

```
error: Uncaught (in promise) ReferenceError: Cannot access '_default$i' before initialization
const Range$2 = __require$g ?? _default$i ?? _mod$2;
```

Also, resetting to the initial commit or commenting out the use of Sharp will result in a successful build.

The issue probably stems from Sharp being written in CJS. Still, it would be awesome if deno install or Fresh or the Fresh vite plugin could issue a warning when encountering CJS, as it is very difficult to debug when a project is working fine in development mode and then fails at the very last step.

jim-fx:
I tried to [replicate](https://github.com/jim-fx/svelte-kit-deno-sharp) it with svelte-kit + vite + deno + sharp, but for some reason this worked flawlessly. So maybe it's somehow related to `@fresh/plugin-vite`?

CAYdenberg:
> I tried to [replicate](https://github.com/jim-fx/svelte-kit-deno-sharp) it with svelte-kit + vite + deno + sharp, but for some reason this worked flawlessly. So maybe it's somehow related to @fresh/plugin-vite?

I mean, it's going to come down to the order things are bundled in. So trying to reproduce it in a different context doesn't tell us much.

CAYdenberg:
Awesome.

PR #3734, opened 2026-03-28
fix: improve Deno module resolution in Vite plugin

### LINKED ISSUE #3437: Published Fresh app on JSR doesn't work
Discord: https://discord.com/channels/684898665143206084/991511118524715139/1417096969473626184

> is it possible to publish a fresh app to jsr? i tried it, but other than index.html all requests return enoent error.
> Thanks, you can check it from [@merato/voyager-ui](https://jsr.io/@merato/voyager-ui)

### LINKED ISSUE #3654: Vite SSR Cannot Resolve Deno Import Map Alias for MongoDB Driver
👋 everyone 

We are building a podcast analytics platform using:
- Deno 2.x
- Fresh 2.x (with Vite integration)
- MongoDB via jsr:@db/mongo@^0.34.0

We are new to this whole ecosystem, but we have really enjoyed the experience so far 🫶
But now, we are stuck on this issue:

## Project Structure

deno.json (relevant parts):
```
{
  "imports": {
    "mongodb-deno": "jsr:@db/mongo@^0.34.0",
    "fresh": "jsr:@fresh/core@^2.2.0",
    "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@^1.0.8",
    // ... other imports
  }
}
```

vite.config.ts:
```

export default defineConfig({
  plugins: [fresh()],
  ssr: {
    external: ["echarts", "mongodb-deno"],
  }
});
```

util/db.ts (database connection):
```
import { Database, MongoClient } from "mongodb-deno";

let client: MongoClient | null = null;

export async function getDb(): Promise<Database> {
  if (!client) {
    client = new MongoClient();
    await client.connect(MONGODB_URI);
  }
  return client.database(MONGODB_DB);
}
```

routes/index.tsx (imports and uses db.ts):
```
import { define } from "@/utils.ts";
import { getDb } from "@/util/db.ts";

export const handler = define.handlers({
  async GET(ctx) {
    const db = await getDb();
    // ... fetch data and render
  },
});
```

## The Problem

Build works fine:
`deno task build`  ✅ Succeeds

Dev server starts, but fails when accessing any route:
`deno task dev`

When accessing http://localhost:5173/:
```
Error: Cannot find module 'mongodb-deno' imported from '/Users/.../util/db.ts'
    at fetchModule (/Users/.../npm/registry.npmjs.org/vite/7.3.0/dist/node/chunks/config.js:33979:34)
```
What we've Tried

1. Adding mongodb-deno to ssr.external in vite.config.ts
  - Result: Still tries to load it
2. Using direct JSR imports instead of import map alias:
`import { Database, MongoClient } from "jsr:@db/mongo@^0.34.0";`
  - Result: Error: Failed to load url jsr:@db/mongo@^0.34.0 ... Does the file exist?
3. Creating a shim file and using Vite alias:
```
resolve: {
  alias: {
    "mongodb-deno": resolve("./util/mongodb-deno-shim.ts"),
  },
}
```
  - Result: Same error, Vite still can't load JSR imports
4. Custom Vite plugin to mark JSR imports as external:
```
{
  name: "jsr-external",
  resolveId(id) {
    if (id.startsWith("jsr:")) {
      return { id, external: true };
    }
  },
}
```
  - Result: Breaks Fresh's own JSR dependencies

## The Core Issue (As far as we understand)

Vite's SSR module runner in dev mode tries to process mongodb-deno even though it's marked as external. It cannot resolve:
- The Deno import map alias (mongodb-deno → jsr:@db/mongo)
- Direct JSR package URLs (jsr:@db/mongo@^0.34.0)

The build works because it only bundles client code and keeps server code external. But in dev mode, when a route handler imports util/db.ts, Vite's SSR tries to process it and fails.

### Questions we are facing:

1. What's the recommended way to use JSR packages (especially database drivers) in Fresh 2.x routes?
2. Should server-only imports like MongoDB drivers be handled differently in Fresh 2.x + Vite?
3. Is there a Vite config or Fresh plugin option we're missing that would allow Deno to handle its own import resolution during SSR?
4. Are there any working examples of Fresh 2.x projects using MongoDB or other JSR database drivers?

## Additional Context

- Only fails during deno task dev when accessing routes
- We also had to work around a broken transitive dependency (@lucsoft/web-bson) which we fixed with an import map override

🙏 Any guidance would be greatly appreciated! Thank you!

bartlomieju:
## Summary

Three targeted fixes to Fresh's Vite plugin for better module resolution, plus new test coverage.

### 1. Workspace import map resolution
Pass the importer's file URL to `loader.resolve()` so bare specifiers from workspace member packages resolve through the correct import map — not just the root `deno.json`.

Fixes #3654

### 2. Remote island resolution
Use `this.resolve(spec)` in the `fresh-island::` and `fresh-client-island::` resolveId handlers (`server_snapshot.ts` / `client_snapshot.ts`) so JSR remote islands get proper `\0deno::` specifiers. Without this, remote islands resolved to bare specifiers that Vite couldn't load.

Fixes #3437

### 3. Babel JSX operator precedence fix
`!ssr && id.endsWith(".tsx") || id.endsWith(".jsx")` → `!ssr && (id.endsWith(".tsx") || id.endsWith(".jsx"))`. The missing parentheses caused `.jsx` files to incorrectly get Preact JSX transforms during SSR.

### Other changes
- Bump `@deno/loader` from ^0.3.10 to ^0.4.0 (needed for workspace resolution fix)
- Bump `@astral/astral` from ^0.5.5 to ^0.5.6 (fixes `@std/async@1.2.0` compat)
- Add dev server tests for CJS module import and maxmind CJS (#3653)

## Test plan
- [ ] Dev server tests pass (including 2 new CJS tests)
- [ ] Build tests pass
- [ ] `deno check` passes

## Note on CJS issues
This PR does **not** fix CJS module compatibility issues in production builds. The new CJS tests verify dev server behavior (which uses Vite's built-in optimizer). Production CJS fixes need improvements to the CJS-to-ESM Babel transform in `commonjs.ts` — see PR #3697.


bartlomieju:
## Note on CJS module compatibility

The CJS production build test was removed because ioredis fails with `Cannot find module './lib/modern'` during Rollup builds. This is a **pre-existing limitation** of Fresh's `deno.ts` plugin — it's not a regression from this PR.

### Root cause

Fresh's `deno.ts` uses `@deno/loader` to resolve and load modules. When Rollup encounters CJS `require()` calls with relative paths (e.g., `require('./lib/modern')`), the resolution fails because:

1. The Deno loader resolves the parent module to a cache URL (e.g., `file:///Users/.../npm/.../ioredis/built/index.js`)
2. Rollup's CJS-to-ESM plugin (`commonjs.ts`) rewrites `require('./lib/modern')` to an ESM import
3. The relative import is resolved against the cached path, not the original `node_modules` path
4. The `./lib/modern` subpath doesn't exist in the resolved location

### How to fix

This could be addressed by PR #3697 which improves the CJS-to-ESM Babel transform, or by enhancing `deno.ts` to resolve CJS relative imports against the original `node_modules` path. The `@deno/vite-plugin` approach handled this natively because it delegated CJS resolution to Vite/Node's resolver, but that approach was abandoned due to esbuild transform conflicts.

In dev mode, CJS packages work fine because Vite's dev server handles module loading differently (through its own dependency optimizer).

PR #3737, opened 2026-03-29
fix: support WebSocket upgrades in Vite dev server

### LINKED ISSUE #3350: Fresh V2 Beta WebSocket upgrade request never reaches server behind vite development environment

bartlomieju:
## Summary

- Fix WebSocket upgrade requests never reaching Fresh handlers in Vite dev mode (#3350)
- Vite's Connect-based server fires `upgrade` events on the underlying `http.Server` rather than routing them through middleware, so `Deno.upgradeWebSocket()` was never called
- Start a `Deno.serve()` on a random port alongside Vite and proxy upgrade requests to it via raw TCP socket forwarding, where `Deno.upgradeWebSocket()` works natively
- Vite's own HMR WebSocket paths (`/__vite_hmr`, `/__vite_ping`) are excluded from proxying

## Test plan

- [x] New test `vite dev - websocket upgrade` verifies echo via WebSocket in dev mode
- [ ] Verify existing dev server tests still pass
- [ ] Manual test with a real WebSocket-based app in dev mode

Fixes #3350

### DIFF
--- a/packages/plugin-vite/demo/routes/tests/ws.tsx
+++ b/packages/plugin-vite/demo/routes/tests/ws.tsx
@@ -1,0 +1,17 @@
+import type { FreshContext } from "fresh";
+
+export const handler = {
+  GET(ctx: FreshContext) {
+    if (ctx.req.headers.get("upgrade") !== "websocket") {
+      return new Response("Not a WebSocket request", { status: 400 });
+    }
+
+    const { socket, response } = Deno.upgradeWebSocket(ctx.req);
+
+    socket.addEventListener("message", (event) => {
+      socket.send(`echo: ${event.data}`);
+    });
+
+    return response;
+  },
+};
--- a/packages/plugin-vite/src/plugins/dev_server.ts
+++ b/packages/plugin-vite/src/plugins/dev_server.ts
@@ -1,5 +1,6 @@
 import type { DevEnvironment, Plugin } from "vite";
 import * as path from "@std/path";
+import { connect } from "node:net";
 import { ASSET_CACHE_BUST_KEY } from "fresh/internal";
 import { createRequest, sendResponse } from "@remix-run/node-fetch-server";
 import { hashCode } from "../shared.ts";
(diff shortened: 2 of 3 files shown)


bartlomieju:
The second `Deno.serve()` + TCP proxy approach here works but is heavier than ideal. The root cause is that `Deno.upgradeWebSocket()` doesn't work with requests from `node:http` upgrade events — filed denoland/deno#33220 to track fixing this at the Deno level.

Once that Deno issue is resolved, this workaround can be replaced with a simple `upgrade` event handler that calls the Fresh handler directly. Until then, this is the only viable approach.

fibibot (changes requested):
Strategy is sensible — Vite's Connect-based server can't run `Deno.upgradeWebSocket()`, so proxying upgrade frames through a separate `Deno.serve` is the right shape.

Blocker: CI is red on `53bc5c82` — nine `vite dev` tests timed out at ~55s each (`loads islands`, `tailwind no _app`, `tailwind _app`, `css modules`, `route css import`, `partial island`, `nested islands`, `inline env vars`, `can import json in npm package`), all going through Puppeteer + SSR. The run is from 2026-03-29; a rebase + re-run will at least rule out drift from main, then it's worth confirming whether the new `Deno.serve` or the upgrade handler is interacting with the browser sessions.

- nit: `connect(wsPort, "127.0.0.1", ...)` reads `wsPort` synchronously inside the upgrade callback, but it's only set in the async `onListen`. If an upgrade fires before listening completes, `connect(0, ...)` fails. Practically rare (the Vite server hasn't started accepting either) but easy to harden with a `Promise.withResolvers()` gate.
- nit: `req.url === \"/__vite_hmr\"` is an exact-string match. Any query string would miss the bypass and proxy to the Deno server instead of letting Vite handle HMR. `req.url?.startsWith(\"/__vite_hmr\")` is safer.

bartlomieju:
Closing in favor of https://github.com/denoland/fresh/pull/3823

PR #3757, opened 2026-04-08
feat: add trustProxy option to respect X-Forwarded-* headers

### LINKED ISSUE #2660: Option to let Fresh respect `X-Forwarded-*` headers?

bartlomieju:
## Summary
- Adds `trustProxy` option to `FreshConfig` (default: `false`)
- When enabled, the handler reads `X-Forwarded-Proto` and `X-Forwarded-Host` headers and rewrites `ctx.url` so it reflects the actual client-facing URL
- Useful for apps running behind nginx, Caddy, or other reverse proxies where `ctx.url` would otherwise return `http://` instead of `https://`

Closes #2660

## Usage
```ts
const app = new App({ trustProxy: true });
```

## Test plan
- [ ] With `trustProxy: true`, send a request with `X-Forwarded-Proto: https` and `X-Forwarded-Host: example.com` — verify `ctx.url` reflects `https://example.com/...`
- [ ] With `trustProxy: false` (default), same headers are ignored — `ctx.url` uses the original request URL
- [x] Type checks pass

### DIFF
--- a/packages/fresh/src/config.ts
+++ b/packages/fresh/src/config.ts
@@ -36,7 +36,8 @@ export interface ResolvedFreshConfig {
   mode: "development" | "production";
   /**
    * When enabled, Fresh will respect `X-Forwarded-Proto` and
-   * `X-Forwarded-Host` headers to construct `ctx.url`.
+   * `X-Forwarded-Host` headers to construct `ctx.url`. Enable
+   * this when running behind a reverse proxy.
    */
   trustProxy: boolean;
 }
--- a/packages/fresh/src/config.ts
+++ b/packages/fresh/src/config.ts
@@ -15,6 +15,9 @@ export interface FreshConfig {
    * When enabled, Fresh will respect `X-Forwarded-Proto` and
    * `X-Forwarded-Host` headers to construct `ctx.url`. Enable
    * this when running behind a reverse proxy.
+   *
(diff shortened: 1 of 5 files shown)


packages/fresh/src/config.ts
```diff
@@ -11,6 +11,13 @@ export interface FreshConfig {
    * The mode Fresh can run in.
    */
   mode?: "development" | "production";
+  /**
+   * When enabled, Fresh will respect `X-Forwarded-Proto` and
+   * `X-Forwarded-Host` headers to construct `ctx.url`. Enable
+   * this when running behind a reverse proxy.
```
bartlomieju, line 17: Add the same comment as in https://github.com/denoland/fresh/pull/3757/changes#diff-869143188eb3aca39968a4ef6c2c1b954b515698eefc7e6887339d562a45fcc6R54

PR #3758, opened 2026-04-08
chore: deprecate legacy Builder API

### LINKED ISSUE #3635: Deprecation of the legacy `Builder` API
In the Fresh docs [the `Builder` API is described as "legacy"](https://fresh.deno.dev/docs/advanced/builder).

> The Builder class was used during the alpha phase of Fresh 2 before the Fresh vite plugin was released. You can skip this page if you’re using vite.

Does it make sense to deprecate this API officially, to discourage new users from this API? It might make sense also to update the OPTIONS when using `@fresh/init` to mention that the `--builder` is legacy, or deprecated if that route is chosen.

```
    --builder    (Legacy) Setup with builder instead of vite
```

I think that unless there's any blockers with the Vite plugin, it's better to deprecate ASAP, to clearly signal that this path should be moved away from.

## Deprecations

If it makes sense to deprecate, I think these are the public API's that should add `@deprecated` JSDoc:

- `@fresh/core/dev` - All public exports
  - `Builder` and all related types
- `@fresh/plugin-tailwind` - All public exports, since this plugin is only used with `Builder`
- `@fresh/plugin-tailwind-v1` - All public exports, since this plugin is only used with `Builder`

In addition, to make it clearer that this is not recommended for new projects, maybe add the following:

- A warning message if `--builder` is used with `@fresh/init`
- Update the option text for `--builder` to mark it as deprecated

bartlomieju:
## Summary
- Adds `@deprecated` JSDoc tags to all public exports from `@fresh/core/dev` (Builder, BuildOptions, ResolvedBuildConfig, OnTransformArgs, OnTransformOptions, TransformFn)
- Adds `@deprecated` JSDoc tags to `@fresh/plugin-tailwind` and `@fresh/plugin-tailwind-v3` exports
- Updates `--builder` flag in `@fresh/init` help text to show "(Deprecated)"
- Prints a warning when `--builder` flag is used during init

All deprecation notices point users to the Vite plugin as the replacement.

Closes #3635

## Test plan
- [ ] Verify IDE shows deprecation strikethrough on `Builder`, `BuildOptions`, etc. when imported from `fresh/dev`
- [ ] Run `deno run -A jsr:@fresh/init --builder .` and verify deprecation warning is printed
- [x] Type checks pass

### DIFF
--- a/packages/fresh/src/dev/builder.ts
+++ b/packages/fresh/src/dev/builder.ts
@@ -109,6 +109,10 @@ export type ResolvedBuildConfig = Required<Omit<BuildOptions, "sourceMap">> & {
   sourceMap?: FreshBundleOptions["sourceMap"];
 };
 
+/**
+ * @deprecated Use the Vite plugin (`@fresh/plugin-vite`) instead.
+ * The Builder class will be removed in a future version of Fresh.
+ */
 // deno-lint-ignore no-explicit-any
 export class Builder<State = any> {
   #transformer: FileTransformer;
--- a/packages/fresh/src/dev/mod.ts
+++ b/packages/fresh/src/dev/mod.ts
@@ -1,10 +1,25 @@
-export {
-  Builder,
-  type BuildOptions,
-  type ResolvedBuildConfig,
-} from "./builder.ts";
-export {
-  type OnTransformArgs,
-  type OnTransformOptions,
-  type TransformFn,
-} from "./file_transformer.ts";
+/**
+ * @deprecated Use the Vite plugin (`@fresh/plugin-vite`) instead.
+ * The Builder API will be removed in a future version of Fresh.
+ */
+export { Builder } from "./builder.ts";
+/**
+ * @deprecated Use the Vite plugin (`@fresh/plugin-vite`) instead.
+ */
+export type { BuildOptions } from "./builder.ts";
+/**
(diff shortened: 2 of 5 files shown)


lunadogbot (approved):
Scope is clean: JSDoc `@deprecated` on `Builder`, the five re-exported types, both Tailwind `tailwind()` entrypoints, and a runtime warning + help-text marker for `--builder` in `@fresh/init`. Lockfile untouched. Closes #3635. CI green; the failing `deploy/deno/fresh` preview is unrelated to a docstring/console-warn change and the sibling `deploy/deno/denoland-fresh` deploy is green.

One thing worth a follow-up: `www/dev.ts:3` and `packages/init/src/init.ts:559` still `import { Builder } from "fresh/dev"`, so the docs site and the init scaffolder will themselves emit deprecation warnings in IDEs after this lands. The init.ts import is required by the `--builder` codegen path so that one can stay; `www/` is the natural candidate to migrate next (or `// deno-lint-ignore deprecation` if migration is deferred).

- nit: the runtime warning at `packages/init/src/init.ts:148` says "Use Vite instead (the default)" but doesn't point at the migration docs — a link would save users a search.
- nit: the test-plan items in the PR body are unchecked; worth running `deno run -A jsr:@fresh/init --builder .` once before merging to confirm the warning renders the way you want.

Issue #3673, opened 2026-02-09
ioredis is inoperable in Fresh 2


ewired:
I am migrating a large Fresh 1 app to Fresh 2 and I am blocked by a guaranteed ETIMEDOUT in ioredis in the vite dev server. https://github.com/ewired/fresh-redis-experiment/tree/ioredis

```
Task dev vite

  VITE v7.3.1  ready in 261 ms

  ➜  Local:   http://127.0.0.1:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
ioredis v5 does not support plugging third-party Promise library anymore. Native Promise will be used.
GET http://127.0.0.1:5173/
[ioredis] Unhandled error event: Error: connect ETIMEDOUT
    at Socket.eval (/home/user/Documents/fresh-redis-experiment/node_modules/.deno/ioredis@5.9.2/node_modules/ioredis/built/Redis.js:216:27)
    at Object.onceWrapper (ext:deno_node/_events.mjs:559:14)
    at Socket.emit (ext:deno_node/_events.mjs:436:20)
    at Socket._onTimeout (node:net:1007:8)
    at cb (ext:deno_node/internal/timers.mjs:85:49)
    at callback (ext:deno_web/02_timers.js:42:7)
    at eventLoopTick (ext:core/01_core.js:223:13)
[ioredis] Unhandled error event: Error: connect ETIMEDOUT
    at Socket.eval (/home/user/Documents/fresh-redis-experiment/node_modules/.deno/ioredis@5.9.2/node_modules/ioredis/built/Redis.js:216:27)
    at Object.onceWrapper (ext:deno_node/_events.mjs:559:14)
    at Socket.emit (ext:deno_node/_events.mjs:436:20)
    at Socket._onTimeout (node:net:1007:8)
    at cb (ext:deno_node/internal/timers.mjs:85:49)
    at callback (ext:deno_web/02_timers.js:42:7)
    at eventLoopTick (ext:core/01_core.js:223:13)
^C
```

In the same environment, `deno run -A test_redis.ts` works fine, and ioredis is used without logging the Promise warning or having ETIMEDOUT.

bartlomieju:
This appears to be a Vite SSR module evaluation issue rather than a Fresh or Deno bug specifically. The key evidence:

- `deno run -A test_redis.ts` works fine with ioredis — Deno's Node compat handles it correctly in isolation
[...]

Issue #3505, opened 2025-09-28
mongoose (CJS) import error


fry69:
I am not sure if this is not already a known issue, but since it got mentioned in #3500 I think it is good to document it properly:

When trying to use Fresh 2 and Vite with `mongoose`, which gets distributed as CommonJS, it will fail when using Deno's automatic CommonJS import. At least in conjunction with Vite:

```shell
$ deno task build
Task build vite build
vite v7.1.7 building for production...
✓ 25 modules transformed.
_fresh/client/.vite/manifest.json                        0.63 kB │ gzip: 0.24 kB
_fresh/client/assets/client-entry-D1ibAoFr.css           6.92 kB │ gzip: 2.27 kB
_fresh/client/assets/fresh-island__Counter-lU_Fxzzr.js   0.82 kB │ gzip: 0.50 kB
_fresh/client/assets/preact.module-B5DG9wGp.js          10.91 kB │ gzip: 4.63 kB
_fresh/client/assets/client-entry-APY_tf7M.js           24.59 kB │ gzip: 9.10 kB
✓ built in 609ms
vite v7.1.7 building SSR bundle for production...
node_modules/.deno/mongodb@6.18.0/node_modules/mongodb/lib/bson.js (1:536): "default" is not exported by "node_modules/.deno/bson@6.10.4/node_modules/bson/lib/bson.node.mjs", imported by "node_modules/.deno/mongodb@6.18.0/node_modules/mongodb/lib/bson.js".
node_modules/.deno/mongodb@6.18.0/node_modules/mongodb/lib/bson.js (1:584): "default" is not exported by "node_modules/.deno/bson@6.10.4/node_modules/bson/lib/bson.node.mjs", imported by "node_modules/.deno/mongodb@6.18.0/node_modules/mongodb/lib/bson.js".
node_modules/.deno/mongoose@8.18.2/node_modules/mongoose/lib/helpers/clone.js (1:1200): "default" is not exported by "node_modules/.deno/bson@6.10.4/node_modules/bson/lib/bson.node.mjs", imported by "node_modules/.deno/mongoose@8.18.2/node_modules/mongoose/lib/helpers/clone.js".
[...]

Issue #3478, opened 2025-09-22
Error on mongodb


kavan-mevada:
```
import { define } from "../../utils.ts";
import { MongoClient } from "npm:mongodb@6.1.0";


export const handler = define.handlers({
  async GET(ctx) {
    const name = ctx.params.name;

    const password = "1Ik1eA38GBcGqVDd"
    const client = new MongoClient(`mongodb+srv://kavanmevada_db_user:${password}@cluster0.jua4ouv.mongodb.net/?retryWrites=true&w=majority&appName=Cluster0`);
    await client.connect();
    interface DinosaurSchema {
      name: string;
      skills: string[];
    }
    const db = client.db("animals");
    const dinosaurs = db.collection<DinosaurSchema>("dinosaurs");
    await dinosaurs.insertOne({
      name: "deno",
      skills: ["dancing", "hiding"],
    });
    const allDinosaurs = await dinosaurs.find({ name: "deno" }).toArray();

    console.log(allDinosaurs);
    client.close();

    return new Response(
      `Hello, ${name.charAt(0).toUpperCase() + name.slice(1)}!`,
    );
  },
});
```

deno build log:
```
Task build vite build
vite v7.1.7 building for production...
✓ 25 modules transformed.
_fresh/client/.vite/manifest.json                        0.63 kB │ gzip: 0.24 kB
_fresh/client/assets/client-entry-PRjVyscb.css           1.83 kB │ gzip: 0.89 kB
_fresh/client/assets/fresh-island__Counter-lU_Fxzzr.js   0.82 kB │ gzip: 0.50 kB
_fresh/client/assets/preact.module-B5DG9wGp.js          10.91 kB │ gzip: 4.63 kB
_fresh/client/assets/client-entry-BnA5gZOx.js           24.55 kB │ gzip: 9.07 kB
✓ built in 1.38s
vite v7.1.7 building SSR bundle for production...
node_modules/.deno/mongodb@6.1.0/node_modules/mongodb/lib/bson.js (1:245): "default" is not exported by "node_modules/.deno/bson@6.10.4/node_modules/bson/lib/bson.node.mjs", imported by "node_modules/.deno/mongodb@6.1.0/node_modules/mongodb/lib/bson.js".
[...]

Issue #3449, opened 2025-09-16
`@supabase/postgres-js` breaks due to commonjs


marvinhagemeister:
```sh
error: Uncaught (in promise) TypeError: Class extends value #<Object> is not a constructor or null
    at eval (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/@supabase+postgrest-js@1.16.1/node_modules/@supabase/postgrest-js/dist/cjs/PostgrestTransformBuilder.js, <anonymous>:6:429)
```

MauScheff:
Full trace:

```
Task start vite

  VITE v7.1.5  ready in 821 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
  ➜  press h + enter to show help
9:55:49 AM [vite] (ssr) Error when evaluating SSR module fresh:server_entry: Class extends value #<Object> is not a constructor or null
      at eval (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/@supabase+postgrest-js@1.16.1/node_modules/@supabase/postgrest-js/dist/cjs/PostgrestTransformBuilder.js:6:429)
      at async ESModulesEvaluator.runInlinedModule (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/module-runner.js:910:3)
      at async SSRCompatModuleRunner.directRequest (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/module-runner.js:1119:59)
      at async SSRCompatModuleRunner.directRequest (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/chunks/dep-M_KD0XSK.js:18863:22)
      at async SSRCompatModuleRunner.cachedRequest (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/module-runner.js:1037:73)
      at async eval (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/@supabase+postgrest-js@1.16.1/node_modules/@supabase/postgrest-js/dist/cjs/index.js:11:31)
      at async ESModulesEvaluator.runInlinedModule (/Users/mau/Development/vibechain/packages/app/node_modules/.deno/vite@7.1.5/node_modules/vite/dist/node/module-runner.js:910:3)
[...]

PR #3765, opened 2026-04-09
fix: externalize CJS-only npm packages in SSR build

### LINKED ISSUE #3653: CJS dependencies cause fatal issues when building/running that go unnoticed in dev mode

bartlomieju:
## Summary

CJS dependencies (Sharp, ioredis, MongoDB, etc.) cause fatal TDZ `ReferenceError` when building for production, even though they work fine in dev mode. The root cause is that Fresh's CJS-to-ESM babel transform hoists `require()` calls to `import` declarations, and Rollup can reorder these `const` bindings in the bundled output, causing references before initialization.

**Fix:** Instead of transforming CJS modules, externalize them in the SSR build. They're loaded at runtime by Deno's Node compat layer, which handles CJS natively.

### How it works

The SSR Rollup config now has an `external` function that checks each imported package:
1. Reads the package's `package.json` from `node_modules`
2. If the package has `"type": "module"`, a `"module"` field, or `"import"` conditions in `"exports"` → **keep bundled** (ESM, no issues)
3. If the package is CJS-only (none of the above) → **externalize** (skip CJS transform, load at runtime)

Framework packages (preact, @preact/signals, fresh) are always bundled regardless, to prevent duplicate module instances.

### Impact

Should fix the entire class of "CJS dependency breaks production build" issues:
- #3653 — Sharp
- #3673 — ioredis  
- #3505 — mongoose
- #3478 — MongoDB
- #3449 — @supabase/postgres-js

Closes #3653

## Test plan
- [x] All 28 existing build tests pass (2 pre-existing pwa failures unrelated)
- [ ] Clone the [reproduction repo](https://github.com/CAYdenberg/fresh-2-cjs-build-repro) and verify `deno task build && deno task start` works
- [ ] Test with ioredis, MongoDB, and other CJS deps
- [ ] Verify preact/signals still work correctly (no duplicate module issues)

### DIFF
--- a/packages/plugin-vite/src/mod.ts
+++ b/packages/plugin-vite/src/mod.ts
@@ -44,37 +44,68 @@ const SSR_BUNDLE_ALLOWLIST = new Set([
 ]);
 
 /**
- * Check if a package is CJS-only (no ESM entry point).
- * Returns true if the package should be externalized in the SSR build.
+ * Scan node_modules for CJS-only packages and return their names.
+ * A package is CJS-only if it has no ESM entry point (no "type": "module",
+ * no "module" field, no "import" condition in "exports").
  */
-function isCjsOnlyPackage(id: string, root: string): boolean {
-  // Extract bare package name (handle scoped packages)
-  const parts = id.startsWith("@") ? id.split("/", 2) : id.split("/", 1);
-  const packageName = parts.join("/");
+function findCjsOnlyPackages(root: string): string[] {
+  const result: string[] = [];
+  const nodeModulesDir = path.join(root, "node_modules");
 
-  if (SSR_BUNDLE_ALLOWLIST.has(packageName) || SSR_BUNDLE_ALLOWLIST.has(id)) {
-    return false;
+  let entries: fs.Dirent[];
+  try {
+    entries = fs.readdirSync(nodeModulesDir, { withFileTypes: true });
+  } catch {
+    return result;
   }
 
-  // Look for the package's package.json
-  const pkgJsonPath = path.join(
-    root,
-    "node_modules",
-    packageName,
-    "package.json",
-  );
+  for (const entry of entries) {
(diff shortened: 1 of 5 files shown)


lunadogbot (changes requested):
CI is red on every test job because `packages/plugin-vite/tests/fixtures/cjs_dependency/routes/index.tsx:3` imports `cjs-test-module` but the fixture has no `deno.json` (or `package.json`) declaring the import, so `deno task check:types` fails with `TS2307: Import "cjs-test-module" not a dependency and not in import map`. The integration test in `build_test.ts:689` also expects `node_modules/cjs-test-module/` to exist for the symlink and prod launch, but no such fixture is committed. Add a `deno.json` to the fixture and commit (or generate during test setup) a minimal `node_modules/cjs-test-module/{package.json,index.js}`.

- nit: `isCjsOnly` (`packages/plugin-vite/src/mod.ts:106`) checks for the `import` condition via `JSON.stringify(pkg.exports).includes('"import"')` — substring-matches any key or value containing the literal `"import"` (e.g. a custom `"importmap"` condition or a path with `import` in its name). Walking the exports object and looking for an exact `import` key is more robust.

PR #3769, opened 2026-04-11
fix: in sortRoutePaths: sort files alphabetically

### LINKED ISSUE #3773: [Bug] Incorrect route sorting order for mixed static and dynamic paths
## Description

The `sortRoutePaths` function does not correctly sort routes when a mix of static paths, dynamic parameters (`:name`), and catch-all segments (`[...]`) are provided. The sorting order does not match the expected hierarchy.

## Steps to Reproduce
You can reproduce this issue by running the following test case:

```ts
  test('fsRoutes - sortRoutePaths', () => {
    let routes = [
      '/foo',
      "/foo/:name",
      '/foo/baz'
    ];
    let sorted = [
      '/foo/baz',
      "/foo/:name",
      '/foo',
    ];
    routes.sort(sortRoutePaths);
    expect(routes).toEqual(sorted);
  });
```

## Expected Behavior

The routes should be sorted by specificity, where static paths have the highest priority, followed by dynamic parameters, and finally catch-all segments.

## Actual Behavior

The sorting fails, and the actual order is incorrect.

Error Log:

```diff
sortRoutePaths › fsRoutes - sortRoute

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Array [
    -   "/foo/baz",
        "/foo/:name",
    +   "/foo/baz",
        "/foo",
      ]
```

will-weiss:
The fix for the bug described was found with the assistance of Claude Opus 4.6. The analysis presented is from the AI, not my own, but I can confirm that the workaround it presented (which it said was based on its findings) did work for my use case.

## What I observed
For our fresh 2.2.0 project I observed that on github actions builds middleware defined in `routes/app/_middleware.tsx` were not running before the handler defined in `routes/app.tsx` after vite build. This was not the same behavior I observed when testing locally on my Mac.

I pulled down the generated `/_fresh` directory from github actions and instructed Claude to look to see if there were any differences in those files generated during the github actions build and those generated by running the same commands on my mac.

## From Claude Opus 4.6

```
@fresh/core's sortRoutePaths function has a bug on lines with return charA <
  charB ? 0 : 1 — when charA < charB, it returns 0 (equal) instead of -1 (a before b). This
  makes the comparator asymmetric: compare(a, b) can return 0 while compare(b, a) returns 1.

  On macOS, Deno.readDir returns files alphabetically, which happens to produce the correct
  final sort even with the broken comparator. On Linux (CI), Deno.readDir returns files in
  hash-table order. The stable TimSort algorithm + inconsistent comparator + different initial
  ordering = /app/_middleware ends up at position 112 instead of position 4, so the auth
  middleware never runs for /app routes.

  Fix: scripts/fix_fresh_route_order.ts re-sorts the fsRoutes array in
  _fresh/server/server-entry.mjs after the build using the correct comparator. It's wired into
  deno task build so it runs automatically on both local and CI.
  ```
  
The referenced workaround script [scripts/fix_fresh_route_order.ts](https://github.com/Virtual-Hospitals-Africa/virtual-hospitals-africa/blob/main/scripts/fix_fresh_route_order.ts) overwrites the route ordering in the generated `_fresh/server/server-entry.mjs` file after that has been built by `vite build`. While I have not evaluated the contents of that script line by line, I did scan it and it does seem to be doing what is suggested and I can confirm that it did fix the issue, suggesting that the overall analysis is sound.

Closes #3773

### DIFF
--- a/packages/fresh/src/fs_routes.ts
+++ b/packages/fresh/src/fs_routes.ts
@@ -247,7 +247,7 @@ export function sortRoutePaths(a: string, b: string) {
       if (scoreA === scoreB) {
         if (charA !== charB) {
           // TODO: Do we need localeSort here or is this good enough?
-          return charA < charB ? 0 : 1;
+          return charA < charB ? -1 : 1;
         }
         continue;
       }
@@ -257,7 +257,7 @@ export function sortRoutePaths(a: string, b: string) {
 
     if (charA !== charB) {
       // TODO: Do we need localeSort here or is this good enough?
-      return charA < charB ? 0 : 1;
+      return charA < charB ? -1 : 1;
     }
 
     // If we're at the end of A or B, then we assume that the longer
@@ -292,8 +292,9 @@ function getRoutePathScore(char: string, s: string, i: number): number {
   }
 
   if (
-    i + 4 === s.length - 1 && char === "i" && s[i + 1] === "n" &&
-    s[i + 2] === "d" && s[i + 3] === "e" && s[i + 4] === "x"
+    char === "i" && s[i + 1] === "n" &&
+    s[i + 2] === "d" && s[i + 3] === "e" && s[i + 4] === "x" &&
+    (i + 4 === s.length - 1 || s[i + 5] === ".")
   ) {
     return 3;
   }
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Thanks!

PR #3770, opened 2026-04-11
fix: handle commas in static file paths

SebF33:
## Problem

After #3659, static file handling correctly supports spaces and encoded paths,
but filenames containing commas are not resolved.

Example:
- File: `foo,bar.css`
- Stored in cache as: `/foo%2Cbar.css`
- Request: `/foo,bar.css`

This causes a mismatch and the static file is not found.

## Cause

`prepareStaticFile()` stores file paths using URL encoding, while
`static_files.ts` compares against `url.pathname` without applying the same
normalization.

## Solution

Normalize the incoming request pathname using the same encoding logic before
reading from the build cache.

## Tests

Added `static files - comma in pathname` to verify that both of these resolve:

- `/foo,bar.css`
- `/foo%2Cbar.css`


bartlomieju (approved):
Thanks!

PR #3775, opened 2026-04-22
feat: refresh landing page for Fresh 2.3

bartlomieju:
Screenshot 2026-04-22 at 20 12 23

- Update HelloBar to announce Fresh 2.3 with feature callouts (WebSockets, View Transitions, Temporal API)
- Rework hero section: new headline ("The framework so simple, you already know it"), descriptive subtitle, moved CTA to left
- Remove old Simple section, add "Take a tour of Fresh" heading before showcase
- Add new showcase sections: View Transitions, API Routes/Handlers, WebSockets, `<Head>` element, OpenTelemetry
- Fix broken Deco showcase image (missing `/showcase/` directory)
- Fix syntax highlighting in Simple section (was using raw `<pre>` instead of Prism)
- Update RenderingSection to use Temporal API instead of Date API
- Fix cursor styles on Forms and Partials interactive demos
- Move WebSocket docs from Examples to Advanced in TOC
- Densify all showcase sections: smaller headings, code blocks, icons, demo boxes, arrows, and tighter spacing


www/components/homepage/SocialProof.tsx
```diff
@@ -16,13 +16,15 @@ export function SocialProof() {
...
-        />
+      <a
+        href="https://deno.com/blog/deco-cx-subhosting-serve-their-clients-storefronts-fast"
+        class="block mx-auto text-center"
+      >
+        <div class="mx-auto max-w-screen-md bg-gradient-to-br from-green-50 to-yellow-50 border border-green-200 rounded-lg p-12 flex items-center justify-center">
+          <span class="text-4xl sm:text-5xl md:text-6xl font-extrabold text-green-700 tracking-tight">
+            deco.cx
+          </span>
+        </div>
```
josh-collinsworth: Looks like this isn't scaling up right on larger screens.

And in any case, I think it's better to use their actual logo, or screenshot, or something, rather than just typing the company's name in our font.

Image

www/routes/index.tsx
```diff
@@ -56,11 +60,18 @@ export default define.page<typeof handler>(function MainPage() {
...
         <IslandsSection />
         <FormsSection />
         <PartialsSection />
+        <TransitionsSection />
+        <APIRoutesSection />
+        <WebSocketSection />
+        <HeadSection />
+        <SecuritySection />
         <SocialProof />
         <DenoSection />
```
josh-collinsworth, line 68: I like all the new info we're giving people, but at a certain point, it becomes a little repetitive. It's a lot of similar content, kinda crammed together.

At the very least, I'd recommend adding some more breathing room between each feature. (On mobile especially, they feel pretty bunched together.) That alone would help a lot, and if that's all we have time for, I could live with that.

Ideally, though: I think we might want to consider which 3–4 of these features warrant star billing, and which ones should be presented in some kind of secondary subsection, rather than having nine nearly identical sections all in a row.

Image

Image

josh-collinsworth (changes requested):
See comments, but tl;dr: at the very least, we should fix the Deco logo issue and add a bit more space between these features.

josh-collinsworth (approved):
Made some visual changes to the MoreFeatures section, now LGTM

crowlKats (approved):

PR #3776, opened 2026-04-22
fix: keep ?fresh-partial an implementation detail

CertainLach:
When encountering redirect during route change with partials enabled, url in address bar (history api) is being left as ?fresh-partial=true, this doesn't seem right


bartlomieju (commented):
Could you add a test for it?

bartlomieju (approved):

PR #3779, opened 2026-04-24
feat: serve content-hashed chunks outside BUILD_ID path

bartlomieju:
## Summary

- Shared chunks and assets (WASM, etc.) produced by esbuild already have content hashes in their filenames (`chunk-XXXX.js`), but were served under `/_fresh/js/{BUILD_ID}/` — meaning every deploy invalidated all cached chunks, even unchanged ones
- Use esbuild's `chunkNames`/`assetNames` options to output shared chunks to `../c/`, which the builder maps to `/_fresh/js/c/` — outside the BUILD_ID directory
- Entry points (fresh-runtime, island modules) remain at `/_fresh/js/{BUILD_ID}/` since they don't have content hashes in their filenames
- The static files middleware now treats `/_fresh/js/c/` as immutable, so browsers cache these files across deploys

**Before:** `/_fresh/js/{BUILD_ID}/chunk-abc123.js` — URL changes every deploy even if content didn't change  
**After:** `/_fresh/js/c/chunk-abc123.js` — URL only changes when content changes

This is especially impactful for large assets like WASM files that rarely change but are expensive to re-download.

## Test plan

- [x] Existing static file tests pass
- [x] New test verifies immutable caching for `/_fresh/js/c/` paths (both JS chunks and WASM assets)
- [ ] Manual test with a project that has shared chunks/WASM to verify browser caching across deploys
- [ ] Verify source maps resolve correctly for relocated chunks

### DIFF
--- a/docs/latest/advanced/builder.md
+++ b/docs/latest/advanced/builder.md
@@ -48,8 +48,12 @@ const builder = new Builder({
   islandDir?: string;
   // Path to routes directory. (Default: `<root>/routes`)
   routeDir?: string;
-  // File paths which should be ignored 
+  // File paths which should be ignored
   ignore?: RegExp[];
+  // Glob patterns for static files that should use content-hash
+  // caching instead of build-ID. Useful for large assets like WASM
+  // that rarely change between deploys.
+  contentAddressedStatic?: string[];
   // Optionally generate production source maps
   // See https://esbuild.github.io/api/#source-maps
   sourceMap?: {
--- a/docs/latest/concepts/static-files.md
+++ b/docs/latest/concepts/static-files.md
@@ -122,6 +122,37 @@ export default function Gallery() {
 }
 ```
 
+## Content-addressed static files
+
+By default, `asset()` appends a cache-bust key based on the current build ID.
+This means **every deploy invalidates every cached static file**, even if the
+file content hasn't changed. For small files this is fine, but for large assets
+like WASM binaries, fonts, or media files, re-downloading unchanged files on
+every deploy is wasteful.
+
+The `contentAddressedStatic` option lets you specify glob patterns for files
+that should use their **content hash** as the cache-bust key instead of the
+build ID. The URL only changes when the file content changes — surviving
+deploys unchanged.
+
+```ts dev.ts
+import { Builder } from "fresh/dev";
+
+const builder = new Builder({
+  contentAddressedStatic: ["**/*.wasm", "**/*.bin"],
+});
+```
+
+With this config, `asset("/module.wasm")` produces a URL like
+`/module.wasm?__frsh_c=<content-hash>` instead of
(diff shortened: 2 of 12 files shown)


bartlomieju (commented):
**Overall**: Clean PR, well-structured. Two distinct but related features: (1) esbuild shared chunks always go to `/_fresh/js/c/` (automatic), (2) user static files opt in via `contentAddressedStatic` globs. Good test coverage and docs.

**Issues:**

1. **Glob matching inconsistency** — In `dev_build_cache.ts:363`, `isContentAddressed(name)` is called where `name` is a URL-style pathname (e.g., `/large.wasm`). In `server_snapshot.ts:284`, `isContentAddressed(relative)` uses a filesystem-relative path (e.g., `large.wasm`, no leading slash). The same user-provided glob pattern like `**/*.wasm` may match one but not the other depending on how `globToRegExp` handles the leading `/`. Worth adding a test that exercises both paths with the same pattern, or normalizing the input before matching.

2. **Source maps for relocated chunks** — The test plan has "Verify source maps resolve correctly for relocated chunks" unchecked. Since chunks move from `/_fresh/js/{BUILD_ID}/` to `/_fresh/js/c/`, the `//# sourceMappingURL=` comment esbuild emits is relative. If the `.map` file also lands under `../c/` it should be fine, but this is worth verifying before merge — broken source maps in production are painful to debug after the fact.

3. **`file.hash !== cacheKey` broadens valid cache keys** (`static_files.ts:68`) — Before this change, only `BUILD_ID` was accepted as a cache key. Now any file's content hash is also accepted. This is correct for content-addressed files, but it also applies to *all* static files — if a request comes in with a matching hash for a non-content-addressed file, it'll get immutable caching too. This is probably fine (correct hash = correct content), but it's a subtle behavioral change worth calling out in the PR description.

**Nits:**

4. The `|| undefined` pattern in `dev_build_cache.ts:366` and `server_snapshot.ts:322` (`immutable: isContentAddressed(name) || undefined`) is a bit subtle — a short comment like `// omit false to keep default behavior` would help.

5. The `"../c/"` string appears in both `esbuild.ts` and `builder.ts` — consider extracting it as a shared constant to keep them in sync.

CertainLach (approved):

lunadogbot (changes requested):
CI's been red since April 24 on a fixable type error: `packages/fresh/src/dev/dev_build_cache_test.ts:16` declares a `ResolvedBuildConfig` literal that wasn't updated when the type was extended. Adding `contentAddressedStatic: []` to that fixture turns the type check green:

```
TS2322 [ERROR]: Type '{ serverEntry: string; ... target: string; }' is not assignable to type 'ResolvedBuildConfig'.
    const config: ResolvedBuildConfig = {
    at packages/fresh/src/dev/dev_build_cache_test.ts:16:11
```

Otherwise no new issues beyond the ones you already flagged in the self-review.

PR #3781, opened 2026-04-25
fix: CSS modules not working in _app/_layout/_error and across routes in non-island components

Hajime-san:
Originally reported:
- https://github.com/denoland/fresh/issues/3633#issuecomment-3556127762

Repro:
https://github.com/honmanyau/fresh-2-layout-css-module-issue/blob/24bf9c597a057c60acdb2c2934a65f22b6276ecd/routes/page-b/_layout.tsx

The PR seems to resolve the island components problem, but not non-island.
- https://github.com/denoland/fresh/pull/3764

This PR introduces CSS Modules handling for non-island components rendered through utility files such as `_app`, `_layout`, `_error`, and `_404`. The core change is that utility-file `mod.css` is now preserved through Fresh’s command and render pipeline, so CSS referenced from app/layout/route/error rendering can be accumulated and emitted in the final response instead of being dropped.

~~On the Vite side, development now also collects CSS exposed through `fresh-route-css::*` virtual modules, which is the path used by server-rendered non-island components in utility files. For production builds, CSS placeholder replacement is no longer limited to route chunks, because this CSS can be hoisted into shared chunks such as `server-entry`.~~

I added new tests and confirmed to work fine about our private project via `links` in deno config locally.


packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -520,19 +523,23 @@ export default ${JSON.stringify(route.css)}
...
+
+            const filePath = path.join(serverOutDir, info.file);
+            const content = await Deno.readTextFile(filePath);
+            if (!content.includes(`["__FRESH_CSS_PLACEHOLDER__"]`)) continue;
+
+            const replaced = content.replace(
+              `["__FRESH_CSS_PLACEHOLDER__"]`,
+              info.css
+                ? JSON.stringify(info.css.map((css) => `/${css}`))
+                : "null",
```
bartlomieju: This uses `content.replace()` which only replaces the **first** occurrence of the placeholder. If two utility-file CSS placeholders (e.g. from both `_app` and `_layout`) get hoisted into the same shared chunk like `server-entry`, only the first one would be replaced.

Should this be `replaceAll` (or a loop) to be safe?

Alternatively, if each route virtual module gets its own unique placeholder, this is fine — but it might be worth a comment explaining why a single replace is sufficient.

packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -501,6 +500,10 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
             return `export default ["__FRESH_CSS_PLACEHOLDER__"];`;
           }
 
+          route.css = server === undefined
+            ? route.css
+            : await collectRouteCss(server, route.filePath);
```
bartlomieju: This re-walks the full SSR import graph and overwrites `route.css` on every `load` call (including HMR invalidations). That seems intentional for picking up changes during dev, but for deep dependency trees the `fetchModule` calls inside `collectRouteCss` could add up.

Might be worth a short comment here noting this is intentional for HMR correctness, so a future reader doesn't try to "optimize" it with caching.

packages/fresh/src/context.ts
```diff
@@ -298,16 +321,18 @@ export class Context<State> {
```
bartlomieju, line 320: Nit: `appDef?.css` evaluates to `undefined` when `appDef` is `null`, which the new `setAdditionalStyles` handles via the `css == null` guard. Works correctly, but it's a subtle null-propagation chain. An explicit guard might be clearer:

```ts
if (appDef !== null) {
  setAdditionalStyles(this, appDef.css);
}
```

Especially since the `appDef !== null` checks already exist a few lines below.

bartlomieju (commented):
Nice fix — the approach of threading `css` through the existing command → segment → context pipeline is clean and consistent with the rest of the architecture. A few things caught my eye below (inline), but overall this looks solid.

The `setAdditionalStyles` refactor from a simple setter to a dedup-accumulator is exactly right, `getSsrModule()` properly handles both real files and virtual module IDs, and the BFS graph walk in `collectRouteCss` is thorough.

Let me know if you'd like a hand getting this over the finish line — happy to help with any of the items below or with testing.

packages/fresh/src/context.ts
```diff
@@ -182,8 +188,24 @@ export class Context<State> {
...
-      ctx.#additionalStyles = css;
+    setAdditionalStyles = <T>(
+      ctx: Context<T>,
+      css: string[] | null | undefined,
+    ) => {
+      if (css == null) return;
+
+      if (ctx.#additionalStyles === null) {
+        ctx.#additionalStyles = css.slice();
+        return;
```
bartlomieju, line 199: Minor: `includes()` is O(n) per insertion. For the small CSS lists in practice this is fine, but a `Set` would be more idiomatic if you ever want to tighten this up. Low priority.

packages/plugin-vite/tests/build_test.ts
```diff
@@ -388,6 +388,28 @@ integrationTest(
...
+
+          const color = await page
+            .locator(".green > h1")
+            // deno-lint-ignore no-explicit-any
+            .evaluate((el) => window.getComputedStyle(el as any).color);
+          expect(color).toEqual("rgb(0, 128, 0)");
+        });
+      },
+    );
+  },
```
bartlomieju, line 469: The tests cover `_layout.tsx` which is great. The PR title also mentions `_app`, `_error`, and `_404` — are those paths already exercised by existing tests, or would it be worth adding a case or two here to lock in coverage for those as well?

Hajime-san:
@bartlomieju 
Thanks!
Upon investigation, I found that our project was exhibiting behavior where the screen was not rendering in dev mode.
Also, it appears that CI is crashing only on Windows. I will re-request review once the items already reviewed and the above problems are resolved.

UPDATE:
>  the screen was not rendering in dev mode.

It seems an another problem of `links`. It works fine with `vendor` and copy-pasting locally.

packages/plugin-vite/src/plugins/server_snapshot.ts
Hajime-san: Nice catch. I notice that it causes requesting to `http://localhost:8000/__FRESH_CSS_PLACEHOLDER__` in build mode. It should be `replaceAll`. Fixed and added a test.

packages/fresh/src/dev/fs_crawl.ts
```diff
@@ -86,7 +86,7 @@ export async function crawlRouteDir<State>(
 
     files.push({
       id,
-      filePath: entry.path,
+      filePath: toPosix(entry.path),
```
Hajime-san, line 89: Suddenly, several tests failed on Windows. It's likely that a potential bug was occured by this PR.

```bash
vite dev - css modules => ./packages/fresh/src/test_utils.ts:198:8
vite dev - css modules in _app/_layout/_error non-island component are injected => ./packages/fresh/src/test_utils.ts:198:8
vite dev - route css import => ./packages/fresh/src/test_utils.ts:198:8
```

<details><summary>debug console on Windows</summary>

```bash
------- post-test output -------
👺 > fresh:route-css > name _index
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
  </head>
  <body>
    <div class="green">
      <h1 class="_root_1mdiz_1">
        green text
      </h1>
    </div>
    <div class="red">
      <h1 class="_root_xlv1v_1">
        red text
      </h1>
    </div>
    <h1>
      ok
    </h1>
  </body>
</html>

[
  "Port 5173 is in use, trying another one...",
  "",
  "  VITE v7.3.1  ready in 1276 ms",
  "",
  "  ➜  Local:   http://127.0.0.1:5174/"
]
[
  "Port 5173 is in use, trying another one...",
  "",
  "  VITE v7.3.1  ready in 1276 ms",
  "",
  "  ➜  Local:   http://127.0.0.1:5174/",
  "\x1b[2m  \x1b[32m➜\x1b[39m  \x1b[1mNetwork\x1b[22m\x1b[2m: use \x1b[22m\x1b[1m--host\x1b[22m\x1b[2m to expose\x1b[22m",
  "👺 > fresh:route-css > route {",
  '  id: "/_app",',
  '  filePath: "D:\\\\a\\\\fresh\\\\fresh\\\\packages\\\\plugin-vite\\\\tests\\\\fixtures\\\\non_island_css_modules\\\\routes\\\\_app.tsx",',
  '  type: "app",',
  '  pattern: "*",',
  '  routePattern: "*",',
  "  lazy: false,",
  "  css: [",
  '    "/@fs/D:/a/fresh/fresh/packages/plugin-vite/demo/components/CssModulesNonIsland.module.css"',
  "  ],",
  "  overrideConfig: undefined",
  "}",
  "👺 > fresh:route-css > name __app",
  "👺 > fresh:route-css > route {",
  '  id: "/_layout",',
  '  filePath: "D:\\\\a\\\\fresh\\\\fresh\\\\packages\\\\plugin-vite\\\\tests\\\\fixtures\\\\non_island_css_modules\\\\routes\\\\_layout.tsx",',
  '  type: "layout",',
  '  pattern: "/",',
  '  routePattern: "/",',
  "  lazy: false,",
  "  css: [",
  '    "/@fs/D:/a/fresh/fresh/packages/plugin-vite/demo/components/CssModulesNonIsland3.module.css"',
  "  ],",
  "  overrideConfig: undefined",
  "}",
  "👺 > fresh:route-css > name __layout",
  "👺 > fresh:route-css > route {",
  '  id: "/_error",',
  '  filePath: "D:\\\\a\\\\fresh\\\\fresh\\\\packages\\\\plugin-vite\\\\tests\\\\fixtures\\\\non_island_css_modules\\\\routes\\\\_error.tsx",',
  '  type: "error",',
  '  pattern: "/",',
  '  routePattern: "/",',
  "  lazy: false,",
  "  css: [",
  '    "/@fs/D:/a/fresh/fresh/packages/plugin-vite/demo/components/CssModulesNonIsland2.module.css"',
  "  ],",
  "  overrideConfig: undefined",
  "}",
  "👺 > fresh:route-css > name __error",
  "👺 > fresh:route-css > route {",
  '  id: "/index",',
  '  filePath: "D:\\\\a\\\\fresh\\\\fresh\\\\packages\\\\plugin-vite\\\\tests\\\\fixtures\\\\non_island_css_modules\\\\routes\\\\index.tsx",',
  '  type: "route",',
  '  pattern: "/",',
  '  routePattern: "/",',
  "  lazy: true,",
  "  css: [],",
  '  overrideConfig: { methods: "ALL" }',
  "}",
  "👺 > fresh:route-css > name _index",
  "👺 > renderRoute > setAdditionalStyles []",
  "👺 > render Compose VNode > setAdditionalStyles []",
  "👺 > render compose App > setAdditionalStyles []"
]
./packages/plugin-vite/tests/dev_server_test.ts => vite dev - css modules in _app/_layout/_error non-island component are injected ...----- post-test output end -----
./packages/plugin-vite/tests/dev_server_test.ts => vite dev - css modules in _app/_layout/_error non-island component are injected ... FAILED (5s)
./packages/plugin-vite/tests/build_test.ts => vite build - tailwind _app ... ok (6s)
./packages/plugin-vite/tests/build_test.ts => vite build - partial island ... ok (1s)
------- post-test output -------
```
</details>

packages/fresh/src/context.ts
Hajime-san: Yep, I added a FIXME comment below. I think it could be use `Set` instead of `css: Array<string>`.It seems that at least all the tests pass in the following branch.

https://github.com/Hajime-san/fresh/compare/fix-ssr-css-modules...Hajime-san:fresh:perf-css

I'm planning it as a follow up PR.

Hajime-san:
I can't make out why the test fails...

```bash
...
vite dev - source mapped stack traces => ./packages/fresh/src/test_utils.ts:202:8

FAILED | 558 passed (14 steps) | 1 failed | 14 ignored (4m11s)
```

It seems to be flaky because all test passes on my local macOS with `deno test -A`, but it fails with `% deno test -A --filter='source mapped stack traces'`.

packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -158,7 +161,7 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
             const route = result.routes[i];
             const name = routeNamer.getUniqueName(route.id);
 
-            routeFileToName.set(route.filePath, name);
+            routeFileToName.set(toPosix(route.filePath), name);
```
Hajime-san, line 175: seems to be same as https://github.com/denoland/fresh/pull/3781/changes#r3143377420

packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -416,7 +419,7 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
             let item: EnvironmentModuleNode | undefined;
             while ((item = queue.pop()) !== undefined) {
               if (item.file !== null) {
-                const normalized = path.normalize(item.file);
+                const normalized = toPosix(path.normalize(item.file));
```
Hajime-san, line 433: ditto

packages/plugin-vite/src/plugins/server_snapshot.ts
Hajime-san: Sorry for confusing, it seems to be my AI slop code. This code is unnecessary for HMR.
I'm not sure if the HMR is working with the changes in this PR. Since it definitely works after a page reload, I'm thinking of putting it out of scope for now.

lunadogbot (approved):
Reviewed the diff — all four substantive points from the earlier review landed on `493f81a`: `replaceAll` at `server_snapshot.ts:536` with a regression test in `build_test.ts:392` covering `_app`/`_layout`/`_error`; explicit `appDef !== null` guard at `context.ts:326`; the over-eager HMR re-walk was removed in `948dbb94`; the `includes()` dedup at `context.ts:205` carries a `FIXME` with a `Set`-based follow-up.

One thing worth keeping in mind: every route's virtual `fresh-route-css::*` module emits the same `["__FRESH_CSS_PLACEHOLDER__"]` literal, so when multiple utility files hoist into a shared chunk, `replaceAll` writes the same combined `info.css` into each placeholder. That's correct today since `setAdditionalStyles` dedups on the render side — but if a future chunk-splitting strategy changes the "every placeholder in a chunk wants the same css list" assumption, this will need a rethink.

CI green on `493f81a`.

bartlomieju:
I feel like this placeholder situation should have better test coverage

packages/fresh/src/context.ts
```diff
@@ -182,8 +188,25 @@ export class Context<State> {
...
+      }
+
+      for (let i = 0; i < css.length; i++) {
+        const href = css[i];
+        // FIXME: consider to use `Set` instead of `css: string[]` for entire codebase
+        if (!ctx.#additionalStyles.includes(href)) {
+          ctx.#additionalStyles.push(href);
+        }
+      }
```
bartlomieju: Two small things on this loop:

1. The `// FIXME: consider to use Set ...` would be better as a tracked issue (or dropped) — inline FIXMEs tend to rot. For typical per-request CSS counts the `Array.includes` dedup is fine in practice, so I'd lean toward just removing the comment.
2. `for (const href of css)` would read a bit cleaner than the indexed loop, since the index isn't used.

packages/plugin-vite/tests/fixtures/non_island_css_modules/components/CssModuleNonIsland2.tsx
```diff
@@ -0,0 +1,9 @@
+import styles from "./CssModulesNonIsland2.module.css";
```
bartlomieju, line 1: These three new `CssModuleNonIsland{2,3,4}.tsx` (+ `.module.css`) files appear to be imported only by the new test fixture via `../../../../demo/components/...`. If that's the case, colocating them under `tests/fixtures/non_island_css_modules/` would keep the demo package leaner and make the fixture self-contained.

bartlomieju (commented):
Thanks for the thorough fix — especially the dev + prod integration tests and the unit test for `replaceFreshCssPlaceholders`. The expanded placeholder replacement (all JS chunks containing `FRESH_CSS_PLACEHOLDER` rather than just `_fresh-route___*`) is the right shape for hoisted CSS, and the prod test asserting no leftover placeholders in any `.mjs` is a solid regression guard. Same for the `toPosix` normalization on both insert and lookup sides.

A couple of things worth double-checking before merging:

- `UiTree.app` and `UiTree.layouts[]` changed shape (now `{ component, css }` wrappers). It's behind `getInternals`/the internal symbol so blast radius should be small, but worth a grep across the monorepo to be sure nothing external reads `internals.app` as a component.
- The semantics of `setAdditionalStyles` changed from "replace" to "accumulate + dedup". Looks correct given the new call sites and per-request `#additionalStyles`, but worth confirming no remaining caller depended on the old replace behavior.
- `_404`: the diff threads `css` into `newNotFoundCmd`, but I didn't see the notFound pipeline pick it up explicitly. The `/non_existent` integration test passes so it presumably flows through `renderRoute` → `setAdditionalStyles(ctx, route.css)` — just calling it out to confirm.

A few smaller suggestions inline.

bartlomieju (approved):

PR #3783, opened 2026-04-25
feat: add client-only islands support

bartlomieju:
## Summary

- Adds support for **client-only islands** via `export const clientOnly = true` in island files
- Islands marked as client-only render an empty placeholder `<div>` during SSR instead of executing the component, avoiding crashes from libraries that reference browser globals (e.g. `document`) at the module top level
- On the client, the component renders normally via Preact's `render()`
- Also includes the earlier refactor replacing Babel env var transforms with Vite's native `define` and removing the CJS→ESM Babel transform

## Test plan

- [x] New SSR test verifies placeholder is rendered and `:c` marker flag is present
- [x] New browser test verifies the island renders correctly on the client
- [x] All 27 existing island tests continue to pass
- [x] `no_client_js` and `head` test suites pass


lunadogbot (changes requested):
Feature core is small and clean: SSR renders `<div></div>` + a `:c` marker (`packages/fresh/src/runtime/server/preact_hooks.ts:300-314`), the reviver picks it up, and the existing `render()` path populates the empty container. Fixture + tests cover both SSR (placeholder + marker) and client (component runs).

Two things to address:

1. **`IslandReq.clientOnly` is dead.** `packages/fresh/src/runtime/client/reviver.ts:24,273-279` parses the `:c` flag and stores it on the request, but no client path reads it — `revive()` at `reviver.ts:115` always calls `render()` unconditionally. The feature works "by accident" because rendering into the empty `<div>` patches it to the component output. Either drop the field and the `:c` marker variant (the empty SSR `<div>` alone is enough; the existing path handles it), or wire `clientOnly` through to a code path that actually branches on it.

2. **Scope.** Title is `feat: add client-only islands support` but the diff also deletes the CJS→ESM and inline-env-vars Babel patches (~1950 LOC across `packages/plugin-vite/src/plugins/patches/commonjs.ts`, `commonjs_test.ts`, `inline_env_vars.ts`, `inline_env_vars_test.ts`) and rewrites `packages/plugin-vite/src/mod.ts` + `plugins/deno.ts` to use Vite `define` + `ssr.noExternal`. The CJS transform was just improved in #3697 (`d6b0a16`); reversing it is a strategy change that deserves its own PR and its own title. Reviewed standalone, the islands change is ~150 LOC.

- nit: `deno.lock` rolls `preact` 10.29.1 → 10.29.0 (and `@preact/signals` / `preact-render-to-string` follow). Not a deps PR; looks accidental.

PR #3790, opened 2026-04-27
feat: add link prefetching support

### LINKED ISSUE #3788: feat: add link prefetching support
## Problem

Fresh supports client-side navigation via `f-client-nav` and Partials, but there's no built-in mechanism for prefetching upcoming navigations. Frameworks like Next.js (`<Link prefetch>`), Remix (`<Link prefetch="intent">`), and Astro (`prefetch`) all provide this, and it makes a noticeable difference in perceived performance.

Currently, the only workaround is manually adding `<link rel="prefetch" href="...">` inside `<Head>`, which is static and doesn't respond to user intent.

## Proposed Solution

Add prefetching support, potentially via one or more of these approaches:

### 1. `prefetch` attribute on `<a>` tags (within `f-client-nav`)

Extend the client-side navigation system to support a `f-prefetch` attribute:

```html
<nav f-client-nav>
  <a href="/about" f-prefetch>About</a>
  <a href="/dashboard" f-prefetch="hover">Dashboard</a>
  <a href="/settings" f-prefetch="viewport">Settings</a>
</nav>
```

Strategies:
- **`hover`** (default) — prefetch when the user hovers over the link (similar to [instant.page](https://instant.page/))
- **`viewport`** — prefetch when the link enters the viewport (via IntersectionObserver)
- **`load`** — prefetch immediately on page load
- **`none`** — opt out (useful when a parent sets a default)

### 2. Container-level default

Allow setting a default prefetch strategy on the `f-client-nav` container:

```html
<nav f-client-nav f-prefetch="hover">
  <!-- All links in here prefetch on hover by default -->
  <a href="/about">About</a>
  <a href="/heavy-page" f-prefetch="none">Heavy Page</a>
</nav>
```

### Implementation considerations

- Prefetched responses should work with the Partials system — prefetch the partial HTML, not the full page
- Use `<link rel="prefetch">` injection or `fetch()` with low priority depending on browser support
- Respect `Save-Data` header / `navigator.connection.saveData`
- Deduplicate prefetch requests (don't re-prefetch already cached URLs)
- Consider a cache/TTL for prefetched content to avoid serving stale partials
- Should integrate with the `cache()` middleware (#3782) — prefetched responses benefit from server-side caching

## Alternatives considered

- **Manual `<Head>` injection** — works but static, no intent-based prefetching
- **Third-party script (instant.page)** — works for full page loads but doesn't integrate with Fresh's Partial system

bartlomieju:
## Summary

- Adds `f-prefetch` attribute for prefetching partial navigations ahead of time
- Supports four strategies: `hover` (default), `viewport`, `load`, `none`
- Container-level defaults via `f-prefetch` on parent elements (e.g. `f-client-nav` containers)
- Integrates with the existing partials system — prefetches the partial HTML, not full pages
- Respects `navigator.connection.saveData` and slow connection types
- Deduplicates in-flight requests and caches responses with 30s TTL
- Uses `MutationObserver` to handle dynamically added links

## Usage

```html
<nav f-client-nav f-prefetch="hover">
  <!-- All links prefetch on hover by default -->
  <a href="/about">About</a>
  <a href="/dashboard" f-prefetch="viewport">Dashboard</a>
  <a href="/heavy-page" f-prefetch="none">Heavy Page</a>
</nav>
```

## Test plan

- [x] All existing 120 tests pass
- [ ] Manual testing with `f-prefetch` attribute on links
- [ ] Verify hover strategy triggers fetch on pointerenter
- [ ] Verify viewport strategy triggers fetch when link scrolls into view
- [ ] Verify load strategy triggers fetch immediately
- [ ] Verify none strategy prevents prefetching
- [ ] Verify container-level defaults propagate to child links
- [ ] Verify Save-Data is respected
- [ ] Verify cache TTL expiry works correctly
- [ ] Verify cached response is consumed on navigation (no double fetch)

Closes #3788

### DIFF
--- a/packages/fresh/src/runtime/client/mod.ts
+++ b/packages/fresh/src/runtime/client/mod.ts
@@ -1,5 +1,6 @@
 import "./polyfills.ts";
 import "./preact_hooks_client.ts";
 import "./partials.ts";
+import "./prefetch.ts";
 export { asset, IS_BROWSER, Partial, type PartialProps } from "../shared.ts";
 export { boot, revive } from "./reviver.ts";
--- a/packages/fresh/src/runtime/client/partials.ts
+++ b/packages/fresh/src/runtime/client/partials.ts
@@ -23,6 +23,7 @@ import { createRootFragment, isCommentNode, isElementNode } from "./reviver.ts";
 import type { PartialStateJson } from "../server/preact_hooks.ts";
 import { parse } from "../../jsonify/parse.ts";
 import { INTERNAL_PREFIX, PARTIAL_SEARCH_PARAM } from "../../constants.ts";
+import { getCachedResponse } from "./prefetch.ts";
 
 export const PARTIAL_ATTR = "f-partial";
 
@@ -359,7 +360,15 @@ async function fetchPartials(
   init.redirect = "follow";
   partialUrl = new URL(partialUrl);
   partialUrl.searchParams.set(PARTIAL_SEARCH_PARAM, "true");
-  const res = await fetch(partialUrl, init);
+
+  // Check prefetch cache for GET requests (no custom init body)
+  let res: Response;
+  const cached = !init.body ? getCachedResponse(partialUrl) : null;
+  if (cached) {
+    res = cached;
+  } else {
+    res = await fetch(partialUrl, init);
+  }
 
   if (res.redirected) {
     const nextUrl = new URL(res.url);
--- a/packages/fresh/src/runtime/client/prefetch.ts
+++ b/packages/fresh/src/runtime/client/prefetch.ts
@@ -1,0 +1,239 @@
+import { PARTIAL_SEARCH_PARAM } from "../../constants.ts";
+import { CLIENT_NAV_ATTR } from "../shared_internal.ts";
+import { PARTIAL_ATTR } from "./partials.ts";
+
+export const PREFETCH_ATTR = "f-prefetch";
+
+type PrefetchStrategy = "hover" | "viewport" | "load" | "none";
+
+interface CacheEntry {
+  response: Response;
+  timestamp: number;
+}
+
+const CACHE_TTL = 30_000; // 30 seconds
+const prefetchCache = new Map<string, CacheEntry>();
+const inflightRequests = new Map<string, Promise<Response>>();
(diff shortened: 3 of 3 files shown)


lunadogbot (changes requested):
Strategy reads well — `closest()` for container defaults, `IntersectionObserver` for viewport, `MutationObserver` for dynamic links, `navigator.connection.saveData` + slow-type opt-out, single-shot cache consumption on click.

One blocker:

1. **In-flight race in `getCachedResponse`**: the function is sync and only checks `prefetchCache`. A user who clicks on a `hover`-prefetched link before its fetch resolves hits the `inflightRequests` window — `prefetchCache.get(url)` returns undefined, so `partials.ts:fetchPartials` falls through and issues a second `fetch(partialUrl, init)`. Two concurrent requests for the same URL is the exact case the inflight map exists to prevent. The fix is to await the inflight promise:

   ```ts
   export async function getCachedResponse(partialUrl: URL): Promise<Response | null> {
     const url = partialUrl.href;
     const inflight = inflightRequests.get(url);
     if (inflight) await inflight;
     // ...existing TTL check + delete-and-return
   }
   ```

   The call site at `partials.ts:363` already lives inside an async function, so `await getCachedResponse(partialUrl)` is a one-character change.

- nit: no tests for the new 239-line module. The eligibility rules (origin, target, fragment-only hrefs), strategy resolution (element vs. container vs. nested overrides), TTL expiry, and single-use consumption are all unit-testable with `jsdom`/a stubbed `IntersectionObserver`.
- nit: `prefetchedUrls.add` followed by `prefetchCache.has(url)` check at `prefetch.ts:79-82` is unreachable in practice — anything in `prefetchCache` was added by a prior `prefetch()` call that also added to `prefetchedUrls`, so the earlier `prefetchedUrls.has(url)` guard already returned. The TTL-staleness branch you intended to hit will never fire.
- nit: \`getStrategy\` returns \`\"none\"\` when neither the element nor any ancestor has \`f-prefetch\`. With the current setup, that means prefetch is off by default unless opt-in — worth a brief doc note since the PR description shows it as default-hover within an \`f-prefetch=\"hover\"\` container.

PR #3795, opened 2026-04-28
docs: remove claim that node_modules is not necessary

tredondo:
This was true for Fresh 1.x. `deno run -Ar jsr:@fresh/init` creates `node_modules`.

### DIFF
--- a/www/components/homepage/Hero.tsx
+++ b/www/components/homepage/Hero.tsx
@@ -14,7 +14,7 @@ export function Hero() {
               The framework so simple, you already know it.
             </h2>
             <p class="mt-6 text-lg sm:text-xl text-gray-700 text-balance max-w-prose">
-              No config files, no build step, no node_modules. Just one file and
+              No config files, no build step. Just one file and
               you have a server with routing, JSX, and islands.
             </p>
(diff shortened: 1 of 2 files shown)


CertainLach:
Legacy builder works without node_modules, so is https://github.com/CertainLach/freshpack

Fresh itself supports working without them, the only problem is vite tooling during the compilation

tredondo:
Since Vite is required, the "no node_modules" claim feels disingenuous now.

Also, most plugins (starting with tailwind) required node_modules.

lunadogbot (commented):
The change is accurate — the init template sets `nodeModulesDir: "manual"` in the generated `deno.json` (`packages/init/src/init.ts:563`), so a `node_modules/` folder is created on first run. The marketing copy claiming "no node_modules" is misleading.

Wording reads naturally with the phrase removed. Thanks!

PR title (`docs: …`) and scope are good; lockfile untouched. Leaving the merge to a core team member.

lunadogbot (changes requested):
`deno fmt --check` fails on Ubuntu — removing \", no node_modules\" leaves the lines short enough that fmt wants to rewrap both `www/components/homepage/Hero.tsx:17-18` and `www/components/homepage/Simple.tsx:47-48`. Running `deno fmt` locally will fix it.

PR #3809, opened 2026-05-14
fix: gate fresh:reload on SSR-relevant changes (plugin-vite)

### LINKED ISSUE #3637: Fresh 2 + Vite keeps rebuilding on Windows
Fresh 2 + Vite keeps rebuilding on Windows, when some tempory files are changed and I cannot disable it:

```ts
import tailwindcss from "@tailwindcss/vite";

export default defineConfig({
  plugins: [
    fresh(),
    tailwindcss(),
  ],
  server: {
    watch: {
      // Don't watch test files, temp files, or hidden files during development
      // to prevent dev server crashes
      ignored: [
        "**\\*_test.ts*",
        "**/*_test.ts*",
        "**\\*.test.ts*",
        "**/*.test.ts*",
        "**\\*.tmp.*",
        "**/*.tmp.*",
        "**\\*.timestamp-*",
        "**/*.timestamp-*",
        "**\\.*",
        "**/.*",
        "**\\encrypted\\**",
        "**/encrypted/**",
        "**\\kv\\**",
        "**/kv/**",
        "vite.config.*",
        "encrypted\\kv\\main-shm",
        "encrypted/kv/main-shm",
        "encrypted\\kv\\main-wal",
        "encrypted/kv/main-wal",
        "node_modules\\**",
        "node_modules/**",
        "**/*",
        "**\\*",
        "**",
        "*",
      ],
    },
  },
});
```

I also tried:

```ts
import tailwindcss from "@tailwindcss/vite";

export default defineConfig({
  plugins: [
    fresh(),
    tailwindcss(),
  ],
  server: {
    watch: null,
  },
});
```

I got it working on wsl, and I've tried with both path type \\ and /, but it keeps rebuilding.

fibibot:
## Summary

Fixes #3637.

`serverSnapshot` attaches a `viteServer.watcher.on(\"all\", ...)` handler that runs alongside Vite's own watcher. After the island-file early-return and the route add/unlink branch, it was calling `viteServer.ws.send(\"fresh:reload\")` unconditionally — regardless of whether the changed file had anything to do with SSR.

Because Fresh attaches directly to Vite's chokidar instance, the user's `server.watch.ignored` config does not filter events before Fresh sees them. On Windows, journaling and delayed-close semantics mean files like Deno KV's `main-shm`/`main-wal`, `*.tmp.*`, and `.timestamp-*` emit watcher events on every write, so each KV write reloaded the page.

This change gates `fresh:reload` so it only fires when:

- the changed file is in the SSR module graph (`ssr.moduleGraph.getModulesByFile(filePath)` non-empty), or
- the changed file is a route file under `options.routeDir`.

Files unknown to either are ignored: Vite's own HMR pipeline already handles client-graph files (the island early-return below covers islands; `dev_server.ts`'s `hotUpdate` covers SSR modules through Vite's normal HMR path which does honour `server.watch.ignored`).

The route add/unlink branch that invalidates the snapshot is preserved.

## Test plan

- [x] New regression test `vite dev - unrelated file changes do not trigger fresh:reload` writes a file (`main-shm`) outside `routes/` and the SSR module graph, then asserts no `fresh:reload` message arrives on the Vite HMR WebSocket, and finally confirms that editing the route file still produces `fresh:reload` so the gate isn't too aggressive.
- [x] Verified the test fails on the pre-fix code (reproduces the bug) and passes after the fix.
- [x] `deno fmt --check`, `deno lint`, `deno task check:types`, and `deno test -A packages/plugin-vite/tests/` (73/73 passing) locally.

Closes bartlomieju/orchid-inbox#64

### DIFF
--- a/packages/plugin-vite/src/plugins/server_snapshot.ts
+++ b/packages/plugin-vite/src/plugins/server_snapshot.ts
@@ -98,28 +98,37 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
             }
           }
 
+          const relRoutes = path.relative(options.routeDir, filePath);
+          const isRouteFile = !relRoutes.startsWith("..") &&
+            !path.isAbsolute(relRoutes) &&
+            !/[\\/]+\(_[^)]+\)[\\/]+/.test(filePath);
+
           // Check for route files. We need to invalidate the snapshot if
           // they are removed or added.
-          if (
-            (ev === "add" || ev === "unlink") &&
-            !/[\\/]+\(_[^)]+\)[\\/]+/.test(filePath)
-          ) {
-            const relRoutes = path.relative(options.routeDir, filePath);
-            if (!relRoutes.startsWith("..")) {
-              const mod = ssr.moduleGraph.getModuleById(`\0${modName}`);
-              if (mod !== undefined) {
-                // Clear state
-                islands.clear();
-                islandsByFile.clear();
-                islandSpecByName.clear();
+          if ((ev === "add" || ev === "unlink") && isRouteFile) {
+            const mod = ssr.moduleGraph.getModuleById(`\0${modName}`);
+            if (mod !== undefined) {
+              // Clear state
+              islands.clear();
+              islandsByFile.clear();
+              islandSpecByName.clear();
 
-                ssr.moduleGraph.invalidateModule(mod);
-              }
+              ssr.moduleGraph.invalidateModule(mod);
             }
           }
 
-          // Finally, notify the client
-          viteServer.ws.send("fresh:reload");
+          // Only notify the client when the changed file is actually
+          // relevant to SSR — either a route file, or something already
+          // tracked in the SSR module graph. Vite attaches us directly to
+          // its chokidar instance, so `server.watch.ignored` does not
+          // filter events before we see them. Without this gate, any
+          // unrelated write (e.g. Deno KV `main-shm`/`main-wal` journals
+          // or editor temp files on Windows) would trigger a full reload.
+          // See https://github.com/denoland/fresh/issues/3637.
+          const ssrMods = ssr.moduleGraph.getModulesByFile(filePath);
+          if (isRouteFile || (ssrMods !== undefined && ssrMods.size > 0)) {
+            viteServer.ws.send("fresh:reload");
+          }
         });
       },
       resolveId: {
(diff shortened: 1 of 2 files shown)


packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -98,28 +98,37 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
...
             }
           }
 
-          // Finally, notify the client
-          viteServer.ws.send("fresh:reload");
+          // Only notify the client when the changed file is actually
+          // relevant to SSR — either a route file, or something already
+          // tracked in the SSR module graph. Vite attaches us directly to
+          // its chokidar instance, so `server.watch.ignored` does not
+          // filter events before we see them. Without this gate, any
```
bartlomieju, line 124: Small nit on phrasing: chokidar's `ignored` option *does* drop events at the source, so `server.watch.ignored` isn't strictly bypassed. The real reason this gate matters is that the user's globs are hard to author correctly (especially on Windows with mixed separators) and shouldn't be load-bearing for avoiding KV temp file reloads. Could we reword to focus on "defense in depth: we shouldn't depend on user ignore globs being perfect" rather than the bypass claim?

bartlomieju (commented):
Diagnosis is correct, fix is well-scoped, test is solid. The watcher callback at server_snapshot.ts:76 was indeed sending `fresh:reload` for every chokidar event after the island early-return — gating on SSR-relevant changes is the right move, and the legitimate SSR-modification path through `dev_server.ts`'s `hotUpdate` hook keeps working.

A couple of optional notes:

- **Pre-existing behavior worth knowing.** Adding a *new* island file (one not yet in `islandsByFile`) no longer triggers `fresh:reload`. The pre-fix reload was effectively useless anyway — the snapshot wouldn't pick up the new island without a route add/unlink invalidating it — so this isn't a meaningful regression, just a behavior to be aware of.

- **Test timing.** The 1500ms negative wait is fine in principle (false negatives are tolerable for "no event arrived" assertions). On a busy CI runner you may occasionally see a spurious event sneak in just after the window — consider bumping to ~2s if it ever flakes.

- **Follow-up.** The original issue also reported `server.watch = null` not stopping reloads. Worth verifying separately whether Vite respects that with our direct `viteServer.watcher` access — if not, that's a Vite-side concern worth filing upstream.

Left one inline note on the comment-block phrasing.

PR #3810, opened 2026-05-14
fix: inject client entry on islands-free pages for HMR

### LINKED ISSUE #3806: Hot module reloading (HMR) doesn't work if a page has no islands
Right now when I use `deno run -Ar jsr:@fresh/init` to generate a new Fresh site/app and remove all of the code generated (the Counter island, etc.) and just have a `routes/index.tsx` with some basic HTML, hot module reloading (HMR) breaks. The page does not auto-refresh when I make changes in my editor.

I see in the Fresh source that this is a TODO:

https://github.com/denoland/fresh/blob/39b5f06f8a7d7fa02dd2e2950f2291d04ef9fea7/packages/plugin-vite/src/plugins/dev_server.ts#L189-L190

## Temporary Fix

I found a workaround that does get it working:

```diff
diff --git a/routes/_app.tsx b/routes/_app.tsx
index 0225605..f00f73b 100644
--- a/routes/_app.tsx
+++ b/routes/_app.tsx
@@ -7,6 +7,9 @@ export default define.page(function App({ Component }) {
         <meta charset="utf-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>My Fresh Site</title>
+        {import.meta.env.DEV && (
+          <script type="module" src="/@id/fresh:client-entry"></script>
+        )}
       </head>
       <body>
         <Component />
```

I'm not sure if that's appropriate as a long-term fix though. So I thought I'd open up this GitHub Issue.

It would be great if this just worked out of the box. Especially for server-side rendered apps that don't have any client-side code.

fibibot:
## Summary

Fixes denoland/fresh#3806: when a Fresh project has no islands, edits to
a route file in dev mode never refreshed the browser.

The dev server already broadcasts `fresh:reload` over Vite's WebSocket
when an SSR-only module changes (`packages/plugin-vite/src/plugins/dev_server.ts`),
but the listener for that event lives in `@fresh/plugin-vite/client`,
which is only loaded transitively via the client entry boot script. The
SSR runtime only emitted that boot script when the page had at least
one island or `hmrClientEntry` was defined. The Vite plugin never set
`hmrClientEntry`, so islands-free pages had no listener and edits were
silently dropped.

## Fix

Wire `hmrClientEntry` through the build snapshot so the SSR runtime
always emits the boot script in dev:

- Add `hmrClientEntry?: string` to `BuildSnapshot` and surface it on
  `ProdBuildCache` (used by the Vite snapshot loader).
- Have `generateSnapshotServer` emit a `hmrClientEntry` export when one
  is provided.
- Set `hmrClientEntry` to the client entry virtual id in the Vite
  plugin's `server_snapshot.ts` whenever the server is in dev mode.

This piggybacks on the existing FreshRuntimeScript branch that was
clearly written with this case in mind (note the existing comment:
\"Full-document boot: islands / partials / client nav, or Vite/HMR
dev\"); it just wasn't being triggered for the Vite plugin.

## Test plan

- [x] New regression test `vite dev - injects client entry on
      islands-free pages for HMR` in
      `packages/plugin-vite/tests/dev_server_test.ts` asserts the
      no-islands fixture's SSR response contains both
      `/@id/fresh:client-entry` and the `import { boot }` call.
- [x] `deno task ok` passes locally (613 tests passed, 0 failed).
- [x] Existing islands-related dev and build tests (`loads islands`,
      `partial island`, `nested islands`, `remote island`,
      `vite build - *`) all still pass.

Closes bartlomieju/orchid-inbox#85

### DIFF
--- a/packages/fresh/src/build_cache.ts
+++ b/packages/fresh/src/build_cache.ts
@@ -17,6 +17,12 @@ export interface FileSnapshot {
 export interface BuildSnapshot<State> {
   version: string;
   clientEntry: string;
+  /**
+   * Pathname for an HMR-only client entry. When defined, the SSR runtime
+   * always emits the boot script so HMR listeners attach to pages that
+   * have no islands. Undefined outside of dev.
+   */
+  hmrClientEntry?: string;
   fsRoutes: FsRouteFile<State>[];
   staticFiles: Map<string, FileSnapshot>;
   islands: ServerIslandRegistry;
@@ -51,6 +57,7 @@ export class ProdBuildCache<State> implements BuildCache<State> {
   #snapshot: BuildSnapshot<State>;
   islandRegistry: ServerIslandRegistry;
   clientEntry: string;
+  hmrClientEntry: string | undefined;
   features = { errorOverlay: false };
 
   constructor(public root: string, snapshot: BuildSnapshot<State>) {
@@ -58,6 +65,7 @@ export class ProdBuildCache<State> implements BuildCache<State> {
     this.#snapshot = snapshot;
     this.islandRegistry = snapshot.islands;
     this.clientEntry = snapshot.clientEntry;
+    this.hmrClientEntry = snapshot.hmrClientEntry;
   }
 
   getEntryAssets(): string[] {
--- a/packages/fresh/src/build_cache.ts
+++ b/packages/fresh/src/build_cache.ts
@@ -18,9 +18,9 @@ export interface BuildSnapshot<State> {
   version: string;
   clientEntry: string;
   /**
-   * Pathname for an HMR-only client entry. When defined, the SSR runtime
-   * always emits the boot script so HMR listeners attach to pages that
-   * have no islands. Undefined outside of dev.
+   * When defined, forces the boot script to be emitted in dev so HMR
+   * listeners attach even on island-free pages. The value itself is
+   * currently unused — only its presence matters. Undefined outside of dev.
    */
   hmrClientEntry?: string;
   fsRoutes: FsRouteFile<State>[];
--- a/packages/fresh/src/runtime/client/dev_hmr.ts
+++ b/packages/fresh/src/runtime/client/dev_hmr.ts
@@ -3,7 +3,7 @@ import { IS_BROWSER } from "../shared.ts";
 let ws: WebSocket;
 let revision = 0;
 
-let reconnectTimer: number;
+let reconnectTimer: ReturnType<typeof setTimeout>;
 const backoff = [
   // Wait 100ms initially, because we could also be
   // disconnected because of a form submit.
--- a/packages/plugin-vite/src/plugins/server_snapshot.ts
+++ b/packages/plugin-vite/src/plugins/server_snapshot.ts
@@ -165,10 +165,18 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
           const staticFiles: PendingStaticFile[] = [];
           let islandMods: IslandModChunk[] = [];
           let clientEntry = "/@id/fresh:client-entry";
+          let hmrClientEntry: string | undefined;
           let buildId = "";
           const entryAssets: string[] = [];
 
           if (isDev && server !== undefined) {
+            // The client entry hosts the HMR listener. Set hmrClientEntry so
+            // the SSR runtime always emits a boot script in dev, even when
+            // a page has zero islands. Without this, edits to islands-free
+            // routes never trigger a browser reload because the
+            // `fresh:reload` WebSocket listener is never attached.
+            hmrClientEntry = clientEntry;
+
             for (const id of islands.keys()) {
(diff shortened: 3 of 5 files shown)


bartlomieju (commented):
Diagnosis and fix look correct — I traced the chain end-to-end:

- SSR-only edits broadcast `fresh:reload` (`plugin-vite/src/plugins/dev_server.ts:228`).
- The listener lives in `plugin-vite/src/client.ts`, which is folded into `fresh:client-entry` via `client_entry.ts:54`.
- The boot script is gated on `needsClientRuntime || hmrClientEntry !== undefined` (`preact_hooks.ts:726`), and the Vite plugin previously never set `hmrClientEntry`, so islands-free pages never loaded the listener.

The legacy esbuild dev path already toggles `hmrClientEntry` at `dev/builder.ts:375`, so this brings the Vite plugin in line. A few small things worth addressing — see inline comments.

packages/fresh/src/build_cache.ts
```diff
@@ -17,6 +17,12 @@ export interface FileSnapshot {
 export interface BuildSnapshot<State> {
   version: string;
   clientEntry: string;
+  /**
+   * Pathname for an HMR-only client entry. When defined, the SSR runtime
+   * always emits the boot script so HMR listeners attach to pages that
+   * have no islands. Undefined outside of dev.
+   */
+  hmrClientEntry?: string;
```
bartlomieju, line 25: The JSDoc calls this a "Pathname for an HMR-only client entry," but in practice the value is never imported — only its presence gates the boot script in `preact_hooks.ts:726` and the preload header in `context.ts:420`. The boot script always imports `buildCache.clientEntry`, never `hmrClientEntry`.

That's fine, but the field is effectively a `bootInDev: boolean` masquerading as a pathname (the legacy esbuild builder happens to assign a real chunk path; the Vite plugin assigns the same string as `clientEntry`). Two options:

1. Keep the field, but rewrite the doc to describe it as a gate ("When defined, forces the boot script to be emitted in dev so HMR listeners attach even on island-free pages. The value itself is currently unused — only its presence matters.").
2. Replace with a boolean (e.g. `forceBootInDev`) in a follow-up cleanup.

Option 1 is enough for this PR.

packages/plugin-vite/src/plugins/server_snapshot.ts
```diff
@@ -165,10 +165,18 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
...
           let buildId = "";
           const entryAssets: string[] = [];
 
           if (isDev && server !== undefined) {
+            // The client entry hosts the HMR listener. Set hmrClientEntry so
+            // the SSR runtime always emits a boot script in dev, even when
+            // a page has zero islands. Without this, edits to islands-free
+            // routes never trigger a browser reload because the
+            // `fresh:reload` WebSocket listener is never attached.
+            hmrClientEntry = clientEntry;
```
bartlomieju, line 179: Assigning `clientEntry` here is functionally a marker — the value isn't read, only the `!== undefined` check matters (see `preact_hooks.ts:726`). Worth a short comment noting that the *presence* is what matters, so a future reader doesn't assume the value is plumbed through to an import.

packages/fresh/src/runtime/client/dev_hmr.ts
```diff
@@ -3,7 +3,7 @@ import { IS_BROWSER } from "../shared.ts";
 let ws: WebSocket;
 let revision = 0;
 
-let reconnectTimer: number;
+let reconnectTimer: ReturnType<typeof setTimeout>;
```
bartlomieju, line 6: Unrelated to the HMR fix — looks like an incidental type tightening. Either mention it in the PR description, or split into its own commit so the fix commit stays minimal.

packages/plugin-vite/tests/dev_server_test.ts
```diff
@@ -69,6 +69,23 @@ integrationTest("vite dev - starts without islands/ dir", async () => {
...
+  async () => {
+    const fixture = path.join(FIXTURE_DIR, "no_islands");
+    await withDevServer(fixture, async (address) => {
+      const res = await fetch(`${address}/`);
+      const text = await res.text();
+      expect(text).toContain("ok");
+      expect(text).toContain("/@id/fresh:client-entry");
+      expect(text).toMatch(/import\s*\{\s*boot\s*\}/);
+    });
+  },
```
bartlomieju, line 86: Good regression test for the missing script tag. Worth considering a follow-up that exercises the actual reload flow end-to-end (edit a route file, observe `fresh:reload` on the WS) if the harness supports it — the current assertion catches the symptom (script not emitted) but not a regression that emits the script while breaking the listener wiring. Not a blocker for this PR.

bartlomieju (approved):
Review feedback addressed. Rebased on main (dropping the duplicate reconnectTimer fix that landed in #3820). Merging.

PR #3812, opened 2026-05-22
feat: Make URL rewriting possible

maurictg:
Currently there is no way built-in way to rewrite routes in Fresh. Other frameworks like [SvelteKit](https://svelte.dev/docs/kit/hooks#Universal-hooks-reroute) and [Astro](https://docs.astro.build/en/guides/middleware/#rewriting) do offer this feature.

There are multiple reasons why rewriting routes is a desirable feature:
- Localized routes (i18n)
- Strangler fig pattern
- Canonical routes

I know there is currently one way to achieve rewriting without modifying Fresh' source code:
```js
import fetcher from './_fresh/server.js';

Deno.serve((req, info) => {
    const url = new URL(req.url);
    // your rewrite logic...

    return fetcher.fetch(new Request(url, req), info);
});
```

However, there are a few disadvantages to this strategy:
- Outside Fresh
- Not working with Vite
- Not possible to add data to the context/state (for instance the language)

All suggestions are welcome! To be honest: this is my first time ever contributing to Deno and/or Fresh, and there might be better approaches. But at least this PR serves as proof it is technically possible to have rewrite middleware.

### DIFF
--- a/docs/latest/concepts/context.md
+++ b/docs/latest/concepts/context.md
@@ -130,6 +130,28 @@ app.get("/old-url", (ctx) => {
 });
 ```
 
+## `.rewrite()`
+
+Rewrite a request internally to another route without redirecting the client.
+The browser URL stays the same, but Fresh rematches and handles the rewritten
+path.
+
+```ts
+app.use((ctx) => {
+  if (ctx.url.pathname.startsWith("/legacy/")) {
+    const pathname = ctx.url.pathname.replace("/legacy", "");
+    return ctx.rewrite(pathname);
+  }
+
+  return ctx.next();
+});
+```
+
+`ctx.rewrite()` only accepts same-origin targets.
+
+When the target is a string without a `?query`, Fresh keeps the current query
+parameters.
+
 ## `.render()`
 
 Render JSX and create a HTML `Response`.
--- a/docs/latest/concepts/context.md
+++ b/docs/latest/concepts/context.md
@@ -130,28 +130,6 @@ app.get("/old-url", (ctx) => {
 });
 ```
 
-## `.rewrite()`
-
-Rewrite a request internally to another route without redirecting the client.
-The browser URL stays the same, but Fresh rematches and handles the rewritten
-path.
-
-```ts
-app.use((ctx) => {
-  if (ctx.url.pathname.startsWith("/legacy/")) {
-    const pathname = ctx.url.pathname.replace("/legacy", "");
-    return ctx.rewrite(pathname);
-  }
-
-  return ctx.next();
-});
-```
-
-`ctx.rewrite()` only accepts same-origin targets.
-
-When the target is a string without a `?query`, Fresh keeps the current query
-parameters.
-
 ## `.render()`
 
 Render JSX and create a HTML `Response`.
--- a/packages/fresh/src/app.ts
+++ b/packages/fresh/src/app.ts
@@ -38,6 +38,47 @@ export const DEFAULT_CONN_INFO: any = {
   remoteAddr: { transport: "tcp", hostname: "localhost", port: 1234 },
 };
 
+const MAX_REWRITE_COUNT = 16;
+
+function normalizeRequestUrl(req: Request, trustProxy: boolean): URL {
+  const url = new URL(req.url);
+  // Prevent open redirect attacks.
+  url.pathname = url.pathname.replace(/\/+/g, "/");
+
+  // Apply X-Forwarded-* headers when behind a reverse proxy.
+  if (trustProxy) {
+    const proto = req.headers.get("x-forwarded-proto");
+    if (proto) {
+      url.protocol = proto + ":";
+    }
+    const host = req.headers.get("x-forwarded-host");
+    if (host) {
+      url.host = host;
+    }
+  }
+
+  return url;
+}
+
+function getRewriteUrl(currentUrl: URL, pathOrUrl: string | URL): URL {
+  const rewritten = pathOrUrl instanceof URL
+    ? new URL(pathOrUrl)
+    : new URL(pathOrUrl, currentUrl);
+
+  if (rewritten.origin !== currentUrl.origin) {
+    throw new Error(
+      `ctx.rewrite() only supports same-origin URLs. Expected "${currentUrl.origin}", got "${rewritten.origin}"`,
+    );
+  }
+
+  // Keep existing query params unless the target explicitly sets a query.
+  if (typeof pathOrUrl === "string" && !pathOrUrl.includes("?")) {
+    rewritten.search = currentUrl.search;
+  }
+
+  return rewritten;
+}
+
 const defaultOptionsHandler = (methods: string[]): () => Promise<Response> => {
   return () =>
     Promise.resolve(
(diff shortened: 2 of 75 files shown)


packages/fresh/src/app.ts
```diff
@@ -38,6 +38,47 @@ export const DEFAULT_CONN_INFO: any = {
...
+    }
+  }
+
+  return url;
+}
+
+function getRewriteUrl(currentUrl: URL, pathOrUrl: string | URL): URL {
+  const rewritten = pathOrUrl instanceof URL
+    ? new URL(pathOrUrl)
+    : new URL(pathOrUrl, currentUrl);
```
bartlomieju: basePath footgun: this resolves the string target against `currentUrl`, which means `ctx.rewrite("/new")` in an app with `basePath: "/base"` produces a URL of `/new` (not `/base/new`) and won't match any registered routes. The basePath test on line 348 works around this by passing a full `URL` object.

Two reasonable options:
- Auto-prefix `this.config.basePath` for string targets that don't already start with it (mirrors how route registration works).
- Document explicitly that string targets must be absolute paths including the basePath, and only `URL` targets bypass that.

Either is fine, but the current behavior is going to surprise users with a non-empty basePath.

packages/fresh/src/app.ts
```diff
@@ -473,6 +501,24 @@ export class App<State> {
...
+          }
+
+          if (req.bodyUsed) {
+            throw new Error(
+              "Cannot rewrite request after its body has already been consumed",
+            );
+          }
+
+          const rewrittenUrl = getRewriteUrl(url, pathOrUrl);
+          const rewrittenReq = new Request(rewrittenUrl, req);
```
bartlomieju: Two small notes on `new Request(rewrittenUrl, req)`:

1. This transfers ownership of the body stream from `req` to the rewritten request. Any middleware that planned to read `ctx.req` after calling `ctx.rewrite()` will see an already-consumed body. Worth a one-liner in the JSDoc on `Context.rewrite`.
2. For streamed request bodies, some runtimes require `{ duplex: "half" }` when constructing a `Request` from a body-bearing init. Deno's current stable runtime appears to tolerate this, but it would be safer to set `duplex: "half"` explicitly when `req.body !== null`.

packages/fresh/src/app.ts
```diff
@@ -38,6 +38,47 @@ export const DEFAULT_CONN_INFO: any = {
   remoteAddr: { transport: "tcp", hostname: "localhost", port: 1234 },
 };
 
+const MAX_REWRITE_COUNT = 16;
```
bartlomieju: 16 feels generous for an internal-rewrite cap — even legitimate rewrite chains rarely go past 2–3 hops. Not a blocker, but `8` would catch buggy middleware sooner with fewer wasted `Request`/`URL` allocations. Up to you.

packages/fresh/src/context.ts
```diff
@@ -195,16 +204,41 @@ export class Context<State> {
...
+   * app.use((ctx) => {
+   *   if (ctx.url.pathname.startsWith("/legacy/")) {
+   *     return ctx.rewrite(ctx.url.pathname.replace("/legacy", ""));
+   *   }
+   *
+   *   return ctx.next();
+   * });
+   * ```
+   */
+  rewrite(pathOrUrl: string | URL): Promise<Response> {
```
bartlomieju, line 293: Worth folding two facts from the prose docs into the JSDoc so they show up in editor tooltips:
- Only same-origin targets are accepted (throws otherwise).
- The rewrite transfers ownership of the request body — `ctx.req` won't be readable afterwards in the calling middleware.

Also consider an `@see` cross-link to `redirect()` since they're conceptually paired.

packages/fresh/src/app_test.tsx
```diff
@@ -235,6 +235,202 @@ Deno.test("App - methods with middleware", async () => {
...
+    .use((ctx) => {
+      const [, first, ...rest] = ctx.url.pathname.split("/");
+
+      if (ctx.state.locale === undefined && LOCALES.has(first)) {
+        ctx.state.locale = first;
+      }
+
+      if (LOCALES.has(first)) {
+        const rewritten = `/${rest.join("/")}`;
+        return ctx.rewrite(rewritten === "/" ? "/" : rewritten);
```
bartlomieju: nit: `rewritten === "/" ? "/" : rewritten` is a no-op — both branches return the same value. Can just be `ctx.rewrite(rewritten)`.

docs/latest/concepts/context.md
```diff
@@ -130,6 +130,28 @@ app.get("/old-url", (ctx) => {
 });
 ```
 
+## `.rewrite()`
```
bartlomieju: `docs/latest/` is the Fresh 1.x docs surface. Fresh 2's docs live under `docs/canary/the-canary-version/`, which (at the moment) doesn't have `context.md` / `middleware.md` equivalents. Worth confirming with maintainers whether this documentation should be mirrored/moved there — otherwise users on Fresh 2 won't find these docs.

bartlomieju (commented):
Nice feature, and well-scoped for a first contribution. The security-relevant pieces (same-origin enforcement, path normalization on every dispatch, loop cap, body-used guard) all look right, and the test coverage is thorough.

A few things to consider, mostly polish — see inline comments. The two that matter most:

1. Behavior of string targets when `basePath` is configured — currently `new URL("/new", ctx.url)` will drop the basePath segment, so `ctx.rewrite("/new")` won't match routes registered under the basePath. The basePath test sidesteps this by passing a full `URL`. Worth deciding whether string targets should be auto-prefixed with basePath, or documenting that they must include it.
2. The docs are added under `docs/latest/concepts/`, which is the Fresh 1.x docs surface. The Fresh 2 canary docs live under `docs/canary/the-canary-version/`. Worth confirming with maintainers where this should live.

Minor nice-to-have: a test that asserts `ctx.params` is repopulated against the new route after a rewrite (e.g. rewrite `/u/123` → `/users/:id` and read `ctx.params.id` in the target). The current tests cover `route` and `url`, but not `params`.

packages/fresh/src/app.ts
maurictg: Let's go with the auto-prefixing to keep it consistent with the route registration

packages/fresh/src/app.ts
maurictg: 1. True, I'll add that
2. I'll change it to half when a request body is present, but I am not sure if Deno supports this fully. I assume so. The `duplex` property is by the way not in de VScode built in types (lib.dom.d.ts) but it is in the MDN docs.

packages/fresh/src/app.ts
maurictg: You're absolutely right. Lets go with 8.

packages/fresh/src/app_test.tsx
maurictg: My bad, it was a hot afternoon

docs/latest/concepts/context.md
maurictg: Allright, I'll create a file there and one might decide to move it to the right place.
But.. Fresh 2 isn't canary anymore right?

maurictg:
@bartlomieju Are you willing to review again? I processed your feedback!

PR #3816, opened 2026-05-23
fix: bypass the request when the `server.proxy` is specified

### LINKED ISSUE #3814: Bug: `server.proxy` not working with `@fresh/plugin-vite`
# Step to Reproduce

- deno.json

```json
{
  "imports": {
    "vite": "npm:vite@7.1.11",
    "preact": "npm:preact@10.29.1",
    "fresh": "jsr:@fresh/core@2.3.3",
    "@fresh/plugin-vite": "jsr:@fresh/plugin-vite@1.1.2"
  },
  "nodeModulesDir": "auto",
  "compilerOptions": {
    "jsx": "precompile",
    "jsxImportSource": "preact",
    "jsxImportSourceTypes": "preact"
  },
  "tasks": {
    "start": "deno task proxy & deno task dev",
    "dev": "deno run -REWS --allow-run --allow-ffi --allow-net=127.0.0.1:3000,jsr.io:443 vite --host 127.0.0.1 --port 3000",
    "proxy": "deno run --allow-net=127.0.0.1:8000 proxy.ts"
  }
}
```

- vite.config.ts

```ts
import { defineConfig } from 'vite';
import { fresh } from '@fresh/plugin-vite';

export default defineConfig({
  plugins: [fresh()],
  server: {
    host: '127.0.0.1',
    port: 3000,
    proxy: {
      '/api': 'http://127.0.0.1:8000',
    },
  },
});
```

- main.ts

```ts
import { App } from 'fresh';

export const app = new App()
  .get('/', () => new Response('ok'));
```

- proxy.ts

```ts
Deno.serve({ hostname: '127.0.0.1', port: 8000 }, (req) => {
  const url = new URL(req.url);
  console.log(`proxy ${req.method} ${url.pathname}${url.search}`);
  return Response.json({
    ok: true,
    pathname: url.pathname,
    search: url.search,
  });
});
```

## Expected behaviour

```bash
% curl -i 'http://127.0.0.1:3000/api/ping?x=1'
HTTP/1.1 200 OK
content-type: application/json
content-length: 50
date: Fri, 22 May 2026 14:44:50 GMT

{"ok":true,"pathname":"/api/ping","search":"?x=1"}%
```

## Acutual behaviour

```bash
% curl -i 'http://127.0.0.1:3000/api/ping?x=1'
HTTP/1.1 404
Vary: Origin
content-type: text/plain;charset=UTF-8
Date: Fri, 22 May 2026 14:41:30 GMT
Connection: keep-alive
Keep-Alive: timeout=5
Transfer-Encoding: chunked
```

Hajime-san:
- fixes https://github.com/denoland/fresh/issues/3814

### DIFF
--- a/packages/plugin-vite/src/plugins/dev_server.ts
+++ b/packages/plugin-vite/src/plugins/dev_server.ts
@@ -13,6 +13,11 @@ function getContentType(ext: string): string {
 /**
  * Handling the user config of proxy
  * https://vite.dev/config/server-options#server-proxy
+ *
+ * Note: per-entry `bypass(req, res, options)` callbacks are not supported here.
+ * Requests matched by a proxy key are always short-circuited to Vite's proxy
+ * handler; a `bypass` that returns `false` would still skip Fresh but then
+ * fall through the proxy without being proxied, resulting in a 404.
  */
 function createProxyUrlMatcher(
   proxy: ResolvedServerOptions["proxy"],
@@ -55,7 +60,7 @@ export function devServer(freshConfig: ResolvedFreshViteConfig): Plugin[] {
         const IGNORE_URLS = new RegExp(
           `^(${base})?/(@(vite|fs|id)|\\.vite)/`,
         );
-        // build proxy url list matcher beofre the request is coming
+        // Precompute the proxy URL matcher; proxy config is fixed at server start.
         const matchesProxyUrl = createProxyUrlMatcher(
           server.config.server.proxy,
         );
--- a/packages/plugin-vite/src/plugins/dev_server.ts
+++ b/packages/plugin-vite/src/plugins/dev_server.ts
@@ -1,4 +1,4 @@
-import type { DevEnvironment, Plugin } from "vite";
+import type { DevEnvironment, Plugin, ResolvedServerOptions } from "vite";
 import * as path from "@std/path";
 import { contentType as getStdContentType } from "@std/media-types/content-type";
 import { ASSET_CACHE_BUST_KEY } from "fresh/internal";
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Approach looks right — matches Vite's own `startsWith`/`^`-prefix semantics, and precomputing the matcher in `configureServer` is nicer than rebuilding regexes per request. Left a few inline notes; once those are addressed I think this is good to go.

packages/plugin-vite/src/plugins/dev_server.ts
```diff
@@ -27,15 +55,25 @@ export function devServer(freshConfig: ResolvedFreshViteConfig): Plugin[] {
         const IGNORE_URLS = new RegExp(
           `^(${base})?/(@(vite|fs|id)|\\.vite)/`,
         );
+        // build proxy url list matcher beofre the request is coming
```
bartlomieju: Typo: `beofre` → `before`. Also worth rewording — the current phrasing is a bit hard to parse.

```suggestion
        // Precompute the proxy URL matcher; proxy config is fixed at server start.
```

packages/plugin-vite/src/plugins/dev_server.ts
```diff
@@ -10,6 +10,34 @@ function getContentType(ext: string): string {
   return getStdContentType(ext) ?? "application/octet-stream";
 }
 
+/**
+ * Handling the user config of proxy
+ * https://vite.dev/config/server-options#server-proxy
+ */
+function createProxyUrlMatcher(
```
bartlomieju, line 22: One behavioral gap worth flagging: Vite's proxy options support a `bypass(req, res, options)` callback per entry that can return `false` to opt out of proxying for a given request. With this matcher, such requests would still be short-circuited away from Fresh and then fall through Vite's proxy without being proxied, ending up as 404s. Probably uncommon in practice — fine to leave for a follow-up, but a code comment noting the limitation would help future readers.

packages/plugin-vite/tests/dev_server_test.ts
```diff
@@ -465,6 +466,85 @@ integrationTest(
...
+      {
+        const res = await fetch(`${address}/api2/pong?y=2`);
+        expect(res.status).toEqual(200);
+        expect(await res.text()).toEqual("api2");
+      }
+      {
+        const res = await fetch(`${address}/api3/pong?z=3`);
+        expect(res.status).toEqual(200);
+        expect(await res.text()).toEqual("api3");
+      }
```
bartlomieju, line 540: Consider adding a negative-case assertion so we prove the bypass is *selective* — e.g. that `/` still hits the Fresh route. Without it, a regression that bypasses every request would still pass these checks.

```suggestion
      {
        const res = await fetch(`${address}/api3/pong?z=3`);
        expect(res.status).toEqual(200);
        expect(await res.text()).toEqual("api3");
      }
      {
        const res = await fetch(`${address}/`);
        expect(res.status).toEqual(200);
        expect(await res.text()).toEqual("ok");
      }
```

Hajime-san: Exactly, thanks!

PR #3822, opened 2026-05-26
docs: websockets do not work with vite dev server pre-Deno 2.8 and Fresh 2.4

will-weiss:
Updating the docs to reflect the fact that upgrading websocket requests isn't currently supported by the vite dev server.

This cost me some time as there isn't any warning in the console on the server, the upgrade request merely hangs. Perhaps that could be fixed as well, but this update to the docs (while verbose) would document the behavior.

### DIFF
--- a/docs/latest/advanced/websockets.md
+++ b/docs/latest/advanced/websockets.md
@@ -6,6 +6,24 @@ description: |
 Fresh provides built-in helpers for upgrading HTTP connections to WebSockets.
 There are two main approaches depending on your use case.
 
+> [warn]: WebSocket upgrades do not work under the Vite dev server (`deno task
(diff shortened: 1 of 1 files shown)


bartlomieju:
This is expected to work with Deno V2.8: https://deno.com/blog/v2.8#miscellaneous

will-weiss:
@bartlomieju thanks for the quick response!

Hm, I am running deno 2.8 and yet I still appear to be encountering this issue.

bartlomieju:
Good catch, I forgot about a counterpart needed in Fresh. I opened https://github.com/denoland/fresh/pull/3823 to fix that and will do a 2.4 release afterwards.

Issue #3596, opened 2025-11-04
JSR: Improve documentation of public symbols


csvn:
This issue is related to #3597, and tracks progress of documentation of public symbols.

At the time of the creation of this issue, this is the percentage of documented symbols for all packages under the `@fresh` scope on JSR:

- [`@fresh/build-id`](https://jsr.io/@fresh/build-id): 0% - ⛔
- [`@fresh/core`](https://jsr.io/@fresh/core): 44% - ⚠️ 
- [`@fresh/examples`](https://jsr.io/@fresh/examples): 0% - ⛔
- [`@fresh/init`](https://jsr.io/@fresh/init): 100% - ✅
- [`@fresh/plugin-tailwind`](https://jsr.io/@fresh/plugin-tailwind): 0% - ⛔
- [`@fresh/plugin-tailwind-v3`](https://jsr.io/@fresh/plugin-tailwind-v3): 0% - ⛔
- [`@fresh/plugin-vite`](https://jsr.io/@fresh/plugin-vite): 0% - ⛔
- [`@fresh/update`](https://jsr.io/@fresh/update): 100% - ✅

I think it would be a good goal to aim for at least 80% to assuage JSR, but being closer to 100% documentation of public symbols would be even better. Not every exported type needs large text or examples, and a small description can often be helpful for users to learn the framework.

## Progress

Tracking progress of packages with >80% documented public symbols.

- [ ] [`@fresh/build-id`](https://jsr.io/@fresh/build-id)
- [ ] [`@fresh/core`](https://jsr.io/@fresh/core)
- [x] [`@fresh/examples`](https://jsr.io/@fresh/examples): #3603
- [x] [`@fresh/init`](https://jsr.io/@fresh/init)
- [ ] [`@fresh/plugin-tailwind`](https://jsr.io/@fresh/plugin-tailwind) #3832
- [ ] [`@fresh/plugin-tailwind-v3`](https://jsr.io/@fresh/plugin-tailwind-v3)
- [x] [`@fresh/plugin-vite`](https://jsr.io/@fresh/plugin-vite): #3604
- [x] [`@fresh/update`](https://jsr.io/@fresh/update)

PR #3832, opened 2026-06-05
docs: add JSDoc to @fresh/plugin-tailwind public symbols

gaurav0107:
## What

Adds JSDoc to the public surface of `@fresh/plugin-tailwind` and corrects the README usage snippets.

- `tailwind()` (mod.ts): module-level `@module` block plus a function-level JSDoc with `@param`, two `@example` blocks, and a description of the production/development `optimize` default.
- `TailwindPluginOptions` / `PluginOptions` (types.ts): JSDoc on every property, plus an `@example` on the public interface.
- `README.md`: the `Basic Usage` and `Option Configuration` snippets called `tailwind(builder, app, ...)`, which does not match the actual `(builder, options)` signature. The `base: null` example also did not match the `base?: string` type, so it was dropped.

## Why

Tracking issue #3596 lists `@fresh/plugin-tailwind` at 0% documented public symbols on JSR. This change brings the v4 plugin to 100% on the public exports without any runtime change.

The README correction is a drive-by fix for a real bug — anyone copy-pasting the existing snippet hits a TypeScript error (extra positional argument).

## Scope and risk

Docs-only. No exported types or values are renamed, removed, or re-typed. Behavior of `tailwind()` is unchanged.

Refs #3596

### DIFF
--- a/packages/plugin-tailwindcss/README.md
+++ b/packages/plugin-tailwindcss/README.md
@@ -11,7 +11,7 @@ import { app } from "./main.ts";
 import { tailwind } from "@fresh/plugin-tailwind";
 
 const builder = new Builder();
-tailwind(builder, app);
+tailwind(builder);
 
 if (Deno.args.includes("build")) {
   builder.build(app);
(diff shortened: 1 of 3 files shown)


gaurav0107:
Gentle nudge on this docs PR — still open and green. Happy to rebase or adjust if there's any feedback. Thanks for your time!

PR #3839, opened 2026-06-09
docs: recommend --lock-write over manual lockfile editing

SisyphusZheng:
### Problem

When the lockfile contains unpinned remote specifiers (e.g. 
efs/heads/main), upstream pushes can cause deno install to fail with an integrity check error. The current AGENTS.md instructs contributors to **manually edit the hash** in deno.lock, which is error-prone and unnecessary.

### Root cause

denoland/deno#32991 was closed as working as intended — Deno's integrity check is correct. deno cache --reload re-fetches content but still validates against the old integrity, so it fails. The proper approach is --lock-write, which tells Deno to accept new content and update the lockfile.

### Solution

Replace the manual hash editing instruction in AGENTS.md with the deno install --lock-write command, which is Deno's official recommended fix.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -52,11 +52,15 @@ with workspace members in `packages/*` and `www/`.
 
 ### Lockfile quirks
 
-The lockfile contains remote specifiers pointing to `refs/heads/main` (e.g.
-`raw.githubusercontent.com/.../refs/heads/main/...`). These hashes go stale when
-upstream pushes. When that happens, manually update the hash in `deno.lock`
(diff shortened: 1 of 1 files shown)


SisyphusZheng:
Closing: \--lock-write\ is not a valid Deno CLI flag. Verified against Deno v2.8.3: \deno install --lock-write\ produces \error: unexpected argument '--lock-write' found\. The original recommendation of manually editing lockfile hashes, while tedious, is at least executable. Reverting this change.

PR #3840, opened 2026-06-09
fix: use namespace imports for packages losing default exports on canary

SisyphusZheng:
﻿### Problem

Deno canary CI fails with TS1192: Module has no default export for two npm packages:

| File | Import | Error |
|---|---|---|
| packages/plugin-tailwindcss/src/mod.ts | import twPostcss from "@tailwindcss/postcss" | TS1192 |
| packages/plugin-vite/src/mod.ts | import prefresh from "@prefresh/vite" | TS1192 |

The default exports exist at runtime but their .d.ts type declarations lose
the default export signature on canary builds.

### Changes

**packages/plugin-tailwindcss/src/mod.ts**
- import twPostcss from "@tailwindcss/postcss" → import * as twPostcss from "@tailwindcss/postcss"
- 	wPostcss({...}) → 	wPostcss.default({...})

**packages/plugin-vite/src/mod.ts**
- import prefresh from "@prefresh/vite" → import * as prefresh from "@prefresh/vite"
- prefresh({...}) → prefresh.default({...})

### Verification

deno task check:types passes on stable Deno v2.7.14.

### DIFF
--- a/packages/plugin-tailwindcss/src/mod.ts
+++ b/packages/plugin-tailwindcss/src/mod.ts
@@ -1,5 +1,5 @@
 import type { Builder } from "fresh/dev";
-import twPostcss from "@tailwindcss/postcss";
+import * as twPostcss from "@tailwindcss/postcss";
 import postcss from "postcss";
 import type { TailwindPluginOptions } from "./types.ts";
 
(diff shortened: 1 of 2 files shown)


SisyphusZheng:
Closing: this has been superseded by #3857 which uses \@ts-ignore TS1192\ with a detailed comment. The \@ts-ignore\ approach is cleaner (no \.default()\ boilerplate at call sites) and self-documents the root cause (upstream packages have self-contradictory CJS types + ESM exports). Both approaches achieve the same result at runtime — #3857 is the preferred fix.

PR #3883, opened 2026-06-26
fix: deleting an island wedges the dev server

aiibe:
## The problem

Delete any file from `islands/` while running `deno task dev` and the app crashes on reload page with a permanent 500 until you restart. Fresh auto discovers that folder so it happens even when the island isn't used anywhere.

Second bug: delete one island while another is still on the page and the surviving island quietly stops reacting after the next reload.

## Steps to reproduce

1. Run `deno task dev` and open `/about`. It renders the `Foo` and `Bar` islands.
2. In `about.tsx` remove `<Bar />` and its import so the page keeps only `Foo`.
3. Delete `islands/Bar.tsx`.
4. Before the fix the page 500s, or reloads with `Foo` rendered but dead so it no longer reacts to clicks.

## What was happening

The watcher only looked at `routes/` so deleting an island never triggered a rebuild. The snapshot map only ever added islands and never dropped deleted ones, leaving a stale static import that choked the server. And on a cold client graph right after a delete, a surviving island's boot import got recorded as the internal `fresh-island::Name` specifier, so the browser got a bare import it couldn't resolve and the island never hydrated.

## The fix

The watcher now tracks islands too. The `load` handler rebuilds the island map from scratch on every reload instead of appending, with a small cache so names stay stable. And it skips the `fresh-island::Name` specifier when recording a chunk, keeping the real `/@id/...` URL so a survivor always boots.

## Testing

One regression test loads a page with a live island, deletes a sibling, then checks the server stays at 200 and the survivor still hydrates and stays clickable with no `fresh-island::` in the HTML.

### DIFF
--- a/packages/plugin-vite/src/plugins/server_snapshot.ts
+++ b/packages/plugin-vite/src/plugins/server_snapshot.ts
@@ -50,11 +50,38 @@ export function serverSnapshot(options: ResolvedFreshViteConfig): Plugin[] {
   const islands = new Map<string, { name: string; chunk: string | null }>();
   const islandsByFile = new Set<string>();
   const islandSpecByName = new Map<string, string>();
+  const islandNameBySpec = new Map<string, string>();
   const routeNamer = new UniqueNamer();
   const routeFileToName = new Map<string, string>();
 
   // deno-lint-ignore no-explicit-any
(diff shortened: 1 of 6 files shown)


aiibe:
Pushed a follow up. Fixing the wedge exposed a related client side bug where a surviving sibling island stopped hydrating. Added a guard and updated the description.

PR #3886, opened 2026-06-29
docs: update define imports to use @/utils.ts alias

eaallen:
## Summary
- Update outdated documentation code examples that imported `define` from relative paths (`../utils.ts` and `../../utils.ts`)
- Replace them with the `@/utils.ts` alias to match current Fresh project conventions

## Test plan
- [x] Verified no remaining `../utils.ts` or `../../utils.ts` imports in `docs/latest/`
- [x] Spot-check updated pages on usefresh.dev after merge (getting started, middleware, layouts, testing, advanced guides)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/latest/advanced/app-wrapper.md
+++ b/docs/latest/advanced/app-wrapper.md
@@ -24,7 +24,7 @@ If you're using [file-based routing](/docs/concepts/file-routing), create a
 ## Basic example
 
 ```tsx routes/_app.tsx
-import { define } from "../utils.ts";
+import { define } from "@/utils.ts";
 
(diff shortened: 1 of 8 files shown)


eaallen (commented):
I just noticed some of the examples use the old ../utils import method instead of @/utils. When I was first learning about fresh having to deal with the relative imports was frustrating. So I am updating the docs to make this smoother for other people. 

I will also mention that some of the docs already used the @/utils, so this is just ensuring we use the same semantics throughout the docs.

eaallen:
crickets for what is so obviously an easy PR to accept or reject.

Issue #3895, opened 2026-07-30
Dev server fails with ENOENT errors when optimizeDeps enabled in Vite config


mryall-mawson:
Enabling `optimizeDeps` in `vite.config.ts` makes the Fresh dev server return 500 for every request.

With optimizeDeps enabled, the Deno loader plugin receives a module id carrying Vite's `?v=<hash>` suffix and passes it to `path.toFileUrl()`, which percent-encodes the `?` into the filename, which throws an exception because the file cannot be found:

    [vite] Internal server error: Unable to load
      /…/preact@10.29.7/…/preact/debug/dist/debug.module.js?v=ff8da874:
      ENOENT: no such file or directory, open
      '/…/preact@10.29.7/…/preact/debug/dist/debug.module.js?v=ff8da874'

Both environments are affected — client and `environments.ssr` fail identically, on the `.module.js` and `.mjs` builds respectively.

The failing module is not the one being optimized: adding `preact` to `optimizeDeps` produces the error for `preact/debug`. Vite versions dependency ids whether or not it pre-bundles them, so excluding one module just moves the error to the next.

### Reproduction

See https://github.com/mryall-mawson/fresh-vite-loader-repro

`app/` is an unmodified `deno run -Ar jsr:@fresh/init@^2` scaffold; the only edit is `vite.config.ts`, switching `optimizeDeps` on from an env var:

    ./run.sh
    without optimizeDeps: HTTP 200
    with optimizeDeps   : HTTP 500

`repro.ts` in the same repo shows the defect with no Vite or Fresh involved.

### Expected behaviour

The page renders with `optimizeDeps` enabled.

### Versions

@fresh/core 2.3.3, @fresh/plugin-vite 1.1.2, vite 7.3.6, deno 2.9.4

### Cause

`packages/plugin-vite/src/plugins/deno.ts:220`

```js
const url = path.toFileUrl(id);
const result = await loader.load(url.href, meta.type);

PR #3896, opened 2026-07-30
fix(plugin-vite): correct module resolution when Vite optimizeDeps is enabled (#3895)

mryall-mawson:
Enabling `optimizeDeps` in Vite config causes the Fresh dev server to fail with file-not-found errors. In this mode, modules URLs include a query string which is not stripped before being passed to Deno's loader, which interprets them as filenames.

See #3895 for the full bug description and reproduction.

This PR provides a simple test case using an existing fixture, and the fix in `plugin-vite` to strip the query strings before loading.

Note there are some existing test failures on `main` which this PR does not attempt to fix.

### DIFF
--- a/packages/plugin-vite/src/plugins/deno.ts
+++ b/packages/plugin-vite/src/plugins/deno.ts
@@ -9,7 +9,7 @@ import {
 import * as path from "@std/path";
 import * as babel from "@babel/core";
 import { httpAbsolute } from "./patches/http_absolute.ts";
-import { JS_REG, JSX_REG } from "../utils.ts";
+import { cleanId, JS_REG, JSX_REG } from "../utils.ts";
 import { builtinModules } from "node:module";
 
 // @ts-ignore Workaround for https://github.com/denoland/deno/issues/30850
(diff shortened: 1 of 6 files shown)


mryall-mawson:
I made a couple more fixes in this PR to get Vite’s `optimizeDeps` fully working in Fresh for client-side modules, with new tests to reproduce the issues I found along the way.

It is now tested and working for me, and I've released a fork here which we will use until this fix is released: https://jsr.io/@mawson/fresh-plugin-vite

We saw a reduction in our functional test execution time from 3.5 min to 2.5 mins against a dev server with optimizeDeps enabled for some large and deep dependency trees.

Edit: the specific error I saw was: `Cannot read properties of undefined (reading '__$f')` This indicated a duplicate copy of Preact was loaded on the front-end.

mryall-mawson:
It looks like another PR, #3898, is trying to fix a related issue in a different way, so it would be good if a maintainer could review and see which approach is more correct.

Btw, the CI checks on this branch were not approved within 30 days, so they are now expired and will probably need to be kicked off manually. The CI feedback would have been handy around the time I was actively working on this.

